PR #2, opened 2019-12-20
 Setup CMake project with unit tests 

### LINKED ISSUE #1: CMake and unit testing stub
- Add CMake main script
- GTest dependency
- Single unit test leb128.
- Circle CI

chfast:
- CMake project setup
- Unit tests based on GTest (using system installed gtest-dev and FindGTest CMake's standard module).
- Clang-format from evmone

All is up to discussion, but make it quick.

Closes #1.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,14 @@
 cmake_minimum_required(VERSION 3.13)
 
+include(CMakeDependentOption)
+
 option(FIZZY_TESTING "Enable Fizzy internal tests" OFF)
+cmake_dependent_option(HUNTER_ENABLED "Enable Hunter package manager" ON
+    "FIZZY_TESTING" OFF)
+
+if(HUNTER_ENABLED)
+    include(cmake/Hunter/init.cmake)
+endif()
 
 project(fizzy LANGUAGES CXX)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,18 @@
+cmake_minimum_required(VERSION 3.13)
+
+option(FIZZY_TESTING "Enable Fizzy internal tests" OFF)
+
+project(fizzy LANGUAGES CXX)
+
+set(include_dir ${PROJECT_SOURCE_DIR}/include)  # Public include directory.
+
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/lib)
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/lib)
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/bin)
+
+add_subdirectory(lib)
+
+if(FIZZY_TESTING)
+    enable_testing()  # Enable CTest. Must be done in main CMakeLists.txt.
+    add_subdirectory(test)
+endif()
--- a/lib/fizzy/CMakeLists.txt
+++ b/lib/fizzy/CMakeLists.txt
@@ -1,0 +1,10 @@
+
+add_library(fizzy)
+add_library(fizzy::fizzy ALIAS fizzy)
+
+target_sources(
+    fizzy PRIVATE
+    dummy.cpp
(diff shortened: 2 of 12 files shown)


CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+cmake_minimum_required(VERSION 3.13)
```
chfast: Using recent CMake versions is just convenient because of nice new features. You just spend less time programming CMake. Also testing older versions is difficult because CIs usually have very recent versions.

CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+
+option(FIZZY_TESTING "Enable Fizzy internal tests")
```
chfast: This is OFF by default, so you should explicitly enable it when developing fizzy, but users can use default options.

CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+
+project(fizzy LANGUAGES CXX)
```
chfast: By default CMake enables both C and C++. Explicitly enabling only C++ makes configuration slightly faster.

CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+
+set(include_dir ${PROJECT_SOURCE_DIR}/include)  # Public include directory.
```
chfast: Not used any more, but should be included in the fizzy lib when we will have any public header. This can be removed from this PR.

CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+
+add_subdirectory(lib)
```
chfast: I used `lib/fizzy` dir structure (based on evmone, based on LLVM), but other structures are possible. E.g. `src/fizzy; src/runner` or `src/lib/fizzy; src/tools/runner`.

lib/fizzy/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+
+add_library(fizzy)
+add_library(fizzy::fizzy ALIAS fizzy)
```
chfast: It is best practice to also define `fizzy::fizzy` alias. It is better to use the alias when specifying link dependencies (the `::` indicates that it is a CMake target, the name `fizzy` can also mean a system lib - no CMake error then).
This is also useful when including fizzy as a CMake subproject.

lib/fizzy/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
...
+
+target_sources(
```
chfast: Using `target_sources()` instead of adding files  to `add_library` is purely pedantic at this stage. I don't believe `PRIVATE`/`PUBLIC` for sources has any useful meaning.

lib/fizzy/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
...
+    fizzy PRIVATE
+    dummy.cpp
+    leb128.hpp
+)
+target_compile_features(fizzy PUBLIC cxx_std_14)
```
chfast: Specify that fizzy lib requires C++14 support.

test/unittests/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+
+include(GoogleTest)
+
+find_package(GTest REQUIRED)
+
+add_executable(fizzy-unittests)
+target_link_libraries(fizzy-unittests PRIVATE fizzy::fizzy GTest::Main)
+target_include_directories(fizzy-unittests PRIVATE ${PROJECT_SOURCE_DIR}/lib/fizzy)
```
chfast: This allows including fizzy's internal headers for testing.

CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+cmake_minimum_required(VERSION 3.13)
+
+option(FIZZY_TESTING "Enable Fizzy internal tests" OFF)
```
chfast: Enables internal testing. When developing fizzy, this should be set to `ON`, but users can use default options values.

chfast:
To be continued:
- Add coverage build (requires 2 modules from Cable).
- Upload to codecov.

PR #3, opened 2019-12-20
LEB128 decoder

gumb0:
https://webassembly.github.io/spec/core/binary/values.html#integers
https://en.wikipedia.org/wiki/LEB128

### DIFF
--- a/lib/fizzy/leb128.hpp
+++ b/lib/fizzy/leb128.hpp
@@ -2,8 +2,18 @@
 
 namespace fizzy
 {
-inline constexpr uint64_t leb128_decode(const uint8_t*) noexcept
+inline constexpr uint64_t leb128u_decode(const uint8_t* input) noexcept
 {
-    return 0;
+    uint64_t result = 0;
+    uint8_t result_shift = 0;
+
+    for (auto byte = input;; ++byte, result_shift += 7)
+    {
+        result |= static_cast<uint64_t>(*byte & 0x7F) << result_shift;
+        if ((*byte & 0x80) == 0)
+            break;
+    }
+
+    return result;
 }
 }  // namespace fizzy
--- a/lib/fizzy/leb128.hpp
+++ b/lib/fizzy/leb128.hpp
@@ -1,19 +1,20 @@
 #include <cstdint>
+#include <stdexcept>
 
 namespace fizzy
 {
-constexpr uint64_t leb128u_decode(const uint8_t* input) noexcept
+uint64_t leb128u_decode(const uint8_t* input, size_t size)
 {
     uint64_t result = 0;
     int result_shift = 0;
 
-    for (auto byte = input;; ++byte, result_shift += 7)
+    for (auto byte = input; byte != input + size && result_shift <= 63; ++byte, result_shift += 7)
     {
         result |= static_cast<uint64_t>(*byte & 0x7F) << result_shift;
         if ((*byte & 0x80) == 0)
-            break;
+            return result;
     }
 
-    return result;
+    throw std::runtime_error("Invalid unsigned integer encoding.");
 }
 }  // namespace fizzy
--- a/lib/fizzy/leb128.hpp
+++ b/lib/fizzy/leb128.hpp
@@ -12,6 +12,9 @@ uint64_t leb128u_decode(bytes_view input)
     for (auto byte = input.cbegin(); byte != input.cend() && result_shift <= 63;
          ++byte, result_shift += 7)
     {
+        // TODO this ignores the bits in the 10th byte other than the least significant one
+        // (when result_shift==63)
+        // So would not reject some invalid encoding with those with bits set.
         result |= static_cast<uint64_t>(*byte & 0x7F) << result_shift;
         if ((*byte & 0x80) == 0)
             return result;
--- a/lib/fizzy/leb128.hpp
+++ b/lib/fizzy/leb128.hpp
@@ -1,14 +1,16 @@
+#include "types.hpp"
 #include <cstdint>
 #include <stdexcept>
 
 namespace fizzy
 {
-uint64_t leb128u_decode(const uint8_t* input, size_t size)
+uint64_t leb128u_decode(bytes_view input)
 {
     uint64_t result = 0;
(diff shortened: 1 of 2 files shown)


gumb0:
@axic I think you meant `0x81 0x80 0x80 0x80 0x80 0x00` as it's little-endian.
Tests added.

lib/fizzy/leb128.hpp
```diff
@@ -2,8 +2,18 @@
 
 namespace fizzy
 {
-inline constexpr uint64_t leb128_decode(const uint8_t*) noexcept
+inline constexpr uint64_t leb128u_decode(const uint8_t* input) noexcept
```
chfast: I think we should remove `inline` and let compiler fully decide in this case.

lib/fizzy/leb128.hpp
```diff
@@ -2,8 +2,18 @@
...
 {
-    return 0;
+    uint64_t result = 0;
+    uint8_t result_shift = 0;
+
+    for (auto byte = input;; ++byte, result_shift += 7)
+    {
+        result |= static_cast<uint64_t>(*byte & 0x7F) << result_shift;
```
chfast: What if we run out of the input buffer (malformed encoding)? Should we care about it now?

chfast: The result value can overflow here. How to handle this?

lib/fizzy/leb128.hpp
```diff
@@ -2,8 +2,18 @@
 
 namespace fizzy
 {
-inline constexpr uint64_t leb128_decode(const uint8_t*) noexcept
+inline constexpr uint64_t leb128u_decode(const uint8_t* input) noexcept
 {
-    return 0;
+    uint64_t result = 0;
+    uint8_t result_shift = 0;
```
chfast: You should avoid using unsigned types smaller than native word size (`size_t`), especially in loops. The unsigned types guarantee wrapping behavior and compiler might need to emulate that with additional instructions. On the opposite, when using signed type the overflow is undefined behavior and compiler can assume the overflow never happens. Moreover, the true overflow can be detected with Undefined Behavior Sanitizer (very useful in fuzzing).

In such case I use `int` type indicating some signed integer value.

By changing type to `int`, we save one instruction in the loop: https://godbolt.org/z/xo7rZY.

lib/fizzy/leb128.hpp
gumb0: It looks like according to the spec it would be also malformed encoding 
https://webassembly.github.io/spec/core/binary/values.html#bytes

>  As an additional constraint, the total number of bytes encoding a value of type uN must not exceed ceil(N/7) bytes.

gumb0:
Rebased.

axic:
> I think you meant 0x81 0x80 0x80 0x80 0x80 0x00 as it's little-endian.
Tests added.

Sorry I meant  `0x80 0x80 0x80 0x80 0x80 0x01` should decode to 1.

test/unittests/leb128_test.cpp
```diff
@@ -1,14 +1,50 @@
...
+    EXPECT_EQ(fizzy::leb128u_decode(encoded_624485_leading_zeroes, 6), 624485);
+
+    uint8_t encoded_bigger_than_int32[] = {0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x00};
+    EXPECT_EQ(fizzy::leb128u_decode(encoded_bigger_than_int32, 8), 562949953421311);
+
+    uint8_t encoded_bigger_than_int32_leading_zeroes[] = {
+        0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x80, 0x80, 0x00};
+    EXPECT_EQ(fizzy::leb128u_decode(encoded_bigger_than_int32_leading_zeroes, 10), 562949953421311);
+
+    uint8_t encoded_max[] = {0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x01};
```
gumb0: This also works
```
    uint8_t encoded_max2[] = {0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x7f};
    EXPECT_EQ(fizzy::leb128u_decode(encoded_max2, 10), std::numeric_limits<uint64_t>::max());
```
1s or any bits in the last byte are ignored except for the least significant bit in that byte, because they don't fit into 64 bit.
Not completely sure it's correct.

test/unittests/leb128_test.cpp
```diff
@@ -1,14 +1,50 @@
...
+    EXPECT_EQ(fizzy::leb128u_decode(encoded_624485_leading_zeroes), 624485);
+
+    fizzy::bytes encoded_bigger_than_int32{0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x00};
+    EXPECT_EQ(fizzy::leb128u_decode(encoded_bigger_than_int32), 562949953421311);
+
+    fizzy::bytes encoded_bigger_than_int32_leading_zeroes{
+        0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x80, 0x80, 0x00};
+    EXPECT_EQ(fizzy::leb128u_decode(encoded_bigger_than_int32_leading_zeroes), 562949953421311);
+
+    fizzy::bytes encoded_max{0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x01};
```
gumb0: This also works
```
    fizzy::bytes  encoded_max2{0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x7f};
    EXPECT_EQ(fizzy::leb128u_decode(encoded_max2), std::numeric_limits<uint64_t>::max());
```

gumb0: added it to code

test/unittests/leb128_test.cpp
gumb0: Probably it's not, added TODO comment and test case for now

PR #4, opened 2019-12-20
cmake: Add Cable modules, configure build type and compilers

chfast:
## Cable

This adds 2 Cable modules. They were converted to standalone CMake modules so no other Cable files are needed. The modules are also versioned independently. I also added the `cable.cmake` tool that helps with installing / updating Cable modules. E.g. `./cable.cmake install CableBuildType` or `./cable.cmake update`.

This is new way of consuming Cable. All 3 files to be uploaded to the Cable repo.

## Coverage

Code coverage setup is done. ~~Just an admin has to enable codecov.~~

LCOV: https://79-224892340-gh.circle-artifacts.com/0/coverage/index.html
Summary: https://79-224892340-gh.circle-artifacts.com/0/coverage/report.txt
Full: https://79-224892340-gh.circle-artifacts.com/0/coverage/full.html
Missing: https://79-224892340-gh.circle-artifacts.com/0/coverage/missing.html

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -92,6 +92,51 @@ jobs:
       - build
       - test
 
+  clang-latest-coverage:
+    executor: linux-clang-9
+    environment:
+      BUILD_TYPE: Coverage
+      TESTS_FILTER: unittests
+    steps:
+      - run:
+          name: "Install lcov"
+          working_directory: /tmp
+          command: |
+            curl -sL http://downloads.sourceforge.net/ltp/lcov-1.14.tar.gz | tar xz --strip=1
+            sudo make install
+      - build
+      - test
+      - run:
+          name: "Collect coverage data"
+          working_directory: ~/build
+          command: |
+            mkdir coverage
+            find -name '*.profraw'
+            llvm-profdata merge *.profraw -o fizzy.profdata
+
+            llvm-cov report -instr-profile fizzy.profdata -Xdemangler llvm-cxxfilt \
+              -object bin/fizzy-unittests -use-color
+            llvm-cov report -instr-profile fizzy.profdata -Xdemangler llvm-cxxfilt > coverage/report.txt \
+              -object bin/fizzy-unittests
+
+            llvm-cov show -format=html -instr-profile fizzy.profdata -Xdemangler llvm-cxxfilt -region-coverage-lt=100 > coverage/missing.html \
+              -object bin/fizzy-unittests
+            llvm-cov show -format=html -instr-profile fizzy.profdata -Xdemangler llvm-cxxfilt > coverage/full.html \
+             -object bin/fizzy-unittests
+
+            llvm-cov export -instr-profile fizzy.profdata -format=lcov > fizzy.lcov \
+              -object bin/fizzy-unittests
+            genhtml fizzy.lcov -o coverage -t Fizzy
+      - store_artifacts:
+          path: ~/build/coverage
+          destination: coverage
+      - run:
+          name: "Upload to Codecov"
+          command: |
+            # Convert to relative paths
+            sed -i 's|$(pwd)/||' ~/build/fizzy.lcov
+            codecov --file ~/build/fizzy.lcov -X gcov
+
 
 workflows:
   version: 2
@@ -100,3 +145,4 @@ workflows:
       - lint
       - release-linux
       - release-macos
+      - clang-latest-coverage
--- a/cmake/CableCompilerSettings.cmake
+++ b/cmake/CableCompilerSettings.cmake
@@ -1,0 +1,166 @@
+# Cable: CMake Bootstrap Library <https://github.com/ethereum/cable>
+# Copyright 2018-2019 Pawel Bylica.
+# Licensed under the Apache License, Version 2.0.
+
+# Cable Compiler Settings, version 1.0.0
+#
+# This CMake module provides default configuration (with some options)
+# for C/C++ compilers. Use cable_configure_compiler().
+
+if(cable_compiler_settings_included)
+    return()
+endif()
+set(cable_compiler_settings_included TRUE)
+
+include(CheckCXXCompilerFlag)
+
+# Adds CXX compiler flag if the flag is supported by the compiler.
+#
+# This is effectively a combination of CMake's check_cxx_compiler_flag()
+# and add_compile_options():
+#
+#    if(check_cxx_compiler_flag(flag))
+#        add_compile_options(flag)
+#
+function(cable_add_cxx_compiler_flag_if_supported FLAG)
+    # Remove leading - or / from the flag name.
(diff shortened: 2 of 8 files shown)


gumb0:
Are codecov.io and LCOV independent tools? Do we need both of them?

cmake/cable.cmake
```diff
@@ -0,0 +1,68 @@
...
+    if(status EQUAL 0)
+        set(msg DONE)
+    else()
+        file(REMOVE "${dst}")
+        set(msg "${status_code} ${error_msg}\n  ${src}")
+    endif()
+    message("Downloading ${MODULE_NAME}: ${msg}")
+endfunction()
+
+set(cmd ${CMAKE_ARGV3})
```
gumb0: why is this `ARGV3`, shouldn't it be `ARGV1`?

chfast: The args look like:
```
cmake -P cable.cmake arg3...
```

chfast:
> ~Are codecov.io and LCOV independent tools? Do we need both of them?~
> Ah I see result of lcov is uploaded to codecov.io

The LCOV is a local tool that can combine GCOV (GCC) data and create e.g. html reports (also available as CI artifact here). It happens that codecov has a LCOV format processor, so exactly the same report should be shown by codecov.

But the original coverage data is coming from clang's source-code level coverage (think: C++ AST coverage) which is much more precise than GCC's debug-info level coverage (think: line bases coverage). This is later converted to LCOV format, where we loose some information (e.g. clang's coverage will report coverage for every template instance independently!). So additional clang's reports are also generated and available as CI artifacts.

cmake/cable.cmake
```diff
@@ -0,0 +1,68 @@
...
+# This CMake script allows installing or updating Cable modules.
+# Commands:
+# - list
+# - install
+# - update
+#
+# You can also include it from CMakeLists.txt to add Cable modules to
+# CMAKE_MODULE_PATH.
+
+if(NOT CMAKE_ARGV0)
```
chfast: Using `CMAKE_SCRIPT_MODE_FILE` here instead (as suggested in https://stackoverflow.com/questions/51427538/cmake-test-if-i-am-in-scripting-mode) might be more readable.

cmake/CableCompilerSettings.cmake
```diff
@@ -0,0 +1,168 @@
...
+            set(test_name have_fsanitize_${SANITIZE})
+            check_cxx_compiler_flag(-fsanitize=${SANITIZE} ${test_name})
+            if(NOT ${test_name})
+                message(FATAL_ERROR "Unsupported sanitizer: ${SANITIZE}")
+            endif()
+            add_compile_options(-fno-omit-frame-pointer -fsanitize=${SANITIZE})
+
+            set(backlist_file ${PROJECT_SOURCE_DIR}/sanitizer-blacklist.txt)
+            if(EXISTS ${backlist_file})
+                check_cxx_compiler_flag(-fsanitize-blacklist=${backlist_file} have_fsanitize-blacklist)
```
gumb0: seems these lines here could have been replaced with `cable_add_cxx_compiler_flag_if_supported`

circle.yml
```diff
@@ -92,6 +92,53 @@ jobs:
...
+      TESTS_FILTER: unittests
+    steps:
+      - run:
+          name: "Install lcov"
+          working_directory: /tmp
+          command: |
+            sudo apt-get -q update
+            # sudo apt-get install -qy lcov --no-install-recommends
+            apt-get download lcov
+            sudo dpkg -i --force-depends lcov*.deb
```
gumb0: why this unusual way to install?

circle.yml
```diff
@@ -92,6 +92,53 @@ jobs:
...
+      - test
+      - run:
+          name: "Collect coverage data"
+          working_directory: ~/build
+          command: |
+            mkdir coverage
+            find -name '*.profraw'
+            llvm-profdata merge *.profraw -o fizzy.profdata
+
+            llvm-cov report -instr-profile fizzy.profdata -Xdemangler llvm-cxxfilt \
```
gumb0: Output of this line doesn't seem to be saved anywhere?

circle.yml
chfast: Although being just a perl script, the lcov debian package depends on gcc package. Installing in a normal way will pull additional 300 MB. I plan to add lcov to the cpp-build-env image, probably downloading official release from http://ltp.sourceforge.net/coverage/lcov.php.

The trick used here is probably not good for docker images as apt-get will start reporting broken packages installed.

circle.yml
chfast: It is meant for CI log.

cmake/CableCompilerSettings.cmake
chfast: I will try that. We can also drop this option, because it comes from old Aleth config and is not used in any project currently.

cmake/cable.cmake
chfast: I will add a comment here.

cmake/CableCompilerSettings.cmake
```diff
@@ -0,0 +1,166 @@
...
+                # The compiler supports stack protection options.
+                if(cable_NO_STACK_PROTECTION)
+                    # Stack protection explicitly disabled.
+                    # Add "no" flag, because in some configuration the compiler has it enabled by default.
+                    add_compile_options(-fno-stack-protector)
+                else()
+                    # Try enabling the "strong" variant.
+                    cable_add_cxx_compiler_flag_if_supported(-fstack-protector-strong have_stack_protector_strong_support)
+                    if(NOT have_stack_protector_strong_support)
+                        # Fallback to standard variant of "strong" not available.
```
gumb0:
```suggestion
                        # Fallback to standard variant if "strong" not available.
```

cmake/CableCompilerSettings.cmake
```diff
@@ -0,0 +1,166 @@
...
+            set(CMAKE_EXE_LINKER_FLAGS "-fsanitize=${SANITIZE} ${CMAKE_EXE_LINKER_FLAGS}")
+
+            set(test_name have_fsanitize_${SANITIZE})
+            check_cxx_compiler_flag(-fsanitize=${SANITIZE} ${test_name})
+            if(NOT ${test_name})
+                message(FATAL_ERROR "Unsupported sanitizer: ${SANITIZE}")
+            endif()
+            add_compile_options(-fno-omit-frame-pointer -fsanitize=${SANITIZE})
+
+            set(backlist_file ${PROJECT_SOURCE_DIR}/sanitizer-blacklist.txt)
```
gumb0: typo
```suggestion
            set(blacklist_file ${PROJECT_SOURCE_DIR}/sanitizer-blacklist.txt)
```

PR #7, opened 2020-01-06
Add basic Wasm types

axic:
Names are based on the Binary Encoding specification


lib/fizzy/types.hpp
```diff
@@ -0,0 +1,61 @@
...
+{
+    local_get = 0x20,
+    local_set = 0x21,
+    local_tee = 0x22,
+    i32_add = 0x6a,
+};
+
+struct code
+{
+    uint32_t localcount;
```
```suggestion
    uint32_t local_count;
```

PR #8, opened 2020-01-06
Add CAVEATS file

axic:


chfast:
Please see the _Syntactic limits_ in the spec: https://webassembly.github.io/spec/core/appendix/implementation.html#syntactic-limits.

axic:
I propose we also enforce some of the limits (such as the memory index in the data section) in the parser, because we are not creating a generic tool for manipulating binaries.

CAVEATS.md
```diff
@@ -0,0 +1,8 @@
+# Caveats
```
chfast, line 1: I'm not sure it is worth putting this in separate file.

axic: Yeah I'm merging the README part into #278. The caveats file/contents is too early to be merged.

axic:
This PR is not useful, moved to #282.

PR #10, opened 2020-01-06
wasm binary parser

chfast:
wasm binary parser.

### DIFF
--- a/lib/fizzy/parser.cpp
+++ b/lib/fizzy/parser.cpp
@@ -32,17 +32,22 @@ module parse(bytes_view input)
     for (auto it = input.begin(); it != input.end();)
     {
         const auto id = static_cast<sectionid>(*it++);
-        const auto [size, new_pos] = leb128u_decode<uint32_t>(it);
-        it = new_pos;
+        uint32_t size;
+        std::tie(size, it) = leb128u_decode<uint32_t>(it);
+        const auto expected_end_pos = it + size;
         switch (id)
         {
         case sectionid::type:
-            std::tie(mod.typesec, std::ignore) = parser<std::vector<functype>>{}(it);
+            std::tie(mod.typesec, it) = parser<std::vector<functype>>{}(it);
             break;
         default:
+            it += size;
             break;
         }
-        it += size;
+
+        if (it != expected_end_pos)
+            throw parser_error{"incorrect section " + std::to_string(static_cast<int>(id)) +
+                               " size, difference: " + std::to_string(it - expected_end_pos)};
     }
 
     return mod;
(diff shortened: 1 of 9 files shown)


lib/fizzy/parser.cpp
```diff
@@ -1,10 +1,29 @@
...
+    {
+        const auto id = static_cast<sectionid>(*it++);
+        const auto [size, new_pos] = leb128u_decode<uint32_t>(it);
+        it = new_pos;
+        switch (id)
+        {
+        default:
+            break;
+        }
+        it += size;
```
axic: I think this should move into the `default` case.

chfast: This is for all sections, also the "implemented" ones.

lib/fizzy/parser.hpp
```diff
@@ -0,0 +1,72 @@
+#pragma once
+
+#include "types.hpp"
+#include <stdexcept>
+#include <tuple>
+
+namespace fizzy
+{
+constexpr uint8_t wasm_prefix_data[] = {0x00, 0x61, 0x73, 0xd6, 0x01, 0x00, 0x00, 0x00};
```
gumb0: Does this need to be in a header?

chfast: Used in tests.

lib/fizzy/parser.hpp
```diff
@@ -0,0 +1,72 @@
...
+        }
+    }
+};
+
+template <>
+struct parser<uint32_t>
+{
+    parser_result<uint32_t> operator()(const uint8_t* pos)
+    {
+        const auto b = *pos++;
```
gumb0: should use `leb128u_decode<uint32_t>`

lib/fizzy/parser.cpp
gumb0: For parsed section we should check that parser consumed the expected number of bytes

lib/fizzy/parser.hpp
chfast: Use and simplified.

lib/fizzy/parser.cpp
chfast: Can it consume less?

chfast: I will add the more restrictive check.

lib/fizzy/parser.cpp
```diff
@@ -0,0 +1,55 @@
...
+        case sectionid::type:
+            std::tie(mod.typesec, it) = parser<std::vector<functype>>{}(it);
+            break;
+        default:
+            it += size;
+            break;
+        }
+
+        if (it != expected_end_pos)
+            throw parser_error{"incorrected section " + std::to_string(static_cast<int>(id)) +
```
gumb0:
```suggestion
            throw parser_error{"incorrect section " + std::to_string(static_cast<int>(id)) +
```

PR #11, opened 2020-01-06
Add basic interpreter framework

axic:


lib/fizzy/execute.cpp
```diff
@@ -0,0 +1,60 @@
...
+    {
+        auto instruction = *pc++;
+        std::cout << "Op: " << static_cast<int>(instruction) << std::endl;
+        switch (instruction)
+        {
+        case instr::end:
+            goto end;
+        case instr::local_get: {
+            auto idx = load<uint32_t>(immediates);
+            immediates += 4;
```
axic: Could also change `load` to do the increment.

test/unittests/execute_test.cpp
```diff
@@ -0,0 +1,49 @@
...
+    auto ret = fizzy::execute(module, 0, {});
+
+    ASSERT_EQ(ret.size(), 1);
+    EXPECT_EQ(ret[0], 0);
+}
+
+TEST(execute, basic_add_with_inputs)
+{
+    fizzy::module module;
+    fizzy::bytes imm;
```
chfast:
```suggestion
    fizzy::bytes imm{0, 0, 0, 0, 1, 0, 0, 0};
```

lib/fizzy/execute.cpp
```diff
@@ -0,0 +1,60 @@
...
```
```suggestion
            immediates += sizeof(uint32_t);
```

lib/fizzy/execute.hpp
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+#include "types.hpp"
+#include <cstdint>
+
+namespace fizzy
+{
+std::vector<uint64_t> execute(
+    struct module& _module, funcidx _function, std::vector<uint64_t> _args);
```
```suggestion
    const module& _module, funcidx _function, std::vector<uint64_t> _args);
```

PR #12, opened 2020-01-07
Width-generic LEB128 decoder returning position in the stream after decode

gumb0:


lib/fizzy/leb128.hpp
```diff
@@ -4,22 +4,27 @@
...
+        result |= static_cast<T>(*input & 0x7F) << result_shift;
+        if ((*input & 0x80) == 0)
+            return {result, input + 1};
     }
 
-    throw std::runtime_error("Invalid unsigned integer encoding.");
+    throw std::runtime_error("Invalid LEB128 encoding: too many bytes.");
 }
+
+inline std::pair<uint64_t, const uint8_t*> leb128_decode_u64(const uint8_t* input)
```
chfast: I don't think this is so much useful. I would rather user the template directly.

test/unittests/leb128_test.cpp
```diff
@@ -1,54 +1,46 @@
...
-    fizzy::bytes encoded_max{0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x01};
-    EXPECT_EQ(fizzy::leb128u_decode(encoded_max), std::numeric_limits<uint64_t>::max());
-
-    // this is not really correct encoding, but is not rejected by our decoder
-    fizzy::bytes encoded_max2{0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x7f};
-    EXPECT_EQ(fizzy::leb128u_decode(encoded_max2), std::numeric_limits<uint64_t>::max());
+    // clang-format off
+    std::vector<std::pair<fizzy::bytes, uint64_t>> testcases = {
+        { {0}, 0 }, // 0
+        { {0x80, 0x80, 0x00}, 0 }, // 0 with leading zeroes
```
chfast: The spacing between `} }` is inconsistent.

PR #15, opened 2020-01-07
Implement support for memory (and grow/size/load/store)

axic:


lib/fizzy/execute.cpp
```diff
@@ -53,6 +53,20 @@ std::vector<uint64_t> execute(const module& _module, funcidx _function, std::vec
...
             goto end;
+        case instr::drop: {
+            stack.pop();
+            break;
+        }
+        case instr::select: {
+            const auto a = static_cast<uint32_t>(stack.pop());
+            const auto b = static_cast<uint32_t>(stack.pop());
+            const auto condition = static_cast<uint32_t>(stack.pop());
+            if (condition == 0)
```
axic: Need to check which one is it. The spec is unclear.

axic: Fixed.

lib/fizzy/execute.cpp
```diff
@@ -147,6 +168,59 @@ execution_result execute(Instance& instance, FuncIdx function, std::vector<uint6
...
             break;
         }
+        case Instr::memory_size:
+        {
+            stack.push(static_cast<uint32_t>(instance.memory.size() / pagesize));
+            break;
+        }
+        case Instr::memory_grow:
+        {
+            const auto a = static_cast<uint32_t>(stack.pop());
```
axic: Renaming `a` to `delta`.

lib/fizzy/execute.cpp
```diff
@@ -79,7 +96,11 @@ inline T rotr(T lhs, T rhs) noexcept
 
 Instance instantiate(const Module& module)
 {
-    return {module};
+    // FIXME: set pages from proper section
+    constexpr unsigned pages = 1;
+    // NOTE: fill it with zeroes
+    bytes memory(pages * pagesize, 0);
+    return {module, memory};
```
```suggestion
    return {module, std::move(memory)};
```

lib/fizzy/parser_expr.cpp
```diff
@@ -61,6 +63,20 @@ parser_result<Code> parse_expr(const uint8_t* pos)
             push(code.immediates, imm);
             break;
         }
+
+        case Instr::i32_load:
+        case Instr::i32_store:
+        {
+            // alignment
+            uint32_t imm;
+            std::tie(imm, pos) = leb128u_decode<uint32_t>(pos);
```
chfast:
```suggestion
            std::tie(std::ignore, pos) = leb128u_decode<uint32_t>(pos);
```

axic:
Also adding parsing of memory section.

lib/fizzy/execute.cpp
```diff
@@ -6,13 +6,10 @@
...
 namespace fizzy
 {
-struct Instance
-{
-    const Module& module;
-};
-
 namespace
 {
+constexpr unsigned pagesize = 65536;
```
gumb0: Maybe camel case for consts?

lib/fizzy/execute.cpp
```diff
@@ -129,7 +140,11 @@ inline uint64_t popcnt64(uint64_t value) noexcept
 
 Instance instantiate(const Module& module)
 {
-    return {module};
+    // FIXME: set pages from proper section
+    constexpr unsigned pages = 1;
+    // NOTE: fill it with zeroes
+    bytes memory(pages * pagesize, 0);
```
gumb0: Could be
```suggestion
    bytes memory(pages * pagesize);
```

lib/fizzy/execute.cpp
```diff
@@ -129,7 +140,11 @@ inline uint64_t popcnt64(uint64_t value) noexcept
...
+    return {module, std::move(memory)};
```
gumb0: I think better without `move` or even just construct inline
```suggestion
    return {module, bytes(pages * pagesize)};
```

lib/fizzy/execute.cpp
axic: I don't mind. @chfast ?

lib/fizzy/execute.cpp
axic: Apparently the constructor requires the explicit filling character.

gumb0: not for `bytes`, sorry

lib/fizzy/execute.cpp
axic: I'd prefer to keep the explicit `memory` name for now.

lib/fizzy/execute.cpp
```diff
@@ -205,6 +220,45 @@ execution_result execute(Instance& instance, FuncIdx function, std::vector<uint6
...
+            stack.push(static_cast<uint32_t>(instance.memory.size() / pagesize));
+            break;
+        }
+        case Instr::memory_grow:
+        {
+            const auto delta = static_cast<uint32_t>(stack.pop());
+            const auto cur_pages = instance.memory.size() / pagesize;
+            assert(cur_pages <= size_t(std::numeric_limits<int32_t>::max()));
+            const auto new_pages = cur_pages + delta;
+            assert(cur_pages >= new_pages);
```
gumb0: condition looks reversed

lib/fizzy/execute.cpp
chfast: I don't care.

lib/fizzy/execute.cpp
```diff
@@ -205,6 +222,85 @@ execution_result execute(Instance& instance, FuncIdx function, std::vector<uint6
...
+            store<uint32_t>(instance.memory, address + offset, value);
+            break;
+        }
+        case Instr::i64_store:
+        {
+            const auto address = static_cast<uint32_t>(stack.pop());
+            // NOTE: alignment is dropped by the parser
+            const auto offset = read<uint32_t>(immediates);
+            const auto value = static_cast<uint64_t>(stack.pop());
+            if ((address + offset + sizeof(uint32_t)) > instance.memory.size())
```
```suggestion
            if ((address + offset + sizeof(uint64_t)) > instance.memory.size())
```

lib/fizzy/execute.cpp
```diff
@@ -205,6 +222,85 @@ execution_result execute(Instance& instance, FuncIdx function, std::vector<uint6
...
+            try
+            {
+                // TODO: remove this once section parsing is done (and enforce limit in parser)
+                if ((new_pages * pagesize) > memorylimit)
+                    throw std::bad_alloc();
+                instance.memory.resize(new_pages * pagesize);
+            }
+            catch (std::bad_alloc const&)
+            {
+                ret = 0xffffffff;  // -1
```
```suggestion
                ret = static_cast<uint32_t>(-1);
```

PR #16, opened 2020-01-07
Implement i32 numerical and bit shifting instructions

gumb0:


lib/fizzy/execute.cpp
```diff
@@ -74,11 +111,66 @@ std::vector<uint64_t> execute(const module& _module, funcidx _function, std::vec
...
+        case instr::i32_sub: {
+            binary_op(stack, std::minus<uint32_t>());
+            break;
+        }
+        case instr::i32_mul: {
+            binary_op(stack, std::multiplies<uint32_t>());
             break;
         }
+        case instr::i32_div_s: {
+            binary_op(stack, std::divides<int32_t>());
```
chfast: Should trap for some cases.

lib/fizzy/execute.cpp
```diff
@@ -33,6 +32,44 @@ inline T read(const uint8_t*& input) noexcept
     return ret;
 }
 
+template <typename Op>
+void binary_op(uint64_stack& stack, Op op)
```
gumb0: will make inline

test/unittests/execute_test.cpp
```diff
@@ -81,3 +81,212 @@ TEST(execute, local_tee)
...
+        {0, 0, 0, 0, 1, 0, 0, 0}});
+
+    const auto [trap, ret] = fizzy::execute(module, 0, {4200, 4242});
+
+    ASSERT_EQ(trap, false);
+    ASSERT_EQ(ret.size(), 1);
+    EXPECT_EQ(ret[0], 42);
+}
+
+TEST(execute, i32_and)
```
axic: Do you want to remove `basic_add` if we have this here?

test/unittests/execute_test.cpp
```diff
@@ -52,32 +65,174 @@ TEST(execute, basic_add_with_inputs)
...
+    ASSERT_EQ(trap, false);
+    ASSERT_EQ(ret.size(), 1);
+    EXPECT_EQ(ret[0], 42);
+}
+
+TEST(execute, i32_rem_u_by_zero)
+{
+    const auto [trap, ret] = execute_binary_operation(fizzy::instr::i32_rem_u, 4242, 0);
+
+    ASSERT_EQ(trap, true);
```
chfast:
```suggestion
    ASSERT_TRUE(trap);
```

PR #18, opened 2020-01-07
Support traps in execute

axic:


lib/fizzy/execute.hpp
```diff
@@ -5,6 +5,6 @@
 
 namespace fizzy
 {
-std::vector<uint64_t> execute(
+std::tuple<bool, std::vector<uint64_t>> execute(
```
gumb0: `#include <tuple>`

lib/fizzy/execute.hpp
```diff
@@ -5,6 +5,14 @@
 
 namespace fizzy
 {
-std::vector<uint64_t> execute(
-    const module& _module, funcidx _function, std::vector<uint64_t> _args);
+struct execution_result
+{
+    // true is execution resulted in a trap
```
```suggestion
    // true if execution resulted in a trap
```

PR #19, opened 2020-01-07
Integer conversion instructions

gumb0:


lib/fizzy/execute.cpp
```diff
@@ -79,6 +79,21 @@ std::vector<uint64_t> execute(const module& _module, funcidx _function, std::vec
             stack.push(a + b);
             break;
         }
+        case instr::i32_wrap_i64: {
+            const auto a = static_cast<uint64_t>(stack.pop());
```
axic: `stack.pop` return `uint64_t` so no need to cast?

lib/fizzy/execute.cpp
```diff
@@ -79,6 +79,21 @@ std::vector<uint64_t> execute(const module& _module, funcidx _function, std::vec
...
+            stack.push(a & 0x00000000ffffffff);
```
axic: Cannot we just use `static_cast<uint32_t>`  to do this?

```suggestion
            stack.push(static_cast<uint32_t>(stack.pop()));
```

that's my preference.

lib/fizzy/execute.cpp
```diff
@@ -79,6 +79,21 @@ std::vector<uint64_t> execute(const module& _module, funcidx _function, std::vec
...
+            break;
+        }
+        case instr::i64_extend_i32_s: {
+            const auto a = static_cast<int64_t>(stack.pop());
```
chfast: This is wrong. You need 3 casts: `int32_t` -> `int64_t` -> `uint64_t`. I.e. you want `0x00000000ffffffff` become `0xffffffffffffffff`.

lib/fizzy/execute.cpp
```diff
@@ -79,6 +79,21 @@ std::vector<uint64_t> execute(const module& _module, funcidx _function, std::vec
...
+            stack.push(static_cast<uint64_t>(a));
+            break;
+        }
+        case instr::i64_extend_i32_u: {
```
chfast: This should be effectively no-op, because we already zero-extend `i32` values when storing on stack. 
We can even drop this instruction in parser.

axic: We still need to meter it (at some point).

lib/fizzy/execute.cpp
```diff
@@ -79,6 +79,19 @@ std::vector<uint64_t> execute(const module& _module, funcidx _function, std::vec
             stack.push(a + b);
             break;
         }
+        case instr::i32_wrap_i64: {
+            stack.push(static_cast<uint32_t>(stack.pop()));
+            break;
+        }
+        case instr::i64_extend_i32_s: {
+            const auto a = static_cast<int32_t>(stack.pop());
+            stack.push(static_cast<uint64_t>(static_cast<int64_t>(a)));
```
```suggestion
            stack.push(static_cast<uint64_t>(int64_t{a}));
```

The `{}` guarantees there is not information loss at conversion.

PR #23, opened 2020-01-07
Implement i32.const and i64.const

axic:
Depends on #29.

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -153,6 +153,12 @@ execution_result execute(Instance& instance, FuncIdx function, std::vector<uint6
             stack.push(value);
             break;
         }
+        case Instr::i64_const:
+        {
+            const auto value = read<uint64_t>(immediates);
+            stack.push(value);
+            break;
+        }
         case Instr::i32_eqz:
         {
             const auto value = static_cast<uint32_t>(stack.pop());
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -147,6 +147,12 @@ execution_result execute(Instance& instance, FuncIdx function, std::vector<uint6
             locals[idx] = stack.back();
             break;
         }
+        case Instr::i32_const:
+        {
+            const auto value = read<uint32_t>(immediates);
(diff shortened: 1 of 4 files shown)


axic:
@gumb0 @chfast ready for review

lib/fizzy/execute.cpp
```diff
@@ -147,6 +147,18 @@ execution_result execute(Instance& instance, FuncIdx function, std::vector<uint6
             locals[idx] = stack.back();
             break;
         }
+        case Instr::i32_const:
+        {
+            const auto value = read<int32_t>(immediates);
```
chfast: Read directly int `uint32_t` type.

lib/fizzy/execute.cpp
```diff
@@ -147,6 +147,18 @@ execution_result execute(Instance& instance, FuncIdx function, std::vector<uint6
...
+            stack.push(static_cast<uint32_t>(value));
+            break;
+        }
+        case Instr::i64_const:
+        {
+            const auto value = read<int64_t>(immediates);
```
chfast: Read directly into `uint64_t` type.

lib/fizzy/parser_expr.cpp
```diff
@@ -76,6 +76,22 @@ parser_result<Code> parse_expr(const uint8_t* pos)
             push(code.immediates, imm);
             break;
         }
+
+        case Instr::i32_const:
+        {
+            int32_t imm;
+            std::tie(imm, pos) = leb128s_decode<int32_t>(pos);
```
chfast: Convert to `uint32_t` here.

lib/fizzy/parser_expr.cpp
```diff
@@ -76,6 +76,22 @@ parser_result<Code> parse_expr(const uint8_t* pos)
...
+            push(code.immediates, imm);
+            break;
+        }
+
+        case Instr::i64_const:
+        {
+            int64_t imm;
+            std::tie(imm, pos) = leb128s_decode<int64_t>(pos);
```
chfast: covert to `uint64_t` here.

PR #24, opened 2020-01-07
Implement drop and select

axic:


test/unittests/execute_test.cpp
```diff
@@ -48,6 +48,45 @@ TEST(execute, nop)
...
 
+TEST(execute, drop)
+{
+    fizzy::module module;
+    module.codesec.emplace_back(fizzy::code{
+        1, {fizzy::instr::local_get, fizzy::instr::drop, fizzy::instr::end}, {0, 0, 0, 0}});
+
+    const auto [trap, ret] = fizzy::execute(module, 0, {});
+
+    ASSERT_EQ(trap, false);
```
chfast: You can use `ASSERT_FALSE()`.

PR #25, opened 2020-01-07
Implement labels

axic:
Requires #48

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -12,6 +12,15 @@ constexpr unsigned page_size = 65536;
 constexpr auto memory_pages_limit = (256 * 1024 * 1024ULL) / page_size;
 
 
+struct LabelContext
+{
+    const Instr* pc = nullptr;
+    const uint8_t* immediate = nullptr;
+    size_t arity = 0;
+    size_t stack_height = 0;
+};
+
+
 template <typename T>
 inline void store(bytes& input, size_t offset, T value) noexcept
 {
@@ -198,6 +207,8 @@ execution_result execute(Instance& instance, FuncIdx function, std::vector<uint6
     // TODO: preallocate fixed stack depth properly
     Stack<uint64_t> stack;
 
+    Stack<LabelContext> labels;
+
     bool trap = false;
 
     const Instr* pc = code.instructions.data();
@@ -213,8 +224,57 @@ execution_result execute(Instance& instance, FuncIdx function, std::vector<uint6
             goto end;
         case Instr::nop:
             break;
+        case Instr::block:
+        {
+            const auto arity = read<uint8_t>(immediates);
+
+            if (arity != 0)
+            {
+                throw std::runtime_error(
+                    "block instruction with non-empty result type not implemented yet");
+            }
+
+            const auto target_pc = read<uint32_t>(immediates);
+            const auto target_imm = read<uint32_t>(immediates);
+            LabelContext label{code.instructions.data() + target_pc,
+                code.immediates.data() + target_imm, arity, stack.size()};
+            labels.emplace_back(label);
+            break;
+        }
+        case Instr::loop:
+        {
+            LabelContext label{pc - 1, immediates, 0, stack.size()};  // Target this instruction.
+            labels.push_back(label);
+            break;
+        }
         case Instr::end:
-            goto end;
+        {
+            if (!labels.empty())
+                labels.pop_back();
+            else
+                goto end;
+            break;
+        }
+        case Instr::br:
+        case Instr::br_if:
+        {
+            const auto label_idx = read<uint32_t>(immediates);
+
+            // Check condition for br_if.
+            if (instruction == Instr::br_if && static_cast<uint32_t>(stack.pop()) == 0)
+                break;
+
+            assert(labels.size() > label_idx);
+            labels.drop(label_idx);  // Drop skipped labels (does nothing for labelidx == 0).
+            const auto label = labels.pop();
+
+            pc = label.pc;
+            immediates = label.immediate;
+
+            // FIXME: handle arity + stack_height ?
+            assert(stack.size() == label.stack_height);
+            break;
+        }
         case Instr::call:
         {
             const auto func_idx = read<uint32_t>(immediates);
@@ -760,6 +820,7 @@ execution_result execute(Instance& instance, FuncIdx function, std::vector<uint6
     }
 
 end:
+    assert(labels.empty());
     // move allows to return derived Stack<uint64_t> instance into base vector<uint64_t> value
     return {trap, std::move(stack)};
 }
--- a/test/unittests/execute_control_test.cpp
+++ b/test/unittests/execute_control_test.cpp
@@ -1,4 +1,6 @@
 #include "execute.hpp"
+#include "parser.hpp"
+#include "utils.hpp"
 #include <gtest/gtest.h>
 
 using namespace fizzy;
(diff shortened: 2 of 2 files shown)


lib/fizzy/parser_expr.cpp
```diff
@@ -10,11 +10,26 @@ inline void push(bytes& b, T value)
     b.resize(b.size() + sizeof(value));
     __builtin_memcpy(&b[b.size() - sizeof(value)], &value, sizeof(value));
 }
+
+struct open_block_possition
```
axic: possition -> position

lib/fizzy/execute.cpp
```diff
@@ -165,8 +175,49 @@ execution_result execute(Instance& instance, FuncIdx function, std::vector<uint6
...
         case Instr::nop:
             break;
+        case Instr::block:
+        {
+            const auto imm_begin = immediates;
+            const auto arity = read<uint8_t>(immediates);
+            const auto instr_diff = read<uint32_t>(immediates);
+            const auto imm_diff = read<uint32_t>(immediates);
+            label_context label{pc + instr_diff, imm_begin + imm_diff, arity, stack.size()};
+            labels.push_back(label);
```
```suggestion
            labels.emplace_back(pc + instr_diff, imm_begin + imm_diff, arity, stack.size());
```

lib/fizzy/parser_expr.cpp
```diff
@@ -92,9 +106,84 @@ parser_result<Code> parse_expr(const uint8_t* pos)
         case Instr::i64_extend_i32_u:
             break;
 
+        case Instr::end:
+        {
+            if (!open_blocks_possitions.empty())
```
axic: Need to also add an assertion at the end of `parse_expr` for `positions.empty()` (so that all of them were processed, e.g. it is balanced).

lib/fizzy/execute.cpp
chfast: I can change it to `emplace_back(label)` if that make a difference. But for debugging purposes it's useful to be able to inspect the create object before it is pushed.

gumb0: Well it should be `emplace_back(std::move(label))` then

lib/fizzy/execute.cpp
```diff
@@ -146,6 +146,14 @@ inline uint64_t popcnt64(uint64_t value) noexcept
 {
     return static_cast<uint64_t>(__builtin_popcountll(value));
 }
+
+struct label_context
```
gumb0: Maybe keep types at the top of the unnamed namespace (not to mix types with functions)

lib/fizzy/execute.cpp
```diff
@@ -227,8 +237,51 @@ execution_result execute(Instance& instance, FuncIdx function, std::vector<uint6
             goto end;
         case Instr::nop:
             break;
+        case Instr::block:
+        {
+            const auto arity = read<uint8_t>(immediates);
+            assert(arity == 0);
```
gumb0: I think it would be better to throw "not implemented"

lib/fizzy/execute.cpp
chfast: Maybe it looks more modern, but I hope you realize it makes no difference in this case.

gumb0: Ok true, it won't make any difference, we can leave it

chfast:
Thanks for reviews. I believe we should put this PR on hold until #48 is merged, because there are changes introduced there.

lib/fizzy/execute.cpp
```diff
@@ -213,8 +224,56 @@ execution_result execute(Instance& instance, FuncIdx function, std::vector<uint6
...
+            if (!labels.empty())
+                labels.pop_back();
+            else
+                goto end;
+            break;
+        }
+        case Instr::br:
+        case Instr::br_if:
+        {
+            const auto labelidx = read<uint32_t>(immediates);
```
axic: In some parts we have used underscore before `idx` (e.g. `label_idx`), but I don't care which one.

axic:
@chfast Approved by me. (I can't approve since I've created the PR)

lib/fizzy/execute.cpp
```diff
@@ -213,8 +224,56 @@ execution_result execute(Instance& instance, FuncIdx function, std::vector<uint6
...
+        case Instr::br:
+        case Instr::br_if:
+        {
+            const auto label_idx = read<uint32_t>(immediates);
+
+            // Check condition for br_if.
+            if (instruction == Instr::br_if && static_cast<uint32_t>(stack.pop()) == 0)
+                break;
+
+            labels.drop(label_idx);  // Drop skipped labels (does nothing for labelidx == 0).
```
gumb0: Maybe `assert(labels.size() >= label_idx + 1)` ?

gumb0:
Looks good, but would be good to add unit tests for
- non-zero immediate of `br`/`br_if`
- empty `block` and/or `block` via `end` (now in tests there're only blocks exiting with `br`)

gumb0:
Another thing is that I noticed that the spec describing function invocation says that `block` is implicitly created https://webassembly.github.io/spec/core/exec/instructions.html#exec-invoke
I think this means that the function with the code like `br 0 end` should work, but here it would crash

gumb0:
> Another thing is that I noticed that the spec describing function invocation says that block is implicitly created https://webassembly.github.io/spec/core/exec/instructions.html#exec-invoke
I think this means that the function with the code like br 0 end should work, but here it would crash

And this applies not only to `call`s but for external invocation, too, because it's spec refer to the same invocation procedure https://webassembly.github.io/spec/core/exec/modules.html#invocation

axic:
Probably that's correct, because then `return` also jumps to the deepest label.

chfast:
That should be implemented together with `return`.

lib/fizzy/execute.cpp
```diff
@@ -213,8 +224,57 @@ execution_result execute(Instance& instance, FuncIdx function, std::vector<uint6
...
+                goto end;
+            break;
+        }
+        case Instr::br:
+        case Instr::br_if:
+        {
+            const auto label_idx = read<uint32_t>(immediates);
+
+            // Check condition for br_if.
+            if (instruction == Instr::br_if && static_cast<uint32_t>(stack.pop()) == 0)
```
gumb0: Why this `static_cast`? And what if `br_if` checks result of `i64.add`, will it ignore high bytes?

axic: Screenshot 2020-01-16 at 15 21 09

gumb0: Maybe a comment in the code would be useful.

axic:
> empty block and/or block via end (now in tests there're only blocks exiting with br)

These cases should still be added here.

lib/fizzy/execute.cpp
chfast: This is complicated and affects many other instructions implementations.

There are two options when putting `i32` values in our implementation of the stack with `uint64_t` types:

1. Zero-extend to `uint64_t`. This is 64-bit memory store instruction.
2. Only store 32-bits, leaving garbage in the high word on the stack.

We are currently doing 1, mostly because this is much easier to do in C/C++ than 2.

So here the `static_cast` is not needed because the high bits are going to be zeros anyway. The `static_cast` would be required if we implement option 2.

Nevertheless, using `static_cast` emits an assembly instruction for 32-bit load instead of 64-bit load. Probably not any faster one (no way 64-bit one can cross cache line here). But you saved one byte in the final code size.

chfast:
> > empty block and/or block via end (now in tests there're only blocks exiting with br)
> 
> These cases should still be added here.

This test has block ended by `end`: https://github.com/wasmx/fizzy/pull/25/files#diff-6c6c1bda2d45defc0c70b79a10a0f90cR104-R131

Not sure what you mean by empty blocks? `(block)`? Then I don't think this is good test because there is no way you can tell if it was executed or not.

gumb0:
> Not sure what you mean by empty blocks? (block)? Then I don't think this is good test because there is no way you can tell if it was executed or not.

I meant something similar to `loop_void_empty` test

chfast:
> > 
> 
> I meant something similar to `loop_void_empty` test

I added  a test for `block` analogous to `loop_void_empty` although both of them are not valid wasm code.

axic:
> Not sure what you mean by empty blocks? (block)? Then I don't think this is good test because there is no way you can tell if it was executed or not.

That's fine, but at least we can it didn't crash anything.

PR #26, opened 2020-01-08
Introduce instance and instantiate (and use it in execute)

axic:


lib/fizzy/execute.cpp
```diff
@@ -6,6 +6,11 @@
 
 namespace fizzy
 {
+struct instance
```
chfast: I think we should name types with CamelCase, because they conflict with natural var names. I'm ok with doing that already in this case: `Instance`.

axic: Why don't you make a PR which changes every existing one and we merge that first?

chfast: This will probably take much longer.

PR #29, opened 2020-01-08
Implement LEB128 signed decoder

gumb0:


lib/fizzy/leb128.hpp
```diff
@@ -24,4 +26,28 @@ std::pair<T, const uint8_t*> leb128u_decode(const uint8_t* input)
...
+    using T_unsigned = typename std::make_unsigned<T>::type;
+    T_unsigned result = 0;
+    int result_shift = 0;
+
+    for (; result_shift < std::numeric_limits<T_unsigned>::digits; ++input, result_shift += 7)
+    {
+        result |= static_cast<T_unsigned>((static_cast<T_unsigned>(*input) & 0x7F) << result_shift);
+        if ((*input & 0x80) == 0)
+        {
+            // sign extend
```
chfast: Sign extension has to be done only once in the end. See https://en.wikipedia.org/wiki/LEB128#Decode_signed_integer.

gumb0: It is done only on the last iteration

PR #32, opened 2020-01-08
Implement i32/i64.clz/ctz/popcnt

axic:


lib/fizzy/execute.cpp
```diff
@@ -75,6 +83,48 @@ inline T rotr(T lhs, T rhs) noexcept
     auto const k = rhs % std::numeric_limits<T>::digits;
     return (lhs >> k) | (lhs << (std::numeric_limits<T>::digits - k));
 }
+
+inline uint32_t clz32(uint32_t value) noexcept
```
axic: Note: Wasm spec says that these operations return the same value type as the input. Hence the cast.

PR #35, opened 2020-01-08
test: Add milestone2 example code

axic:


axic:
> We will have more of such tests for some time, so it might be good to move them to a separate file.

Good idea, perhaps call them "integration" or "end-to-end" tests?

chfast:
> > We will have more of such tests for some time, so it might be good to move them to a separate file.
> 
> Good idea, perhaps call them "integration" or "end-to-end" tests?

Can you check now?

PR #37, opened 2020-01-09
Implement globals

gumb0:


lib/fizzy/execute.cpp
```diff
@@ -154,11 +154,18 @@ Instance instantiate(const Module& module)
...
     // NOTE: fill it with zeroes
     bytes memory(pages * pagesize, 0);
-    return {module, std::move(memory)};
+    // TODO: add imported globals first
+    std::vector<uint64_t> globals;
+    globals.reserve(module.globalsec.size());
+    for (auto const& global : module.globalsec)
+        globals.emplace_back(global.init_value);
+
+    return {module, std::move(memory), globals};
```
axic: Move globals?

lib/fizzy/execute.cpp
```diff
@@ -154,11 +154,25 @@ Instance instantiate(const Module& module)
...
+    // TODO: add imported globals first
+    std::vector<uint64_t> globals;
+    globals.reserve(module.globalsec.size());
+    for (auto const& global : module.globalsec)
+    {
+        if (global.init_type == GlobalInitType ::constant)
+            globals.emplace_back(global.init.value);
+        else
+        {
+            // TODO: initialize by imported global
```
axic: Can you throw a runtime exception here?

lib/fizzy/execute.cpp
```diff
@@ -172,11 +172,27 @@ Instance instantiate(const Module& module)
     }
     // NOTE: fill it with zeroes
     bytes memory(memory_min * page_size, 0);
-    return {module, std::move(memory), memory_max};
+
+    // TODO: add imported globals first
+    std::vector<uint64_t> globals;
+    globals.reserve(module.globalsec.size());
```
axic: Can't you just use `std::vector<uint64_t> globals(module.globalsec.size())` ?

test/unittests/execute_test.cpp
```diff
@@ -117,6 +117,42 @@ TEST(execute, local_tee)
     EXPECT_EQ(ret[0], 42);
 }
 
+TEST(execute, global_get)
+{
+    fizzy::Module module;
+    fizzy::Global g{true, fizzy::GlobalInitType::constant, {42}};
+    module.globalsec.emplace_back(g);
```
axic: Can you squash these two?

lib/fizzy/execute.cpp
gumb0: there's no constructor that reserves but not resizes

test/unittests/execute_test.cpp
gumb0: It won't compile, because it can't deduce the type of `{42}`

axic:
```
module.globalsec.emplace_back(fizzy::Global{true, fizzy::GlobalInitType::constant, {42}});
``

won't compile ?

Comment (gumb0) [reply test/unittests/execute_test.cpp]: this one will, thanks

File: lib/fizzy/parser.cpp [hunk 5]
```diff
@@ -21,6 +21,68 @@ struct parser<FuncType>
...
 };
 
+template <>
+struct parser<Global>
+{
+    parser_result<Global> operator()(const uint8_t* pos)
+    {
+        ValType type;
+        std::tie(type, pos) = parser<ValType>{}(pos);
+        if (type != ValType::i32 && type != ValType::i64)
```
chfast: parser<ValType> will throw this error

gumb0: removed

PR #39, opened 2020-01-09
Properly use the limits from the memory section

axic:


lib/fizzy/execute.hpp
```diff
@@ -10,6 +10,7 @@ struct Instance
 {
     const Module& module;
     bytes memory;
+    size_t memory_max_pages;
```
```suggestion
    size_t memory_max_pages = 0;
```

lib/fizzy/execute.cpp
```diff
@@ -150,11 +150,29 @@ inline uint64_t popcnt64(uint64_t value) noexcept
...
+    {
+        // FIXME: better error handling
+        throw std::runtime_error("Cannot support more than 1 memory section.");
+    }
+    else if (module.memorysec.size() > 0)
+    {
+        memory_min = module.memorysec[0].limits.min;
+        memory_max = module.memorysec[0].limits.max;
+        // FIXME: better error handling
+        if (memory_max > (memorylimit / pagesize))
```
gumb0: seems it was simpler to define `memorylimit` in pages

axic: Yes, but since this is part instantiation I think it's fine for simplicity (e.g. it is clear the VM will only use up to 256MB memory)

lib/fizzy/execute.cpp
```diff
@@ -10,7 +10,7 @@ namespace
 {
 constexpr unsigned pagesize = 65536;
 // Set hard limit of 256MB of memory.
-constexpr auto memorylimit = 256 * 1024 * 1024ULL;
+constexpr auto memory_limit = (256 * 1024 * 1024ULL) / pagesize;
```
```suggestion
constexpr auto memory_pages_limit = (256 * 1024 * 1024ULL) / pagesize;
```

PR #41, opened 2020-01-09
Implement support for Data section

axic:
Depends on #56.

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -178,8 +178,6 @@ Instance instantiate(const Module& module)
         else
             throw std::runtime_error("data initialization by imported global is not supported yet");
 
-        // FIXME: assert in parser
-        assert(data.memory_index == 0);
         // NOTE: these instructions can overlap
         assert((offset + data.init.size()) <= (memory_max * page_size));
         std::memcpy(memory.data() + offset, data.init.data(), data.init.size());
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -2,6 +2,7 @@
 #include "stack.hpp"
 #include "types.hpp"
 #include <cassert>
+#include <cstring>
 
 namespace fizzy
 {
@@ -168,6 +169,22 @@ Instance instantiate(const Module& module)
     // NOTE: fill it with zeroes
     bytes memory(memory_min * page_size, 0);
 
+    // Fill out memory based on data segments
+    for (const auto& data : module.datasec)
+    {
+        uint64_t offset;
+        if (data.offset.kind == ConstantExpression::Kind::Constant)
+            offset = data.offset.value.constant;
+        else
+            throw std::runtime_error("data initialization by imported global is not supported yet");
+
+        // FIXME: assert in parser
+        assert(data.memory_index == 0);
+        // NOTE: these instructions can overlap
+        assert((offset + data.init.size()) <= (memory_max * page_size));
+        std::memcpy(memory.data() + offset, data.init.data(), data.init.size());
+    }
+
     // TODO: add imported globals first
     std::vector<uint64_t> globals;
     globals.reserve(module.globalsec.size());
@@ -180,7 +197,8 @@ Instance instantiate(const Module& module)
             // TODO: initialize by imported global
             // Wasm spec section 3.3.7 constrains initialization by another global to imports only
             // https://webassembly.github.io/spec/core/valid/instructions.html#expressions
-            throw std::runtime_error("global initialization by imported global is not supported");
+            throw std::runtime_error(
+                "global initialization by imported global is not supported yet");
         }
     }
 
--- a/lib/fizzy/parser.cpp
+++ b/lib/fizzy/parser.cpp
@@ -182,6 +182,24 @@ struct parser<Export>
     }
 };
 
+template <>
+struct parser<Data>
+{
+    parser_result<Data> operator()(const uint8_t* pos)
+    {
+        MemIdx memory_index;
+        std::tie(memory_index, pos) = leb128u_decode<uint32_t>(pos);
+
+        ConstantExpression offset;
+        std::tie(offset, pos) = parser<ConstantExpression>{}(pos);
+
+        std::vector<uint8_t> init;
+        std::tie(init, pos) = parser<std::vector<uint8_t>>{}(pos);
+
+        return {{memory_index, offset, bytes(init.data(), init.size())}, pos};
+    }
+};
+
 Module parse(bytes_view input)
 {
     if (input.substr(0, wasm_prefix.size()) != wasm_prefix)
@@ -222,10 +240,12 @@ Module parse(bytes_view input)
         case SectionId::code:
             std::tie(module.codesec, it) = parser<std::vector<Code>>{}(it);
             break;
+        case SectionId::data:
+            std::tie(module.datasec, it) = parser<std::vector<Data>>{}(it);
+            break;
         case SectionId::custom:
         case SectionId::table:
         case SectionId::element:
-        case SectionId::data:
             // These sections are ignored for now.
             it += size;
             break;
--- a/lib/fizzy/parser.cpp
+++ b/lib/fizzy/parser.cpp
@@ -189,6 +189,8 @@ struct parser<Data>
     {
         MemIdx memory_index;
         std::tie(memory_index, pos) = leb128u_decode<uint32_t>(pos);
+        if (memory_index != 0)
+            throw parser_error{"unexpected memidx value " + std::to_string(memory_index)};
 
         ConstantExpression offset;
         std::tie(offset, pos) = parser<ConstantExpression>{}(pos);
@@ -196,7 +198,7 @@ struct parser<Data>
         std::vector<uint8_t> init;
         std::tie(init, pos) = parser<std::vector<uint8_t>>{}(pos);
 
-        return {{memory_index, offset, bytes(init.data(), init.size())}, pos};
+        return {{offset, bytes(init.data(), init.size())}, pos};
     }
 };
 
--- a/lib/fizzy/types.hpp
+++ b/lib/fizzy/types.hpp
@@ -42,6 +42,9 @@ using TypeIdx = uint32_t;
 // https://webassembly.github.io/spec/core/binary/modules.html#binary-funcidx
 using FuncIdx = uint32_t;
 
+// https://webassembly.github.io/spec/core/syntax/modules.html#syntax-memidx
+using MemIdx = uint32_t;
+
 /// Function locals.
 /// https://webassembly.github.io/spec/core/binary/modules.html#binary-local
 struct Locals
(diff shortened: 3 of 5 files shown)


lib/fizzy/execute.cpp
```diff
@@ -176,6 +176,16 @@ Instance instantiate(const Module& module)
     // NOTE: fill it with zeroes
     bytes memory(memory_min * page_size, 0);
 
+    // Fill out memory based on data segments
+    for (auto const& data : module.datasec)
+    {
+        // FIXME: assert in parser
+        assert(data.memidx == 0);
```
axic: I would do it this way for now, but when we have spec tests and clean up the code, we can move a lot of these "wasm version 1" assumptions into the parser.

lib/fizzy/execute.cpp
```diff
@@ -176,6 +176,16 @@ Instance instantiate(const Module& module)
...
+        // NOTE: these instructions can overlap
+        assert((data.offset + data.init.size()) <= (memory_max * page_size));
+        __builtin_memcpy(memory.data() + data.offset, data.init.data(), data.init.size());
```
axic: I guess could just use `std::memcpy` here?

test/unittests/execute_test.cpp
```diff
@@ -1450,7 +1450,7 @@ TEST(execute, start_section)
     // In this test the start function (index 1) writes a i32 value to the memory
     // and the same is read back in the "main" function (index 0).
     fizzy::Module module;
-    module.startfunc = 1;
+    module.startfunc = fizzy::TypeIdx{1};
```
axic: typo: `FuncIdx`

lib/fizzy/parser.hpp
```diff
@@ -140,4 +140,22 @@ struct parser<Code>
...
+        MemIdx memidx;
+        std::tie(memidx, pos) = leb128u_decode<uint32_t>(pos);
+
+        unsigned offset;
+        std::tie(offset, pos) = leb128u_decode<uint32_t>(pos);
+
+        std::vector<uint8_t> init;
+        std::tie(init, pos) = parser<std::vector<uint8_t>>{}(pos);
+
+        return {{memidx, offset, bytes{init.data(), init.size()}}, pos};
```
gumb0: I would use parentheses here not to look like it creates `bytes` with two elements
(actually I'm not sure why it doesn't)
```suggestion
        return {{memidx, offset, bytes(init.data(), init.size())}, pos};
```

lib/fizzy/parser.hpp
```diff
@@ -140,4 +140,22 @@ struct parser<Code>
...
+    parser_result<Data> operator()(const uint8_t* pos)
+    {
+        MemIdx memidx;
+        std::tie(memidx, pos) = leb128u_decode<uint32_t>(pos);
+
+        unsigned offset;
+        std::tie(offset, pos) = leb128u_decode<uint32_t>(pos);
+
+        std::vector<uint8_t> init;
+        std::tie(init, pos) = parser<std::vector<uint8_t>>{}(pos);
```
gumb0: I don't think this is correct because initializer is a constant expression (sequence of instructions).
I think it would be good to reuse here expression parser of `Global`.

axic: The initialiser is a byte array and not an expression.

gumb0: Sorry, I meant this not for initializer but for the offset

axic: Hm, the `offset` might be an expression.

lib/fizzy/execute.cpp
```diff
@@ -176,6 +176,16 @@ Instance instantiate(const Module& module)
     // NOTE: fill it with zeroes
     bytes memory(memory_min * page_size, 0);
 
+    // Fill out memory based on data segments
+    for (auto const& data : module.datasec)
```
```suggestion
    for (const auto& data : module.datasec)
```

test/unittests/execute_test.cpp
```diff
@@ -1441,3 +1441,24 @@ TEST(execute, start_section)
     EXPECT_EQ(ret[0], 42);
     EXPECT_EQ(instance.memory.substr(0, 4), from_hex("2a000000"));
 }
+
+TEST(execute, data_section)
```
gumb0: Probably better to move the test to instantiate_tests.cpp (and then it doesn't need `execute`)

lib/fizzy/execute.cpp
axic: Which one is the preferred?

```
$ grep "const auto" * |wc -l
      48
$ grep "auto const" * |wc -l
      18
```

lib/fizzy/execute.cpp
gumb0: I think `memcpy` or `std::copy` would be better

lib/fizzy/execute.cpp
gumb0: I think it was mostly me who committed `auto const`, and it was a mistake.

lib/fizzy/parser.hpp
axic: Fixed.

test/unittests/execute_test.cpp
axic: I can do that. @chfast any comments?

chfast: Yes, it makes sense to move it.

axic: Done.

lib/fizzy/parser.cpp
```diff
@@ -182,6 +182,24 @@ struct parser<Export>
     }
 };
 
+template <>
+struct parser<Data>
+{
+    parser_result<Data> operator()(const uint8_t* pos)
+    {
+        MemIdx memidx;
```
gumb0:
```suggestion
        MemIdx mem_idx;
```

lib/fizzy/types.hpp
```diff
@@ -212,6 +215,14 @@ struct Code
     bytes immediates;
 };
 
+// https://webassembly.github.io/spec/core/binary/modules.html#data-section
+struct Data
+{
+    MemIdx memidx;
```
```suggestion
    MemIdx memory_index;
```
(we have `function_type_index` in another struct)

test/unittests/parser_test.cpp
```diff
@@ -478,6 +478,25 @@ TEST(parser, code_section_with_basic_instructions)
     EXPECT_EQ(module.codesec[0].immediates, from_hex("010000000200000003000000"));
 }
 
+TEST(parser, data_section)
```
gumb0: Perhaps also a test with `GlobalGet`-inited offset?

lib/fizzy/execute.cpp
```diff
@@ -159,6 +159,22 @@ Instance instantiate(const Module& module)
...
+        if (data.offset.kind == ConstantExpression::Kind::Constant)
+            offset = data.offset.value.constant;
+        else
+            throw std::runtime_error("data initialization by imported global is not supported yet");
+
+        // FIXME: assert in parser
+        assert(data.memidx == 0);
+        // NOTE: these instructions can overlap
+        assert((offset + data.init.size()) <= (memory_max * page_size));
+        __builtin_memcpy(memory.data() + offset, data.init.data(), data.init.size());
```
gumb0: I would vote for using `memcpy` to minimize non-standard code, if it doesn't make performance/correctness difference

axic: You mean C memcpy or C++ std::copy ?

lib/fizzy/execute.cpp
axic: @chfast @gumb0 I am changing it to this, unless you guys object.

chfast: "to this" mean moving it to parser?

lib/fizzy/execute.cpp
```diff
@@ -159,6 +159,22 @@ Instance instantiate(const Module& module)
     // NOTE: fill it with zeroes
     bytes memory(memory_min * page_size, 0);
 
+    // Fill out memory based on data segments
+    for (const auto& data : module.datasec)
+    {
+        uint64_t offset;
+        if (data.offset.kind == ConstantExpression::Kind::Constant)
```
chfast: How many cases are going to be here? Can you rewrite it in a way `uint64_t offset` is initialized when declared.

lib/fizzy/execute.cpp
chfast: `std::copy_n` should be fine.

chfast: BTW, the `__builtin_memcpy()` and `memcpy()` is the same, just you don't need to include `<cstring>` for the first one.

lib/fizzy/execute.cpp
axic: Yeah.

lib/fizzy/execute.cpp
axic: Two cases. Constant or global and both evaluate to a single offset.

lib/fizzy/parser.cpp
axic: The type is called MemIdx, why not mirror that in the var name?

test/unittests/parser_test.cpp
axic: Added.

lib/fizzy/parser.cpp
gumb0: I don't insist, but in other places (in execute.cpp) we use underscore with `_idx`

axic:
@chfast @gumb0 made all these changes I believe (just haven't squashed them)

lib/fizzy/types.hpp
```diff
@@ -218,7 +218,7 @@ struct Code
 // https://webassembly.github.io/spec/core/binary/modules.html#data-section
 struct Data
 {
-    MemIdx memidx;
+    // Memory index is ensured to be 0 in the parser
```
chfast: This comment probably should be before `struct Data`.

lib/fizzy/execute.cpp
```diff
@@ -168,6 +169,20 @@ Instance instantiate(const Module& module)
     // NOTE: fill it with zeroes
     bytes memory(memory_min * page_size, 0);
 
+    // Fill out memory based on data segments
+    for (const auto& data : module.datasec)
+    {
+        uint64_t offset;
```
```suggestion
        const uint64_t offset = (data.offset.kind == ConstantExpression::Kind::Constant) ? data.offset.value.constant : throw std::runtime_error("data initialization by imported global is not supported yet");
```

axic: I'm not sure I like this :)

lib/fizzy/types.hpp
chfast: This one is left to fix.

axic: Why?

chfast: This way you are documenting `ConstantExpression offset` field.

axic: Ok, that makes sense, reworded and moved it.

lib/fizzy/parser.cpp
```diff
@@ -182,6 +182,26 @@ struct parser<Export>
...
 
+template <>
+struct parser<Data>
+{
+    parser_result<Data> operator()(const uint8_t* pos)
+    {
+        MemIdx memory_index;
+        std::tie(memory_index, pos) = leb128u_decode<uint32_t>(pos);
+        if (memory_index != 0)
+            throw parser_error{"unexpected memidx value " + std::to_string(memory_index)};
```
axic: It is memidx in the spec.

PR #43, opened 2020-01-09
Throw error on totally unknown section id

axic:


lib/fizzy/parser.cpp
```diff
@@ -46,9 +46,20 @@ Module parse(bytes_view input)
...
+        case SectionId::global:
+        case SectionId::export_:
+        case SectionId::start:
+        case SectionId::element:
+        case SectionId::data:
+            // These sections are ignored for now.
             it += size;
             break;
+        default:
+            throw parser_error{"unknown section encountered " + std::to_string(static_cast<int>(id))};
```
gumb0: lint complains about formatting

PR #44, opened 2020-01-09
Implement parsing export section

gumb0:


lib/fizzy/types.hpp
```diff
@@ -132,6 +132,23 @@ struct Memory
...
+    Table,
+    Memory,
+    Global
+};
+
+
+// https://webassembly.github.io/spec/core/binary/modules.html#export-section
+struct Export
+{
+    bytes name;
```
axic: I think this can be `string` given it must be an UTF-8 string.

lib/fizzy/execute.cpp
```diff
@@ -753,4 +753,13 @@ execution_result execute(const Module& module, FuncIdx function, std::vector<uin
...
     return execute(instance, function, args);
 }
+
+FuncIdx find_exported_function(const Module& module, std::string_view name)
+{
+    for (auto const& export_ : module.exportsec)
+        if (export_.type == ExportType::Function && name == export_.name)
+            return export_.index;
+
+    throw std::runtime_error("Exported function not found.");
```
axic: I think this function should return `optional` given it is an external API used prior to execution.

gumb0: changed

lib/fizzy/parser.cpp
```diff
@@ -81,6 +81,42 @@ struct parser<Global>
...
+{
+    parser_result<Export> operator()(const uint8_t* pos)
+    {
+        std::vector<uint8_t> name;
+        std::tie(name, pos) = parser<std::vector<uint8_t>>{}(pos);
+
+        Export result;
+        result.name.assign(name.begin(), name.end());
+
+        uint8_t const type = *pos++;
```
axic: `const uint8_t` as that's what we're using in other places?

lib/fizzy/execute.cpp
```diff
@@ -753,4 +753,13 @@ execution_result execute(const Module& module, FuncIdx function, std::vector<uin
     auto instance = instantiate(module);
     return execute(instance, function, args);
 }
+
+std::optional<FuncIdx> find_exported_function(const Module& module, std::string_view name)
+{
+    for (const auto& export_ : module.exportsec)
```
chfast: 1. I suggest following the coding rule to enclose a body of `for`, `if,` etc. with `{}` if body has more than 2 lines of code. This is enforceable by clang-tidy.
2. You can use `std::find_if`.

lib/fizzy/parser.cpp
```diff
@@ -81,6 +81,42 @@ struct parser<Global>
...
+            result.type = ExportType::Table;
+            break;
+        case 0x02:
+            result.type = ExportType::Memory;
+            break;
+        case 0x03:
+            result.type = ExportType::Global;
+            break;
+        default:
+            throw parser_error{"unexpected export type value " + std::to_string(*(pos - 1))};
```
chfast:
```suggestion
            throw parser_error{"unexpected export type value " + std::to_string(type)};
```

test/unittests/api_test.cpp
```diff
@@ -0,0 +1,27 @@
+#include "execute.hpp"
+#include <gtest/gtest.h>
+
+TEST(api, find_exported_function)
```
chfast: Please add a negative case.

gumb0: done

test/unittests/parser_test.cpp
```diff
@@ -285,6 +285,42 @@ TEST(parser, global_multi_const_inited)
...
+        bytes{wasm_prefix} + uint8_t{0x07} + uint8_t(section_contents.size()) + section_contents;
+
+    const auto module = parse(bin);
+    ASSERT_EQ(module.exportsec.size(), 1);
+    EXPECT_EQ(module.exportsec[0].name, "abc");
+    EXPECT_EQ(module.exportsec[0].type, ExportType::Function);
+    EXPECT_EQ(module.exportsec[0].index, 0x42);
+}
+
+TEST(parser, export_multiple)
```
axic: Can you add a test where the section contains `(export "memory" (memory 1))`  and trying to find "memory" as a function?

gumb0: added checks to api.find_exported_function test, is that ok?

lib/fizzy/types.hpp
```diff
@@ -154,6 +154,23 @@ struct Global
...
+    Memory = 0x02,
+    Global = 0x03
+};
+
+
+// https://webassembly.github.io/spec/core/binary/modules.html#export-section
+struct Export
+{
+    std::string name;
+    ExportType type = ExportType::Function;
```
gumb0: Added default initializers here

gumb0:
Rebased.

test/unittests/parser_test.cpp
axic: Yes, I just made the comment at the wrong place.

PR #47, opened 2020-01-10
Implement parsing function section

gumb0:


lib/fizzy/parser.hpp
```diff
@@ -26,6 +26,12 @@ struct parser
 {
 };
 
+template <>
+struct parser<uint32_t>
```
chfast: Why is this needed?

gumb0: parser of vector requires parser of an element

PR #48, opened 2020-01-10
Parse control instructions

chfast:


axic:
Rebased.

lib/fizzy/parser_expr.cpp
```diff
@@ -94,6 +94,14 @@ parser_result<Code> parse_expr(const uint8_t* pos)
         case Instr::i64_extend_i32_u:
             break;
 
+        case Instr::loop:
+        {
+            const auto type = *pos++;
```
axic: I think this should be statically typed for clarity: `const uint8_t type = *pos++;`

lib/fizzy/parser_expr.cpp
```diff
@@ -94,11 +101,61 @@ parser_result<Code> parse_expr(const uint8_t* pos)
         case Instr::i64_extend_i32_u:
             break;
 
+        case Instr::end:
+        {
+            if (!label_positions.empty())
+            {
+                const auto label_pos = label_positions.back();
+                label_positions.pop_back();
```
axic: Hm, why don't we introduce a `stack` type at some point? :)

lib/fizzy/parser_expr.cpp
```diff
@@ -10,23 +10,30 @@ inline void push(bytes& b, T value)
     b.resize(b.size() + sizeof(value));
     __builtin_memcpy(&b[b.size() - sizeof(value)], &value, sizeof(value));
 }
+
+/// The immediates offset for block instructions.
+using label_position = size_t;
+
+/// Sentinel constant to loop labels. They are not interested for parser.
```
gumb0:
```suggestion
/// Sentinel constant to loop labels. They are not interesting for parser.
```

lib/fizzy/types.hpp
```diff
@@ -18,6 +18,9 @@ enum class ValType : uint8_t
     i64 = 0x7e,
 };
 
+// The byte meaning an empty type list.
+constexpr uint8_t EmptyType = 0x40;
```
gumb0: `BlockTypeEmpty` would be closer to the spec, if this is only for block types

lib/fizzy/parser_expr.cpp
```diff
@@ -10,23 +10,30 @@ inline void push(bytes& b, T value)
...
+constexpr auto loop_label = label_position(-1);
+
 }  // namespace
 
 parser_result<Code> parse_expr(const uint8_t* pos)
 {
     Code code;
-    Instr instr;
-    do
+    std::vector<label_position> label_positions;
```
gumb0: I would prefer a comment here, too, saying that these are positions in immediates

lib/fizzy/parser_expr.cpp
```diff
@@ -94,11 +101,71 @@ parser_result<Code> parse_expr(const uint8_t* pos)
...
+                const auto label_pos = label_positions.back();
+                label_positions.pop_back();
+                if (label_pos != loop_label)  // If end of block instruction.
+                {
+                    const auto target_pc = static_cast<uint32_t>(code.instructions.size() + 1);
+                    const auto target_imm = static_cast<uint32_t>(code.immediates.size());
+
+                    // Set the imm values for block instruction.
+                    auto* block_imm = code.immediates.data() + label_pos;
+                    __builtin_memcpy(block_imm, &target_pc, sizeof(target_pc));
```
gumb0: Why can't you use `push` helper here?

gumb0: Ah got it, because it doesn't push, but modifies existing values.

But still would be better to extract this to another helper.

test/unittests/parser_test.cpp
```diff
@@ -465,3 +465,63 @@ TEST(parser, milestone1)
...
+    EXPECT_THROW(parse_expr(loop_without_imm.data()), parser_error);
+
+    const auto loop_void_empty = from_hex("03400b0b");
+    const auto [code1, pos1] = parse_expr(loop_void_empty.data());
+    EXPECT_EQ(code1.instructions, (std::vector{Instr::loop, Instr::end, Instr::end}));
+    EXPECT_EQ(code1.immediates.size(), 0);
+
+    const auto loop_i32_empty = from_hex("037f0b0b");
+    EXPECT_THROW(parse_expr(loop_i32_empty.data()), parser_error);  // loop arity != 0.
+}
```
gumb0: Add test for unbalanced `end`s

test/unittests/parser_test.cpp
```diff
@@ -465,3 +465,63 @@ TEST(parser, milestone1)
...
+TEST(parser, instr_block)
+{
+    const auto wrong_type = from_hex("0200");
+    EXPECT_THROW(parse_expr(wrong_type.data()), parser_error);
+
+    const auto empty = from_hex("010102400b0b");
+    const auto [code1, pos1] = parse_expr(empty.data());
+    EXPECT_EQ(code1.instructions,
+        (std::vector{Instr::nop, Instr::nop, Instr::block, Instr::end, Instr::end}));
+    EXPECT_EQ(hex(code1.immediates), "000400000009000000");
```
gumb0: would prefer this formatting
```suggestion
    EXPECT_EQ(hex(code1.immediates), "00" + "04000000" + "09000000");
```

lib/fizzy/types.hpp
gumb0: Also the constants in execute.cpp still use snake case https://github.com/wasmx/fizzy/blob/e7a14a73fbf6a8bba7f93b42fb0dd649adab1750/lib/fizzy/execute.cpp#L11

lib/fizzy/parser_expr.cpp
chfast: Not a generic one.

lib/fizzy/types.hpp
chfast: But things related to wasm spec in types.hpp use CamelCase.

gumb0: True, I would rather rename those other ones

test/unittests/parser_test.cpp
```diff
@@ -476,3 +476,72 @@ TEST(parser, milestone1)
...
+        (std::vector{Instr::nop, Instr::nop, Instr::block, Instr::end, Instr::end}));
+    EXPECT_EQ(hex(code1.immediates),
+        "00"
+        "04000000"
+        "09000000");
+
+    const auto block_i64 = from_hex("027e0b0b");
+    const auto [code2, pos2] = parse_expr(block_i64.data());
+    EXPECT_EQ(code2.instructions, (std::vector{Instr::block, Instr::end, Instr::end}));
+    EXPECT_EQ(hex(code2.immediates), "010200000009000000");
```
gumb0: The string could have been also split into separate immediates

PR #49, opened 2020-01-10
Implement imported functions

gumb0:


lib/fizzy/types.hpp
```diff
@@ -154,6 +154,29 @@ struct Global
     } init;
 };
 
+enum class ImportType : uint8_t
```
gumb0: Maybe this should better be merged with `ExportType` (they are identical) into something like `ExternalType` (or `ExternalKind` to disambiguate from types of type section)

lib/fizzy/execute.hpp
```diff
@@ -2,30 +2,35 @@
...
+struct execution_result
+{
+    // true if execution resulted in a trap
+    bool trapped;
+    // the resulting stack (e.g. return values)
+    // NOTE: this can be either 0 or 1 items
+    std::vector<uint64_t> stack;
+};
+
+using ImportedFunction = execution_result (*)(const std::vector<uint64_t>&);
```
axic: Should pass the `instance`, because many imported functions likely need access to the memory.

lib/fizzy/execute.hpp
```diff
@@ -2,30 +2,37 @@
...
+    // true if execution resulted in a trap
+    bool trapped;
+    // the resulting stack (e.g. return values)
+    // NOTE: this can be either 0 or 1 items
+    std::vector<uint64_t> stack;
+};
+
+struct Instance;
+
+using ImportedFunction = execution_result (*)(Instance&, const std::vector<uint64_t>&);
```
gumb0: Added `Instance` here, but I think there needs to be a safer way, passing only exported memory/globals/tables. `Instance` has all of them, not only exported.
Maybe not for this PR...

lib/fizzy/types.hpp
```diff
@@ -154,6 +154,29 @@ struct Global
...
+    Function = 0x00,
+    Table = 0x01,
+    Memory = 0x02,
+    Global = 0x03
+};
+
+// https://webassembly.github.io/spec/core/binary/modules.html#import-section
+struct Import
+{
+    std::string module;
```
gumb0: We probably don't need module and name for execution

gumb0:
I plan to rebase and add a test for `call` to imported function.

lib/fizzy/execute.cpp
```diff
@@ -192,7 +193,19 @@ Instance instantiate(const Module& module)
         }
     }
 
-    Instance instance = {module, std::move(memory), memory_max, std::move(globals)};
+    std::vector<TypeIdx> imported_function_types;
```
gumb0: Here we collect all types of the imported functions from `importsec` vector, which contains everything (functions, tables, memories, globals).

We could do it in parser, but then `importsec` would not be `vector<Import>` anymore.

lib/fizzy/parser.cpp
```diff
@@ -81,6 +81,47 @@ struct parser<Global>
...
+            result.type = ImportType::Table;
+            throw parser_error{"importing Tables is not implemented"};
+            break;
+        case 0x02:
+            result.type = ImportType::Memory;
+            std::tie(result.desc.memory, pos) = parser<Memory>{}(pos);
+            break;
+        case 0x03:
+            result.type = ImportType::Global;
+            std::tie(result.desc.global, pos) = parser<Global>{}(pos);
```
gumb0: I realized this is not correct, as imported globals don't have initializer expression, so we can't reuse `Global` parser here. I'll change this.

gumb0: Fixed by overwriting commits.

gumb0:
Rebased.

gumb0:
Rebased.

lib/fizzy/execute.cpp
```diff
@@ -131,7 +131,7 @@ inline uint64_t popcnt64(uint64_t value) noexcept
 }
 }  // namespace
 
-Instance instantiate(const Module& module)
+Instance instantiate(const Module& module, std::vector<ImportedFunction> imported_functions)
```
gumb0: I think at some point we might introduce a struct to pass all imports as one parameter, like
```
struct Imports
{
  std::vector<ImportedFunction> imported_functions
  std::vector<uint64_t> globals;
  std::optional<bytes> memory; // not sure yet if it's data bytes or just limits that can be imported 
  // table
};
```
In this PR I'd leave it as only vector of functions.

lib/fizzy/execute.cpp
```diff
@@ -189,8 +190,10 @@ Instance instantiate(const Module& module)
 
 execution_result execute(Instance& instance, FuncIdx function, std::vector<uint64_t> args)
 {
-    // TODO: handle the case when function index points to import
-    const auto& code = instance.module.codesec[function];
+    if (function < instance.imported_functions.size())
+        return instance.imported_functions[function](instance, std::move(args));
+
+    const auto& code = instance.module.codesec[function - instance.imported_functions.size()];
```
axic: Shouldn't we also add an assert that this value is less than `codesec.size()` ?

gumb0: I did in later commit

axic: Yes that is confusing some of that is moved around in later commits. Would it be possible to clean that up?

gumb0: I can squash first three commits

lib/fizzy/execute.cpp
```diff
@@ -214,8 +226,13 @@ Instance instantiate(const Module& module)
 
 execution_result execute(Instance& instance, FuncIdx function, std::vector<uint64_t> args)
 {
-    // TODO: handle the case when function index points to import
-    const auto& code = instance.module.codesec[function];
+    if (function < instance.imported_functions.size())
```
chfast: Can you rename `function` to `func_idx`? This name is confusing.

lib/fizzy/execute.cpp
```diff
@@ -214,8 +226,13 @@ Instance instantiate(const Module& module)
...
+        return instance.imported_functions[function](instance, std::move(args));
```
chfast: You move args here, but `ImportedFunction` takes args by reference.

lib/fizzy/execute.cpp
axic: But please do it on its own commit :)

lib/fizzy/execute.cpp
gumb0: Yeah, nice catch, I think I'll change `ImportedFunction` to take value

gumb0: I think it still worked correctly due to the rule "reference to temporary extends the lifetime to lifetime of the reference", but it's better not to rely on this obscure rule here, and for symmetry with `execute` pass vector of args by value to `ImportedFunction`, too

chfast: It was not a bug. The `std::move` is a cast to `&&`, but because of the function argument type is `const&` it becomes `const&&`. The last reference type is not good match for the move constructor, so the copy constructor is called.

gumb0: It copied into a newly constructed temporary indeed

chfast:
:joystick:

PR #51, opened 2020-01-11
Make max optional in Limits

axic:


lib/fizzy/types.hpp
```diff
@@ -29,7 +29,7 @@ struct FuncType
 struct Limits
 {
     uint32_t min;
-    uint32_t max;
+    std::optional<uint32_t> max;
```
axic: This field is used twice in the spec, for `memory` and `tables`. Therefore only used during instantiation and the overhead of optional is acceptable there.

test/unittests/parser_test.cpp
```diff
@@ -35,15 +35,16 @@ TEST(parser, limits_min)
     const auto input = from_hex("007f");
     const auto [limits, pos] = parser<Limits>{}(input.data());
     EXPECT_EQ(limits.min, 0x7f);
-    EXPECT_EQ(limits.max, std::numeric_limits<uint32_t>::max());
+    EXPECT_EQ(limits.max.has_value(), false);
```
chfast: `EXPECT_FALSE`

test/unittests/parser_test.cpp
```diff
@@ -35,15 +35,16 @@ TEST(parser, limits_min)
...
 }
 
 TEST(parser, limits_minmax)
 {
     const auto input = from_hex("01207f");
     const auto [limits, pos] = parser<Limits>{}(input.data());
     EXPECT_EQ(limits.min, 0x20);
-    EXPECT_EQ(limits.max, 0x7f);
+    EXPECT_EQ(limits.max.has_value(), true);
```
chfast: `EXPECT_TRUE`

axic:
> This looks ok, but I don't understand what is the motivation. You are setting the same value, just in other place.

In the parser it is left empty if it is not provided.

In the instantiation max is either set to what was provided or to the numerical limit. Finally we enforce our own limit.

PR #52, opened 2020-01-13
Templatize i32.eqz and i64.eqz to match the other opcodes

axic:
Not sure this is useful, but looks nice.


axic:
Chill out :wink:

Where was it detected by the unit test btw?

chfast:
> Where was it detected by the unit test btw?

It wasn't.

axic:
Fixed and added tests to actually catch it.

PR #53, opened 2020-01-13
Move some constants to limits.hpp

axic:


lib/fizzy/limits.hpp
```diff
@@ -0,0 +1,9 @@
+namespace fizzy
+{
+namespace
+{
+constexpr unsigned page_size = 65536;
+// Set hard limit of 256MB of memory.
+constexpr unsigned memory_pages_limit = (256 * 1024 * 1024ULL) / page_size;
```
axic: @gumb0 @chfast we could also camelcase these in this PR if you want.

gumb0: I would like to, I think it's useful to see that a variable is a global const without jumping to definition

axic:
@gumb0 @chfast how about this one? Keep it as it is, move to `limits.h`, fold into `types.h`?

gumb0:
Keeping these in `limits.hpp` looks good.
Is unnamed namespace needed?

lib/fizzy/limits.hpp
```diff
@@ -0,0 +1,9 @@
+namespace fizzy
+{
+namespace
```
chfast: Anonymous namespace should never be in a header.

axic: That was a typo, not sure why it even compiled.

lib/fizzy/limits.hpp
axic: @gumb0 @chfast how about this one then?

lib/fizzy/limits.hpp
chfast: The effect on `constexpr` variables are neglectable - they have internal linkage anyway.

axic:
On a second though `PageSize` probably belongs to `types.h`. Should we move both there?

chfast:
> On a second though `PageSize` probably belongs to `types.h`. Should we move both there?

I don't know.

axic:
Merging for now, just to get the camcelcase change in and make @gumb happy :wink:

gumb0:
Thanks, I'll go happily rebase my stuff now.

gumb0:
We should use it in `instantiate_test.cpp`, too

PR #54, opened 2020-01-14
Using namespace fizzy in api_test, execute_test and leb128_test

gumb0:


axic:
I liked the explicit names, but if you import the namespace, don't forget about `api_test.cpp` and `leb128_test.cpp`.

PR #55, opened 2020-01-14
Add unit tests for instantiate

gumb0:


test/unittests/instantiate_test.cpp
```diff
@@ -0,0 +1,72 @@
...
+    auto instance = instantiate(module);
+
+    ASSERT_EQ(instance.memory.size(), PageSize);
+    EXPECT_EQ(instance.memory_max_pages, 1);
+}
+
+TEST(instantiate, memory_single_unlimited)
+{
+    Module module;
+    module.memorysec.emplace_back(Memory{{1, {}}});
```
axic: Can use `std::nullopt` here.

test/unittests/CMakeLists.txt
```diff
@@ -13,6 +13,7 @@ target_sources(
     end_to_end_test.cpp
     execute_control_test.cpp
     execute_test.cpp
+    instantiate_test.cpp
```
axic: Should this be separate or part of `api_test.cpp`?

test/unittests/instantiate_test.cpp
```diff
@@ -0,0 +1,72 @@
+#include "execute.hpp"
+#include "utils.hpp"
+#include <gtest/gtest.h>
+
+using namespace fizzy;
+
+namespace
+{
+constexpr unsigned PageSize = 65536;
```
axic:
```suggestion
constexpr unsigned page_size = 65536;
```

I would prefer to rename this consistently in the source if we decide so.

PR #56, opened 2020-01-14
Split out ConstantExpression parsing from Global

axic:


lib/fizzy/types.hpp
```diff
@@ -136,15 +136,15 @@ struct Memory
     Limits limits;
 };
 
-enum class GlobalInitType : uint8_t
+enum class ConstantInitType : uint8_t
```
axic: Could also scope this enum within `ConstantExpression`.

chfast: Unless it is named like that in the spec, the `ConstantInitKind` is preferable name. In general, in language parsing context "kind" is better name than "type" for something that is not a type in language sense (`i32`, `i64`, etc).

axic: It is called type in the spec.

Screenshot 2020-01-14 at 16 18 06

gumb0:
Looks good, but I agree with renaming to `ConstantInitKind`
(I would need then to rename `ExportType` and `ImportType` in other PRs, too...)

lib/fizzy/types.hpp
axic: Can we scope it within `ConsantExpression`? It is specific to it.

axic: And I suppose we rename `init_type` to `init_kind`.

axic: I would propose the following renaming:
```
ConstantExpression expression;
expression.kind = ConstantExpression::Kind::Constant;
expression.value.constant = 1;
```
and
```
ConstantExpression expression;
expression.kind = ConstantExpression::Kind::GlobalGet;
expression.value.global_index = 1;
```

Can we use anonymous unions or some compilers have issues with that?

chfast: 1. The proposal looks good.
2. The nested structs and enums can be annoying to use, but we can start with it anyway.
3. Anonymous unions are not C++ compliant. I believe only MSVC is complaining in some pedantic mode.
4. You can also check `std::variant` type.

lib/fizzy/types.hpp
```diff
@@ -136,23 +136,27 @@ struct Memory
...
 
-enum class GlobalInitType : uint8_t
+struct ConstantExpression
 {
-    constant,
-    global
+    enum class Kind : uint8_t
+    {
+        constant,
+        global_get
```
axic: Should we make these camelcase?

PR #57, opened 2020-01-14
Enforce memory limits properly

axic:
Fixes #51.

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -164,7 +164,7 @@ Instance instantiate(const Module& module)
         else
             memory_max = std::numeric_limits<uint32_t>::max();
         // FIXME: better error handling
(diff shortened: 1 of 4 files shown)


lib/fizzy/execute.cpp
```diff
@@ -162,9 +162,9 @@ Instance instantiate(const Module& module)
         if (module.memorysec[0].limits.max)
             memory_max = *module.memorysec[0].limits.max;
         else
-            memory_max = std::numeric_limits<uint32_t>::max();
+            memory_max = memory_pages_limit;
```
axic: @chfast actually this is what I meant doing in #51.

axic:
> Is the parser checking if memory_min is less than memory_max?

No, that could be added. Just checked the spec but that doesn't have this explicitly stated.

PR #59, opened 2020-01-14
Stack type

chfast:


lib/fizzy/stack.hpp
```diff
@@ -4,20 +4,30 @@
...
+        const auto res = back();
         pop_back();
         return res;
     }
 
-    uint64_t peek(difference_type depth = 1) const noexcept { return *(end() - depth); }
+    T peek(difference_type depth = 1) const noexcept { return *(end() - depth); }
 };
+
+using uint64_stack = stack<uint64_t>;
```
axic: Why not just use `stack<uint64_t>` in the correct place?

lib/fizzy/stack.hpp
```diff
@@ -0,0 +1,31 @@
...
+{
+template <typename T>
+class stack : public std::vector<T>
+{
+public:
+    using difference_type = typename std::vector<T>::difference_type;
+
+    using std::vector<T>::vector;
+
+    using std::vector<T>::back;
```
gumb0: Are these required? (I think they are not with public inheritance?)

And with these here can we use private inheritance?

test/unittests/leb128_test.cpp
```diff
@@ -1,4 +1,5 @@
 #include "leb128.hpp"
+#include "stack.hpp"
```
gumb0: Not sure why it's needed in this file?

lib/fizzy/stack.hpp
chfast: I don't know why, but somehow they are required. Otherwise you cannot use vector's method from within the stack class definition.

There are many other methods used from outside so private inheritance will not work.

gumb0: I see, maye something like `std::vector<T>::back()` would have worked, too, but these might look better.

chfast: This is strange and unexpected compilation error, looks like related to which methods depend on T type in `std::vector`. Also `vector::vector` does not work any more, you need `std::vector<T>::vector`. I didn't bother to investigate longer than 15 mins.

axic:
Needs rebase though.

lib/fizzy/stack.hpp
axic: Perhaps https://github.com/wasmx/fizzy/blob/master/lib/fizzy/execute.cpp#L283 replacing this with `stack.peek()` would have helped?

chfast: The `peek()` looks weird. The `peek(1)` means the top item and it unnecessary uses `end()` instead of just using `[]` operator. I might touch it in the "labels" PR.

axic: Can you just create another clean up PR for stack?

PR #60, opened 2020-01-15
Implement parsing import section

gumb0:


lib/fizzy/parser.cpp
```diff
@@ -77,18 +91,50 @@ struct parser<Global>
...
+    }
+};
 
-        Global result;
-        result.is_mutable = (*pos == 0x01);
-        ++pos;
+template <>
+struct parser<Import>
+{
+    parser_result<Import> operator()(const uint8_t* pos)
```
axic: Ref: wasm spec 5.5.5. And it seems to be correct.

lib/fizzy/parser.cpp
```diff
@@ -77,18 +91,50 @@ struct parser<Global>
...
+    {
+        std::vector<uint8_t> module;
+        std::tie(module, pos) = parser<std::vector<uint8_t>>{}(pos);
```
axic: We should add a parser helper for names. Which must be utf8. @gumb0 can you create a new PR for that?

gumb0:
Rebased.

PR #62, opened 2020-01-15
Add parser helper for strings

gumb0:


lib/fizzy/parser.cpp
```diff
@@ -95,15 +95,26 @@ struct parser<Global>
 };
 
 template <>
-struct parser<Export>
+struct parser<std::string>
 {
-    parser_result<Export> operator()(const uint8_t* pos)
+    parser_result<std::string> operator()(const uint8_t* pos)
     {
         std::vector<uint8_t> name;
```
axic: Name? What about value or something like that?

gumb0: changed

PR #64, opened 2020-01-16
 Merge ImportKind and ExportKind enums into ExternalKind

gumb0:


lib/fizzy/types.hpp
```diff
@@ -190,7 +190,7 @@ struct Import
     } desc;
 };
 
-enum class ExportType : uint8_t
+enum class ExportKind : uint8_t
```
axic: This is actually called export type in the spec, isn't it? But I think kind is a better term.

gumb0: There's no name for it in Binary Format, but in Structure there's a notion of "External type" (the actual type of function/global etc.) https://webassembly.github.io/spec/core/syntax/types.html#external-types
which suggests maybe we should merge `ImportKind` and `ExportKind` into `ExternalKind`

axic: Yes, sounds good.

lib/fizzy/types.hpp
```diff
@@ -203,7 +203,7 @@ enum class ExportType : uint8_t
 struct Export
 {
     std::string name;
-    ExportType type = ExportType::Function;
+    ExportKind type = ExportKind::Function;
```
axic: Shouldn't this also use `kind`? `struct Import` names it kind.

PR #65, opened 2020-01-16
Load CMake 3.15.6 on lgtm

axic:
Based on https://discuss.lgtm.com/t/update-cmake-to-3-14-to-allow-fetch-content/2048/16

Apparently this works: https://lgtm.com/logs/0940aadb80a435c4499775daaa8a39b7fe87d60e/lang:cpp

### DIFF
--- a/lgtm.yml
+++ b/lgtm.yml
@@ -1,0 +1,8 @@
+extraction:
+  cpp:
+    after_prepare:
+      - mkdir custom_cmake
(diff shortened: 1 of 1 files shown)


chfast:
Is there a report we can see?

axic:
> Is there a report we can see?

Unfortunately not until we merge this and let it run properly on the repo.

Here's a report for evmone: https://lgtm.com/projects/g/ethereum/evmone?mode=list

axic:
Triggered builds. Should be here when ready: https://lgtm.com/projects/g/wasmx/fizzy?mode=list

I didn't enable PR integration so it is only built once a day or so, but we could enable integration and get reports here.

PR #67, opened 2020-01-16
Implement stack cleanup in br/br_if instructions

chfast:
Requires #25.

When branch is taken, additional stack items must be dropped. I.e. when stack height at the label was `X` branch instruction must make it `X` again (`loop`, `block` with type `[]`) or `X+1` (`block` with type `[t]`).


gumb0:
Looks good, I would a test with non-zero immediate of `br`

chfast:
> Looks good, I would a test with non-zero immediate of `br`

Added new test where `br` jumps through `loop` to outer `block`. The final check is different there — we just check if there is nothing on the stack in the end.

PR #68, opened 2020-01-16
Implement blocks with results (non-zero arity)

chfast:
Requires #25, #67.

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -227,13 +227,6 @@ execution_result execute(Instance& instance, FuncIdx function, std::vector<uint6
         case Instr::block:
         {
             const auto arity = read<uint8_t>(immediates);
-
-            if (arity != 0)
-            {
-                throw std::runtime_error(
-                    "block instruction with non-empty result type not implemented yet");
-            }
-
             const auto target_pc = read<uint32_t>(immediates);
             const auto target_imm = read<uint32_t>(immediates);
             LabelContext label{code.instructions.data() + target_pc,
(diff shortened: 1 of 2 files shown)


lib/fizzy/execute.cpp
```diff
@@ -271,12 +264,16 @@ execution_result execute(Instance& instance, FuncIdx function, std::vector<uint6
...
             immediates = label.immediate;
 
-            // FIXME: handle arity
-            assert(label.arity == 0);
-
             // When branch is taken, additional stack items must be dropped.
-            assert(stack.size() >= label.stack_height);
-            stack.resize(label.stack_height);
+            assert(stack.size() >= label.stack_height + label.arity);
+            if (label.arity)
```
```suggestion
            if (label.arity > 0)
```

axic: Why not `!= 0` instead?

chfast: Probably the most correct one is `== 1`. Maybe I will just add an assert inside.

lib/fizzy/execute.cpp
```diff
@@ -271,12 +264,17 @@ execution_result execute(Instance& instance, FuncIdx function, std::vector<uint6
...
-            // FIXME: handle arity
-            assert(label.arity == 0);
-
             // When branch is taken, additional stack items must be dropped.
-            assert(stack.size() >= label.stack_height);
-            stack.resize(label.stack_height);
+            assert(stack.size() >= label.stack_height + label.arity);
+            if (label.arity != 0)
+            {
+                assert(label.arity == 1);
```
chfast: Fine now?

PR #69, opened 2020-01-17
Implementation of iNN.store/load with shorter types (8/16/32)

rodiazet:
store and load instructions for 8,16,32 bit types signed and unsigned.


lib/fizzy/execute.cpp
```diff
@@ -196,6 +196,48 @@ Instance instantiate(const Module& module)
     return instance;
 }
 
+namespace
+{
+template <typename DstT, typename SrcT>
+DstT extend(SrcT in)
```
chfast: Alternative implementation (a bit better because `static_cast` is only used when needed).

```
template <typename DstT, typename SrcT>
DstT extend(SrcT in)
{
    if constexpr (std::is_signed<SrcT>::value)
    {
        using SignedDstT = typename std::make_signed<DstT>::type;
        return static_cast<DstT>(SignedDstT{in});
    }
    else
        return DstT{in};
}
```

lib/fizzy/execute.cpp
```diff
@@ -196,6 +196,48 @@ Instance instantiate(const Module& module)
...
+inline bool load_from_memory(Instance& instance, Stack<uint64_t>& stack, const uint8_t* immediates)
+{
+    const auto address = static_cast<uint32_t>(stack.pop());
+    // NOTE: alignment is dropped by the parser
+    const auto offset = read<uint32_t>(immediates);
+    if ((address + offset + sizeof(SrcT)) > instance.memory.size())
+        return false;
+
+    const auto ret = load<SrcT>(instance.memory, address + offset);
+    stack.push(extend<DstT, SrcT>(ret));
```
```suggestion
    stack.push(extend<DstT>(ret));
```
You can omit `SrcT` because it is going to be deduced.

lib/fizzy/execute.cpp
```diff
@@ -358,64 +400,159 @@ execution_result execute(Instance& instance, FuncIdx function, std::vector<uint6
...
             break;
         }
-        // FIXME: make this into a template?
         case Instr::i32_load:
         {
-            const auto address = static_cast<uint32_t>(stack.pop());
-            // NOTE: alignment is dropped by the parser
-            const auto offset = read<uint32_t>(immediates);
-            if ((address + offset + sizeof(uint32_t)) > instance.memory.size())
+            if (!load_from_memory<uint32_t, uint32_t>(instance, stack, immediates))
```
chfast: In the declaration of the template you can do `template <typename DstT, typename SrcT = DstT>` and omit one type here: `load_from_memory<uint32_t>`.

lib/fizzy/execute.cpp
```diff
@@ -196,6 +196,50 @@ Instance instantiate(const Module& module)
...
+    {
+        using SignedDstT = typename std::make_signed<DstT>::type;
+        return static_cast<DstT>(SignedDstT{in});
+    }
+    else
+        return DstT{in};
+}
+
+template <typename DstT, typename SrcT = DstT>
+inline bool load_from_memory(Instance& instance, Stack<uint64_t>& stack, const uint8_t* immediates)
```
gumb0: `Instance` can be passed by const reference, I guess?

I think I would actually pass only memory here instead of the whole instance.

lib/fizzy/execute.cpp
```diff
@@ -196,6 +196,50 @@ Instance instantiate(const Module& module)
     return instance;
 }
 
+namespace
+{
+template <typename DstT, typename SrcT>
+DstT extend(SrcT in)
```
```suggestion
inline constexpr DstT extend(SrcT in) noexcept
```

test/unittests/execute_test.cpp
```diff
@@ -320,6 +320,346 @@ TEST(execute, i64_load_trap)
...
+{
+    Module module;
+    module.memorysec.emplace_back(Memory{{1, 1}});
+    module.codesec.emplace_back(
+        Code{0, {Instr::local_get, Instr::i32_load8_s, Instr::end}, {0, 0, 0, 0, 0, 0, 0, 0}});
+
+    auto instance = instantiate(module);
+    instance.memory[0] = 0x80;
+    instance.memory[1] = 0xf1;
+    const auto [trap, ret] = execute(instance, 0, {65537});
```
gumb0: I suggest to merge each pair of non-trap and trap test into one, as they differ only by passed argument

test/unittests/execute_test.cpp
```diff
@@ -320,6 +320,346 @@ TEST(execute, i64_load_trap)
...
+    instance.memory[0] = 0x01;
+    instance.memory[1] = 0x00;
+    instance.memory[2] = 0x00;
+    instance.memory[3] = 0x80;
+    instance.memory[4] = 0xf1;
+    const auto [trap, ret] = execute(instance, 0, {0});
+
+    ASSERT_FALSE(trap);
+    ASSERT_EQ(ret.size(), 1);
+    ASSERT_EQ(static_cast<uint64_t>(ret[0]), 2147483649);
```
gumb0: I think it would be better to write expectation in hex, at least for unsigned tests.

gumb0: Also `ret[0]` is already `uint64_t`, casting is not needed

lib/fizzy/execute.cpp
```diff
@@ -196,6 +196,50 @@ Instance instantiate(const Module& module)
...
+    if ((address + offset + sizeof(SrcT)) > memory.size())
+        return false;
+
+    const auto ret = load<SrcT>(memory, address + offset);
+    stack.push(extend<DstT>(ret));
+    return true;
+}
+
+template <typename DstT>
+inline bool store_into_memory(Instance& instance, Stack<uint64_t>& stack, const uint8_t* immediates)
```
gumb0: Here, too, `bytes& memory` for symmetry instead of `instance`

rodiazet: Damn! Too fast. Sorry. Fixed.

lib/fizzy/execute.cpp
```diff
@@ -196,6 +196,50 @@ Instance instantiate(const Module& module)
     return instance;
 }
 
+namespace
+{
+template <typename DstT, typename SrcT>
+inline DstT extend(SrcT in)
```
```suggestion
inline DstT extend(SrcT in) noexcept
```

lib/fizzy/execute.cpp
```diff
@@ -362,64 +406,159 @@ execution_result execute(Instance& instance, FuncIdx function, std::vector<uint6
...
-            if ((address + offset + sizeof(uint64_t)) > instance.memory.size())
+            if (!store_into_memory<uint64_t>(instance.memory, stack, immediates))
+            {
+                trap = true;
+                goto end;
+            }
+            break;
+        }
+        case Instr::i32_store8:
+        case Instr::i64_store8:
```
chfast: Note: If both instructions remain the same, parser can emit just one of them.

rodiazet: I assume that I don't have to change anything. yes?

chfast: No, you don't have to.

axic:
Can you use capitalised commit messages? Also perhaps list the opcodes, such as :
> Tests for iNN.store/load with shorter types (8/16/32)

rodiazet:
> Can you use capitalised commit messages? Also perhaps list the opcodes, such as :
> 
> > Tests for iNN.store/load with shorter types (8/16/32)

Done.

lib/fizzy/execute.cpp
axic: I'd suggest we only merge instructions when everything is implemented and we pass the test suite.

lib/fizzy/execute.cpp
```diff
@@ -212,6 +212,50 @@ Instance instantiate(const Module& module)
     return instance;
 }
 
+namespace
```
axic: Damn, these helpers should have been next to the other load/store helpers and not between instantiate/execute.

rodiazet: Should I move it in an additional PR?

PR #70, opened 2020-01-17
Return instruction

chfast:


lib/fizzy/execute.cpp
```diff
@@ -223,13 +224,17 @@ execution_result execute(Instance& instance, FuncIdx function, std::vector<uint6
...
     bool trap = false;
 
     const Instr* pc = code.instructions.data();
     const uint8_t* immediates = code.immediates.data();
 
+    Stack<LabelContext> labels;
+    //    bool have_result =
+    //    !instance.module.typesec[instance.module.funcsec[function]].outputs.empty();
+    labels.emplace_back(LabelContext{
+        &code.instructions[code.instructions.size() - 1], nullptr, have_result, stack.size()});
```
axic: I don't like that bool is casted to size_t here.

This can be deduced from the type section too. I think there should be a PR which enforces the type section on execute.

chfast:
This requires some cleanups, but I need to take a break so this is good time for first review.

lib/fizzy/execute.cpp
```diff
@@ -323,6 +326,26 @@ execution_result execute(Instance& instance, FuncIdx function, std::vector<uint6
...
             break;
         }
+        case Instr::return_:
+        case_return:
+        {
+            // TODO: Not needed, but satisfies the assert in the end of the main loop.
+            labels.clear();
+
+            const bool have_result =
+                !instance.module.typesec[instance.module.funcsec[function]].outputs.empty();
```
axic: Add an assert for `instance.module.typesec.size() >= instance.module.funcsec[function]`. At least we did that so far, even though the indexing should also throw.

lib/fizzy/execute.cpp
```diff
@@ -388,6 +388,27 @@ execution_result execute(Instance& instance, FuncIdx func_idx, std::vector<uint6
                 stack.push(ret.stack[0]);
             break;
         }
+        case Instr::return_:
+        {
+            // TODO: Not needed, but satisfies the assert in the end of the main loop.
+            labels.clear();
+
+            assert(func_idx < instance.module.funcsec.size());
```
axic: This assert should go to the top of the function.

lib/fizzy/execute.cpp
```diff
@@ -388,6 +388,27 @@ execution_result execute(Instance& instance, FuncIdx func_idx, std::vector<uint6
...
+            const auto type_idx = instance.module.funcsec[func_idx];
+            assert(type_idx < instance.module.typesec.size());
```
axic: I think actually these two could also be moved up if they are used multiple times?

lib/fizzy/execute.cpp
chfast: I don't think it should. Functype is only needed for `return` instruction, no point requiring it for every execution.

lib/fizzy/execute.cpp
chfast: This is used at most once.

lib/fizzy/execute.cpp
axic: We should validate the number of input arguments based on the functype.

chfast: Not sure what this is needed all the time, but probably should be discussed as a part of another PR.

gumb0:
Test suggestion:  non-zero `br` ending the function, something like   
```
(func (result i32)
      (loop
        i32.const 1
        br 1
      )
      i32.const 0
)
```

chfast:
> Since imported functions are now supported, instead of `func_idx` you should use here `func_idx - imported_functions.size()`. This value is also already defined at the top of `execute` as `code_idx`

I added suggested unit tests that also has an imported function.

PR #71, opened 2020-01-17
Add function's implict block

chfast:
There is implicit `block` instruction at the beginning of every function. This label can be targeted with `return` instruction or simply `br` with correct `labelidx`.

There are at least 2 ways how to implement this:

### Hacky

As this example - handle case where `labelidx` is out of the label stack range. The hacky part is to skip the stack unwinding. Therefore, information about function's output type is not needed (if the function has a result or not).

### Clean

Just insert the label at the start of `execute()` as shown in #70. But to make it work properly we need the function type information (i.e. label's arity). This is theoretically available in `funcsec` / `typesec`, but in most execution tests now provided.


chfast:
Closed in favor of #70

PR #72, opened 2020-01-17
test: introduce make_section helper in parser_test

axic:


test/unittests/parser_test.cpp
```diff
@@ -9,6 +9,14 @@ static const auto functype_void_to_void = from_hex("600000");
 static const auto functype_i32i64_to_i32 = from_hex("60027f7e017f");
 static const auto functype_i32_to_void = from_hex("60017f00");
 
+namespace
+{
+bytes push_section(uint8_t id, const bytes& section)
```
axic: `bytes_view section` wasn't working (needed to be converted), so this seemed simpler.

axic: The reason for this helper is that it makes the tests cleaner when multiple sections are pushed. I have a change in progress which enforces some checks in the parser (e.g. function and code section size must match), which require adding more sections in some tests.

axic: Could also name `add_section`.

chfast: I would name it `make_section` because it does not push nor add it anywhere.

test/unittests/parser_test.cpp
```diff
@@ -9,6 +9,14 @@ static const auto functype_void_to_void = from_hex("600000");
...
+{
+    return bytes{} + id + uint8_t(section.size()) + section;
```
```suggestion
    return bytes{id} + uint8_t(section.size()) + section;
```
I don't like that in previous version there are `uint8_t` connected with `+`.
Also add assert that `section.size() < 0x80`.

test/unittests/parser_test.cpp
```suggestion
bytes push_section(uint8_t id, const bytes& content)
```

axic: Done.

test/unittests/parser_test.cpp
axic: `ASSERT_TRUE` doesn't work in the helper. Any suggestions?

test/unittests/parser_test.cpp
```diff
@@ -9,6 +9,15 @@ static const auto functype_void_to_void = from_hex("600000");
 static const auto functype_i32i64_to_i32 = from_hex("60027f7e017f");
 static const auto functype_i32_to_void = from_hex("60017f00");
 
+namespace
+{
+bytes make_section(uint8_t id, const bytes& content)
+{
+    assert(content.size() < 0x80);
+    return bytes{id} + uint8_t(content.size()) + content;
```
axic:
```suggestion
    return bytes{id, uint8_t(content.size()} + content;
```

@chfast do you want to merge it like this?

chfast: Yes, this is fine.
I would also use `static_cast<uint8_t>` explicitly here.

test/unittests/parser_test.cpp
chfast: For that to work you would need to create GTest fixture and implement `make_section` as a method of it. https://github.com/google/googletest/blob/master/googletest/docs/primer.md#test-fixtures-using-the-same-data-configuration-for-multiple-tests-same-data-multiple-tests

PR #74, opened 2020-01-21
Fizzy benchmark tool

chfast:


test/benchmarks/micro/spinner.inputs
```diff
@@ -0,0 +1,18 @@
+1
```
axic: What is the format of this file?

axic: Can you leave a brief README in the benchmarking directory to explain the format and how it works?

test/benchmarks/README.md
```diff
@@ -0,0 +1,41 @@
+# Fizzy Benchmarking Suite
+
+## Inputs files
+
+For each wasm file there may exist an `.inputs` file matching the `.wasm` file
+name.
+
+The inputs file is need to benchark execution and multiple execution case may be
```
axic:
```suggestion
The inputs file is need to benchmark execution and multiple execution case may be
```

gumb0:
```suggestion
The inputs file is needed to benchmark execution and multiple execution case may be
```

test/bench/bench.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+            case InputsReadingState::FuncName:
+                benchmark_case.func_name = std::move(l);
+                st = InputsReadingState::FuncArguments;
+                break;
+
+            case InputsReadingState::FuncArguments:
+            {
+                std::istringstream iss{std::move(l)};
+                while (!iss.eof())
```
gumb0: did you consider doing it with stream iterators something like
```
benchmark_case.func_args.assign(std::istream_iterator<uint64_t>(iss), std::istream_iterator<uint64_t>());
```

test/benchmarks/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+Execution case is specified by 6 lines:
+
+1. The case name. Must not be empty.
+2. The exported wasm function name to be executed.
+3. The function arguments as space-separated list of integers. May be empty.
+   E.g. `1 2 3`.
+4. The hex-encoded bytes of the initial memory. May be empty.
+5. The expected result as an integer. Empty line means no result is expected.
+6. The hex-encoded bytes of the expected memory after execution.
```
gumb0: May be empty?

test/bench/bench.cpp
```diff
@@ -0,0 +1,242 @@
...
+struct ExecutionBenchmarkCase
+{
+    std::shared_ptr<const fizzy::bytes> wasm_binary;
+    std::string func_name;
+    std::vector<uint64_t> func_args;
+    fizzy::bytes memory;
+    std::optional<uint64_t> expected_result;
+    fizzy::bytes expected_memory;
+
+    void operator()(benchmark::State& state)
```
gumb0: Maybe for nice consistency this could be free function `void benchmark_execute(benchmark::State& state, const ExecutionBenchmarkCase& execution_case)` and then move `ExecutionBenchmarkCase` instance into lambda passed to `RegisterBenchmark`

test/bench/bench.cpp
```diff
@@ -0,0 +1,242 @@
...
+                else if (result.stack[0] != *expected_result)
+                    return state.SkipWithError("Incorrect result");
+            }
+            else if (!result.stack.empty())
+                return state.SkipWithError("Unexpected result");
+
+            if (instance.memory.size() < expected_memory.size())
+                return state.SkipWithError("Result memory is shorted than expected");
+
+            if (!std::equal(std::begin(expected_memory), std::end(expected_memory),
```
gumb0: Could be just `!=`

test/bench/bench.cpp
```diff
@@ -0,0 +1,242 @@
...
+                    std::begin(instance.memory)))
+                return state.SkipWithError("Incorrect result memory");
+        }
+
+        for (auto _ : state)
```
gumb0: What does it do in this loop?

gumb0: Ah I guess the first `execute` above is just to check the correctness and this loop is running many iterations of it to actually benchmark.

test/bench/bench.cpp
```diff
@@ -0,0 +1,242 @@
...
+        {
+            std::copy(std::begin(memory), std::end(memory), std::begin(instance.memory));
```
gumb0: This actually doesn't seem correct in case `execute` of previous iteration writes to the bytes above initial memory
(you'll reset only the first bytes, not the whole page)

test/bench/bench.cpp
```diff
@@ -0,0 +1,242 @@
...
+    state.counters["rate"] =
+        benchmark::Counter(static_cast<double>(num_bytes_parsed), benchmark::Counter::kIsRate);
+}
+
+void benchmark_instantiate(benchmark::State& state, const fizzy::bytes& wasm_binary)
+{
+    const auto module = fizzy::parse(wasm_binary);
+    for (auto _ : state)
+    {
+        auto instance = fizzy::instantiate(module, {});
```
gumb0: Can be now
```suggestion
        auto instance = fizzy::instantiate(module, {});
```

CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 cmake_minimum_required(VERSION 3.15)
 
 cmake_policy(SET CMP0077 NEW)
+set(CMAKE_FIND_PACKAGE_PREFER_CONFIG TRUE)
```
gumb0: what does this do?

test/bench/bench.cpp
chfast: True. I will have to copy initial memory from the instance restore it for every loop iteration.

test/bench/bench.cpp
```diff
@@ -0,0 +1,242 @@
...
+                break;
+
+            case InputsReadingState::FuncName:
+                benchmark_case.func_name = std::move(l);
+                st = InputsReadingState::FuncArguments;
+                break;
+
+            case InputsReadingState::FuncArguments:
+            {
+                std::istringstream iss{l};
```
axic: This fails on my xcode:
```
/Users/alex/Projects/fizzy/test/bench/bench.cpp:160:36: error: implicit instantiation of undefined template 'std::__1::basic_istringstream<char, std::__1::char_traits<char>,
      std::__1::allocator<char> >'
                std::istringstream iss{l};
                                   ^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/iosfwd:134:32: note: template is declared here
    class _LIBCPP_TEMPLATE_VIS basic_istringstream;
                               ^
```

axic: Needs explicit `<sstream>` include.

test/benchmarks/README.md
```diff
@@ -0,0 +1,41 @@
...
+2. The exported wasm function name to be executed.
+3. The function arguments as space-separated list of integers. May be empty.
+4. The hex-encoded bytes of the initial memory. May be empty.
+5. The expected result as an integer. Empty line means no result is expected.
+6. The hex-encoded bytes of the expected memory after execution. 
+   If empty, result memory will not be checked.
+
+Additional empty lines are allowed before each case.
+
+FIXME: Only beginning of memory should be set/checked in both cases.
```
gumb0: This looks fixed already?

test/bench/bench.cpp
```diff
@@ -0,0 +1,243 @@
...
+                if (!l.empty())
+                    benchmark_case.expected_result = std::stoull(l);
+                st = InputsReadingState::ExpectedMemory;
+                break;
+
+            case InputsReadingState::ExpectedMemory:
+                benchmark_case.expected_memory = fizzy::from_hex(strip_space(l));
+
+                benchmark::RegisterBenchmark(("execute/" + base_name + '/' + input_name).c_str(),
+                    [benchmark_case](
```
gumb0: I think you could move into lambda like
```
[case = std::move(benchmark_case)](benchmark::State& state) { benchmark_execute(state, case); }
```
And then reset it right after `RegisterBenchmark`
```
benchmark_case = {};
```

test/bench/bench.cpp
```diff
@@ -0,0 +1,243 @@
...
+            else if (result.stack[0] != *benchmark_case.expected_result)
+                return state.SkipWithError("Incorrect result");
+        }
+        else if (!result.stack.empty())
+            return state.SkipWithError("Unexpected result");
+
+        if (instance.memory.size() < benchmark_case.expected_memory.size())
+            return state.SkipWithError("Result memory is shorted than expected");
+
+        if (!std::equal(std::begin(benchmark_case.expected_memory),
```
gumb0: Maybe add a comment that we check only the starting segment of memory

test/bench/bench.cpp
```diff
@@ -0,0 +1,245 @@
...
+    std::optional<uint64_t> expected_result;
+    fizzy::bytes expected_memory;
+};
+
+void benchmark_execute(benchmark::State& state, const ExecutionBenchmarkCase& benchmark_case)
+{
+    const auto module = fizzy::parse(*benchmark_case.wasm_binary);
+    const auto func_idx = fizzy::find_exported_function(module, benchmark_case.func_name);
+    if (!func_idx)
+        return state.SkipWithError(
```
gumb0: braces {}

chfast:
The example output:
```
-------------------------------------------------------------------------------------
Benchmark                              Time           CPU Iterations UserCounters...
-------------------------------------------------------------------------------------
parse/mul256_opt0                 179953 ns     179895 ns       3555 rate=4.65826M/s size=838
instantiate/mul256_opt0            63547 ns      63542 ns      12140
execute/mul256_opt0/input0       1774267 ns    1774175 ns        385
execute/mul256_opt0/input1       1753808 ns    1753662 ns        405
parse/micro/spinner                37503 ns      37491 ns      19376 rate=1.70706M/s size=64
instantiate/micro/spinner            697 ns        697 ns    1019623
execute/micro/spinner/1             3458 ns       3458 ns     207499
execute/micro/spinner/1000        691329 ns     691331 ns        973
execute/micro/spinner/1000000  767400200 ns  767401651 ns          1
```

The naming convention is (except engine name is omitted):
```
engine_name/benchmark_kind/case_name/input_name
```

I can change the order is there are any suggestions.

We might also want to "normalize" `benchmark_kind` to have the same number of characters for readability. How about
- `parse`
- `execu`
- `insta`
?

PR #75, opened 2020-01-21
Add parser for table section

axic:


test/unittests/parser_test.cpp
```diff
@@ -257,6 +257,48 @@ TEST(parser, function_section_with_multiple_functions)
     EXPECT_EQ(module.funcsec[3], 0xff);
 }
 
+TEST(parser, table_single_min_limit)
+{
+    const auto section_contents =
+        bytes{} + uint8_t{0x01} + uint8_t{0x70} + uint8_t{0x00} + uint8_t{0x7f};
```
axic: The file uses `uint8_t{}` allover the place, I suggest it is fixed in a follow up PR. Currently these tests mirror the memory section tests (as they are very similar).

lib/fizzy/parser.cpp
```diff
@@ -259,6 +277,9 @@ Module parse(bytes_view input)
         if (it != expected_end_pos)
             throw parser_error{"incorrect section " + std::to_string(static_cast<int>(id)) +
                                " size, difference: " + std::to_string(it - expected_end_pos)};
+
+        if (module.tablesec.size() > 1)
```
axic: I have another set of changes which updates the parser for the memory section, etc.

test/unittests/parser_test.cpp
```suggestion
        bytes{0x01, 0x70, 0x00, 0x7f};
```
Not sure what do you want to fix here exactly.

test/unittests/parser_test.cpp
```diff
@@ -257,6 +257,48 @@ TEST(parser, function_section_with_multiple_functions)
...
+    const auto bin = bytes{wasm_prefix} + make_section(4, section_contents);
```
```suggestion
    const auto bin = wasm_prefix + make_section(4, section_contents);
```

test/unittests/parser_test.cpp
axic: I want to replace the entire style in one commit. Similar changes apply to half of the file.

lib/fizzy/types.hpp
```diff
@@ -22,6 +22,8 @@ enum class ValType : uint8_t
 /// https://webassembly.github.io/spec/core/binary/types.html#result-types
 constexpr uint8_t BlockTypeEmpty = 0x40;
 
+constexpr uint8_t FuncRef = 0x70;
```
gumb0: Maybe add link to https://webassembly.github.io/spec/core/binary/types.html#table-types

test/unittests/parser_test.cpp
axic: In #80.

PR #76, opened 2020-01-21
Add check into CMake for bigendian platforms

axic:
And reject compilation if so.

I know this was for a later milestone, but wanted to get this clear.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,6 +19,12 @@ cable_set_build_type(DEFAULT Release CONFIGURATION_TYPES Debug;Release;Coverage)
 
 project(fizzy LANGUAGES CXX)
 
+include(TestBigEndian)
+test_big_endian(is_big_endian)
+if(is_big_endian)
+    message(FATAL_ERROR "${PROJECT_NAME} currently does not support big endian systems.")
+endif()
+
 cable_configure_compiler()
 
 set(include_dir ${PROJECT_SOURCE_DIR}/include)  # Public include directory.


CMakeLists.txt
```diff
@@ -19,6 +19,12 @@ cable_set_build_type(DEFAULT Release CONFIGURATION_TYPES Debug;Release;Coverage)
 
 project(fizzy LANGUAGES CXX)
 
+include(TestBigEndian)
+TEST_BIG_ENDIAN(IS_BIG_ENDIAN)
```
chfast:
```suggestion
test_big_endian(is_big_endian)
```

CMakeLists.txt
```diff
@@ -19,6 +19,12 @@ cable_set_build_type(DEFAULT Release CONFIGURATION_TYPES Debug;Release;Coverage)
...
+if (IS_BIG_ENDIAN)
```
chfast:
```suggestion
if(is_big_endian)
```

CMakeLists.txt
axic: This is a copy paste from Solidity and it works (checked with `NOT IS_BIG_ENDIAN`).

What is the difference between lower and upper case?

chfast: No functional difference, just naming convention. CMake in general started using lower case for function names. Moreover, I prefer using lower case private variables.

axic: So are cmake variables/function names case insensitive or `TestBigEndian` exports both `test_big_endian` and `TEST_BIG_ENDIAN`?

chfast: It is case insensitive but only for function/macro names to my understanding. Variable names are case sensitive.

axic: Thanks for the clarification. Updated both.

CMakeLists.txt
```diff
@@ -19,6 +19,12 @@ cable_set_build_type(DEFAULT Release CONFIGURATION_TYPES Debug;Release;Coverage)
 
 project(fizzy LANGUAGES CXX)
 
+include(TestBigEndian)
+test_big_endian(is_big_endian)
+if (is_big_endian)
```
chfast:
```suggestion
if(is_big_endian)
```

axic: Changed. I always wondered why there isn't a space there

CMakeLists.txt
```diff
@@ -19,6 +19,12 @@ cable_set_build_type(DEFAULT Release CONFIGURATION_TYPES Debug;Release;Coverage)
 
 project(fizzy LANGUAGES CXX)
 
+include(TestBigEndian)
+test_big_endian(is_big_endian)
+if(is_big_endian)
+        message(FATAL_ERROR "${PROJECT_NAME} currently does not support big endian systems.")
```
chfast:
```suggestion
    message(FATAL_ERROR "${PROJECT_NAME} currently does not support big endian systems.")
```

CMakeLists.txt
chfast: Because
```
if ()
else()
endif()
```

CMakeLists.txt
axic: Can you add a cmake-format for cmakefiles? :)

chfast: It actually exists: https://github.com/cheshirekow/cmake_format
But my editor is good enough.

PR #77, opened 2020-01-21
Implement imported globals passed by pointer

gumb0:
Some relevant links

- Import matching described in the spec https://webassembly.github.io/spec/core/exec/modules.html#import-matching
- How instantiating and importing globals looks from the JS side https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/Global#Examples

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -18,6 +18,71 @@ struct LabelContext
 };
 
 
+void match_imported_functions(const std::vector<TypeIdx>& module_imported_types,
+    const std::vector<ImportedFunction>& imported_functions)
+{
+    if (module_imported_types.size() != imported_functions.size())
+    {
+        throw std::runtime_error("Module requires " + std::to_string(module_imported_types.size()) +
+                                 " imported functions, " +
+                                 std::to_string(imported_functions.size()) + " provided");
+    }
+}
+
+void match_imported_globals(const std::vector<bool>& module_imports_mutability,
+    const std::vector<ImportedGlobal>& imported_globals)
+{
+    if (module_imports_mutability.size() != imported_globals.size())
+    {
+        throw std::runtime_error(
+            "Module requires " + std::to_string(module_imports_mutability.size()) +
+            " imported globals, " + std::to_string(imported_globals.size()) + " provided");
+    }
+
+    for (size_t i = 0; i < imported_globals.size(); ++i)
+    {
+        if (imported_globals[i].is_mutable != module_imports_mutability[i])
+        {
+            throw std::runtime_error("Global " + std::to_string(i) +
+                                     " mutability doesn't match module's global mutability");
+        }
+        if (imported_globals[i].value == nullptr)
+        {
+            throw std::runtime_error(
+                "Global " + std::to_string(i) + " has a null pointer to value");
+        }
+    }
+}
+
+std::vector<TypeIdx> match_imports(const Module& module,
+    const std::vector<ImportedFunction>& imported_functions,
+    const std::vector<ImportedGlobal>& imported_globals)
+{
+    std::vector<TypeIdx> imported_function_types;
+    std::vector<bool> imported_globals_mutability;
+    for (auto const& import : module.importsec)
+    {
+        switch (import.kind)
+        {
+        case ExternalKind::Function:
+            imported_function_types.emplace_back(import.desc.function_type_index);
+            break;
+        case ExternalKind::Global:
+            imported_globals_mutability.emplace_back(import.desc.global_mutable);
+            break;
+        default:
+            throw std::runtime_error("Import of type " +
+                                     std::to_string(static_cast<uint8_t>(import.kind)) +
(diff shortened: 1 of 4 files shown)


lib/fizzy/execute.cpp
```diff
@@ -141,7 +198,8 @@ inline uint64_t popcnt64(uint64_t value) noexcept
 }
 }  // namespace
 
-Instance instantiate(const Module& module, std::vector<ImportedFunction> imported_functions)
+Instance instantiate(const Module& module, std::vector<ImportedFunction> imported_functions,
+    std::vector<ImportedGlobal> imported_globals)
```
```suggestion
    std::vector<ImportedGlobal> imported_globals = {})
```

gumb0: Yeah I didn't do it on purpose, because I'm not a fan of default arguments in general, because they implicitly pass some values... But I agree here it gets very annoying in tests. But I should first add default to `imported_functions`, too, in a separate PR.

lib/fizzy/execute.cpp
```diff
@@ -22,6 +22,63 @@ struct LabelContext
...
+                "Global " + std::to_string(i) + " has a null pointer to value");
+    }
+}
+
+std::vector<TypeIdx> matchImports(const Module& module,
+    const std::vector<ImportedFunction>& imported_functions,
+    const std::vector<ImportedGlobal>& imported_globals)
+{
+    std::vector<TypeIdx> imported_function_types;
+    std::vector<bool> imported_globals_mutability;
```
gumb0: Using of `vector<bool>` is sometimes discouraged because of its `operator[]` not returning a value in a contiguous segment of memory like for regular vector... But here it's a simple usage, I think it's acceptable.

gumb0:
Rebased.

lib/fizzy/execute.cpp
```diff
@@ -18,6 +18,63 @@ struct LabelContext
 };
 
 
+void match_imported_functions(const std::vector<TypeIdx>& module_imported_types,
+    const std::vector<ImportedFunction>& imported_functions)
+{
+    if (module_imported_types.size() != imported_functions.size())
+        throw std::runtime_error("Module requires " + std::to_string(module_imported_types.size()) +
```
chfast: This expression (and some similar ones later) should be rather wrapped with `{}`.

test/unittests/instantiate_test.cpp
```diff
@@ -24,6 +24,75 @@ TEST(instantiate, imported_functions)
...
+}
+
+TEST(instantiate, imported_globals_nullptr)
+{
+    Module module;
+    module.importsec.emplace_back(Import{"mod", "g1", ExternalKind::Global, {true}});
+    module.importsec.emplace_back(Import{"mod", "g2", ExternalKind::Global, {false}});
+
+    ImportedGlobal g{nullptr, true};
+    EXPECT_THROW(instantiate(module, {}, {g}), std::runtime_error);
```
chfast: I believe this fails because of the number of globals mismatch (https://github.com/wasmx/fizzy/pull/77/files#diff-da8232128fc88a65268ab30bf8e8abdcR33), not because of the `nullptr`.

Unfortunately, GTest do not provide a macro for testing exception message, but it might be good time to add one ourselves. The https://github.com/wasmx/fizzy/pull/77/files#diff-da8232128fc88a65268ab30bf8e8abdcR33 provides some possible implementations of this.

gumb0: Yeah, good catch. In this PR I would just pass the right number of globals here.

(you mistyped the second link)

Also not sure if it makes sense to test the exception messages at all, but if/when we add the error codes, it certainly makes sense to test those.

gumb0:
Fixed the issues, rebased and also fixed the order of tests in instantiate_test.cpp.

lib/fizzy/execute.cpp
```diff
@@ -18,6 +18,71 @@ struct LabelContext
...
+    const std::vector<ImportedFunction>& imported_functions,
+    const std::vector<ImportedGlobal>& imported_globals)
+{
+    std::vector<TypeIdx> imported_function_types;
+    std::vector<bool> imported_globals_mutability;
+    for (auto const& import : module.importsec)
+    {
+        switch (import.kind)
+        {
+        case ExternalKind::Function:
```
axic: Why are these supported in this PR?

gumb0: This was moved from instantiate to a separate function https://github.com/wasmx/fizzy/pull/77/files#diff-da8232128fc88a65268ab30bf8e8abdcL203

axic: I'm not sure that link clears it up for me, it seems to link to the `global_set` case?

gumb0: It links to the line 203 of the old version

test/unittests/instantiate_test.cpp
gumb0: https://github.com/wasmx/fizzy/pull/94

PR #80, opened 2020-01-22
Some parser test cleanups

axic:


test/unittests/parser_test.cpp
```diff
@@ -259,7 +270,7 @@ TEST(parser, function_section_with_multiple_functions)
 
 TEST(parser, memory_single_min_limit)
 {
-    const auto section_contents = bytes{} + uint8_t{0x01} + uint8_t{0x00} + uint8_t{0x7f};
+    const auto section_contents = bytes{} + bytes{0x01, 0x00, 0x7f};
```
```suggestion
    const auto section_contents = bytes{0x01, 0x00, 0x7f};
```

test/unittests/parser_test.cpp
```diff
@@ -119,31 +130,31 @@ TEST(parser, custom_section_nonempty)
 
 TEST(parser, functype_wrong_prefix)
 {
-    const auto section_contents = uint8_t{0x01} + from_hex("610000");
+    const auto section_contents = bytes{0x01} + from_hex("610000");
```
axic: Do we really need `from_hex` if we use `bytes{0x00, 0x0b}` or do we prefer to use `from_hex` everywhere instead?

test/unittests/parser_test.cpp
axic: I was sure I didn't miss any.

test/unittests/parser_test.cpp
chfast: I don't mind mixing them. Hex are easier to read and work with.

If needed we can also add literals for `bytes`, e.g. `"610000"_hex`.

axic: Perhaps the literals would be the best for the tests?

Also thinking about introducing `make_vec` that could help in the above case.

test/unittests/parser_test.cpp
```diff
@@ -390,7 +398,7 @@ TEST(parser, export_multiple)
 
 TEST(parser, start)
 {
-    const auto section_contents = bytes{} + uint8_t{0x07};
+    const auto section_contents = bytes{0x07};
```
chfast: You can shorten many cases like this:

```suggestion
    const auto bin = wasm_prefix + make_section(8, {0x07});
```

test/unittests/parser_test.cpp
```diff
@@ -17,6 +17,15 @@ bytes prefix_with_size(const bytes& content)
     return bytes{static_cast<uint8_t>(content.size())} + content;
 }
 
+bytes make_vec(std::vector<bytes_view> contents)
```
```suggestion
bytes make_vec(std::initializer_list<bytes_view> contents)
```

chfast: As most of the sections are vectors, you can add overload `make_section(uint8_t id, std::initializer_list<bytes_view> contents)`.

axic: Add or replace?

chfast: If `make_vec` is not used in any other context, I would remove it.

test/unittests/parser_test.cpp
```diff
@@ -11,10 +11,30 @@ static const auto functype_i32_to_void = from_hex("60017f00");
...
+        ret.append(content);
+    return ret;
+}
+
+bytes make_section(uint8_t id, const bytes& content)
+{
+    return bytes{id} + prefix_with_size(content);
+}
+
+bytes make_invalid_section(uint8_t id, size_t size, const bytes& content)
```
gumb0: maybe
```suggestion
bytes make_invalid_size_section(uint8_t id, size_t size, const bytes& content)
```

test/unittests/parser_test.cpp
```diff
@@ -11,10 +11,30 @@ static const auto functype_i32_to_void = from_hex("60017f00");
 
 namespace
 {
-bytes make_section(uint8_t id, const bytes& content)
+bytes prefix_with_size(const bytes& content)
 {
     assert(content.size() < 0x80);
-    return bytes{id, static_cast<uint8_t>(content.size())} + content;
+    return bytes{static_cast<uint8_t>(content.size())} + content;
```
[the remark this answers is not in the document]
axic: @chfast wanted to specifically go against using `uint8_t`  with the plus operator in these cases.

test/unittests/parser_test.cpp
```diff
@@ -11,10 +11,30 @@ static const auto functype_i32_to_void = from_hex("60017f00");
 
 namespace
 {
-bytes make_section(uint8_t id, const bytes& content)
+bytes prefix_with_size(const bytes& content)
```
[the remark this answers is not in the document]
axic: I don't have a strong preference on the naming, @chast?

test/unittests/parser_test.cpp
axic: Same here, no strong preference myself. @chfast ?

test/unittests/parser_test.cpp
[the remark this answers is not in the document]
chfast: Although is probably correct, I also had issues with reading "prefix_with_size". I'd go with `add_size_prefix` or even `size_prefix`.

test/unittests/parser_test.cpp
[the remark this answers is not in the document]
chfast: I mostly wanted avoid having `uint8_t + uint8_t` producing bytes. Here is fine, because not this case and inside a helper. But the previous version is fine as well.

test/unittests/parser_test.cpp
chfast: I'm strongly in favor of `make_vec` because this is how the wasm spec reference it (i.e. we mean wasm `vec` not C++ `std::vector`).

In the other matter, I'm against adding anything that might be used some day. If it cannot be used now, I prefer not to adding it.

test/unittests/parser_test.cpp
chfast: I line suggested name better. But considering number of uses, I'm not convinced it is worth for add this helper in general.

axic: I think we'll need to add more tests with invalid sizes to ensure the parser is good, so it will be needed.

test/unittests/parser_test.cpp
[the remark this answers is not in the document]
axic: Changed.

test/unittests/parser_test.cpp
axic: Changed.

test/unittests/parser_test.cpp
axic: Changed from `vector` to `initializer_list`. Kept the name as it reflects the spec.

test/unittests/parser_test.cpp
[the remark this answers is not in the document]
axic: I'll just keep it as is for now.

test/unittests/parser_test.cpp
```diff
@@ -119,31 +139,35 @@ TEST(parser, custom_section_nonempty)
...
     const auto bin = bytes{wasm_prefix} + make_section(1, section_contents);
     EXPECT_THROW(parse(bin), parser_error);
 }
 
 TEST(parser, type_section_larger_than_expected)
 {
-    const auto section_contents = uint8_t{0x01} + functype_void_to_void;
-    const auto bin = bytes{wasm_prefix} + uint8_t{0x01} + uint8_t(section_contents.size() - 1) +
-                     section_contents;
+    const auto section_contents = "01"_bytes + functype_void_to_void;
```
axic: There are still a bunch of cases where the `"01"_bytes` is the vector size and we could just use `make_vec`. Is that our preference? Lets sort it in this PR.

gumb0: Yeah I think it would be good.

chfast: I don't think refactoring existing test brings much added value. It's only worth if you need to modify it anyway or it fails and you don't know why. Otherwise, you will not want to read a test again.

gumb0: Ok I'm fine with not doing it in this PR.

gumb0: (there's also a school of thought that teaches to "read the tests when you need to understand what the code does")

test/unittests/parser_test.cpp
```diff
@@ -119,31 +139,35 @@ TEST(parser, custom_section_nonempty)
...
+    const auto bin =
+        bytes{wasm_prefix} +
```
```suggestion
        wasm_prefix +
```

test/unittests/parser_test.cpp
```diff
@@ -119,31 +139,35 @@ TEST(parser, custom_section_nonempty)
...
 }
 
 TEST(parser, type_section_smaller_than_expected)
 {
-    const auto section_contents = uint8_t{0x01} + functype_void_to_void + uint8_t{0xfe};
-    const auto bin = bytes{wasm_prefix} + uint8_t{0x01} + uint8_t(section_contents.size() + 1) +
-                     section_contents;
+    const auto section_contents = "01"_bytes + functype_void_to_void + "fe"_bytes;
+    const auto bin =
+        bytes{wasm_prefix} +
```
```suggestion
        wasm_prefix +
```

axic:
@gumb0 @chfast I'm giving up on this would like to merge as is :)

PR #81, opened 2020-01-22
Hex literals for tests

chfast:


test/unittests/parser_test.cpp
```diff
@@ -446,9 +446,9 @@ TEST(parser, code_section_with_2_trivial_codes)
 
 TEST(parser, code_section_with_basic_instructions)
 {
-    const auto func_bin = from_hex(
+    const auto func_bin =
         "00"  // vec(locals)
-        "2001210222036a01000b");
+        "2001210222036a01000b"_b;
```
chfast: For disjoint literals only one need to be with `_b`.

test/utils/hex.hpp
```diff
@@ -28,3 +28,17 @@ inline std::string hex(bytes_view data)
...
+/// Operator for 0x integer literals, e.g. 0x0a0b0c0d_b.
+inline fizzy::bytes operator""_b(const char* literal)
+{
+    if (literal[0] != '0' || literal[1] != 'x')
+        throw std::invalid_argument{"Only 0x literals are supported"};
+
+    return fizzy::from_hex(&literal[2]);
+}
+
+inline fizzy::bytes operator""_b(const char* literal, size_t /*length*/)
```
chfast: This is for `""` literals. It might be confusing that hex digits are expected as the content, so this might be renamed to `_hb` or something else.

axic: I suggest we go for a single form only. And perhaps the `_hex` postfix is nicer?

test/utils/hex.hpp
```diff
@@ -28,3 +28,9 @@ inline std::string hex(bytes_view data)
     return hex(data.data(), data.size());
 }
 }  // namespace fizzy
+
+/// Operator for "" literals, e.g. "0a0b0c0d"_hex.
+inline fizzy::bytes operator""_hex(const char* literal, size_t /*length*/)
```
gumb0: Can't it be defined inside `namespace fizzy` ?

chfast: Can be, but then `using namespace fizzy` is required.

chfast:
Names options are:
- `_b` - shortest possible, maps Pythons `b""` literals, suggests bytes, ambiguous.
- `_hex` - comes from hex utils, suggests it encodes to hex instead of decoding from.
- `_bytes` - hides the fact that hex encoded literal is expected (constexpr would help here, but it's not worth it)
- `_hexbytes` - the most correct, but not much shorter than `from_hex()`.

test/utils/hex.hpp
gumb0: I think it's good, it's similar to like standard custom literals are in `namespace std` (and even in the nested inline namespaces). And we do `using namespace fizzy` in tests anyway.

gumb0:
I'm still in favor of `_bytes`.
Would accept `_hexbytes` but it's damn long.

axic:
I'm against `_b`. Fine with the other three, but prefer `_hex` or `_hexbytes`, unless it is constexpr and get a compile time failure if it not is valid, then `_bytes` is good.

test/unittests/parser_test.cpp
```diff
@@ -630,18 +631,18 @@ TEST(parser, block_br)
...
-    EXPECT_EQ(hex(code.immediates), hex(from_hex("00"
-                                                 "08000000"
-                                                 "1d000000"
-                                                 "0a000000"
-                                                 "01000000"
-                                                 "00000000"
-                                                 "0b000000"
-                                                 "01000000"
-                                                 "01000000")));
+    EXPECT_EQ(hex(code.immediates), hex("00"
```
chfast: This one id weird :)

test/unittests/parser_test.cpp
```diff
@@ -567,39 +567,40 @@ TEST(parser, milestone1)
     EXPECT_EQ(c.instructions,
         (std::vector{Instr::local_get, Instr::local_get, Instr::i32_add, Instr::local_get,
             Instr::i32_add, Instr::local_tee, Instr::local_get, Instr::i32_add, Instr::end}));
-    EXPECT_EQ(c.immediates, from_hex("00000000"
-                                     "01000000"
-                                     "02000000"
-                                     "02000000"
-                                     "00000000"));
+    EXPECT_EQ(c.immediates,
```
axic: In `block_br` you prefer the other way around:
```
EXPECT_EQ(hex(c.immediates), "00000000"...);
```

chfast: The one in `block_br` is easier to debug because you see hex in the test failure output. But is probably less readable in the source code. I can convert these back to the compact form.

PR #83, opened 2020-01-22
Fix load/store instructions not updating immediates pointer

chfast:


chfast:
This requires regression tests

PR #84, opened 2020-01-22
Implement imported memory

gumb0:
Some relevant links
- Matching memory imports described in spec https://webassembly.github.io/spec/core/exec/modules.html#memories
https://webassembly.github.io/spec/core/exec/modules.html#limits
-  how memory importing looks from the JS side https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/Memory#Examples
- more about memories from JS point of view (note the last paragraph about imported memory usage) https://webassembly.org/getting-started/js-api/#memory

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -28,6 +28,59 @@ void match_imported_functions(const std::vector<TypeIdx>& module_imported_types,
     }
 }
 
+
+void match_limits(const Limits& external_limits, const Limits& module_limits)
+{
+    if (external_limits.min < module_limits.min)
+    {
+        throw instantiate_error(
+            "Provided memory's min is below imported memory min defined in module.");
+    }
+
+    if (!module_limits.max.has_value())
+        return;
+
+    if (external_limits.max.has_value() && *external_limits.max <= *module_limits.max)
+        return;
+
+    throw instantiate_error(
+        "Provided memory's max is above imported memory max defined in module.");
+}
+
+void match_imported_memories(const std::vector<Memory>& module_imported_memories,
+    const std::vector<ImportedMemory>& imported_memories)
+{
+    assert(module_imported_memories.size() <= 1);
+
+    if (imported_memories.size() > 1)
+        throw instantiate_error("Only 1 imported memory is allowed.");
+
+    if (module_imported_memories.empty())
+    {
+        if (!imported_memories.empty())
+        {
+            throw instantiate_error(
+                "Trying to provide imported memory to a module that doesn't define one.");
+        }
+    }
+    else
+    {
+        if (imported_memories.empty())
+            throw instantiate_error("Module defines an imported memory but none was provided.");
+
+        match_limits(imported_memories[0].limits, module_imported_memories[0].limits);
+
+        if (imported_memories[0].data == nullptr)
+            throw instantiate_error("Provided imported memory has a null pointer to data.");
+
+        const auto size = imported_memories[0].data->size();
+        const auto min = imported_memories[0].limits.min;
+        const auto& max = imported_memories[0].limits.max;
+        if (size < min * PageSize || (max.has_value() && size > *max * PageSize))
+            throw instantiate_error("Provided imported memory doesn't fit provided limits");
+    }
+}
+
 void match_imported_globals(const std::vector<bool>& module_imports_mutability,
     const std::vector<ImportedGlobal>& imported_globals)
 {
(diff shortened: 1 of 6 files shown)


lib/fizzy/execute.hpp
```diff
@@ -19,17 +20,28 @@ struct Instance;
...
+
 // The module instance.
 struct Instance
 {
     const Module& module;
-    bytes memory;
+    // Memory is either allocated and owned by the instance or imported as already allocated bytes
+    // and owned externally.
+    // For these cases unique_ptr would either have a normal deleter or noop deleter respectively
+    bytes_ptr memory;
```
gumb0: I guess when we're going to have a real external API, this will need to be replaced with something more complicated like an interface allowing to access and grow external memory.

gumb0:
Rebsased.

lib/fizzy/execute.cpp
```diff
@@ -376,6 +455,7 @@ execution_result execute(Instance& instance, FuncIdx func_idx, std::vector<uint6
     assert(code_idx < instance.module.codesec.size());
 
     const auto& code = instance.module.codesec[code_idx];
+    auto& memory = *instance.memory;
```
axic: How about making a commit/PR for this refactoring?

gumb0: Extracted it to a separate first commit here.

gumb0:
clang-tidy complains about potential memory leak around moving of `unique_ptr` with custom deleter.

I double checked that deleter is moved by move constructor, too and the memory is indeed deleted, so I think it's a false positive. I haven't found other way to workaround it other that NOLINT

lib/fizzy/execute.cpp
```diff
@@ -368,9 +447,10 @@ Instance instantiate(const Module& module, std::vector<ImportedFunction> importe
 
         // NOTE: these instructions can overlap
         assert((offset + data.init.size()) <= (memory_max * PageSize));
-        std::memcpy(memory.data() + offset, data.init.data(), data.init.size());
+        std::memcpy(memory->data() + offset, data.init.data(), data.init.size());
     }
 
+    // NOLINTNEXTLINE(clang-analyzer-cplusplus.NewDeleteLeaks)
```
gumb0: I could format it in a way to have `NOLINT` only on a single move, but weirdly the warning itself points to moving of a `table` (while the allocation in question is the one of `memory`)
So this would help
```
    Instance instance = {module, std::move(memory), memory_max,
        std::move(table),  // NOLINT(clang-analyzer-cplusplus.NewDeleteLeaks)
        std::move(globals), std::move(imported_functions), std::move(imported_function_types),
        std::move(imported_globals)};

```
but not this:
```
    Instance instance = {module,
        std::move(memory),  // NOLINT(clang-analyzer-cplusplus.NewDeleteLeaks)
        memory_max, std::move(table), std::move(globals), std::move(imported_functions),
        std::move(imported_function_types), std::move(imported_globals)};
```

chfast: Is this the same bug? https://bugs.llvm.org/show_bug.cgi?id=32234

gumb0:
Alternative approaches could be to get rid of weird `unique_ptr` and try instead `std::variant` or union of local memory/imported memory

(BTW the same mechanism will be needed for a table, which is similarly can be either local or imported)

lib/fizzy/execute.cpp
gumb0: I don't know, probably it's not. I can't really reproduce it outside of this our code.

chfast: How it is done right now is fine, but add a FIXME comment to investigate later (I'm willing to take care of it).

gumb0:
Added FIXME comment and rebased.

gumb0:
Weirdly no such warning in https://github.com/wasmx/fizzy/pull/109, while it's very similar.

gumb0:
Rebased.

gumb0:
Now clang-tidy warning disappeared (after changing `Module` reference to value?), I removed `NOLINT`.

lib/fizzy/execute.cpp
```diff
@@ -77,11 +131,61 @@ std::vector<TypeIdx> match_imports(const Module& module,
...
+        if ((memory_min > MemoryPagesLimit) || (memory_max > MemoryPagesLimit))
+            throw std::runtime_error("Imported memory limits cannot exceed hard memory limit of " +
+                                     std::to_string(MemoryPagesLimit * PageSize) + " bytes.");
+
+        bytes_ptr memory{imported_memories[0].data, null_delete};
+        return {std::move(memory), memory_max};
+    }
+    else
+    {
+        bytes_ptr memory{new bytes{}, bytes_delete};
```
gumb0: Apparently this is not correct - no memory should be allocated if none is defined or imported. WebAssembly Studio issues in this case an error like `i32.load requires an imported or defined memory.`
See also for example this test https://github.com/WebAssembly/spec/blob/3b2489d7e570ca0f8d32e9d0b1a1e8255c4212b7/test/core/memory.wast#L22-L25

I would leave it like this here and fix it in another PR.

gumb0:
Rebased, changed `std:;runtime_error`s thrown during instantiation to `instantiate_error`.

gumb0:
Rebased.

gumb0:
Clang-tidy warning is back! :confounded:

gumb0:
Added `NOLINT` back.

gumb0:
Rebased and added default-initializer to `Instance::memory` (required to default-construct `Instance` in `FizzyEngine`)

PR #86, opened 2020-01-23
cmake: Prefer CONFIG mode for find_package

chfast:
Docs: https://cmake.org/cmake/help/v3.16/variable/CMAKE_FIND_PACKAGE_PREFER_CONFIG.html

This makes the CONFIG mode of `find_package()` being the first to try if not explicitly specified. Without setting the variable, the MODULE mode is the first to try.

The CONFIG mode should be always preferable in "modern" CMake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,7 @@
 cmake_minimum_required(VERSION 3.15)
 
 cmake_policy(SET CMP0077 NEW)
+set(CMAKE_FIND_PACKAGE_PREFER_CONFIG TRUE)
 
 include(cmake/cable.cmake)
 include(CableBuildType)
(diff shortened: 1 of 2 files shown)


axic:
The documentation seems to huge, but what is the difference between config and module mode?

chfast:
The module mode loads `Find<PACKAGE_NAME>.cmake` module which contains a script that finds the dependency for you. CMake provides some of these for most common libraries, e.g. Boost, GTest, etc. You can also write your own.

The config mode looks for `<PACKAGE_NAME>Config.cmake` file. The difference is the Config files come from and are maintained by the package itself (i.e. are created by the dependency library). Many CMake-based project export these, e.g. LLVM, GTest (although not included in Ubuntu/Debian packages), EVMC :)

Issue #63, opened 2020-01-15
Parser and validation design


chfast:
## Levels of parsing / decoding / validation

0. "Unsafe" parser.
   It assumes the wasm module must be valid and happily reads it without checkout out-of-buffer access. Providing invalid module can crash the parser.
1. "???" parser.
   It assumes the wasm module must be valid but have additional check preventing invalid memory access. Providing invalid module will produce invalid but deterministic execution results. The parser never crashes. Hopefully, this one is not much slower than the "unsafe" one.
2. Validator.
   It parses the wasm module and also fully validates it.

What we currently have is mixture of 0 and 2 with partial validation.

## Design

1. wasm spec allows lazy validation of functions (executing invalid function traps). So the good option is to implement the parser for wasm this way - functions are validated before being called for the first time.
2. Validation can be a template param for Parser type (separating code for reader/parser and validator seems impractical).

```cpp
#include <cstdint>

enum class ValType : uint8_t
{
    i32 = 0x7f,
    i64 = 0x7e,
};

template <bool Validate>
struct Parser
{
    ValType parseValType(uint8_t b)
    {
        if constexpr (Validate)
        {
            if (b != 0x7f && b != 0x7e)
                throw "invalid type";
        }

        return (b == 0x7e) ? ValType::i64 : ValType::i32;
    }
};
```

axic:
The algorithm seems to validate the code sections, but nothing else really.

axic:
@chfast @gumb0 can we close this?

chfast:
Some of the parts are not implemented.

PR #87, opened 2020-01-23
Validate number of imported memories and tables in parser

gumb0:
Depends on parsing imported tables https://github.com/wasmx/fizzy/pull/112

### DIFF
--- a/lib/fizzy/parser.cpp
+++ b/lib/fizzy/parser.cpp
@@ -298,18 +298,45 @@ Module parse(bytes_view input)
         }
 
         if (it != expected_end_pos)
+        {
             throw parser_error{"incorrect section " + std::to_string(static_cast<int>(id)) +
                                " size, difference: " + std::to_string(it - expected_end_pos)};
-
-        if (module.tablesec.size() > 1)
-            throw parser_error{"too many table sections (at most one is allowed)"};
-
-        if (module.memorysec.size() > 1)
-            throw parser_error{"too many memory sections (at most one is allowed)"};
-
-        if (!module.elementsec.empty() && module.tablesec.empty())
-            throw parser_error("element section encountered without a table section");
+        }
     }
+
+    // Validation checks
+    if (module.tablesec.size() > 1)
+        throw parser_error{"too many table sections (at most one is allowed)"};
+
+    if (module.memorysec.size() > 1)
+        throw parser_error{"too many memory sections (at most one is allowed)"};
+
+    const auto imported_mem_count = std::count_if(module.importsec.begin(), module.importsec.end(),
+        [](const auto& import) noexcept { return import.kind == ExternalKind::Memory; });
+
+    if (imported_mem_count > 1)
+        throw parser_error{"too many imported memories (at most one is allowed)"};
+
+    if (!module.memorysec.empty() && imported_mem_count > 0)
+    {
+        throw parser_error{
+            "both module memory and imported memory are defined (at most one of them is allowed)"};
+    }
+
+    const auto imported_tbl_count = std::count_if(module.importsec.begin(), module.importsec.end(),
+        [](const auto& import) noexcept { return import.kind == ExternalKind::Table; });
+
+    if (imported_tbl_count > 1)
+        throw parser_error{"too many imported tables (at most one is allowed)"};
+
+    if (!module.tablesec.empty() && imported_tbl_count > 0)
+    {
+        throw parser_error{
+            "both module table and imported table are defined (at most one of them is allowed)"};
+    }
+
+    if (!module.elementsec.empty() && module.tablesec.empty() && imported_tbl_count == 0)
+        throw parser_error("element section encountered without a table section");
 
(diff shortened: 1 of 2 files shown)


lib/fizzy/parser.cpp
```diff
@@ -275,14 +276,41 @@ Module parse(bytes_view input)
         }
 
         if (it != expected_end_pos)
+        {
             throw parser_error{"incorrect section " + std::to_string(static_cast<int>(id)) +
                                " size, difference: " + std::to_string(it - expected_end_pos)};
+        }
+    }
+
+    if (module.tablesec.size() > 1)
```
gumb0: Bringing these two existing checks out of the loop, too.

lib/fizzy/parser.cpp
```diff
@@ -275,14 +276,41 @@ Module parse(bytes_view input)
...
+        [](const auto& import) noexcept { return import.kind == ExternalKind::Memory; });
 
-        if (module.tablesec.size() > 1)
-            throw parser_error{"too many table sections (at most one is allowed)"};
+    if (imported_memories > 1)
+        throw parser_error{"too many imported memories (at most one is allowed)"};
 
-        if (module.memorysec.size() > 1)
-            throw parser_error{"too many memory sections (at most one is allowed)"};
+    if (!module.memorysec.empty() && imported_memories > 0)
```
axic: Why not `module.memorysec.size() + imported_memories <= 1` since we ensure `memorysec.size() <= 1` ?

gumb0: Just becasue `empty()` presumably could be more efficient in some implementations than `size()`. But here probably it's not that important, perhaps your variant looks nicer.

gumb0: It would require a cast, so won't look particularly nice either
```suggestion
    if (module.memorysec.size() + static_cast<size_t>(imported_memories) > 1)
```

gumb0:
Actually the code here seems to differ from the snippet in #63 only in template parameter. I could try adding it, if it's decided to be the way to go.

chfast:
> Actually the code here seems to differ from the snippet in #63 only in template parameter. I could try adding it, if it's decided to be the way to go.

I think the good way forward for now is to mark validation checks with a comment.

test/unittests/parser_test.cpp
```diff
@@ -269,6 +269,29 @@ TEST(parser, import_multiple)
...
+    const auto bin = bytes{wasm_prefix} + make_section(2, section_contents);
+
+    EXPECT_THROW_MESSAGE(
+        parse(bin), parser_error, "too many imported memories (at most one is allowed)");
+}
+
+TEST(parser, memory_and_imported_memory)
+{
+    const auto bin =
+        "0061736d01000000"                                // wasm prefix
```
axic: Why not follow the same model as above? Could even just do
```
```
// (import "js" "mem" (memory 1))
const auto import_section = "020b01026a73036d656d0200010008046e616d65020100"_bytes;
const auto memory_section = "05030100010008046e616d65020100"_bytes; 
const auto bin = bytes{wasm_prefix} + import_section + memory_section;
```

This would remove the need for some comments. Perhaps not that much of an improvement.

File: lib/fizzy/parser.cpp (hunk 1 above)
Comment (axic) [reply lib/fizzy/parser.cpp]: I think they were in there because a strict order is expected and there would be safe to check and abort early. However we do not check the ordering (yet).

Comment (axic) [reply lib/fizzy/parser.cpp]: Would an early abort be beneficial btw?

File: lib/fizzy/parser.cpp [hunk 4]
```diff
@@ -298,19 +298,47 @@ Module parse(bytes_view input)
...
+    if (imported_mem_count > 1)
+        throw parser_error{"too many imported memories (at most one is allowed)"};
 
-        if (module.memorysec.size() > 1)
-            throw parser_error{"too many memory sections (at most one is allowed)"};
+    if (!module.memorysec.empty() && imported_mem_count > 0)
+    {
+        throw parser_error{
+            "both module memory and imported memory are defined (at most one of them is "
+            "allowed)"};
```
axic: Is this split due to clang-format?

lib/fizzy/parser.cpp
```diff
@@ -298,19 +298,47 @@ Module parse(bytes_view input)
...
+    }
+
+    const auto imported_tbl_count = std::count_if(module.importsec.begin(), module.importsec.end(),
```
axic: Do we prefer `tbl` here?

axic:
Apart from these comments, the logic and change itself looks good.

lib/fizzy/parser.cpp
gumb0: Might be benificial, but the way it was it's for the price of doing the checks after each section.

lib/fizzy/parser.cpp
gumb0: Actually looks like can fit into one line, will fix.

test/unittests/parser_test.cpp
gumb0: Changed.

PR #88, opened 2020-01-23
Small tests changes

chfast:


test/unittests/parser_test.cpp
```diff
@@ -63,7 +63,7 @@ TEST(parser, DISABLED_limits_min_invalid_too_short)
     EXPECT_THROW(parser<Limits>{}(input.data()), parser_error);
 }
 
-TEST(parser, DISABLED_limits_minmax_invalid_too_short)
+TEST(parser, limits_minmax_invalid_too_short)
```
axic: How is this detected now?

chfast: No idea.

axic: Is ASan running properly? Shouldn't that detect overreads?

chfast: Now it does.

test/unittests/parser_test.cpp
```diff
@@ -86,8 +86,8 @@ TEST(parser, DISABLED_limits_min_invalid_too_short)
 
 TEST(parser, DISABLED_limits_minmax_invalid_too_short)
 {
-    const auto input = "0120"_bytes;
-    EXPECT_THROW(parser<Limits>{}(input.data()), parser_error);
+    const uint8_t input[] = {0x01, 0x02};
```
axic: Does this also means that ASan is useless for every other test we have?

test/unittests/execute_control_test.cpp
```diff
@@ -178,6 +178,7 @@ TEST(execute_control, nested_blocks)
             i32.const 4
             i32.or
             set_local 0
+            i32.const 1
```
axic: What is the reason for this? To make the block arity 1 and support jumping to the func?

chfast: This is for `br 3` (aka `return`) to work, because the function result has type `[i32]`.

test/unittests/parser_test.cpp
chfast: It depends if the `std::string` implementation is properly instrumented.

test/unittests/execute_control_test.cpp
axic: Yeah, just wanted to confirm I understand it properly.

chfast:
Can we merge this?

I can investigate ASan + containers separately. Switching to macos for this build might help.
Also applying https://en.cppreference.com/w/cpp/string/basic_string/shrink_to_fit should also help, but this requires discipline.

axic:
I'm not happy with the `DISABLED_limits_minmax_invalid_too_short` change. We don't know if either of those options are detectable by Asan (because the CI doesn't trigger it due to it being disabled), but apparently the new version is detectable by Asan. However if Asan cannot detect it in the `bytes` case, then we need to change every test case to ensure Asan makes any sense.

I'd propose to merge this for the `br` test case improvement only.

PR #89, opened 2020-01-23
Fix UB in leb128s_decode

chfast:
For types smaller than `int`, there is a promotion to `int` for `~T_unsigned{0}` and the end result is `-1`. It is undefined behavior to shift left a negative value.


gumb0:
How is there no promotion to `int` for `all_ones` with the same result?

gumb0:
Ah it was a promotion before `~` ?

chfast:
> How is there no promotion to `int` for `all_ones` with the same result?

Before: `uint8(0) -> int(0) -> ~ -> int(-1)`
After: `uint8(0xff) -> int(0xff)`

PR #90, opened 2020-01-24
Check fizzy-bench with ASan

chfast:
- Fix fizzy-bench linking with stdc++fs — details explained in comments.
- Run all benchmarks with ASan to detect issues with the tool.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -58,6 +58,14 @@ commands:
         working_directory: ~/build
         command: ctest -R ${TESTS_FILTER:-'.*'} -j4 --schedule-random --output-on-failure
 
+  benchmark:
+    description: "Run benchmarks"
+    steps:
+      - run:
+          name: "Run benchmarks"
+          working_directory: ~/build
+          command: bin/fizzy-bench ~/project/test/benchmarks
+
 jobs:
 
   lint:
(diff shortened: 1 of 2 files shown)


circle.yml
```diff
@@ -58,6 +58,14 @@ commands:
         working_directory: ~/build
         command: ctest -R ${TESTS_FILTER:-'.*'} -j4 --schedule-random --output-on-failure
 
+  benchmark:
```
axic: This won't run with Release mode?

chfast: Not needed.

axic: In that case can you add a first commit which adds benchmarking on ci?

axic: Never mind, this is within the asan case.

chfast: I updated description.

Using CI for benchmarking is actually much more complicated. To be discussed some other time.

circle.yml
```diff
@@ -136,11 +144,12 @@ jobs:
   clang-latest-asan:
     executor: linux-clang-9
     environment:
-      BUILD_TYPE: Release
+      BUILD_TYPE: RelWithDebInfo
```
chfast: Having debug symbols makes sanitizer reports better.

PR #92, opened 2020-01-24
Add wabt to fizzy-bench

chfast:


axic:
How is this different to #5 ?

test/utils/wasm_engine.hpp
```diff
@@ -10,6 +10,8 @@ namespace fizzy::test
 class WasmEngine
 {
 public:
+    using FuncRef = uintptr_t;
```
axic: Why `uintptr_t` and not `intptr_t`? I found no definitive answer.

test/utils/wabt_engine.cpp
```diff
@@ -0,0 +1,110 @@
...
+    };
+
+    wabt::Errors errors;
+    wabt::Result result =
+        wabt::ReadBinaryInterp(&m_env, input.data(), input.size(), options, &errors, &m_module);
+
+    (void)result;
+    // FIXME: parse() should return bool.
+    assert(result == wabt::Result::Ok);
+}
```
axic:
```suggestion
    assert(m_env.GetMemoryCount() ==1);
}
```

test/utils/wabt_engine.cpp
```diff
@@ -0,0 +1,110 @@
...
+{
+    auto& dst = *m_env.GetMemory(0);
+    const auto begin = reinterpret_cast<const char*>(memory.data());
+    dst.data.assign(begin, begin + memory.size());
+}
+
+std::optional<WasmEngine::FuncRef> WabtEngine::find_function(std::string_view name) const
+{
+    const wabt::interp::Export* e = m_module->GetExport({name.data(), name.size()});
+    return e ? reinterpret_cast<WasmEngine::FuncRef>(e) : std::optional<WasmEngine::FuncRef>{};
```
axic: Should also check for `e->kind == ExternalKind::Func`

test/utils/wabt_engine.cpp
```diff
@@ -0,0 +1,110 @@
...
+#include "src/feature.h"
+#include "src/interp/binary-reader-interp.h"
+#include "src/interp/interp.h"
+#include "src/literal.h"
+#include "src/option-parser.h"
+#include "src/resolve-names.h"
+#include "src/stream.h"
+#include "src/validator.h"
+#include "src/wast-lexer.h"
+#include "src/wast-parser.h"
```
axic: Any way to fix these and use something like `<wabt/`> ?

chfast: Only by some workarounds. These headers include others this way.

test/bench/bench.cpp
```diff
@@ -6,7 +6,6 @@
 #include <iostream>
 #include <iterator>
 #include <memory>
-#include <sstream>
```
axic: This is needed for `istringstream` on macos.

test/utils/wasm_engine.hpp
chfast: I wasn't thinking a lot about this. In the end `uintptr_t` helps in fizzy a bit because `uint32_t` -> `uintptr_t` conversion is lossless.

test/utils/wabt_engine.cpp
chfast: Should not be needed for benchmarks not using memory.

test/utils/wasm_engine.hpp
chfast: Will be enough for wasm3?

axic: It is a pointer in wasm3 so I guess it doesn't matter too much.

test/utils/wabt_engine.cpp
```diff
@@ -37,17 +27,9 @@ std::unique_ptr<WasmEngine> create_wabt_engine()
 
 void WabtEngine::parse(bytes_view input)
 {
-    wabt::ReadBinaryOptions options{
```
axic: If you revert to default, what are the default settings for these?

test/utils/wabt_engine.cpp
axic: I didn't see bench has a test for memory not being present.

test/bench/CMakeLists.txt
```diff
@@ -7,7 +7,7 @@ set(fizzy_include_dir ${PROJECT_SOURCE_DIR}/lib/fizzy)
 
 add_executable(fizzy-bench bench.cpp)
 target_compile_features(fizzy-bench PRIVATE cxx_std_17)
-target_link_libraries(fizzy-bench PRIVATE fizzy::fizzy fizzy::test-utils benchmark::benchmark)
+target_link_libraries(fizzy-bench PRIVATE fizzy::fizzy fizzy::test-utils wabt::wabt benchmark::benchmark)
```
axic: This `wabt::wabt` here is unneeded.

test/utils/wabt_engine.cpp
```diff
@@ -58,13 +58,19 @@ void WabtEngine::instantiate() {}
...
     auto& env = const_cast<wabt::interp::Environment&>(m_env);
     const auto& memory = env.GetMemory(0);
     return {reinterpret_cast<uint8_t*>(memory->data.data()), memory->data.size()};
 }
 
 void WabtEngine::set_memory(bytes_view memory)
 {
+    if (memory.empty())
+        return;
+
```
axic: Shouldn't here be an assert for `GetMemoryCount() != 0` ?

test/utils/wabt_engine.cpp
```diff
@@ -58,13 +58,19 @@ void WabtEngine::instantiate() {}
 
 bytes_view WabtEngine::get_memory() const
 {
+    if (m_env.GetMemoryCount() == 0)
+        return {};
```
axic: Can you add a comment to `wasm_engine.h` saying if empty memory is returned it means no memory is available (exported)?

test/utils/wabt_engine.cpp
axic: Default:
```
  Features features;
  Stream* log_stream = nullptr;
  bool read_debug_names = false;
  bool stop_on_first_error = true;
  bool fail_on_custom_section_error = true;
```

axic:
@chfast @gumb0 this should be ready now

test/utils/wabt_engine.cpp
chfast: There is the assert like this inside `GetMemory()`.

test/utils/wabt_engine.cpp
```diff
@@ -0,0 +1,99 @@
...
+    wabt::interp::Executor executor{&m_env};
+
+    const auto* e = reinterpret_cast<const wabt::interp::Export*>(func_ref);
+
+    wabt::interp::TypedValues typed_args;
+    for (const auto arg : args)
+    {
+        wabt::interp::Value v{};
+        v.i32 = static_cast<uint32_t>(arg);
+        typed_args.push_back(wabt::interp::TypedValue{wabt::Type::I32, v});
```
gumb0: This could be also `I64` though? I guess you'll need to get function type from module somehow to find the right types here.

chfast: True. I forgot about it.

PR #94, opened 2020-01-24
Add macro to test the thrown exception message

gumb0:
Failure messages look like this:

- Exception thrown, but wrong type
```
[ RUN      ] instantiate.imported_functions_not_enough
/home/andrei/dev/fizzy/test/unittests/instantiate_test.cpp:64: Failure
        Failed
Unexpected exception type thrown.
```
- Correct type, wrong message
```
[ RUN      ] instantiate.imported_functions_not_enough
/home/andrei/dev/fizzy/test/unittests/instantiate_test.cpp:64: Failure
        Expected equality of these values:
exception.what()
Which is: "Module requires 1 imported functions, 0 provided"
"incorrect message"
```
- Exception not thrown
```
[ RUN      ] instantiate.imported_functions_not_enough
/home/andrei/dev/fizzy/test/unittests/instantiate_test.cpp:68: Failure
        Failed
Exception of type std::runtime_error is expected, but none was thrown.
```

### DIFF
--- a/test/utils/asserts.hpp
+++ b/test/utils/asserts.hpp
@@ -1,0 +1,18 @@
+#pragma once
+
+#include <gtest/gtest.h>
+
+#define EXPECT_THROW_MESSAGE(stmt, ex_type, expected)                                        \
+    try                                                                                      \
+    {                                                                                        \
+        stmt;                                                                                \
+        ADD_FAILURE() << "Exception of type " #ex_type " is expected, but none was thrown."; \
+    }                                                                                        \
+    catch (const ex_type& exception)                                                         \
+    {                                                                                        \
+        EXPECT_STREQ(exception.what(), expected);                                            \
+    }                                                                                        \
+    catch (...)                                                                              \
+    {                                                                                        \
+        ADD_FAILURE() << "Unexpected exception type thrown.";                                \
+    }
--- a/test/unittests/instantiate_test.cpp
+++ b/test/unittests/instantiate_test.cpp
@@ -1,5 +1,6 @@
 #include "execute.hpp"
 #include <gtest/gtest.h>
+#include <test/utils/asserts.hpp>
 #include <test/utils/hex.hpp>
 
 using namespace fizzy;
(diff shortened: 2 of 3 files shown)


test/utils/asserts.hpp
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+#include <gtest/gtest.h>
+
+#define EXPECT_THROW_WITH_MESSAGE(stmt, ex_type, expected)                                   \
```
chfast: This is fine, but `EXPECT_THROW_WITH_MESSSAGE(...) << "hint"` will not work. The alternative solution proposed in https://github.com/google/googletest/issues/952#issuecomment-521361666 seems to fix that and is it leaves type checking to original `EXPECT_THROW`.

gumb0:
> I would also consider any shorter name.

I think `EXPECT_THROW_MESSAGE` would be fine.

test/utils/asserts.hpp
gumb0: I tried it, failure output is more ugly for the case of incorrect exception thrown
```
[ RUN      ] instantiate.imported_functions_not_enough
/home/andrei/dev/fizzy/test/unittests/instantiate_test.cpp:58: Failure
Expected: try { instantiate(module, {}); } catch (const std::bad_alloc& ex) { switch (0) case 0: default: if (const ::testing::AssertionResult gtest_ar = (::testing::internal::CmpHelperSTREQ("ex.what()", "\"Module requires 1 imported functions, 0 provided\"", ex.what(), "Module requires 1 imported functions, 0 provided"))) ; else ::testing::internal::AssertHelper(::testing::TestPartResult::kNonFatalFailure, "/home/andrei/dev/fizzy/test/unittests/instantiate_test.cpp", 58, gtest_ar.failure_message()) = ::testing::Message(); throw; } throws an exception of type std::bad_alloc.
  Actual: it throws a different type.
```

gumb0: For not throwing, too
```
[ RUN      ] instantiate.imported_functions_not_enough
/home/andrei/dev/fizzy/test/unittests/instantiate_test.cpp:62: Failure
Expected: try { instantiate(module1, {}); } catch (const std::runtime_error& ex) { switch (0) case 0: default: if (const ::testing::AssertionResult gtest_ar = (::testing::internal::CmpHelperSTREQ("ex.what()", "\"Module requires 1 imported functions, 0 provided\"", ex.what(), "Module requires 1 imported functions, 0 provided"))) ; else ::testing::internal::AssertHelper(::testing::TestPartResult::kNonFatalFailure, "/home/andrei/dev/fizzy/test/unittests/instantiate_test.cpp", 62, gtest_ar.failure_message()) = ::testing::Message(); throw; } throws an exception of type std::runtime_error.
  Actual: it throws nothing.
```

gumb0:
Renamed and rebased.

PR #95, opened 2020-01-24
Implement br_table instruction

gumb0:


lib/fizzy/parser_expr.cpp
```diff
@@ -185,6 +185,20 @@ parser_result<Code> parse_expr(const uint8_t* pos)
...
         }
 
+        case Instr::br_table:
+        {
+            std::vector<uint32_t> label_indices;
+            std::tie(label_indices, pos) = parser<std::vector<uint32_t>>{}(pos);
+            uint32_t last_label_idx;
+            std::tie(last_label_idx, pos) = leb128u_decode<uint32_t>(pos);
+
+            push(code.immediates, label_indices.size());
```
axic: What is the type of this? Can you enforce an explicit type to be pushed as immediate?

lib/fizzy/parser_expr.cpp
```diff
@@ -185,6 +185,20 @@ parser_result<Code> parse_expr(const uint8_t* pos)
...
+            for (const auto idx : label_indices)
+                push(code.immediates, idx);
+            push(code.immediates, last_label_idx);
```
axic: I think this should be called `default_label_idx` or something similar.

lib/fizzy/parser_expr.cpp
gumb0: Yeah this probably works incorrectly without explicit cast.

lib/fizzy/execute.cpp
```diff
@@ -334,24 +356,25 @@ execution_result execute(Instance& instance, FuncIdx func_idx, std::vector<uint6
...
 
-            assert(labels.size() > label_idx);
-            labels.drop(label_idx);  // Drop skipped labels (does nothing for labelidx == 0).
-            const auto label = labels.pop();
+            branch(label_idx, labels, stack, pc, immediates);
+            break;
+        }
+        case Instr::br_table:
+        {
+            const auto br_table_size = read<uint32_t>(immediates);
```
axic: Would adda note here that the last immediate is the default label.

lib/fizzy/execute.cpp
```diff
@@ -17,6 +17,28 @@ struct LabelContext
     size_t stack_height = 0;
 };
 
+void branch(uint32_t label_idx, Stack<LabelContext>& labels, Stack<uint64_t>& stack,
```
axic: I guess we could benchmark now whether this should be inlined or not.

test/unittests/execute_control_test.cpp
```diff
@@ -480,6 +480,51 @@ TEST(execute_control, br_1_out_of_function_and_imported_function)
...
+      )
+      (i32.const 104)
+    )
+   */
+    const auto bin = from_hex(
+        "0061736d0100000001060160017f017f030201000a330131000240024002"
+        "400240024020000e04030201000441e3000f0b41e4000f0b41e5000f0b41"
+        "e6000f0b41e7000f0b41e8000b000c046e616d6502050100010000");
+
+    for (const auto param : {0u, 1u, 2u, 3u, 4u})
```
chfast: I would add one more input value for default label with value other than `4u`.

gumb0: This I test below outside of the loop, if I understand you correctly.

chfast: Oh yes, right. But I think you are overusing value `42`. The value `5` would be better because it is on the edge of being out of table index.

gumb0: Ok I'll add another case to the loop then

test/unittests/parser_test.cpp
```diff
@@ -664,3 +664,50 @@ TEST(parser, block_br)
         "01000000"
         "01000000"_bytes);
 }
+
+TEST(parser, instr_br_table)
+{
+    // (block
```
axic: Can you use `/* */` comment block here too?

gumb0: done

axic:
> Interesting test would br_table with empty vector (only index for default jump).

@gumb0 can you add this to the parser too?

lib/fizzy/execute.cpp
axic: `mul256_opt0` uses both `br` and `br_if`.

Run both master and this a few times, there's some variance but they seem to be very close.

PR #96, opened 2020-01-24
ci: Clang analyzer

chfast:


gumb0:
How to run it locally?

test/bench/bench.cpp
```diff
@@ -103,14 +103,27 @@ void benchmark_execute(benchmark::State& state, const ExecutionBenchmarkCase& be
...
         const auto result = fizzy::execute(instance, *func_idx, benchmark_case.func_args);
         benchmark::DoNotOptimize(result);
     }
 }
 
+template <typename Lambda>
+void register_benchmark(const std::string& name, Lambda&& fn)
+{
+#ifdef __clang_analyzer__
+    // Clang analyzer reports potential memory leak in benchmark::RegisterBenchmark().
```
gumb0: Not possible to disable it with `NOLINT` ?

chfast: No, because the main error is in `benchmark.hpp`. There is option `HeaderFilterRegex` but does not seem to work. Maybe this option is only for clang-tidy checks – clang analyzer is different beast, just available in the same tool.

chfast:
> How to run it locally?

1. Pass `-DCMAKE_CXX_CLANG_TIDY=clang-tidy`. Then just build. You probably want different build config for this as it will be 10x slower. It is done this way on CI (I have not found a way to only run clang-tidy easily so I mixed it with UBSan build). Docs: https://cmake.org/cmake/help/latest/prop_tgt/LANG_CLANG_TIDY.html
2. CLion has clang-tidy integration — look for clang-tidy in settings and check what "inspections" you have enabled. It works pretty ok, but I noticed it can give different results. Probably clang-analyzer is forced disabled there, can be for performance.
3. Third option is to export "compile commands" from cmake configuration: -DCMAKE_EXPORT_COMPILE_COMMANDS=YES`. Then you can use `clang-tidy -p <build-dir> <file>.

For many warnings `clang-tidy` (and also CLion) is able to automatically fix them (and reformat fix with clang-format). Although I never used that at scale. Docs: https://cmake.org/cmake/help/latest/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html.

As mentioned in the commit, I have not enabled any fancy check causing regressions. Clang analyzer was the main goal.

PR #97, opened 2020-01-26
Script wat2wasm4cpp

chfast:
1. Script is written in Python. Can be done in CMake someday, but this requires substantially  more work.
2. It takes a C++ source file as an argument.
3. It searches for `/* wasm ... */` comments containing WAT code. I think a match like `/*\n(... */` would also work. We can discuss if there should be a common name for this variable, e.g. `bin`, `wasm`, `wasm_binary`.
4. The result wasm binary is used to update the `const auto <any_name> = from_hex("...");`. This was used as a start because it was like that in the file already. There is some issue with formatting `""_bytes"` automatically. If this pattern is not found, the result wasm is ignored.
5. It expects `wat2wasm` (wabt) in the PATH.
6. The modified source file could be additionally formatted with clang-format, but this is not implemented.
7. There is a way to report WAT errors with line numbers matching the source file. But not implemented yet.
8. Fun fact: if you omit `(module)` in WAT, wat2wasm skips at least some custom sections. That's why the updated wasm binaries are smaller.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -82,6 +82,17 @@ jobs:
         name: "Run codespell"
         command: |
           codespell --quiet-level=4
+    - run:
+        name: "Install wabt"
+        working_directory: ~/bin
+        command: curl -L https://github.com/WebAssembly/wabt/releases/download/1.0.13/wabt-1.0.13-linux.tar.gz | tar xz --strip=1
+    - run:
+        name: "Check wat2wasm4cpp"
+        command: |
+          export PATH=$PATH:~/bin
+          ./wat2wasm4cpp.py test/unittests/execute_control_test.cpp
+          git diff --color --exit-code
+
 
   release-linux:
     executor: linux-gcc-9
--- a/test/unittests/execute_control_test.cpp
+++ b/test/unittests/execute_control_test.cpp
@@ -1,6 +1,7 @@
 #include "execute.hpp"
 #include "parser.hpp"
 #include <gtest/gtest.h>
+#include <test/utils/asserts.hpp>
 #include <test/utils/hex.hpp>
 
 using namespace fizzy;
@@ -150,9 +151,9 @@ TEST(execute_control, blocks_without_br)
     EXPECT_EQ(ret[0], 4);
 }
 
-TEST(execute_control, nested_blocks)
+TEST(execute_control, nested_blocks_0)
 {
-    /* wat2wasm --no-check
+    /* wat2wasm
     (func (result i32)
       (local i32)
       (block
@@ -171,7 +172,7 @@ TEST(execute_control, nested_blocks)
             i32.or
             set_local 0
             i32.const 1
-            br <<br_imm>>
+            br 0
             get_local 0
             i32.const 8
             i32.or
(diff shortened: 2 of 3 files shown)


axic:
> Fun fact: if you omit (module) in WAT, wat2wasm skips at least some custom sections. That's why the updated wasm binaries are smaller.

It probably is skipping the names section.

I would actually prefer if the examples were real wat files, e.g. contain the `(module` part.

chfast:
> > Fun fact: if you omit (module) in WAT, wat2wasm skips at least some custom sections. That's why the updated wasm binaries are smaller.
> 
> It probably is skipping the names section.
> 
> I would actually prefer if the examples were real wat files, e.g. contain the `(module` part. At the minimum the script should wrap them.

Actually the reason is the difference between wabt and webassembly.studio. The `(module)` can be omitted in both cases.

axic:
In section 6.6.13:
Screenshot 2020-01-27 at 12 23 36

So apparently this changed, it wasn't the case in 2017 :wink:

chfast:
> So apparently this changed, it wasn't the case in 2017 wink

I planned to add `(module)` in comments, but it worked without it, so I didn't.

chfast:
Notable changes:
1. The script can now add wasm binary hex if the previous one does not exist.
2. The code if formatted with clang-format.
3. No special comment tag is needed, just second line must start with `(`.
4. Additional options can be passed to wat2wasm, e.g. `wat2wasm --no-check` to skip validation. Useful for creating invalid modules for tests.
5. Errors in WAT comments are now mapped to the original C++ source lines.

wat2wasm4cpp.py
```diff
@@ -0,0 +1,131 @@
...
+
+It searches for following block comments:
+
+    /* Additional description or wat2wasm options.
+    (module
+      (func ...)
+    )
+    */
+
+Inportantly, the first line of the comment can contain additional wat2wasm
```
gumb0:
```suggestion
Importantly, the first line of the comment can contain additional wat2wasm
```

gumb0: Not quite clear whether I can combine a regular comment like in the previous snippet with `wat2wasm` comment line?

test/unittests/execute_control_test.cpp
```diff
@@ -28,33 +35,23 @@ TEST(execute_control, nop)
...
-                 "0a000000" /* i32.const 0xa */
-                 "01000000" /* local.set 1 */
-                 "00000000" /* br 0 */
-                 "0b000000" /* i32.const 0xb */
-                 "01000000" /* local.set 1 */
-                 "01000000" /* local.get 1 */
-            )});
-
-    const auto [trap, ret] = execute(module, 0, {});
+    /* wat2wasm --no-check
```
gumb0: This code doesn't need `--no-check` though, right?

chfast: It does because function result type is `[]` but the instructions in the functions give `[i32]`

test/unittests/execute_control_test.cpp
```diff
@@ -195,14 +192,10 @@ TEST(execute_control, nested_blocks)
...
     */
-
-    auto bin = from_hex(
-        "0061736d010000000105016000017f030201000a41013f01017f"
-        "0240200041017221000240200041027221000240200041047221000c"
-        "77"  // <-- br's immediate value.
-        "200041087221000b200041107221000b200041207221000b200041c000720b000c046e"
-        "616d6502050100010000");
-    constexpr auto br_imm_offset = 54;
+    auto wasm = from_hex(
```
chfast: Here the `const` is omitted and the script will leave it as it when updating.

wat2wasm4cpp.py
```diff
@@ -0,0 +1,131 @@
...
+
+It searches for following block comments:
+
+    /* Additional description or wat2wasm options.
+    (module
+      (func ...)
+    )
+    */
+
+Importantly, the first line of the comment can contain additional wat2wasm
```
gumb0: Maybe then
```suggestion
Importantly, the first line instad of a comment can contain additional wat2wasm
```

axic:
What is left on this?

axic:
@chfast @gumb0 can we agree on this?

chfast:
> @chfast @gumb0 can we agree on this?

Agree on what? To my understanding there is nothing left here for the script. Before merge I need to clean up the `execute_control_test.cpp`.

axic:
To agree on the syntax. Are there any comments on that?

test/unittests/execute_control_test.cpp
axic: Why is that needed if you have the comment above with `--no-check`? That is used to ensure the example still compiles, but the proper bytecode is actually different (to fill in the imm)?

Instead of all this mess why not have 2-3 examples with different immediates?

chfast:
> Can you add this to the CI and fail if the files differ after running the script? (To confirm it was updated properly in every commit).

Sounds like a lot of work.

test/unittests/execute_control_test.cpp
chfast: Added 4 examples.

axic:
> > Can you add this to the CI and fail if the files differ after running the script? (To confirm it was updated properly in every commit).
> >
> Sounds like a lot of work.

```
$ wget <url to wabt release>
$ tar xzvf <wabt release>
$ wat2wasm4cpp execute_control_test.cpp
$ git diff --exit-code
```

circle.yml
```diff
@@ -82,6 +82,17 @@ jobs:
...
           codespell --quiet-level=4
+    - run:
+        name: "Install wabt"
+        working_directory: ~/bin
+        command: curl -L https://github.com/WebAssembly/wabt/releases/download/1.0.13/wabt-1.0.13-linux.tar.gz | tar xz --strip=1
+    - run:
+        name: "Check wat2wasm4cpp"
+        command: |
+          export PATH=$PATH:~/bin
+          ./wat2wasm4cpp.py test/unittests/execute_control_test.cpp
```
axic: Just verified locally and this works.

test/unittests/execute_control_test.cpp
```diff
@@ -1,22 +1,30 @@
...
 TEST(execute_control, nop)
 {
+    /*
+    (func
+      nop
+    )
+    */
+    const auto wasm = from_hex("0061736d01000000010401600000030201000a05010300010b");
     Module module;
     module.codesec.emplace_back(Code{0, {Instr::nop, Instr::end}, {}});
```
axic: This case is weird, has wasm and module.

circle.yml
chfast: I checked on CI.

test/unittests/execute_control_test.cpp
chfast: Fixed.

axic:
Not sure what is the best place to put the script:
- root
- `scripts/`
- `tools/`

I think for this PR it is fine as is and we can handle layout later.

chfast:
There seems to be one issue, if you have a `/* */` comment with WAT code, there is no way to tell wat2wasm4cpp to ignore it. I start thinking it might be actually good to have `wat2wasm` in the comment block mandatory.

axic:
> I start thinking it might be actually good to have wat2wasm in the comment block mandatory.

I think explicitness is useful, but

> if you have a /* */ comment with WAT code

`/* */` is invalid for comments, one should use `;;`

chfast:
> > if you have a /* */ comment with WAT code
> 
> `/* */` is invalid for comments, one should use `;;`

How is wat comment syntax related?

axic:
> How is wat comment syntax related?

Sorry, I've misread your text. I'm happy to have it explicit and `/* wat2wasm` sounds good.

axic:
@chfast can you squash the new commits into the original ones? Looks good to merge.

chfast:
> @chfast can you squash the new commits into the original ones? Looks good to merge.

Yes, that is planned. But I wanted to show what has changed.

PR #98, opened 2020-01-27
Pass module to instance by pointer

chfast:
The instance object captures the reference to the module object. This is bad API design, because the owner of the module must remember to keep it alive as long as the instance object is used. Anyway, this change makes it only slightly better. Using temporary module objects is more difficult now, e.g. `instantiate(parse(wasm))` will not compile any more. Also, instance objects are now assignable. This is useful, because they are returned by value from instantiate().

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -202,21 +202,21 @@ inline uint64_t popcnt64(uint64_t value) noexcept
 }
 }  // namespace
 
-Instance instantiate(const Module& module, std::vector<ImportedFunction> imported_functions,
+Instance instantiate(const Module* module, std::vector<ImportedFunction> imported_functions,
     std::vector<ImportedGlobal> imported_globals)
 {
     size_t memory_min, memory_max;
-    if (module.memorysec.size() > 1)
+    if (module->memorysec.size() > 1)
     {
         // FIXME: turn this into an assert if instantiate is not exposed externally and it only
         // takes validated modules
         throw std::runtime_error("Cannot support more than 1 memory section.");
     }
-    else if (module.memorysec.size() == 1)
+    else if (module->memorysec.size() == 1)
     {
-        memory_min = module.memorysec[0].limits.min;
-        if (module.memorysec[0].limits.max)
-            memory_max = *module.memorysec[0].limits.max;
+        memory_min = module->memorysec[0].limits.min;
+        if (module->memorysec[0].limits.max)
+            memory_max = *module->memorysec[0].limits.max;
         else
             memory_max = MemoryPagesLimit;
         // FIXME: better error handling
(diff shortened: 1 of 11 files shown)


gumb0:
It looks ok, but what about going one step further and making it a `shared_ptr`?

axic:
> This is bad API design

It was only a temporary measure and not by design. Before `instantiate` was added, it was safe to pass it by reference to `execute`, but as part of the split out we didn't review the API, just moved it around to get things going.

chfast:
I understand why it happened this way, no blaming here.
The `shared_ptr` seems to be natural choice at this point, but I didn't want to open a discussion about API design. 
To my understanding, you can have mutiple instances of the same module. But I'm not sure if that's really useful feature in blockchain application.

Anyway, I'm happy to try `shared_ptr`. Will that be accepted?

gumb0:
Woah it became huge

It looks ok, but I would be fine if we step back with a `shared_ptr`, as it looks somewhat disruptive, at least to test code.

(Maybe argument against changing now to `shared_ptr`: probably we'll need to change it again anyway if we want to provide an external API that would allow sharing modules among instances, won't we?

although it could probably still stay some kind of a smart pointer on the internal (C++) implementation side...)

chfast:
At least this is correct now.

1. It would be easier if we don't have one-to-many relation between module and instance. We don't use it anyway.
2. The workaround for not exposing `shared_ptr` in `Instance` is to keep both reference and private `shared_ptr`. This will still require writing custom copy constructors.

gumb0:
Now I'm thinking, since Module is supposed to be immutable type, I don't see much necessity in sharing the same module object among instances.
So maybe we should move `Module` into `instantiate`, and for the rare case when the same module is instantiated several times, we can require to copy module on the user's side.

This way we'd make the most common use case the most convenient and efficient, and the rare use case to require additional step.

chfast:
I don't care about API design at this point. But I cannot do anything while the `Instance` is not assignable.

axic:
> So maybe we should move Module into instantiate, and for the rare case when the same module is instantiated several times, we can require to copy module on the user's side.

I was hoping that we can actually copy the parts needed and throw away the rest, e.g.
- we already use the elements/data sections to build up the memory/table (the sections from module are useless after that)
- need to copy code and immediates (and types, perhaps)

However I think it makes sense to do all this when we have feature parity and would probably suggest just to expect Module by value and move it into the instance. And let the caller decide if they need to keep a copy.

chfast:
I was considering copying only required parts, but they are changing frequently.

axic:
For keeping this simple, can we agree to taking `Module` value in `instantiate` and moving the entire object into the `Instance`?

gumb0:
> For keeping this simple, can we agree to taking Module by value in instantiate and moving the entire object into the Instance?

Sounds good to me

chfast:
> For keeping this simple, can we agree to taking `Module` by value in `instantiate` and moving the entire object into the `Instance`?

Sounds good for me too, but there is no way to express that a module should be moved to instance, so it will be up to the user to remember that.

chfast:
I don't thing there is anything useful in this PR for the alternative solution, so it will be best to just replace it with a new one.

gumb0:
> Sounds good for me too, but there is no way to express that a module should be moved to instance, so it will be up to the user to remember that.

That's fine, it's similar to passing `args` to `execute`, they will be copied unless a user remembers to move.

axic:
Closing this according to @chfast hint.

PR #99, opened 2020-01-27
Implement support for if and else instructions

rodiazet:


lib/fizzy/parser_expr.cpp
```diff
@@ -170,6 +172,53 @@ parser_result<Code> parse_expr(const uint8_t* pos)
...
+
+            push(code.immediates, uint32_t{0});  // Diff to the else instruction
+            push(code.immediates, uint32_t{0});  // Diff for the immediates.
+
+            break;
+        }
+
+        case Instr::else_:
+        {
+            assert(!label_positions.empty());
```
gumb0: It should throw an error instead, it means invalid module provided to parse.

lib/fizzy/execute.cpp
```diff
@@ -378,6 +378,55 @@ execution_result execute(Instance& instance, FuncIdx func_idx, std::vector<uint6
             labels.push_back(label);
             break;
         }
+        case Instr::if_:
+        {
+            if (static_cast<uint32_t>(stack.pop() != 0))
```
```suggestion
            if (static_cast<uint32_t>(stack.pop()) != 0)
```

lib/fizzy/execute.cpp
```diff
@@ -378,6 +378,55 @@ execution_result execute(Instance& instance, FuncIdx func_idx, std::vector<uint6
...
+            {
+                const auto arity = read<uint8_t>(immediates);
+                const auto target_pc = read<uint32_t>(immediates);
+                const auto target_imm = read<uint32_t>(immediates);
+                read<uint32_t>(immediates);  // Omit the immediates for else instruction.
```
gumb0: I would instead just do `immediates += 2 * sizeof(uint32_t)`, there's no need to call `memcpy` here.

(also better to call it "skip" instead of "omit")

lib/fizzy/execute.cpp
```diff
@@ -378,6 +378,55 @@ execution_result execute(Instance& instance, FuncIdx func_idx, std::vector<uint6
...
+            }
+            else
+            {
+                const auto arity = read<uint8_t>(immediates);
+                const auto target_pc = read<uint32_t>(immediates);
+                const auto target_imm = read<uint32_t>(immediates);
+                const auto target_else_pc = read<uint32_t>(immediates);
+                const auto target_else_imm = read<uint32_t>(immediates);
+
+                if (target_else_pc != 0 && target_else_imm != 0)  // If else block defined.
```
gumb0: would be enough to check just one?

gumb0:
I would also change these comments to mention that it's now about block/if/else instructions
https://github.com/wasmx/fizzy/blob/3a24b71fd6dc2541719274c93216cf6ebd91c6f8/lib/fizzy/parser_expr.cpp#L33-L34
https://github.com/wasmx/fizzy/blob/3a24b71fd6dc2541719274c93216cf6ebd91c6f8/lib/fizzy/parser_expr.cpp#L121

lib/fizzy/execute.cpp
chfast: One should be enough. This assumes `else` cannot be first instruction.

lib/fizzy/execute.cpp
```diff
@@ -378,6 +378,55 @@ execution_result execute(Instance& instance, FuncIdx func_idx, std::vector<uint6
...
+                {
+                    LabelContext label{code.instructions.data() + target_pc,
+                        code.immediates.data() + target_imm, arity, stack.size()};
+                    labels.emplace_back(label);
+                    pc = code.instructions.data() + target_else_pc;
+                    immediates = code.immediates.data() + target_else_imm;
+                }
+                else  // If else block not defined go to end of if.
+                {
+                    assert(arity == 0);  // if without else cannot have type signature.
```
gumb0: Could you point where is this in the spec?

test/unittests/execute_control_test.cpp
```diff
@@ -541,3 +541,146 @@ TEST(execute_control, return_from_block_stack_cleanup)
...
+{
+    /*
+    (func  (param $x i32) (result i32)
+      get_local $x
+      (if (result i32)
+        (then
+          i32.const -1
+          i32.const 21
+          br 0
+        )
```
gumb0: Better add another `i32.const` after `br` to ensure it really breaks.

lib/fizzy/execute.cpp
rodiazet: WebAssembly Studio returns the error when parsing `if` instructions. Unfortunately I cannot find any information in spec. I can remove the assert if you decide so.

chfast: In simplistic view, both `if` and `else` blocks must have the same type. If `else` is omitted the type is `[]` therefore `if`'s type must be `[]`.
https://webassembly.github.io/spec/core/valid/instructions.html#valid-if

lib/fizzy/execute.cpp
```diff
@@ -396,6 +396,55 @@ execution_result execute(Instance& instance, FuncIdx func_idx, std::vector<uint6
             labels.push_back(label);
             break;
         }
+        case Instr::if_:
+        {
+            if (static_cast<uint32_t>(stack.pop()) != 0)
+            {
+                const auto arity = read<uint8_t>(immediates);
+                const auto target_pc = read<uint32_t>(immediates);
+                const auto target_imm = read<uint32_t>(immediates);
```
axic: The first 3 fields are shared. @gumb0 @chfast Would it make sense moving them above the if?

lib/fizzy/parser_expr.cpp
```diff
@@ -1,6 +1,8 @@
 #include "parser.hpp"
 #include "stack.hpp"
 
+#include <cassert>
```
axic: Why is this include needed here?

lib/fizzy/execute.cpp
chfast: Yep.

lib/fizzy/parser_expr.cpp
rodiazet: No. Fixed.

lib/fizzy/parser_expr.cpp
```diff
@@ -170,6 +170,54 @@ parser_result<Code> parse_expr(const uint8_t* pos)
...
+
+            push(code.immediates, uint32_t{0});  // Diff to the else instruction
+            push(code.immediates, uint32_t{0});  // Diff for the immediates.
+
+            break;
+        }
+
+        case Instr::else_:
+        {
+            if (label_positions.empty())
```
axic: Wouldn't this still allow `block else`?

rodiazet: It would. I can add verification that we it's `if` label like it's done for `loop` and `LoopLabel` or just not to check this because it's verifier's job.

chfast: To resolve "verification" here, the best would be to add "type" filed to `Label`. This will also help a bit with `loop` case where we use sentinel `Label` constant.

But I don't think this is good direction to more checks to parser, which is not a validator. Parser should not crash on invalid input (and this check prevents this).

I would first construct the `(block (else ...))` example to see what happens during execution.

test/unittests/execute_control_test.cpp
```diff
@@ -609,3 +609,149 @@ TEST(execute_control, return_from_block_stack_cleanup)
...
+        "650209010002000178010179");
+
+    for (const auto param : {0u, 1u})
+    {
+        constexpr uint64_t expected_results[]{
+            0,  // no if branch.
+            4,  // if branch.
+        };
+
+        const auto [trap, ret] = execute(parse(bin), 0, {param});
```
gumb0: I see some existing tests are doing this, but I think better not to parse in the loop.

rodiazet: Fixed.

lib/fizzy/parser_expr.cpp
rodiazet: Any decision on that? Should I try to create bytecode with `(block (else ...))` by hand?

chfast: I will handle this change quickly myself.

test/unittests/execute_control_test.cpp
```diff
@@ -613,3 +613,157 @@ TEST(execute_control, return_from_block_stack_cleanup)
...
+          i32.const 2
+        )
+      )
+    )
+    */
+    const auto bin = from_hex(
+        "0061736D0100000001060160017F017F030201000A0E010C002000047F41010541020B0B000D046E616D650206"
+        "010001000178");
+
+    auto module = parse(bin);
```
chfast:
```suggestion
    const auto module = parse(bin);
```
I believe `module` can be `const`. If I'm right, update also other cases.

rodiazet: Done.

lib/fizzy/parser_expr.cpp
chfast: See https://github.com/wasmx/fizzy/pull/120.

axic: I agree to do most of the instruction validation as per #63. However I think if it is not too hard, we should do this simple validation here. Based on #120 it seems to be trivial now.

chfast: Implemented.

lib/fizzy/parser_expr.cpp
```diff
@@ -170,6 +170,56 @@ parser_result<Code> parse_expr(const uint8_t* pos)
...
+                throw parser_error{"unexpected else instruction"};
```
axic: Can you also add a test case triggering this?
```
(func
  else  
)
```

PR #100, opened 2020-01-27
Implement initialization of data section offsets from global constant

gumb0:
1. Move data section processing to after globals initialization in `instantiate`.
2. Extract getting the value of constant expression to a separate function `eval_constant_expression`.
3. Use `eval_constant_expression` for both globals inited by globals and data offsets inited by globals. Should be also later used for element section offsets.

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -297,21 +297,18 @@ Instance instantiate(const Module& module, std::vector<ImportedFunction> importe
     globals.reserve(module.globalsec.size());
     for (auto const& global : module.globalsec)
     {
-        if (global.expression.kind == ConstantExpression::Kind::Constant)
-            globals.emplace_back(global.expression.value.constant);
-        else
+        // Wasm spec section 3.3.7 constrains initialization by another global to const imports only
+        // https://webassembly.github.io/spec/core/valid/instructions.html#expressions
+        if (global.expression.kind == ConstantExpression::Kind::GlobalGet &&
+            global.expression.value.global_index >= imported_globals.size())
         {
-            // initialize by imported global
-            const auto global_idx = global.expression.value.global_index;
-            // Wasm spec section 3.3.7 constrains initialization by another global to const imports
-            // only https://webassembly.github.io/spec/core/valid/instructions.html#expressions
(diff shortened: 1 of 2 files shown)


lib/fizzy/execute.cpp
```diff
@@ -83,6 +83,28 @@ std::vector<TypeIdx> match_imports(const Module& module,
...
+    if (expr.kind == ConstantExpression::Kind::Constant)
+        return expr.value.constant;
+
+    assert(expr.kind == ConstantExpression::Kind::GlobalGet);
+
+    const auto global_idx = expr.value.global_index;
+    const bool is_mutable = (global_idx < imported_globals.size() ?
+                                 imported_globals[global_idx].is_mutable :
+                                 global_types[global_idx - imported_globals.size()].is_mutable);
+    if (is_mutable)
```
gumb0: This validation should perhaps be done in parser (constant expression is invalid if it uses mutable globals) . But currently it seems much easier to do it in instantiate.

lib/fizzy/execute.cpp
```diff
@@ -232,43 +254,36 @@ Instance instantiate(const Module& module, std::vector<ImportedFunction> importe
...
+            throw std::runtime_error(
+                "Global can be initialized by another const global only if it's imported.");
         }
+
+        const auto value = eval_constant_expression(
+            global.expression, imported_globals, module.globalsec, globals);
+        globals.emplace_back(value);
+    }
+
+    // Fill out memory based on data segments
```
axic: Can you create a PR (or commit) which reorders the file (according to the instantiation spec)?

Could also move the `load_from_memory `helpers above `instantiate()`.

gumb0: Did the first reorder in the first commit here.

gumb0:
Reordered and squashed.

lib/fizzy/execute.cpp
gumb0: Did the second reorder in https://github.com/wasmx/fizzy/pull/104

gumb0:
Rebased.

Note the changes in the first commit ordering new instantiate steps from `master`: table is allocated after globals; table elements are initialized before filling out the memory.

PR #101, opened 2020-01-27
Disable tests causing buffer overflow

chfast:


test/unittests/parser_test.cpp
```diff
@@ -595,27 +596,29 @@ TEST(parser, milestone1)
...
     EXPECT_EQ(code1.instructions, (std::vector{Instr::loop, Instr::end, Instr::end}));
     EXPECT_EQ(code1.immediates.size(), 0);
 
     const auto loop_i32_empty = "037f0b0b"_bytes;
-    EXPECT_THROW(parse_expr(loop_i32_empty.data()), parser_error);  // loop arity != 0.
+    EXPECT_THROW_MESSAGE(
+        parse_expr(loop_i32_empty.data()), parser_error, "loop can only have type arity 0");
+}
+
+TEST(parser, DISABLED_instr_loop_input_buffer_overflow)
```
axic: Why is this causing a buffer overflow?

chfast: Because the end opcode is missing, and it continues until finds random byte which is a valid opcode.

axic: As a reminder, `03400b` is consumed as `loop <0x40> end` and should fail with "loop can only have type arity 0". Why isn't that happening?

axic: Ah it's missing the real `end` opcode.

axic: Can you leave that as a comment?

chfast: What exactly?

axic: Don't know, tried to describe but its long: "The `loop` instruction is well formed as it contains the closing `end`, but the code block doesn't contain the implicit closing `end`."

Perhaps just add "`03400b0b`" would be valid.

test/utils/asserts.hpp
```diff
@@ -15,4 +15,5 @@
     catch (...)                                                                              \
     {                                                                                        \
         ADD_FAILURE() << "Unexpected exception type thrown.";                                \
-    }
+    }                                                                                        \
+    (void)0
```
gumb0: what does this do?

axic: Ensure you need to add a trailing semicolon.

test/unittests/parser_test.cpp
```diff
@@ -635,6 +639,13 @@ TEST(parser, instr_block)
         "09000000"_bytes);
 }
 
+TEST(parser, DISABLED_instr_block_input_buffer_overflow)
+{
+    // The function end opcode 0b is missing causing reading out of input buffer.
```
gumb0:
```suggestion
    // The block end opcode 0b is missing causing reading out of input buffer.
```

gumb0: Ah no, I misunderstood it, sorry

PR #102, opened 2020-01-28
Implement call_indirect instruction

gumb0:
Depends on #82

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -17,54 +17,6 @@ struct LabelContext
     size_t stack_height = 0;
 };
 
-void branch(uint32_t label_idx, Stack<LabelContext>& labels, Stack<uint64_t>& stack,
-    const Instr*& pc, const uint8_t*& immediates) noexcept
-{
-    assert(labels.size() > label_idx);
-    labels.drop(label_idx);  // Drop skipped labels (does nothing for labelidx == 0).
-    const auto label = labels.pop();
-
-    pc = label.pc;
-    immediates = label.immediate;
-
-    // When branch is taken, additional stack items must be dropped.
-    assert(stack.size() >= label.stack_height + label.arity);
-    if (label.arity != 0)
-    {
-        assert(label.arity == 1);
-        const auto result = stack.peek();
-        stack.resize(label.stack_height);
-        stack.push(result);
-    }
-    else
-        stack.resize(label.stack_height);
-}
-
-bool invoke_function(
-    uint32_t type_idx, uint32_t func_idx, Instance& instance, Stack<uint64_t>& stack)
-{
-    const auto num_inputs = instance.module.typesec[type_idx].inputs.size();
-    assert(stack.size() >= num_inputs);
-    std::vector<uint64_t> call_args(
-        stack.rbegin(), stack.rbegin() + static_cast<ptrdiff_t>(num_inputs));
-    stack.resize(stack.size() - num_inputs);
-
-    const auto ret = execute(instance, func_idx, call_args);
-    // Bubble up traps
-    if (ret.trapped)
-        return false;
-
-    const auto num_outputs = instance.module.typesec[type_idx].outputs.size();
-    // NOTE: we can assume these two from validation
-    assert(ret.stack.size() == num_outputs);
-    assert(num_outputs <= 1);
-    // Push back the result
-    if (num_outputs != 0)
-        stack.push(ret.stack[0]);
-
-    return true;
-}
-
 void match_imported_functions(const std::vector<TypeIdx>& module_imported_types,
     const std::vector<ImportedFunction>& imported_functions)
 {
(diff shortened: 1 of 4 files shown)


lib/fizzy/parser_expr.cpp
```diff
@@ -178,6 +178,7 @@ parser_result<Code> parse_expr(const uint8_t* pos)
         case Instr::br:
         case Instr::br_if:
         case Instr::call:
+        case Instr::call_indirect:
```
axic: I believe this is wrong, it should also have:
```suggestion
        case Instr::call_indirect:
        {
            uint32_t imm;
            std::tie(imm, pos) = leb128u_decode<uint32_t>(pos);
            push(code.immediates, imm);

                        const uint8_t tableidx{*pos++};
            if (tableidx != 0)
                throw parser_error{"invalid tableidx encountered with call_indirect"};
            break;
 }
```

test/unittests/execute_test.cpp
```diff
@@ -85,6 +85,40 @@ TEST(execute, call_with_arguments)
     EXPECT_EQ(ret[0], 0x2);
 }
 
+TEST(execute, call_indirect)
+{
+    /*
+      (type $out-i32 (func (result i32)))
```
axic: I'd refrain from using dashes in names.

gumb0: That's taken from spec tests (a lot of dashes there), but I can change it.

axic: Ah ok, I had a feeling those would be invalid (as they are in most languages). But yeah, since it is s-expressions/lisp inspired, dashes are accepted there.

lib/fizzy/parser_expr.cpp
```diff
@@ -199,6 +199,17 @@ parser_result<Code> parse_expr(const uint8_t* pos)
             break;
         }
 
+        case Instr::call_indirect:
+        {
+            uint32_t imm;
+            std::tie(imm, pos) = leb128u_decode<uint32_t>(pos);
+            push(code.immediates, imm);
+            const uint8_t tableidx{*pos++};
```
axic: Nitpick: can you add an empty line above this?

lib/fizzy/execute.cpp
```diff
@@ -487,27 +512,55 @@ execution_result execute(Instance& instance, FuncIdx func_idx, std::vector<uint6
...
+            if (expected_type.inputs != actual_type.inputs ||
+                expected_type.outputs != actual_type.outputs)
+            {
+                trap = true;
+                goto end;
+            }
+
+            const auto num_inputs = instance.module.typesec[actual_type_idx].inputs.size();
             assert(stack.size() >= num_inputs);
             std::vector<uint64_t> call_args(
```
axic: Why is this done here but not transmitted to `invoke_function`?

gumb0: Oh this shouldn't be here!
some tests are missing...

gumb0:
I should add a test with call_indirect taking some argumens

lib/fizzy/execute.cpp
```diff
@@ -127,6 +127,31 @@ uint64_t eval_constant_expression(ConstantExpression expr,
         return globals[global_idx - imported_globals.size()];
 }
 
+bool invoke_function(
```
axic: I think this should be grouped with `branch()`. Which is actually above the instantiate-helpers.

gumb0: Should I better bring `branch` down here?

axic: Probably, but would prefer that as a separate reordering commit. (We need to be more thorough reviewing so we don't end up needing to move things around all the time.)

lib/fizzy/execute.cpp
gumb0: Added a test that would catch this

lib/fizzy/execute.cpp
gumb0: Added it after branch in "implement commit" and added another reorder commit.

test/unittests/execute_test.cpp
```diff
@@ -85,6 +85,91 @@ TEST(execute, call_with_arguments)
...
+        (type $bin_func (func (param i32 i32) (result i32)))
+        (table anyfunc (elem $f1 $f2 $f3))
+
+        (func $f1 (param i32 i32) (result i32) (i32.add (get_local 0) (get_local 1)))
+        (func $f2 (param i32 i32) (result i32) (i32.sub (get_local 0) (get_local 1)))
+        (func $f3 (param i32) (result i32) (i32.mul (get_local 0) (get_local 0)))
+
+        (func (param i32) (result i32)
+          (i32.const 1)
+          (i32.const 2)
```
axic: So the input param is not taking into account?

lib/fizzy/execute.cpp
chfast: This should return `true` on success. Compare with `load_from_memory()`.

gumb0:
Toggled `invoke_result` return value and rebased.

PR #103, opened 2020-01-28
Add parser validation of the start function

axic:


lib/fizzy/execute.cpp
```diff
@@ -300,7 +300,9 @@ Instance instantiate(const Module& module, std::vector<ImportedFunction> importe
     // Run start function if present
     if (module.startfunc)
     {
-        if (execute(instance, *module.startfunc, {}).trapped)
+        const auto funcidx = *module.startfunc;
+        assert(funcidx <= module.funcsec.size());
```
axic: This is an assert as it is ensured in the parser. However probably we'll need to make this a proper error, depending on the design of parser vs. instantiate.

lib/fizzy/parser.cpp
```diff
@@ -285,6 +285,9 @@ Module parse(bytes_view input)
             throw parser_error{"too many memory sections (at most one is allowed)"};
     }
 
+    if (module.startfunc && (*module.startfunc >= module.funcsec.size()))
```
gumb0: I believe this is not correct, because it doesn't take into account the number of imported functions

axic:
Rebased. @gumb0 do you perhaps want to take this over?

gumb0:
Ok I can finish this.

gumb0:
Ready for review @chfast @axic

gumb0:
Rebased.

axic:
Looks good to me (but can't approve).

PR #106, opened 2020-01-29
cmake: Switch back classic Hunter setup

chfast:


axic:
What is the difference?

chfast:
This was not possible previously: https://github.com/wasmx/fizzy/pull/106/commits/6c23076813b4d16ab58b326c83c62a97b4dcec8f.
You can read more here: https://github.com/cpp-pm/gate.

PR #109, opened 2020-01-30
Implement imported table

gumb0:


axic:
Please rebase.

axic:
@gumb0 can you pull out the first two parser commits? I think we can merge them right away. The rest needs to be discussed.

gumb0:
Rebased.

gumb0:
Rebased, changed `std:;runtime_error`s thrown during instantiation to `instantiate_error`.

gumb0:
Rebased.

gumb0:
> Can you modify the table after instantiation?

I think a host can, if it's imported or exported table.

gumb0:
Rebased and added default initializer to `Instance::table` (required to default-construct `Instance` in `FizzyEngine`)

lib/fizzy/execute.cpp
```diff
@@ -120,6 +151,9 @@ std::vector<TypeIdx> match_imports(const Module& module,
         case ExternalKind::Function:
             imported_function_types.emplace_back(import.desc.function_type_index);
             break;
+        case ExternalKind::Table:
```
chfast: Now the default case cannot happen.

gumb0: Should I replace it with or assertion or better remove it?

chfast: I would leave an assertion. I think this is UB to have an enum object with a value out of definition, but I'm not sure how to exactly handle that. Moreover, there is at least one more case like that with `SectionId`.

gumb0: `SectionId` switch in parser loop (and also switches to for `ExternalKind` in parser) are somewhat different case, because there default case can really happen when parsing and invalid module. Here I'll change it to assertion.

test/unittests/instantiate_test.cpp
```diff
@@ -1,6 +1,7 @@
 #include "execute.hpp"
 #include "limits.hpp"
 #include <gtest/gtest.h>
+#include <lib/fizzy/types.hpp>
```
axic: The other headers are imported via `"execute.hpp"`.

lib/fizzy/execute.cpp
```diff
@@ -451,26 +507,23 @@ Instance instantiate(Module module, std::vector<ImportedFunction> imported_funct
...
-    if (!module.tablesec.empty())
-        table.resize(module.tablesec[0].limits.min);
+    auto table = allocate_table(module.tablesec, imported_tables);
 
     // Allocate memory
     auto [memory, memory_max] = allocate_memory(module.memorysec, imported_memories);
 
     // Fill the table based on elements segment
-    assert(module.elementsec.empty() || !module.tablesec.empty());
+    assert(module.elementsec.empty() || !module.tablesec.empty() || !imported_tables.empty());
```
axic:
```suggestion
    assert(module.elementsec.empty() || table->data() != nullptr);
 ```

gumb0: good idea, changed

PR #111, opened 2020-01-30
Change instantiate() to take Module by value

axic:
Replaces #98.

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -311,7 +311,7 @@ inline uint64_t popcnt64(uint64_t value) noexcept
 }
 }  // namespace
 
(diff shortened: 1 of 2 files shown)


axic:
I kept the following as is because it is only used in tests and I hope we just remove it?

```
// TODO: remove this helper
execution_result execute(const Module& module, FuncIdx func_idx, std::vector<uint64_t> args);
```

chfast:
> I kept the following as is because it is only used in tests and I hope we just remove it?
> 
> ```
> // TODO: remove this helper
> execution_result execute(const Module& module, FuncIdx func_idx, std::vector<uint64_t> args);
> ```

This can at least be move to tests. I usually make it a method of `testing::Test` for clarity. May be done separately.

PR #113, opened 2020-02-02
Add wasm3 to fizzy-bench

axic:
Depends on #127 and #128.

### DIFF
--- a/cmake/ProjectWasm3.cmake
+++ b/cmake/ProjectWasm3.cmake
@@ -1,0 +1,42 @@
+if(ProjectWasm3Included)
+    return()
+endif()
+set(ProjectWasm3Included TRUE)
+
+include(ExternalProject)
+
+set(prefix ${CMAKE_BINARY_DIR}/_deps)
+set(source_dir ${prefix}/src/wasm3)
+set(binary_dir ${prefix}/src/wasm3-build)
+set(include_dir ${source_dir})
+set(wasm3_library ${binary_dir}/source/${CMAKE_STATIC_LIBRARY_PREFIX}m3${CMAKE_STATIC_LIBRARY_SUFFIX})
+
+# Should use a release, like https://github.com/wasm3/wasm3/archive/v0.4.6.tar.gz, once changes are upstreamed.
+
+ExternalProject_Add(wasm3
+    EXCLUDE_FROM_ALL 1
+    PREFIX ${prefix}
+    DOWNLOAD_NAME wasm3-ext-api.tar.gz
+    DOWNLOAD_DIR ${prefix}/downloads
+    SOURCE_DIR ${source_dir}
+    BINARY_DIR ${binary_dir}
+    URL https://github.com/axic/wasm3/archive/ext-api.tar.gz
+    URL_HASH SHA256=bef4e5534737f6054614b80c986dc6c5330178a8d288f1d338aec3296dbd990c
+    CMAKE_ARGS
+    -DCMAKE_INSTALL_PREFIX=<INSTALL_DIR>
+    -DCMAKE_BUILD_TYPE=Release
+    INSTALL_COMMAND ""
+    BUILD_BYPRODUCTS ${wasm3_library}
+)
+
+add_library(wasm3::wasm3 STATIC IMPORTED)
+
+set_target_properties(
+    wasm3::wasm3
+    PROPERTIES
+    IMPORTED_CONFIGURATIONS Release
+    IMPORTED_LOCATION_RELEASE ${wasm3_library}
+    INTERFACE_INCLUDE_DIRECTORIES "${include_dir}"
+)
+
+add_dependencies(wasm3::wasm3 wasm3)
--- a/test/utils/wasm3_engine.cpp
+++ b/test/utils/wasm3_engine.cpp
@@ -1,0 +1,107 @@
+#include "source/wasm3.h"
+#include <test/utils/wasm_engine.hpp>
+#include <cassert>
+#include <cstring>
+
+namespace fizzy::test
+{
+static_assert(sizeof(IM3Function) <= sizeof(WasmEngine::FuncRef));
+
+class Wasm3Engine : public WasmEngine
+{
+    IM3Environment m_env{nullptr};
+    IM3Runtime m_runtime{nullptr};
+
+public:
+    Wasm3Engine() : m_env{m3_NewEnvironment()} {}
+    ~Wasm3Engine()
+    {
+        if (m_runtime)
+            m3_FreeRuntime(m_runtime);
+        m3_FreeEnvironment(m_env);
+    }
+
+    bool parse(bytes_view input) final;
+    std::optional<FuncRef> find_function(std::string_view name) const final;
+    bool instantiate() final;
+    fizzy::bytes_view get_memory() const final;
+    void set_memory(fizzy::bytes_view memory) final;
+    Result execute(FuncRef func_ref, const std::vector<uint64_t>& args) final;
+};
+
+std::unique_ptr<WasmEngine> create_wasm3_engine()
+{
+    return std::make_unique<Wasm3Engine>();
+}
+
+bool Wasm3Engine::parse(bytes_view input)
+{
+    // Replace runtime (e.g. instance + module)
+    if (m_runtime != nullptr)
+        m3_FreeRuntime(m_runtime);
+
+    // The 64k stack size comes from wasm3/platorms/app
+    m_runtime = m3_NewRuntime(m_env, 64 * 1024, nullptr);
+    if (m_runtime == nullptr)
+        return false;
+
+    IM3Module module;
+    if (m3_ParseModule(m_env, &module, input.data(), uint32_t(input.size())) != m3Err_none)
+        return false;
+
+    // Transfers ownership to runtime.
+    if (m3_LoadModule(m_runtime, module) != m3Err_none)
+    {
+        m3_FreeModule(module);
+        return false;
+    }
+
+    return true;
+}
+
+bool Wasm3Engine::instantiate()
+{
+    // Already done in parse (with owernship transfer).
+    return true;
+}
+
+fizzy::bytes_view Wasm3Engine::get_memory() const
+{
+    uint32_t size;
+    auto data = m3_GetMemory(m_runtime, &size, 0);
+    if (data == nullptr)
+        return {};
+    return {data, size};
+}
+
+void Wasm3Engine::set_memory(fizzy::bytes_view memory)
+{
+    if (memory.empty())
+        return;
+
+    uint32_t size;
+    const auto data = m3_GetMemory(m_runtime, &size, 0);
+    std::memcpy(data, memory.data(), size);
+}
+
+std::optional<WasmEngine::FuncRef> Wasm3Engine::find_function(std::string_view name) const
+{
+    IM3Function function;
+    if (m3_FindFunction(&function, m_runtime, name.data()) == m3Err_none)
+        return reinterpret_cast<WasmEngine::FuncRef>(function);
+    return std::nullopt;
+}
+
+WasmEngine::Result Wasm3Engine::execute(
+    WasmEngine::FuncRef func_ref, const std::vector<uint64_t>& args)
+{
+    unsigned ret_valid;
+    uint64_t ret_value;
+    IM3Function function = reinterpret_cast<IM3Function>(func_ref);
+    auto const result = m3_CallProper(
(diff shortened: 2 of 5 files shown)


cmake/ProjectWasm3.cmake
```diff
@@ -0,0 +1,47 @@
...
+endif()
+set(ProjectWasm3Included TRUE)
+
+include(ExternalProject)
+
+set(prefix ${CMAKE_BINARY_DIR}/_deps)
+set(source_dir ${prefix}/src/wasm3)
+set(binary_dir ${prefix}/src/wasm3-build)
+# TODO: make sure to only copy wasm3.h
+set(include_dir ${source_dir}/source)
```
axic: After adding this `/source` here it fails on CI, but it still works (properly) on macos.

cmake/ProjectWasm3.cmake
```diff
@@ -0,0 +1,48 @@
...
+# Should use a release, like https://github.com/wasm3/wasm3/archive/v0.4.6.tar.gz, once changes are upstreamed.
+
+ExternalProject_Add(wasm3
+    EXCLUDE_FROM_ALL 1
+    PREFIX ${prefix}
+    DOWNLOAD_NAME wasm3-ext-api.tar.gz
+    DOWNLOAD_DIR ${prefix}/downloads
+    SOURCE_DIR ${source_dir}
+    BINARY_DIR ${binary_dir}
+    URL https://github.com/axic/wasm3/archive/ext-api.tar.gz
```
axic: This needed API changes to work. Found here: https://github.com/axic/wasm3/tree/ext-api

test/utils/fizzy_engine.cpp
```diff
@@ -11,7 +11,7 @@ class FizzyEngine : public WasmEngine
 
 public:
     void parse(bytes_view input) final;
-    std::optional<uint32_t> find_function(std::string_view name) const final;
+    std::optional<uint32_t> find_function(std::string_view name) final;
```
axic: These won't be needed if #92 is merged.

test/utils/wasm3_engine.cpp
```diff
@@ -0,0 +1,100 @@
...
+}
+
+void Wasm3Engine::parse(bytes_view input)
+{
+    // Replace runtime (e.g. instance + module)
+    if (m_runtime != nullptr)
+        m3_FreeRuntime(m_runtime);
+
+    // The 64k stack size comes from wasm3/platorms/app
+    m_runtime = m3_NewRuntime(m_env, 64 * 1024, NULL);
```
axic: Could move Runtime+LoadModule to instantiate.

cmake/ProjectWasm3.cmake
axic: Apparently there is a C++ API: https://github.com/wasm3/wasm3/blob/master/platforms/cpp/wasm3_cpp/include/wasm3_cpp.h

Could try using it.

axic: Never mind, it is not exposing memory.

cmake/ProjectWasm3.cmake
```diff
@@ -0,0 +1,45 @@
...
+    URL_HASH SHA256=bef4e5534737f6054614b80c986dc6c5330178a8d288f1d338aec3296dbd990c
+    CMAKE_ARGS
+    -DCMAKE_INSTALL_PREFIX=<INSTALL_DIR>
+    -DCMAKE_BUILD_TYPE=Release
+    -DCMAKE_POSITION_INDEPENDENT_CODE=OFF
```
chfast: Not needed.

cmake/ProjectWasm3.cmake
```diff
@@ -0,0 +1,45 @@
...
+    -DCMAKE_CXX_FLAGS=-fvisibility=hidden
```
chfast: Not needed.

cmake/ProjectWasm3.cmake
```diff
@@ -0,0 +1,45 @@
...
+    -DCMAKE_C_FLAGS=-fvisibility=hidden
```
chfast: Not needed.

cmake/ProjectWasm3.cmake
```diff
@@ -0,0 +1,45 @@
...
+    INSTALL_COMMAND ""
```
chfast: Have you tried without this?

cmake/ProjectWasm3.cmake
```diff
@@ -0,0 +1,45 @@
...
+)
+
+add_library(wasm3::wasm3 STATIC IMPORTED)
+
+set_target_properties(
+    wasm3::wasm3
+    PROPERTIES
+    IMPORTED_CONFIGURATIONS Release
+    IMPORTED_LOCATION_RELEASE ${wasm3_library}
+    INTERFACE_INCLUDE_DIRECTORIES "${include_dir};${binary_dir}"
```
chfast: Have you tried without `${binary_dir}`

test/utils/wasm3_engine.cpp
```diff
@@ -0,0 +1,106 @@
+#include "source/wasm3.h"
```
chfast: Have you tried to use include_dir with "source" and then include `<wasm3.h>`?

test/utils/wasm3_engine.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+namespace fizzy::test
+{
+class Wasm3Engine : public WasmEngine
+{
+    IM3Environment m_env{nullptr};
+    IM3Runtime m_runtime{nullptr};
+
+public:
+    Wasm3Engine() : m_env(m3_NewEnvironment()) {}
```
chfast:
```suggestion
    Wasm3Engine() : m_env{m3_NewEnvironment()} {}
```

test/utils/wasm3_engine.cpp
```diff
@@ -0,0 +1,106 @@
...
+}
+
+bool Wasm3Engine::parse(bytes_view input)
+{
+    // Replace runtime (e.g. instance + module)
+    if (m_runtime != nullptr)
+        m3_FreeRuntime(m_runtime);
+
+    // The 64k stack size comes from wasm3/platorms/app
+    m_runtime = m3_NewRuntime(m_env, 64 * 1024, NULL);
```
chfast:
```suggestion
    m_runtime = m3_NewRuntime(m_env, 64 * 1024, nullptr);
```

test/utils/wasm3_engine.cpp
chfast: The `m3_ParseModule()` looks like parsing only. It might be good to only leave this one here.

On the other hand it seems impossible to split parsing and instantiation in wabt, but we might need to merge `parse()` and `instantiate()`.

test/utils/wasm3_engine.cpp
```diff
@@ -0,0 +1,106 @@
...
+bool Wasm3Engine::instantiate()
+{
+    // Already done in parse (with owernship transfer).
+    return true;
+}
+
+fizzy::bytes_view Wasm3Engine::get_memory() const
+{
+    uint32_t size;
+    auto data = m3_GetMemory(m_runtime, &size, 0);
```
chfast:
```suggestion
    const auto data = m3_GetMemory(m_runtime, &size, 0);
```

test/utils/wasm3_engine.cpp
```diff
@@ -0,0 +1,106 @@
...
+}
+
+void Wasm3Engine::set_memory(fizzy::bytes_view memory)
+{
+    if (memory.empty())
+        return;
+
+    uint32_t size;
+    auto data = m3_GetMemory(m_runtime, &size, 0);
+    assert(size == memory.size());
```
chfast: This will not work if the memory grows during execution.

test/utils/wasm3_engine.cpp
```diff
@@ -0,0 +1,106 @@
...
+    WasmEngine::FuncRef func_ref, const std::vector<uint64_t>& args)
+{
+    unsigned ret_valid;
+    uint64_t ret_value;
+    IM3Function function = reinterpret_cast<IM3Function>(func_ref);
+    auto const result = m3_CallProper(
+        function, static_cast<uint32_t>(args.size()), args.data(), &ret_valid, &ret_value);
+    if (result == m3Err_none)
+        return {false, ret_valid ? ret_value : std::optional<uint64_t>{}};
+    return {true, std::optional<uint64_t>{}};
```
chfast:
```suggestion
    return {true, std::nullopt};
```
or just `{}`.

test/utils/wasm3_engine.cpp
axic: Btw I tried this split, but it was crashing. Didn't bother figuring out why it crashed.

test/utils/wasm3_engine.cpp
axic: I had `set(include_dir ${source_dir}/source)` but that caused a cmake configuration error claiming no such directory. Initially it worked but not after a while.

test/utils/wasm3_engine.cpp
axic: Hm I think I'll go with nullopt for clarity.

cmake/ProjectWasm3.cmake
axic:
```
[ 38%] Performing install step for 'wasm3'
make[3]: *** No rule to make target `install'.  Stop.
```

cmake/ProjectWasm3.cmake
axic: Seems to work without.

cmake/ProjectWasm3.cmake
axic: Seems to work without.

cmake/ProjectWasm3.cmake
axic: Seems to work without.

cmake/ProjectWasm3.cmake
axic: Seems to work without.

test/utils/wasm3_engine.cpp
axic:
```
-- Configuring done
CMake Error in test/utils/CMakeLists.txt:
  Imported target "wasm3::wasm3" includes non-existent path

    "/Users/alex/Projects/fizzy/b/_deps/src/wasm3/source"

  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

  * The path was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and references files it does not
  provide.
```

test/utils/wasm3_engine.cpp
```diff
@@ -0,0 +1,105 @@
...
+    m_runtime = m3_NewRuntime(m_env, 64 * 1024, nullptr);
+    if (m_runtime == nullptr)
+        return false;
+
+    IM3Module module;
+    if (m3_ParseModule(m_env, &module, input.data(), uint32_t(input.size())) != m3Err_none)
+        return false;
+
+    // Transfers ownership to runtime.
+    if (m3_LoadModule(m_runtime, module) != m3Err_none)
```
gumb0: Don't creating runtime and loading module belong more to `instantiate`?

axic: See https://github.com/wasmx/fizzy/pull/113#discussion_r375894251

But for some reason it is crashing. Likely some ownership issue internally.

test/utils/wasm3_engine.cpp
chfast: I will try it later.

test/utils/wasm3_engine.cpp
```diff
@@ -0,0 +1,105 @@
...
+    return std::nullopt;
+}
+
+WasmEngine::Result Wasm3Engine::execute(
+    WasmEngine::FuncRef func_ref, const std::vector<uint64_t>& args)
+{
+    unsigned ret_valid;
+    uint64_t ret_value;
+    IM3Function function = reinterpret_cast<IM3Function>(func_ref);
+    auto const result = m3_CallProper(
```
gumb0: I can't find definition of `m3_CallProper`...

axic: See https://github.com/wasmx/fizzy/pull/113#discussion_r375840944

gumb0:
Would `static_assert(sizeof(IM3Function) <= sizeof(FuncRef))` be useful?

axic:
> Would static_assert(sizeof(IM3Function) <= sizeof(FuncRef)) be useful?

Yeah, good idea.

axic:
@gumb0 added

PR #114, opened 2020-02-02
Use find_exported_function in end_to_end_test

axic:


test/unittests/end_to_end_test.cpp
```diff
@@ -125,8 +126,11 @@ TEST(end_to_end, nested_loops_in_c)
 
     const auto module = parse(bin);
 
+    const auto funcidx = find_exported_function(module, "test");
```
chfast: I still have a question, why not `func_idx` which I find more readable.

axic: Changed.

PR #115, opened 2020-02-03
Wasm engine abstraction for benchmarking

chfast:


test/utils/wasm_engine.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+namespace fizzy
+{
+struct execution_result2
+{
+    bool trapped = false;
+    std::optional<uint64_t> value;
+};
+
+class WasmEngine
```
gumb0: Needs virtual destructor

gumb0: I feel a bit uneasy that it actually encapsulates one instance, but there's no indication of it in the name...

test/utils/wasm_engine.hpp
```diff
@@ -0,0 +1,28 @@
+#pragma once
+
+#include "types.hpp"
+#include <memory>
+#include <optional>
+
+namespace fizzy
+{
+struct execution_result2
```
gumb0: Maybe this should be `execution_result` in another (nested?) namespace?

test/bench/bench.cpp
```diff
@@ -132,11 +138,16 @@ void load_benchmark(const fs::path& path, const std::string& name_prefix)
     const auto wasm_binary = std::make_shared<const fizzy::bytes>(
         std::istreambuf_iterator<char>{wasm_file}, std::istreambuf_iterator<char>{});
 
-    register_benchmark("parse/" + base_name,
-        [wasm_binary](benchmark::State& state) { benchmark_parse(state, *wasm_binary); });
+    auto& engine = *g_engines[0];
```
gumb0: Also it doesn't feel very nice that all of the benchmarks operate on the same global data. Maybe better do a separate `create_fizzy_engine()` for each and move into lambdas?

gumb0:
Also note my changes to make memory optional in Instance https://github.com/wasmx/fizzy/pull/117

I thinks it shouldn't be very difficult to reconcile with this PR.

test/utils/wasm_engine.hpp
chfast: Is `TestInstance` a better name?

test/utils/wasm_engine.hpp
chfast: Number of options here, in order of my preferences:
1. Use `std::optional` in the original `execute_result`.
2. Use original `execute_result` here.
3. Rename to something else. Can be `fizzy::test::execution_result` but I expect it might be confusing to have `fizzy::execution_result` and `fizzy::test::execution_result`.

Still, adding all test utils, including this one into `fizzy::test` sound like good idea.

test/bench/bench.cpp
chfast: This will not scale with multiple engines. We need global registry of these. This is the simplest implementation, the global variable does not leave this file.

The change I'd like to make to it is to add factory functions (e.g. `create_fizzy_engine`) to it, instead of engine objects. The we can make the registry compile-time constant. Will that work for you?

gumb0: Yeah array of factory functions would be ok :+1:

test/utils/wasm_engine.hpp
gumb0: I think it's ok

test/utils/fizzy_engine.cpp
```diff
@@ -0,0 +1,64 @@
...
+}
+
+fizzy::bytes_view FizzyEngine::get_memory() const
+{
+    return {m_instance.memory.data(), m_instance.memory.size()};
+}
+
+void FizzyEngine::set_memory(fizzy::bytes_view memory)
+{
+    m_instance.memory = memory;
```
axic: You'll find that in many cases this will require an exported memory section from the benchmark to have the memory actually accessible in a VM.

I think we'll ensure that in `instantiate`, so have to complaints about the API.

test/utils/wasm_engine.hpp
axic: Why not nest it inside `WasmEngine`? `WasmEngine::Result` ?

test/utils/fizzy_engine.cpp
```diff
@@ -0,0 +1,64 @@
...
+{
+    return "fizzy";
+}
+
+void FizzyEngine::parse(bytes_view input)
+{
+    m_instance.module = fizzy::parse(input);
+}
+
+void FizzyEngine::instantiate()
```
axic: What happens when `instantiate()` is called twice?

test/utils/fizzy_engine.cpp
chfast: I don't understand what you mean here.

axic: Memory is not exported (made available) by wasm engines, unless the bytecode has an export memory statement. Fizzy just exposes things unconditionally currently.

chfast: What's the problem though?

test/utils/wasm_engine.hpp
chfast: I'm not convinced though.

test/utils/wasm_engine.hpp
```diff
@@ -0,0 +1,48 @@
...
+    struct Result
+    {
+        bool trapped = false;
+        std::optional<uint64_t> value;
+    };
+
+    virtual ~WasmEngine() noexcept = default;
+
+    /// Parses input bytes and creates internal wasm module out of it.
+    /// Consecutive invokes replaces the internal module with a new one.
```
axic:
```suggestion
    /// Consecutive invokes replaces the internal module (but not the instance) with a new one.
```

chfast: I guess we should say it also invalidates the instance. Some things will not work as expected.

test/utils/fizzy_engine.cpp
axic: Nothing with `FizzyEngine`.

test/utils/fizzy_engine.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+std::optional<uint32_t> FizzyEngine::find_function(std::string_view name) const
+{
+    return fizzy::find_exported_function(m_instance.module, name);
+}
+
+WasmEngine::Result FizzyEngine::execute(FuncIdx func_idx, const std::vector<uint64_t>& args)
+{
+    const auto [trapped, result_stack] = fizzy::execute(m_instance, func_idx, args);
+    assert(!trapped);
```
axic: Why have `trapped` in result if we assert here?

test/utils/fizzy_engine.cpp
```diff
@@ -0,0 +1,58 @@
...
+    assert(result_stack.size() <= 1);
+    return {trapped, !result_stack.empty() ? result_stack.back() : std::optional<uint64_t>{}};
```
axic: `std::nullopt` ?

test/utils/fizzy_engine.cpp
chfast: This has predated the `Result` type with `trapped.`

test/utils/fizzy_engine.cpp
chfast: Will not work for `?:` operator.

test/utils/wasm_engine.hpp
```diff
@@ -0,0 +1,48 @@
...
+    struct Result
+    {
+        bool trapped = false;
+        std::optional<uint64_t> value;
+    };
+
+    virtual ~WasmEngine() noexcept = default;
+
+    /// Parses input bytes and creates internal wasm module out of it.
+    /// Consecutive invokes replace the internal module and invalidate the internal instance.
```
axic: Did you actually mean "invalidate the internal instance" ? FizzyEngine doesn't seem to do that.

chfast: Yes. There is nothing to do here. It means you are not allowed to use it any more and you need to call `instantiate()` again.

axic: Right. We agreed on in that in a previous comment.

test/utils/wasm_engine.hpp
```diff
@@ -0,0 +1,48 @@
...
+    virtual void parse(bytes_view input) = 0;
+
+    /// Finds an exported function in the internal instance.
+    /// Requires parse().
+    virtual std::optional<FuncIdx> find_function(std::string_view name) const = 0;
+
+    /// Instantiates the internal module creating an internal instance.
+    /// Consecutive invokes replaces the internal instance with a new one.
```
axic:
```suggestion
    /// Consecutive invokes replace the internal instance with a new one.
```

test/bench/bench.cpp
```diff
@@ -59,54 +73,60 @@ struct ExecutionBenchmarkCase
...
-                std::end(benchmark_case.expected_memory), std::begin(instance.memory)))
+                std::end(benchmark_case.expected_memory), std::begin(mem)))
             return state.SkipWithError("Incorrect result memory");
     }
 
     for (auto _ : state)  // NOLINT(clang-analyzer-deadcode.DeadStores)
     {
-        instance.memory = initial_memory;  // Restore initial memory.
-        const auto result = fizzy::execute(instance, *func_idx, benchmark_case.func_args);
+        engine->set_memory(initial_memory);
```
axic: Perhaps a note here would be useful: "Resetting" the instance. This works, as long as we're not using globals or imports.

axic:
@chfast left two more comments and I think this should be squashed somewhat. Otherwise looks good to me to get this moving.

test/bench/bench.cpp
```diff
@@ -59,54 +73,65 @@ struct ExecutionBenchmarkCase
...
+                return state.SkipWithError("Missing result value");
+            else if (*result.value != *benchmark_case.expected_result)
                 return state.SkipWithError("Incorrect result");
         }
-        else if (!result.stack.empty())
+        else if (result.value)
             return state.SkipWithError("Unexpected result");
 
-        if (instance.memory.size() < benchmark_case.expected_memory.size())
+        const auto mem = engine->get_memory();
```
axic: Final nit: perhaps use `memory` (as we use long form mostly in the codebase). But I don't care.

axic:
Tried implement wasm3 in this and there's one issue (up til now): `find_function` doesn't return an index, but an opaque struct. We should change the API in a separate PR though, I'd prefer to merge this as is and then do adjustments after you did wabt and I did wasm3.

test/bench/bench.cpp
```diff
@@ -59,54 +73,65 @@ struct ExecutionBenchmarkCase
...
     if (!func_idx)
     {
         return state.SkipWithError(
             ("Function \"" + benchmark_case.func_name + "\" not found").c_str());
     }
 
-    auto instance = fizzy::instantiate(module, {});
+    engine->instantiate();
+
+    auto initial_memory = fizzy::bytes{engine->get_memory()};
```
gumb0: Can be `const`

test/utils/wasm_engine.hpp
gumb0: "invocations" is probably more correct substantive

test/bench/bench.cpp
gumb0: Ah no, not true, sorry

test/bench/bench.cpp
```diff
@@ -59,54 +73,65 @@ struct ExecutionBenchmarkCase
...
                 return state.SkipWithError("Incorrect result");
         }
-        else if (!result.stack.empty())
+        else if (result.value)
             return state.SkipWithError("Unexpected result");
 
-        if (instance.memory.size() < benchmark_case.expected_memory.size())
+        const auto mem = engine->get_memory();
+        if (mem.size() < benchmark_case.expected_memory.size())
             return state.SkipWithError("Result memory is shorted than expected");
```
gumb0: BTW typo here - "shorter"

test/bench/bench.cpp
```diff
@@ -132,12 +157,20 @@ void load_benchmark(const fs::path& path, const std::string& name_prefix)
     const auto wasm_binary = std::make_shared<const fizzy::bytes>(
         std::istreambuf_iterator<char>{wasm_file}, std::istreambuf_iterator<char>{});
 
-    register_benchmark("parse/" + base_name,
-        [wasm_binary](benchmark::State& state) { benchmark_parse(state, *wasm_binary); });
-
-    register_benchmark("instantiate/" + base_name,
-        [wasm_binary](benchmark::State& state) { benchmark_instantiate(state, *wasm_binary); });
+    for (const auto entry : engine_registry)  // Register parse benchmarks.
```
gumb0: Could be reference, but perhaps not big difference
```suggestion
    for (const auto& entry : engine_registry)  // Register parse benchmarks.
```

test/utils/wasm_engine.hpp
```diff
@@ -0,0 +1,48 @@
...
+    {
+        bool trapped = false;
+        std::optional<uint64_t> value;
+    };
+
+    virtual ~WasmEngine() noexcept = default;
+
+    /// Parses input bytes and creates internal wasm module out of it.
+    /// Consecutive invocations replace the internal module and invalidate the internal instance.
+    virtual void parse(bytes_view input) = 0;
```
axic: Is this different to `fizzy::bytes_view`? If not I'd prefer the explicit path to match the rest of the header.

chfast:
> There seems to be a crash when using multiple engines.

Fixed.

axic:
@chfast can you squash some commits?

test/utils/wasm_engine.hpp
axic: @chfast ping

chfast: It's the same. It should rather be `bytes_view` in other places.

axic: Are you fixing the others then?

chfast: Changed.

PR #117, opened 2020-02-03
Don't allocate empty memory when none is defined or imported

gumb0:
Module is allowed to not have any memory (then all memory instructions are invalid)
https://github.com/WebAssembly/spec/blob/3b2489d7e570ca0f8d32e9d0b1a1e8255c4212b7/test/core/memory.wast#L22-L25

### DIFF
--- a/test/utils/fizzy_engine.cpp
+++ b/test/utils/fizzy_engine.cpp
@@ -51,11 +51,18 @@ bool FizzyEngine::instantiate()
 
 bytes_view FizzyEngine::get_memory() const
 {
+    if (!m_instance.memory)
+        return {};
+
     return {m_instance.memory->data(), m_instance.memory->size()};
 }
 
 void FizzyEngine::set_memory(bytes_view memory)
 {
+    if (memory.empty())
+        return;
+
+    assert(m_instance.memory != nullptr);
     *m_instance.memory = memory;
 }
 
--- a/test/utils/wasm3_engine.cpp
+++ b/test/utils/wasm3_engine.cpp
@@ -81,6 +81,7 @@ void Wasm3Engine::set_memory(fizzy::bytes_view memory)
 
     uint32_t size;
     const auto data = m3_GetMemory(m_runtime, &size, 0);
+    assert(data != nullptr);
     std::memcpy(data, memory.data(), size);
 }
 
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -241,7 +241,7 @@ std::tuple<bytes_ptr, size_t> allocate_memory(const std::vector<Memory>& module_
     }
     else
     {
-        bytes_ptr memory{new bytes{}, bytes_delete};
+        bytes_ptr memory{nullptr, null_delete};
         return {std::move(memory), MemoryPagesLimit};
     }
 }
--- a/test/unittests/instantiate_test.cpp
+++ b/test/unittests/instantiate_test.cpp
@@ -335,8 +335,7 @@ TEST(instantiate, memory_default)
 
     auto instance = instantiate(module);
 
-    ASSERT_EQ(instance.memory->size(), 0);
-    EXPECT_EQ(instance.memory_max_pages * PageSize, 256 * 1024 * 1024);
+    EXPECT_FALSE(instance.memory);
 }
 
 TEST(instantiate, memory_single)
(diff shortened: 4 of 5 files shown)


test/utils/fizzy_engine.cpp
```diff
@@ -51,11 +51,17 @@ bool FizzyEngine::instantiate()
...
 {
+    if (!m_instance.memory)
+        return {};
+
     return {m_instance.memory->data(), m_instance.memory->size()};
 }
 
 void FizzyEngine::set_memory(bytes_view memory)
 {
+    if (!m_instance.memory)
```
axic: The other implementations only have a check for `!memory.empty()`.

gumb0: Why? I think I'll add this check to other engines, too.

axic: Because I think it assumes (and this is how bench is implemented) that if `get_memory` failed then it aborts.

test/utils/wasm3_engine.cpp
```diff
@@ -81,6 +81,9 @@ void Wasm3Engine::set_memory(fizzy::bytes_view memory)
 
     uint32_t size;
     const auto data = m3_GetMemory(m_runtime, &size, 0);
+    if (data == nullptr)
+        return;
```
axic: I consider these a failure. Non-empty memory was supplied but it can't be overriden.

axic: I don't mind agreeing on this though, @chfast ?

gumb0: Properly failing would be better than crashing, so we could make it return `false` in this case.

chfast: I think it should work like this:
1. If the instance has memory, it should be reset accordingly to the `memory` argument. This should be done also for `memory` being empty (current checks are wrong here).
2. If the instance has no memory AND the `memory` argument is not empty -> failure.

gumb0: Hm looks like wasm3's API doesn't allow to reset the memory to empty... Best i can do is overwrite it with zeroes.

test/utils/wasm3_engine.cpp
```diff
@@ -74,14 +74,18 @@ fizzy::bytes_view Wasm3Engine::get_memory() const
...
-    if (memory.empty())
-        return;
-
     uint32_t size;
     const auto data = m3_GetMemory(m_runtime, &size, 0);
-    std::memcpy(data, memory.data(), size);
+    if (data == nullptr)
+        return memory.empty();
+
+    if (memory.data() == nullptr)
```
axic: Why `data() == nullptr` here, while `empty()` elsewhere?

Also not sure if `empty()` means to zero out memory, to be honest.

gumb0: `memory.data() == nullptr` is supposed to check whether it was called like `set_memory(bytes_view{})`. But it can be `memory.empty()`, too.

`empty()` doesn't zero put (it's a view anyway, can't be zeroed out), it checks whether it's empty.
(BTW both `bytes_view{}` and `bytes_view(ptr, 0)` are empty, but second one has non-null `data()`)

test/utils/wasm3_engine.cpp
```diff
@@ -74,14 +74,18 @@ fizzy::bytes_view Wasm3Engine::get_memory() const
...
+        std::memset(data, 0, size);
+    else
+        std::memcpy(data, memory.data(), size);
```
gumb0: This is not correct though when `memory` has fewer when `size` bytes.
Will rewrite it a bit.

gumb0:
Changed `Wasm3Engine::set_memory` implementation. API doesn't allow to change the memory size, so we zero out all of it and then copy given bytes.

test/utils/fizzy_engine.cpp
chfast: In context of benchmarking, I'm not sure returning `bool` is useful. It can only help when adding new benchmarks to detect incompatibilities with this API. But that would be noticed anyway by comparing final results.

In this PR, I would avoid fixing issues with `.set_memory()` provided that current benchmarks works. Maybe we should rework the interface instead of fixing current one.

gumb0: I think we can leave in this PR just assertions in `set_memory` that memory exists.

gumb0:
Now I don't change semantics of `set_memory`, empty input means "don't change" and doesn't fail if memory is absent. When input is non-empty, assertion checks assumption that memory exists.

PR #119, opened 2020-02-03
Make instantiation fail in case element or data section doesn't fit table or memory

gumb0:
This is required by the instantiation algorithm in the spec https://webassembly.github.io/spec/core/exec/modules.html#instantiation

### DIFF
--- a/test/unittests/instantiate_test.cpp
+++ b/test/unittests/instantiate_test.cpp
@@ -263,6 +263,19 @@ TEST(instantiate, element_section_offset_from_mutable_global)
         "Constant expression can use global_get only for const globals.");
 }
 
+TEST(instantiate, element_section_offset_too_large)
+{
+    Module module;
+    module.tablesec.emplace_back(Table{{3, std::nullopt}});
+    module.elementsec.emplace_back(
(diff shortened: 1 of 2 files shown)


test/unittests/instantiate_test.cpp
```diff
@@ -263,6 +263,19 @@ TEST(instantiate, element_section_offset_from_mutable_global)
         "Constant expression can use global_get only for const globals.");
 }
 
+TEST(instantiate, element_section_too_big)
```
axic: `element_section_offset_too_large`?

test/unittests/instantiate_test.cpp
```diff
@@ -318,6 +331,17 @@ TEST(instantiate, data_section_offset_from_mutable_global)
         "Constant expression can use global_get only for const globals.");
 }
 
+TEST(instantiate, data_section_too_big)
```
axic: `data_section_offset_too_large`?

test/unittests/instantiate_test.cpp
gumb0: renamed

axic:
Oops, I thought @chfast reviewed this already. Please shout if there's something wrong with it.

PR #120, opened 2020-02-04
Make LabelPosition a struct with instruction field

chfast:
This should help with `if/else` implementation - allows inspecting what instruction has created a label.

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -11,10 +11,10 @@ namespace
 {
 struct LabelContext
 {
-    const Instr* pc = nullptr;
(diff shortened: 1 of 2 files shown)


lib/fizzy/execute.cpp
```diff
@@ -11,10 +11,10 @@ namespace
 {
 struct LabelContext
 {
-    const Instr* pc = nullptr;
-    const uint8_t* immediate = nullptr;
-    size_t arity = 0;
-    size_t stack_height = 0;
+    const Instr* pc = nullptr;           ///< The jump target instruction
```
```suggestion
    const Instr* pc = nullptr;           ///< The jump target instruction.
```

Issue #110, opened 2020-01-30
Running spec test suite


gumb0:
Spec test suite: https://github.com/WebAssembly/spec/tree/master/test/core

Converting *.wast files with a [wast2json tool](https://webassembly.github.io/wabt/doc/wast2json.1.html) produces wasm files + json files like these:
https://gist.github.com/gumb0/a826ed24a6f16e5a6e88c7a9a88a8b3b
https://gist.github.com/gumb0/f7f3e443ced081fd1902c67881d0ac31

Basically for each command in json file we need to do one of the following:
- [x] `module` - instantiate module from file (it is then used in the following commands)
  - [x] when instantiate failed, skip following tests
  - [x] provide imports to `instantiate` when required (link exports of registered modules)
  - [x] `spectest` module provided by test harness for importing in tests
- [x] `register` - register module to be used in the following commands for linking exports-imports of several modules (used in linking.wast, imports.wast, elem.wast)
- actions
  - [x] `invoke` an exported function (by name)
  - [x] `get` exported global value (by name)
  - [x] choosing module for action by name (many cases in linking.json)
- assertions - check result of an action or module instantiation
  - [x] `assert_return`
    - [x] support case when no returned value expected (see break-drop.wast)
  - [x] `assert_trap`
  - [x] `assert_exhaustion`
  - [x] `assert_malformed` (support needed only for `"module_type": "binary"`)
  - [x] `assert_invalid`
  - [x] `assert_unlinkable`
  - [x] `assert_uninstantiable` (in start.json and linking.json; in original wast it's `assert_trap` with `module` inside)
- [x] `action` without assert (see memory_redundancy.wast)

The spec for all possible asserts/actions/commands is at https://github.com/WebAssembly/spec/tree/master/interpreter#scripts
(all of the assertions mentioned there are used in test suite)

PR #121, opened 2020-02-04
Add spectest runner tool

axic:
Part of #110

Commands that can be used to convert all *.wast files in current dir to json/wasm files in separate subdirs:
```
mkdir json
for f in *.wast; do mkdir json/${f%.*}; wast2json $f -o json/${f%.*}/${f%.*}.json; done
```

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -1,3 +1,6 @@
+set(fizzy_include_dir ${PROJECT_SOURCE_DIR}/lib/fizzy)
+
 add_subdirectory(utils)
 add_subdirectory(bench)
+add_subdirectory(spectests)
 add_subdirectory(unittests)
--- a/test/spectests/CMakeLists.txt
+++ b/test/spectests/CMakeLists.txt
@@ -1,0 +1,17 @@
+if(HUNTER_ENABLED)
+    hunter_add_package(nlohmann_json)
+endif()
+find_package(nlohmann_json REQUIRED)
+
+add_executable(fizzy-spectests spectests.cpp)
+target_compile_features(fizzy-spectests PRIVATE cxx_std_17)
+target_link_libraries(fizzy-spectests PRIVATE fizzy::fizzy fizzy::test-utils nlohmann_json::nlohmann_json)
+target_include_directories(fizzy-spectests PRIVATE ${fizzy_include_dir})
+
+if(UNIX AND NOT APPLE)
+    # For libstdc++ up to version 8 (included) this is needed for proper <filesystem> support.
+    # Otherwise, the result program will crash on first use (no linker errors).
+    # For libstdc++ 9, this is not needed, but cause no harm.
+    # Clang compiler on linux may select libstdc++8 in some configurations.
+    target_link_libraries(fizzy-spectests PRIVATE stdc++fs)
+endif()
--- a/test/spectests/spectests.cpp
+++ b/test/spectests/spectests.cpp
@@ -18,134 +18,168 @@ uint64_t json_to_value(const json& v)
     return static_cast<uint64_t>(static_cast<T>(std::stoull(v.get<std::string>())));
 }
 
-void run_tests_from_file(const fs::path& path)
+class test_runner
 {
-    std::cout << "Running tests from " << path << "\n";
-
-    std::ifstream test_file{path};
-    const json j = json::parse(test_file);
-
-    fizzy::Instance instance;
-    for (const auto& cmd : j.at("commands"))
+public:
+    void run_from_file(const fs::path& path)
     {
-        const auto type = cmd.at("type").get<std::string>();
+        std::cout << "Running tests from " << path << "\n";
 
-        std::cout << "Line " << cmd.at("line").get<int>() << ": " << type << " " << std::flush;
+        std::ifstream test_file{path};
+        const json j = json::parse(test_file);
 
-        if (type == "module")
+        for (const auto& cmd : j.at("commands"))
         {
-            const auto filename = cmd.at("filename").get<std::string>();
-            std::cout << "Instantiating " << filename << "\n";
+            const auto type = cmd.at("type").get<std::string>();
 
-            std::ifstream wasm_file{fs::path{path}.replace_filename(filename)};
+            std::cout << "Line " << cmd.at("line").get<int>() << ": " << type << " " << std::flush;
 
-            const auto wasm_binary = fizzy::bytes(
-                std::istreambuf_iterator<char>{wasm_file}, std::istreambuf_iterator<char>{});
+            if (type == "module")
+            {
+                const auto filename = cmd.at("filename").get<std::string>();
+                std::cout << "Instantiating " << filename << "\n";
 
-            try
-            {
-                instance = fizzy::instantiate(fizzy::parse(wasm_binary));
-            }
-            catch (const fizzy::parser_error& ex)
-            {
-                std::cout << "FAILED Parsing failed with error: " << ex.what() << "\n";
-                continue;
-            }
-            catch (const fizzy::instantiate_error& ex)
-            {
-                std::cout << "FAILED Instantiation failed with error: " << ex.what() << "\n";
-                continue;
-            }
-        }
-        else if (type == "assert_return")
-        {
-            const auto& action = cmd.at("action");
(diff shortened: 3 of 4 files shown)


test/spectests/spectests.cpp
```diff
@@ -0,0 +1,183 @@
...
+                std::istreambuf_iterator<char>{wasm_file}, std::istreambuf_iterator<char>{});
+
+            try
+            {
+                instance = fizzy::instantiate(fizzy::parse(wasm_binary));
+            }
+            catch (fizzy::parser_error const&)
+            {
+                std::cout << "FAILED Parsing failed.\n";
+                continue;
```
gumb0: I would rethrow here, to stop the execution, because the following tests might not make sense in this case

gumb0: Better to skip the following tests till next module, I think that will be easy when we support instantiating more than one module for some tests.

test/spectests/spectests.cpp
```diff
@@ -0,0 +1,281 @@
...
+class test_runner
+{
+public:
+    void run_from_file(const fs::path& path)
+    {
+        std::cout << "Running tests from " << path << "\n";
+
+        std::ifstream test_file{path};
+        json j;
+        test_file >> j;
```
axic: Perhaps if there's no non-stream version of this, I'd introduce a tiny helper `json parse_json_from_file(path)` for readability. But this is just a nitpick.

test/spectests/spectests.cpp
```diff
@@ -0,0 +1,290 @@
...
+
+        for (const auto& cmd : j.at("commands"))
+        {
+            const auto type = cmd.at("type").get<std::string>();
+
+            std::cout << "Line " << cmd.at("line").get<int>() << ": " << type << " " << std::flush;
```
axic: Since there are pass/skip/fail helpers maybe it is worth considering a `log` or `info` helper so that the output stream handling is done at a single place.

test/spectests/spectests.cpp
```diff
@@ -0,0 +1,290 @@
...
+                        skip("Unsupported expected type '" + expected_type + "'.");
+                        continue;
+                    }
+
+                    if (expected_value != actual_value)
+                    {
+                        std::stringstream message;
+                        message << "Incorrect returned value. Expected: " << expected_type << " (0x"
+                                << std::hex << expected_value << ") Actual: " << std::dec
+                                << actual_value << " (0x" << std::hex << actual_value << std::dec
```
axic: Why not use the `to_hex` helpers?

test/spectests/spectests.cpp
```diff
@@ -0,0 +1,290 @@
...
+                    fail("Function expected to trap, but it didn't.");
+                    continue;
+                }
+
+                pass();
+            }
+            else if (type == "assert_invalid")
+            {
+                const auto filename = cmd.at("filename").get<std::string>();
+                std::ifstream wasm_file{fs::path{path}.replace_filename(filename)};
```
axic: This wasm file loading is done in multiple places so perhaps a helper would be due: `bytes_view load_wasm_file(filename)`?

test/spectests/spectests.cpp
gumb0: They are for `bytes`, not for values

test/spectests/spectests.cpp
axic: Maybe even including the parser stage?

test/spectests/spectests.cpp
axic: We actually have a helper for bytes_view: `string hex(bytes_view)`

test/spectests/spectests.cpp
```diff
@@ -0,0 +1,290 @@
...
+                }
+                catch (const fizzy::parser_error& ex)
+                {
+                    std::cout << "Parsing failed with error: " << ex.what() << "\n";
+                    continue;
+                }
+                catch (const fizzy::instantiate_error& ex)
+                {
+                    std::cout << "Instantiation failed with error: " << ex.what() << "\n";
+                    continue;
```
axic: As you said probably makes sense to make both of these `fail()`.

test/spectests/spectests.cpp
gumb0: Added this, but now this requires some additional `to_string` calls

axic:
@gumb0 can you squash some of the commits? I think we should merge it with the current feature set.

gumb0:
Should be ready for review.

test/spectests/spectests.cpp
```diff
@@ -0,0 +1,294 @@
...
+                catch (const fizzy::parser_error& ex)
+                {
+                    fail(std::string{"Parsing failed with error: "} + ex.what());
+                    continue;
+                }
+                catch (const fizzy::instantiate_error& ex)
+                {
+                    fail(std::string{"Instantiation failed with error: "} + ex.what());
+                    continue;
+                }
```
gumb0: Forgot to count it here as `pass()`, will fix.

test/spectests/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+if(HUNTER_ENABLED)
+    hunter_add_package(nlohmann_json)
+endif()
+find_package(nlohmann_json REQUIRED)
+
+set(fizzy_include_dir ${PROJECT_SOURCE_DIR}/lib/fizzy)
```
chfast: Set this in `test/CMakeList.txt`.

test/spectests/spectests.cpp
```diff
@@ -0,0 +1,294 @@
...
+{
+constexpr auto JsonExtension = ".json";
+
+template <typename T>
+uint64_t json_to_value(const json& v)
+{
+    return static_cast<uint64_t>(static_cast<T>(std::stoull(v.get<std::string>())));
+}
+
+fizzy::bytes load_wasm_file(const fs::path& path, std::string_view filename)
```
chfast: Parameters names are confusing. What is `path`?

test/spectests/spectests.cpp
```diff
@@ -0,0 +1,294 @@
...
+    int failed = 0;
+    int skipped = 0;
+};
+
+void run_tests_from_dir(const fs::path& path)
+{
+    std::vector<fs::path> subdirs;
+    std::vector<fs::path> files;
+
+    for (const auto& e : fs::directory_iterator{path})
```
chfast: Have you considered `recursive_directory_iterator`?

test/spectests/spectests.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+class test_runner
+{
+public:
+    void run_from_file(const fs::path& path)
+    {
+        log("Running tests from " + path.string());
+
+        std::ifstream test_file{path};
+        json j = json::parse(test_file);
```
chfast: Can be const?

test/spectests/spectests.cpp
chfast: Maybe `copy_if(recursive_directory_iterator{}, vector, is_json_file)`, then `sort()`.

test/CMakeLists.txt
```diff
@@ -1,3 +1,4 @@
 add_subdirectory(utils)
 add_subdirectory(bench)
-add_subdirectory(unittests)
\ No newline at end of file
+add_subdirectory(unittests)
+add_subdirectory(spectests)
```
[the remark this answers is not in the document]
gumb0: I put `spectests` before `unittests` and didn't move `utils`, because supposedly `bench` depends on it. But actually putting `utils` on the last line works for me, too.

test/spectests/CMakeLists.txt
gumb0: Moved to test `test/CMakeList.txt` and also removed this line from `bench`

test/spectests/spectests.cpp
gumb0: renamed

test/spectests/spectests.cpp
gumb0: With `copy_if` it's `begin(iterator)` + `end(iterator)` + `back_inserter` + lambda, so I preferred three lines of good old range for.

test/spectests/spectests.cpp
gumb0: Yes, changed.

PR #122, opened 2020-02-05
ci: Add benchmark job

chfast:
This uses virtual machine based executor (Ubuntu 16.04) and takes binaries from linux release build.
Using virtual machine gives more stable results comparing to docker jobs.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -90,6 +90,10 @@ jobs:
     steps:
       - build
       - test
+      - persist_to_workspace:
+          root: ~/build
+          paths:
+            - bin/fizzy-bench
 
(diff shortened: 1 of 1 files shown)


axic:
Is this only to ensure it runs (and are visible on ci)? Or are the results stored and compared?

chfast:
> Is this only to ensure it runs (and are visible on ci)? Or are the results stored and compared?

This is only single run to check stddev values.

Storing and comparing does not really make sense, because values differ a lot between runs. But what I want to implement later is to also build the PR base commit and compare results.

This can be merged later.

PR #123, opened 2020-02-05
Fix shr_s

chfast:


axic:
rotl/rotr is likely affected by the same bug

gumb0:
> rotl/rotr is likely affected by the same bug

They are not, because they are used only with unsigned types, but better to change it in `rotl` \ `rotr` functions, too, to make them correct templates.

chfast:
> > rotl/rotr is likely affected by the same bug
> 
> They are not, because they are used only with unsigned types, but better to change it in `rotl` \ `rotr` functions, too, to make them correct templates.

I started rewriting those. They are not affected because they only make sense for unsigned types, but I will still make them looks similar and will add more tests.

chfast:
It would be nice to move arithmetic tests to separate file. The `execute_test.cpp` is too big and I have issues with editing it. Any proposal how to perform the moving?

test/unittests/execute_test.cpp
```diff
@@ -1513,29 +1514,43 @@ TEST(execute, i32_xor)
...
-    const auto [trap, ret] = execute_binary_operation(Instr::i32_shr_s, uint64_t(-84), 1);
-
-    ASSERT_FALSE(trap);
-    ASSERT_EQ(ret.size(), 1);
-    EXPECT_EQ(ret[0], uint64_t(-42));
+    EXPECT_RESULT(execute_binary_operation(Instr::i32_shr_s, uint64_t(-84), 1), uint64_t(-42));
+    EXPECT_RESULT(execute_binary_operation(Instr::i32_shr_s, 0xffffffff, 0), int32_t(0xffffffff));
+    EXPECT_RESULT(execute_binary_operation(Instr::i32_shr_s, 0xffffffff, 1), int32_t(0xffffffff));
+    EXPECT_RESULT(execute_binary_operation(Instr::i32_shr_s, 0xffffffff, 31), int32_t(0xffffffff));
+    EXPECT_RESULT(execute_binary_operation(Instr::i32_shr_s, 0xffffffff, 32), int32_t(0xffffffff));
```
axic: Spectests also have cases like 0x80000000 >> 32 = 1

chfast: Many others can be added, but I think it is fine to leave it as is until next bug.

PR #126, opened 2020-02-06
benchmarks: export memory in spinner

axic:


chfast:
No comments here, but if this is for benchmarking spinner in wabt, I have fixed this case in #92.

PR #127, opened 2020-02-06
wasm_engine: change parse() and instantiate() to return bool

axic:
This is beneficial for wabt and wasm3.


test/bench/bench.cpp
```diff
@@ -40,6 +40,12 @@ void benchmark_parse(
 {
     const auto engine = create_fn();
 
+    // Pre-run for validation
```
axic: Not sure if this is preferred or just a check in the for loop.

chfast: I think it is fine for now. Just you can omit {}.

test/utils/fizzy_engine.cpp
```diff
@@ -23,9 +23,10 @@ std::unique_ptr<WasmEngine> create_fizzy_engine()
     return std::make_unique<FizzyEngine>();
 }
 
-void FizzyEngine::parse(bytes_view input)
+bool FizzyEngine::parse(bytes_view input)
 {
     m_instance.module = fizzy::parse(input);
```
chfast: You should catch exception here and return false then.

test/bench/bench.cpp
axic: Copied the style from `if (!func_idx)` but I guess because had that 2 lines of message.

axic: Changed.

test/utils/wasm_engine.hpp
```diff
@@ -32,10 +33,12 @@ class WasmEngine
     virtual void instantiate() = 0;
 
     /// Returns the memory of the internal instance.
+    /// Can return an empty view if no memory is available (exported).
```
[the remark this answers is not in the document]
axic: Added.

test/utils/fizzy_engine.cpp
axic: Added.

axic:
> Can memory in wasm have size 0?

Yes, it can start with 0 pages and be grown.

test/utils/fizzy_engine.cpp
```diff
@@ -23,9 +23,21 @@ std::unique_ptr<WasmEngine> create_fizzy_engine()
...
-    m_instance.module = fizzy::parse(input);
+    try
+    {
+        m_instance.module = fizzy::parse(input);
+    }
+    catch (const fizzy::parser_error&)
+    {
+        return false;
+    }
+    catch (const fizzy::instantiate_error&)
```
axic: So this needs to be moved to instantiate. Are we also changing instantiate to return bool? Or should it do an `abort()`?

gumb0: I think it would be good if it returns `bool`, too

axic: Added.

PR #128, opened 2020-02-06
wasm_engine: replace FuncIdx with FuncRef

axic:
Split off #92.

### DIFF
--- a/test/utils/wasm_engine.hpp
+++ b/test/utils/wasm_engine.hpp
@@ -10,6 +10,8 @@ namespace fizzy::test
 class WasmEngine
 {
 public:
+    using FuncRef = uintptr_t;
+
     struct Result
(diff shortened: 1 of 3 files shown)


test/utils/wasm_engine.hpp
```diff
@@ -1,6 +1,5 @@
 #pragma once
 
-#include "types.hpp"
```
axic: Oh this is actually needed for `fizzy_bytes` too, but is only visible if trying to compile wabt/wasm3.

gumb0:
Maybe it's worth also renaming `func_idx` variable in bench.cpp

axic:
> Maybe it's worth also renaming func_idx variable in bench.cpp

Good idea. Did that.

PR #129, opened 2020-02-06
Fix rem_s instructions causing SIGFPE

gumb0:
I guess this case is not mentioned in the spec, because unlike for `div_s` which traps, `rem_s` should return a valid result for this case (0), so there's nothing special to mention.

It's just an implementation detail that `%` is not working for these inputs.


axic:
Raised the question here to confirm: https://github.com/WebAssembly/spec/issues/1126

PR #130, opened 2020-02-06
Move numeric instructions tests into a separate file

gumb0:


chfast:
Is the numeric name taken from the spec?

gumb0:
> Is the numeric name taken from the spec?

It is https://webassembly.github.io/spec/core/exec/instructions.html#numeric-instructions

PR #133, opened 2020-02-07
Fix memory address calculation to catch overflows

axic:


axic:
> This requires regression tests.

Yep, that's why it is in draft. The cases need to be added from spectest/memory.wast

chfast:
> The cases need to be added from spectest/memory.wast

Don't copy tests from there, create new ones.

lib/fizzy/execute.cpp
```diff
@@ -301,7 +301,8 @@ inline DstT extend(SrcT in) noexcept
 template <typename DstT, typename SrcT = DstT>
 inline bool load_from_memory(bytes_view memory, Stack<uint64_t>& stack, const uint8_t*& immediates)
 {
-    const auto address = static_cast<uint32_t>(stack.pop());
+    // Addressing is 32-bit, but we keep the value as 64-bit to detect overflows.
+    const auto address = stack.pop() & 0xffffffff;
```
chfast: I don't think the mask is needed at all. If I'm wrong, can you add a test demonstrating this case?

gumb0: If the point of the change is to change the type of `address`, maybe makes sense to state the type explicitly instead of `auto`?

gumb0: Or maybe it should be instead solved with a cast at the point of comparison with `memory.size()`

axic: > If the point of the change is to change the type of address, maybe makes sense to state the type explicitly instead of auto?

The calculation should be done on `uin64_t` to detect overflows. Both `address` and `offset` are limited to 32-bits.

gumb0: I think I would do
```
 if ((static_cast<uint64_t>(address) + offset + sizeof(SrcT)) > memory.size())
```

lib/fizzy/execute.cpp
```diff
@@ -316,7 +317,8 @@ template <typename DstT>
 inline bool store_into_memory(bytes& memory, Stack<uint64_t>& stack, const uint8_t*& immediates)
 {
     const auto value = static_cast<DstT>(stack.pop());
-    const auto address = static_cast<uint32_t>(stack.pop());
+    // Addressing is 32-bit, but we keep the value as 64-bit to detect overflows.
+    const auto address = stack.pop() & 0xffffffff;
     // NOTE: alignment is dropped by the parser
     const auto offset = read<uint32_t>(immediates);
     if ((address + offset + sizeof(DstT)) > memory.size())
```
axic: I think this is also bad, because we should check both `address+offset` is not overflowing and that `address+offset+sizeof` is no exceeding the memory.

lib/fizzy/execute.cpp
axic: > I don't think the mask is needed at all. If I'm wrong, can you add a test demonstrating this case?

I don't remember our stance on the stack. Did we say that we perform no validation (except casting) and rely on the validation algorithm to ensure the stack is sound (e.g. always the same type or with explicit conversion opcodes)?

gumb0: Yes, my understanding is that the validation should ensure that the stack has always correct type for the operation.

In particular, the spec of `load` execution states:

image

chfast: In current version it is assumed the types match and moreover `i32` are zero extended to 64 bits when placed on the stack.

There is an alternative design we can try later.

I recommend not to go defensive here (and every other place) to make sure it works as designed. You can always add an assert though.

However, there is a catch. It might be ok to produce wrong results when executing invalid wasm. But might not be ok for the interpreter to crash in this case.

axic:
Changed to @gumb0's suggestion. @chfast can you check? I'll squash if this is acceptable.

axic:
TODO: optimise the code in the future

axic:
TODO: optimise the code in the future

PR #135, opened 2020-02-07
spectest: support "register" command

gumb0:
1. Implements `register` command, which saves the name of the module to be used in imports.
2. ~~Adds support for modules importing functions (other imports not supported yet)~~ 
Included in https://github.com/wasmx/fizzy/pull/182
3. ~~(requires https://github.com/wasmx/fizzy/pull/170) Adds the `spectest` module, available for all test files.~~
Moved to https://github.com/wasmx/fizzy/pull/215

### DIFF
--- a/test/spectests/spectests.cpp
+++ b/test/spectests/spectests.cpp
@@ -89,6 +89,23 @@ public:
                 }
                 pass();
             }
+            else if (type == "register")
+            {
+                const auto module_name =
+                    (cmd.find("name") != cmd.end() ? cmd["name"] : UnnamedModule);
+
+                const auto it_instance = instances.find(module_name);
+                if (it_instance == instances.end())
+                {
+                    skip("Module not found.");
+                    continue;
+                }
+
+                const auto registered_name = cmd.at("as").get<std::string>();
+
+                registered_names[registered_name] = module_name;
+                pass();
+            }
             else if (type == "assert_return" || type == "action")
             {
                 const auto& action = cmd.at("action");
@@ -283,6 +300,7 @@ private:
 
     test_settings settings;
     std::unordered_map<std::string, fizzy::Instance> instances;
+    std::unordered_map<std::string, std::string> registered_names;
     test_results results;
 };
 


axic:
I think this won't work because functions are done via import (func) index. I see in the tests there are different kind of imports, so they won't match. Have a look at `imports.wast`.

Maybe the API should be changed so that each import is a `(namespace, name, function)` tuple and instantiate matches and orders it properly for execute.

gumb0:
I don't claim this will work correctly in complicated tests like `imports.wast` and `linking.wast` where functions from several modules are linked together. But this helps in tests where either imports are not called at all (like `binary-leb128.wast`) or where they don't care which function is imported, for example  https://github.com/WebAssembly/spec/blob/c9355bf6141d3299f0679d6dac6ba458c564b00e/test/core/start.wast#L80-L84

axic:
@gumb0 can you explain what is going on it this PR? It seems to be a lot.

gumb0:
@axic Added to the PR description.

Only the 3rd part directly requires floating point (shouldn't be merged before #170), I could extract it to a separate PR.

test/spectests/spectests.cpp
```diff
@@ -15,6 +16,31 @@ namespace
 constexpr auto JsonExtension = ".json";
 constexpr auto UnnamedModule = "_unnamed";
 
+/* wat2wasm
```
axic: Can you add these links here (you've listed):
WebAssembly/spec#1111
https://github.com/WebAssembly/spec/blob/e3a3c12f51221daaa901691e674d7ff990029122/interpreter/host/spectest.ml#L33
https://github.com/WebAssembly/spec/blob/f85f6e6b70fff9ee0f6ba21ebcdad2183bd67874/test/harness/sync_index.js#L79

test/spectests/spectests.cpp
```diff
@@ -88,6 +88,23 @@ class test_runner
...
             }
+            else if (type == "register")
+            {
+                const auto module_name =
+                    (cmd.find("name") != cmd.end() ? cmd["name"] : UnnamedModule);
+
+                const auto it_instance = instances.find(module_name);
+                if (it_instance == instances.end())
+                {
+                    skip("Module not found.");
```
axic: Is this a skip or a fail?

test/spectests/spectests.cpp
```diff
@@ -282,6 +310,55 @@ class test_runner
         return fizzy::execute(instance, *func_idx, std::move(args));
     }
 
+    std::optional<imports> create_imports(const fizzy::Module& module)
```
axic: Is this a const function?

test/spectests/spectests.cpp
```diff
@@ -282,6 +310,55 @@ class test_runner
...
+            {
+                fail("Module not instantiated.");
+                return std::nullopt;
+            }
+
+            auto& instance = it_instance->second;
+
+            if (import.kind == fizzy::ExternalKind::Function)
+            {
+                const auto func_idx = fizzy::find_exported_function(instance.module, import.name);
```
axic: Should use `find_export_function(instance, name)` and also simplify the code below.

axic:
@gumb0 I think this could be merged without the last commit, at least it introduce the `register` command. The last commit is the one depending on #170

test/spectests/spectests.cpp
gumb0: Yeah this is done already in https://github.com/wasmx/fizzy/pull/182

I will split `spectest` module into a separate PR, leave this one with `register` command only and #182 with all the imports support.

test/spectests/spectests.cpp
gumb0: I think skip is better, because typically it would mean than another module failed to instantiate, and then any following tests using it are skipped.

gumb0:
@chfast @axic I left only the first commit here (`register` command)

test/spectests/spectests.cpp
gumb0: Added in https://github.com/wasmx/fizzy/pull/215

PR #136, opened 2020-02-07
Parser rework

chfast:
This is a proposal to change template struct specializations to template function specializations.

1. The types that are uses as vector element require template specialization of `parse()`.
2. All other types are implemented as `parse_xxx()`. Including `parse_vec`.

This is only a syntactic change.

### DIFF
--- a/lib/fizzy/parser.cpp
+++ b/lib/fizzy/parser.cpp
@@ -35,56 +35,52 @@ std::tuple<bool, const uint8_t*> parseGlobalType(const uint8_t* pos)
     return {is_mutable, pos};
 }
 
-template <>
-struct parser<ConstantExpression>
+inline parser_result<ConstantExpression> parse_constant_expression(const uint8_t* pos)
 {
-    parser_result<ConstantExpression> operator()(const uint8_t* pos)
+    ConstantExpression result;
+
+    Instr instr;
+    do
     {
-        ConstantExpression result;
-
-        Instr instr;
-        do
+        instr = static_cast<Instr>(*pos++);
+        switch (instr)
         {
-            instr = static_cast<Instr>(*pos++);
-            switch (instr)
-            {
-            default:
-                throw parser_error{"unexpected instruction in the global initializer expression: " +
-                                   std::to_string(*(pos - 1))};
+        default:
+            throw parser_error{"unexpected instruction in the global initializer expression: " +
+                               std::to_string(*(pos - 1))};
 
-            case Instr::end:
-                break;
+        case Instr::end:
+            break;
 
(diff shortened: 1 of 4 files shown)


lib/fizzy/parser.hpp
```diff
@@ -21,125 +21,106 @@ using parser_result = std::tuple<T, const uint8_t*>;
...
+        return {ValType::i64, pos};
+    default:
+        throw parser_error{"invalid valtype " + std::to_string(b)};
     }
-};
+}
 
-template <>
-struct parser<Limits>
+inline parser_result<Limits> parse_limits(const uint8_t* pos)
```
chfast: This is not a template specialization as it is never used in `parse_vec()`. But we can make it so if easier to read.

lib/fizzy/parser.cpp
```diff
@@ -35,208 +32,182 @@ std::tuple<bool, const uint8_t*> parseGlobalType(const uint8_t* pos)
     return {is_mutable, pos};
 }
 
-template <>
-struct parser<ConstantExpression>
+inline parser_result<ConstantExpression> parse_constant_expression(const uint8_t* pos)
```
chfast: This is not a template specialization as it is never used in `parse_vec()`. But we can make it so if easier to read.

lib/fizzy/parser.cpp
```diff
@@ -35,208 +32,182 @@ std::tuple<bool, const uint8_t*> parseGlobalType(const uint8_t* pos)
...
 
-        return {{limits}, pos};
-    }
-};
+    return {{limits}, pos};
+}
 
-template <>
-struct parser<std::string>
+inline parser_result<std::string> parse_string(const uint8_t* pos)
```
chfast: This is not a template specialization as it is never used in `parse_vec()`. But we can make it so if easier to read.

lib/fizzy/parser.cpp
```diff
@@ -35,208 +32,182 @@ std::tuple<bool, const uint8_t*> parseGlobalType(const uint8_t* pos)
...
-struct parser<Global>
+inline parser_result<Global> parse(const uint8_t* pos)
 {
-    parser_result<Global> operator()(const uint8_t* pos)
-    {
-        Global result;
-        std::tie(result.is_mutable, pos) = parseGlobalType((pos));
-        std::tie(result.expression, pos) = parser<ConstantExpression>{}(pos);
+    Global result;
+    std::tie(result.is_mutable, pos) = parseGlobalType((pos));
```
chfast: The `parseGlobalType()` is some old function name. Might be renamed if desired.

PR #139, opened 2020-02-07
Show nicer error message in ValType-parser for floating point

axic:


test/unittests/parser_test.cpp
```diff
@@ -45,8 +45,14 @@ TEST(parser, valtype)
     EXPECT_EQ(std::get<0>(parse<ValType>(&b)), ValType::i64);
     b = 0x7f;
     EXPECT_EQ(std::get<0>(parse<ValType>(&b)), ValType::i32);
+    b = 0x7c;
+    EXPECT_THROW_MESSAGE(
+        std::get<0>(parse<ValType>(&b)), parser_error, "unsupported valtype (floating point)");
```
gumb0: These don't need `std::get<0>`

axic: How would you rewrite them?

gumb0: EXPECT_THROW_MESSAGE(parse<ValType>(&b), parser_error, "unsupported valtype (floating point)");

axic: Oh right, for the failing case :)

I thought you also meant for the success cases.

axic: Fixed.

PR #140, opened 2020-02-07
Add test for failing start function

axic:


test/unittests/instantiate_test.cpp
```diff
@@ -550,3 +550,15 @@ TEST(instantiate, globals_initialized_from_imported)
     EXPECT_THROW_MESSAGE(instantiate(module_invalid2, {}, {}), instantiate_error,
         "Global can be initialized by another const global only if it's imported.");
 }
+
+TEST(execute, start_unreachable)
+{
+    Module module;
+    module.startfunc = FuncIdx{0};
+    // TODO: add type section (once enforced)
```
gumb0:
```suggestion
    module.typesec.emplace_back(FuncType{{}, {}});
```

axic: I actually want to keep it so that its caught when we implement that check.

gumb0: Maybe create a separate test for this?

axic: Would you have noticed if i didn't had the comment there? :)

gumb0: I wouldn't. Just not sure why leave some TODOs that will make future changes more complicated. But not a big deal.

PR #144, opened 2020-02-11
Hide some parser implementations

chfast:


lib/fizzy/parser.hpp
```diff
@@ -24,40 +24,6 @@ parser_result<Code> parse_expr(const uint8_t* input);
...
-    case 0x7F:
-        return {ValType::i32, pos};
-    case 0x7E:
-        return {ValType::i64, pos};
-    default:
-        throw parser_error{"invalid valtype " + std::to_string(b)};
-    }
-}
-
 inline parser_result<Limits> parse_limits(const uint8_t* pos)
```
axic: How about this?

chfast: This is used in unit tests directly.

axic: Yes, but will the test be changed (in a new PR)?

chfast: Can be done, but adding bounds checking to parser seems more important at the moment.

lib/fizzy/parser.hpp
```diff
@@ -19,19 +19,6 @@ parser_result<Code> parse_expr(const uint8_t* input);
...
-template <>
-inline parser_result<uint8_t> parse(const uint8_t* pos)
-{
-    const auto result = *pos;
-    ++pos;
-    return {result, pos};
-}
-
 template <>
 inline parser_result<uint32_t> parse(const uint8_t* pos)
```
axic: How about this?

chfast: This is also used in `parser_expr.cpp`.

PR #147, opened 2020-02-11
test: add parser test for unknown sections

axic:


test/unittests/parser_test.cpp
```diff
@@ -137,6 +137,19 @@ TEST(parser, custom_section_nonempty)
     EXPECT_EQ(module.codesec.size(), 0);
 }
 
+TEST(parser, unknown_section_empty)
+{
+    const auto bin = bytes{wasm_prefix} + make_section(0x42, bytes{});
```
chfast: Stop with this `42` constant. Use the first invalid value.

axic: I was going to do that but figured it is better to use a higher one.

chfast: Why is better?

PR #148, opened 2020-02-11
test: Replace parser locals tests with full wasm modules

chfast:
This replaces the unit test for `Locals` with a one using full wasm module. This removes the need of exposing `parse<Locals>()` function in `fizzy` library.

### DIFF
--- a/test/unittests/parser_test.cpp
+++ b/test/unittests/parser_test.cpp
@@ -95,12 +95,43 @@ TEST(parser, limits_invalid)
     EXPECT_THROW(parse_limits(input.data()), parser_error);
 }
 
-TEST(parser, locals)
+TEST(parser, code_locals)
 {
-    const auto input = "81017f"_bytes;
-    const auto [l, p] = parse<Locals>(input.data());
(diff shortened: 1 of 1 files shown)


test/unittests/parser_test.cpp
```diff
@@ -95,12 +95,33 @@ TEST(parser, limits_invalid)
...
-TEST(parser, locals)
+TEST(parser, code_locals)
 {
-    const auto input = "81017f"_bytes;
-    const auto [l, p] = parse<Locals>(input.data());
-    EXPECT_EQ(l.count, 0x81);
-    EXPECT_EQ(l.type, ValType::i32);
+    const auto wasm_locals = "81017f"_bytes;  // 0x81 x i32.
+    const auto wasm =
+        bytes{wasm_prefix} + make_section(uint8_t(SectionId::code),
```
axic: I would rather keep using static section numbers and not exposing `types.h`

chfast: Ok, no problem.

PR #150, opened 2020-02-11
test: Change code parser tests to use full wasm modules

chfast:


test/unittests/parser_test.cpp
```diff
@@ -495,6 +495,13 @@ TEST(parser, global_multi_const_inited)
     EXPECT_EQ(module.globalsec[1].expression.value.constant, uint32_t(-1));
 }
 
+TEST(parser, global_invalid_mutability)
+{
+    const auto wasm = bytes{wasm_prefix} + make_section(6, make_vec({"7f02"_bytes}));
+    EXPECT_THROW_MESSAGE(parse(wasm), parser_error,
+        "unexpected byte value 2, expected 0x00 or 0x01 for global mutability");
```
axic: Perhaps the parser could be changed to display the values uniformly. I don't care too much.

chfast: Some other time.

PR #151, opened 2020-02-11
More improvements to spectest runner 

gumb0:


test/spectests/spectests.cpp
```diff
@@ -286,12 +287,15 @@ void run_tests_from_dir(const fs::path& path)
         catch (const std::exception& ex)
         {
             std::cerr << "Exception: " << ex.what() << "\n\n";
+            some_file_excepted = true;
```
axic:
```suggestion
            exception_thrown = true;
```

axic:
Can you prefix the commit log with `spectest:` ?

test/spectests/spectests.cpp
```diff
@@ -179,6 +193,11 @@ class test_runner
                     pass();
                     continue;
                 }
+                catch (std::runtime_error const&)
```
axic: Hm isn't the goal to have every such case covered by `parser_error`?

gumb0: Ah yeah, this got here by accident, will remove.

LEB128 decoder throws `runtime_error` currently, I think it should be fixed to throw `parser_error` in another PR.

axic: Yes, I think changing that makes sense.

test/spectests/spectests.cpp
```diff
@@ -27,10 +27,17 @@ fizzy::bytes load_wasm_file(const fs::path& json_file_path, std::string_view fil
...
+    bool skip_validation = false;
+};
+
 class test_runner
 {
 public:
-    void run_from_file(const fs::path& path)
+    explicit test_runner(const test_settings& ts) : settings{ts} {}
+
+    std::tuple<int, int, int> run_from_file(const fs::path& path)
```
chfast: I would create a `struct` for the result type.

gumb0: Changed.

test/spectests/spectests.cpp
```diff
@@ -268,17 +291,30 @@ void run_tests_from_dir(const fs::path& path)
...
     std::sort(std::begin(files), std::end(files));
 
+    test_results total;
+    bool exception_thrown = false;
     for (const auto& f : files)
     {
         try
         {
-            test_runner{}.run_from_file(f);
+            const auto res = test_runner{settings}.run_from_file(f);
```
axic: Since the internal `test_results` are not reset in `run_from_file`, one could just instantiate the runner once and collect the results once.

gumb0: There's also `test_runner::instance` and at some point it might become a number of instances for test files that link several modules together. Therefore I'd like to leave `test_runner` a class that handles running tests from one file.

test/spectests/spectests.cpp
```diff
@@ -287,19 +323,33 @@ int main(int argc, char** argv)
...
         {
-            std::cerr << "Too many arguments\n";
+            std::cerr << "Missing DIR argument\n";
             return -1;
         }
 
-        run_tests_from_dir(argv[1]);
-        return 0;
+        const bool res = run_tests_from_dir(dir, settings);
+        return res ? 0 : -1;
```
chfast: I think you should return `1` in case of failure. Negative values mean something really exceptional. GTest returns `1` in case of failures.

gumb0: Changed to 1.

PR #152, opened 2020-02-11
circleci: run fizzy-spectest

axic:


circle.yml
```diff
@@ -249,6 +247,24 @@ jobs:
           paths:
             - ~/build/corpus
 
+  spectest:
```
chfast: I don't see benefits of running this as a separate container. Just add it to "sanitizers" or any other.

circle.yml
```diff
@@ -249,6 +247,24 @@ jobs:
...
+          command: |
+            git clone https://github.com/wasmx/wasm-spec --branch vanilla-json --single-branch
+      - attach_workspace:
+          at: "~"
+      - run:
+          name: "Run spectest"
+          command: |
+            set +e
+            expected="  PASSED 2857, FAILED 142, SKIPPED 8814."
+            result=`~/bin/fizzy-spectests --skip-validation wasm-spec/test/core/json | tail -1`
```
chfast: More portable (and more readable IMHO).
```suggestion
            result=$(~/bin/fizzy-spectests --skip-validation wasm-spec/test/core/json | tail -1)
```

circle.yml
axic: I like the immediate feedback on the github ci list.

circle.yml
axic: Fixed.

circle.yml
chfast: I don't think we can afford this for every job. Is this one going to be a special one? You also will not notice if there is sanitizer warnings caused by executing a spectest. Because it runs instantly, I would add it there too.

circle.yml
```diff
@@ -249,6 +247,24 @@ jobs:
           paths:
             - ~/build/corpus
 
+  spectest:
+    executor: linux-gcc-9
+    steps:
+      - run:
+          name: "Download spectest files"
+          command: |
+            git clone https://github.com/wasmx/wasm-spec --branch vanilla-json --single-branch
```
chfast:
```suggestion
            git clone https://github.com/wasmx/wasm-spec --branch vanilla-json --single-branch --depth 1
```

axic: Manual says it also implies `--single-branch`. Should remove that then?

circle.yml
axic: So what are you exactly saying/proposing?

chfast: At least add it to "sanitizers" too.

PR #153, opened 2020-02-11
Replace runtime_error with parser_error in leb128

axic:
Follow up: https://github.com/wasmx/fizzy/pull/151#discussion_r377773462

### DIFF
--- a/lib/fizzy/execute.hpp
+++ b/lib/fizzy/execute.hpp
@@ -1,16 +1,12 @@
 #pragma once
 
+#include "exceptions.hpp"
 #include "types.hpp"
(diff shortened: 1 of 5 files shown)


lib/fizzy/execute.hpp
```diff
@@ -1,16 +1,12 @@
 #pragma once
 
+#include "exceptions.hpp"
```
axic: @chfast @gumb0 `exceptions.hpp` or `errors.hpp`?

chfast: `exceptions.hpp`.

PR #158, opened 2020-02-12
Parser bounds checking

chfast:
This adds input buffer bounds checking in wasm binary parser, except for `parse_expr` which parses instructions (to be done later).

### DIFF
--- a/lib/fizzy/leb128.hpp
+++ b/lib/fizzy/leb128.hpp
@@ -30,7 +30,7 @@ std::pair<T, const uint8_t*> leb128u_decode(const uint8_t* input, const uint8_t*
 }
 
 template <typename T>
-std::pair<T, const uint8_t*> leb128s_decode(const uint8_t* input)
+std::pair<T, const uint8_t*> leb128s_decode(const uint8_t* input, const uint8_t* end)
 {
     static_assert(std::numeric_limits<T>::is_signed);
 
@@ -40,6 +40,9 @@ std::pair<T, const uint8_t*> leb128s_decode(const uint8_t* input)
 
     for (; result_shift < std::numeric_limits<T_unsigned>::digits; ++input, result_shift += 7)
     {
+        if (input == end)
+            throw parser_error("Unexpected EOF");
+
         result |= static_cast<T_unsigned>((static_cast<T_unsigned>(*input) & 0x7F) << result_shift);
         if ((*input & 0x80) == 0)
         {
--- a/lib/fizzy/leb128.hpp
+++ b/lib/fizzy/leb128.hpp
@@ -7,7 +7,7 @@
 namespace fizzy
 {
 template <typename T>
-std::pair<T, const uint8_t*> leb128u_decode(const uint8_t* input)
+std::pair<T, const uint8_t*> leb128u_decode(const uint8_t* input, const uint8_t* end)
 {
     static_assert(!std::numeric_limits<T>::is_signed);
 
@@ -16,6 +16,9 @@ std::pair<T, const uint8_t*> leb128u_decode(const uint8_t* input)
 
     for (; result_shift < std::numeric_limits<T>::digits; ++input, result_shift += 7)
     {
+        if (input == end)
+            throw parser_error("Unexpected EOF");
+
         // TODO this ignores the bits in the last byte other than the least significant one
         // So would not reject some invalid encoding with those bits set.
         result |= static_cast<T>((static_cast<T>(*input) & 0x7F) << result_shift);
--- a/lib/fizzy/parser.cpp
+++ b/lib/fizzy/parser.cpp
@@ -292,8 +292,7 @@ Module parse(bytes_view input)
             std::tie(module.exportsec, it) = parse_vec<Export>(it);
             break;
         case SectionId::start:
-            std::tie(module.startfunc, it) =
-                leb128u_decode<uint32_t>(it, it + 4);  // FIXME: Bounds checking.
+            std::tie(module.startfunc, it) = leb128u_decode<uint32_t>(it, input.end());
             break;
         case SectionId::element:
             std::tie(module.elementsec, it) = parse_vec<Element>(it);
--- a/lib/fizzy/parser.cpp
+++ b/lib/fizzy/parser.cpp
@@ -8,10 +8,10 @@ namespace fizzy
 template <>
 inline parser_result<uint8_t> parse(const uint8_t* pos, const uint8_t* end)
 {
-    (void)end;  // FIXME: Bounds checking.
-    const auto result = *pos;
-    ++pos;
-    return {result, pos};
+    if (pos == end)
+        throw parser_error{"Unexpected EOF"};
+
+    return {*pos, pos + 1};
 }
 
 template <>
@@ -126,7 +126,6 @@ inline parser_result<Memory> parse(const uint8_t* pos, const uint8_t* end)
 
 inline parser_result<std::string> parse_string(const uint8_t* pos, const uint8_t* end)
 {
-    // FIXME: Add bounds checking test, the implementation is likely to be changed.
     std::vector<uint8_t> value;
     std::tie(value, pos) = parse_vec<uint8_t>(pos, end);
 
--- a/lib/fizzy/parser.cpp
+++ b/lib/fizzy/parser.cpp
@@ -267,7 +267,11 @@ Module parse(bytes_view input)
         const auto id = static_cast<SectionId>(*it++);
         uint32_t size;
         std::tie(size, it) = leb128u_decode<uint32_t>(it, input.end());
-        const auto expected_end_pos = it + size;
+
+        const auto expected_section_end = it + size;
+        if (expected_section_end > input.end())
+            throw parser_error("Unexpected EOF");
+
         switch (id)
         {
         case SectionId::type:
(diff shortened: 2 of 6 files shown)


chfast:
Performance regression is significant, up to 7%.
```
/usr/share/benchmark/compare.py benchmarks bin/fizzy-bench-master bin/fizzy-bench ../../test/benchmarks --benchmark_filter=fizzy/parse --benchmark_min_time=3
RUNNING: bin/fizzy-bench-master ../../test/benchmarks --benchmark_filter=fizzy/parse --benchmark_min_time=3 --benchmark_out=/tmp/tmpmwumKT
2020-02-12 17:12:08
Running bin/fizzy-bench-master
Run on (8 X 4400 MHz CPU s)
CPU Caches:
  L1 Data 32K (x4)
  L1 Instruction 32K (x4)
  L2 Unified 256K (x4)
  L3 Unified 8192K (x1)
Load Average: 0.90, 0.44, 0.22
------------------------------------------------------------------------------------
Benchmark                          Time             CPU   Iterations UserCounters...
------------------------------------------------------------------------------------
fizzy/parse/mul256_opt0         4468 ns         4468 ns       948670 rate=187.563M/s size=838
fizzy/parse/micro/spinner        594 ns          594 ns      7043855 rate=107.777M/s size=64
RUNNING: bin/fizzy-bench ../../test/benchmarks --benchmark_filter=fizzy/parse --benchmark_min_time=3 --benchmark_out=/tmp/tmpnJadV3
2020-02-12 17:12:18
Running bin/fizzy-bench
Run on (8 X 4400 MHz CPU s)
CPU Caches:
  L1 Data 32K (x4)
  L1 Instruction 32K (x4)
  L2 Unified 256K (x4)
  L3 Unified 8192K (x1)
Load Average: 0.91, 0.46, 0.23
------------------------------------------------------------------------------------
Benchmark                          Time             CPU   Iterations UserCounters...
------------------------------------------------------------------------------------
fizzy/parse/mul256_opt0         4778 ns         4778 ns       856135 rate=175.39M/s size=838
fizzy/parse/micro/spinner        626 ns          626 ns      6716505 rate=102.264M/s size=64
Comparing bin/fizzy-bench-master to bin/fizzy-bench
Benchmark                            Time             CPU      Time Old      Time New       CPU Old       CPU New
-----------------------------------------------------------------------------------------------------------------
fizzy/parse/mul256_opt0           +0.0694         +0.0694          4468          4778          4468          4778
fizzy/parse/micro/spinner         +0.0539         +0.0539           594           626           594           626
```

lib/fizzy/parser.cpp
```diff
@@ -261,42 +286,46 @@ Module parse(bytes_view input)
...
+
+        const auto section_end = it + size;
+        if (section_end > input.end())
+            throw parser_error("Unexpected EOF");
+
         switch (id)
         {
         case SectionId::type:
-            std::tie(module.typesec, it) = parse_vec<FuncType>(it);
+            std::tie(module.typesec, it) = parse_vec<FuncType>(it, input.end());
```
chfast: For every section kind, this could be `section_end`.

lib/fizzy/leb128.hpp
```diff
@@ -27,7 +30,13 @@ std::pair<T, const uint8_t*> leb128u_decode(const uint8_t* input)
...
 
 template <typename T>
-std::pair<T, const uint8_t*> leb128s_decode(const uint8_t* input)
+std::pair<T, const uint8_t*> leb128u_decode(bytes_view input)
+{
+    return leb128u_decode<T>(input.data(), input.data() + input.size());
+}
+
+template <typename T>
+std::pair<T, const uint8_t*> leb128s_decode(const uint8_t* input, const uint8_t* end)
```
gumb0: Why no `leb128s_decode(bytes_view input)` helper then?

chfast: There is longer story for it. I'm not sure what should be the final version.

1. The helper `leb128u_decode(bytes_view input)` was added to limit changes in unit tests.
2. Using `bytes_view` has bad side effect that `(bytes{}.data())` still works, because `const uint8_t*` is convertible to `bytes_view`. Although, looks like clang-tidy might have a check for it.
3. So when I reached  `leb128s_decode()` in the very end, I decided to rather rewrite tests.
4. The `const bytes&` is safer than `bytes_view`. I'm currently using that for `parse_expr()` in incoming PR.

gumb0: Maybe put the helper into tests then, if it's used only there.

lib/fizzy/parser.hpp
```diff
@@ -64,16 +70,16 @@ inline parser_result<Limits> parse_limits(const uint8_t* pos)
 }
 
 template <typename T>
-parser_result<std::vector<T>> parse_vec(const uint8_t* pos)
+parser_result<std::vector<T>> parse_vec(const uint8_t* pos, const uint8_t* end)
```
gumb0: Other functions are `inline`

lib/fizzy/parser.cpp
```diff
@@ -261,42 +286,46 @@ Module parse(bytes_view input)
     {
         const auto id = static_cast<SectionId>(*it++);
         uint32_t size;
-        std::tie(size, it) = leb128u_decode<uint32_t>(it);
-        const auto expected_end_pos = it + size;
+        std::tie(size, it) = leb128u_decode<uint32_t>(it, input.end());
+
+        const auto section_end = it + size;
```
gumb0: I would still call it `expected_section_end`

lib/fizzy/parser.cpp
```diff
@@ -261,42 +286,46 @@ Module parse(bytes_view input)
     {
         const auto id = static_cast<SectionId>(*it++);
         uint32_t size;
-        std::tie(size, it) = leb128u_decode<uint32_t>(it);
-        const auto expected_end_pos = it + size;
+        std::tie(size, it) = leb128u_decode<uint32_t>(it, input.end());
```
gumb0: Not sure if it's worth to create `input_end` variable instead of repeated `end()` calls

lib/fizzy/leb128.hpp
chfast: > Maybe put the helper into tests then, if it's used only there.

Done.

lib/fizzy/parser.hpp
chfast: Mostly `inline` is required for functions and full template specializations if they are in headers. I was also using `inline` for full template specializations in `.cpp` file to check if used anywhere else. I will make this `inline` as well.

lib/fizzy/parser.cpp
chfast: Rather not.

lib/fizzy/parser.cpp
chfast: Done.

test/unittests/parser_test.cpp
```diff
@@ -40,23 +41,25 @@ bytes make_invalid_size_section(uint8_t id, size_t size, const bytes& content)
...
+        parse<ValType>(b.begin(), b.end()), parser_error, "unsupported valtype (floating point)");
+    b[0] = 0x7a;
+    EXPECT_THROW_MESSAGE(parse<ValType>(b.begin(), b.end()), parser_error, "invalid valtype 122");
 }
 
 TEST(parser, valtype_vec)
 {
     const auto input = "037f7e7fcc"_bytes;
-    const auto [vec, pos] = parse_vec<ValType>(input.data());
+    const auto [vec, pos] = parse_vec<ValType>(input.data(), input.data() + input.size());
```
gumb0: Can't be `(input.begin(), input.end())` because iterators are not pointers there?

chfast: Exactly. They are not pointers for `bytes`, but are for `bytes_view` (not guaranteed though).

test/unittests/parser_test.cpp
```diff
@@ -138,7 +138,7 @@ TEST(parser, code_locals_invalid_type)
     EXPECT_THROW_MESSAGE(parse(wasm), parser_error, "invalid valtype 123");
 }
 
-TEST(parser, empty_module)
+TEST(parser, module_empty)
```
axic: Perhaps you want to merge this single-line-change commit into the next commit changing parser tests.

lib/fizzy/parser.cpp
```diff
@@ -116,10 +116,12 @@ inline parser_result<Global> parse(const uint8_t* pos, const uint8_t* end)
...
 inline parser_result<Table> parse(const uint8_t* pos, const uint8_t* end)
 {
-    (void)end;  // FIXME: Bounds checking.
-    const uint8_t kind = *pos++;
-    if (kind != FuncRef)
-        throw parser_error{"unexpected table elemtype: " + std::to_string(kind)};
+    if (pos == end)
+        throw parser_error{"Unexpected EOF"};
+
+    const auto elemtype = *pos++;
```
axic: Why not keep the `uint8_t` type here or do a static_cast for explicitness? I know the input pointer is `uint8_t`.

test/unittests/parser_test.cpp
chfast: Done.

lib/fizzy/parser.cpp
chfast: Changed.

test/unittests/parser_test.cpp
```diff
@@ -198,7 +226,8 @@ TEST(parser, type_section_smaller_than_expected)
     const auto section_contents = "01"_bytes + functype_void_to_void + "fe"_bytes;
     const auto bin =
         bytes{wasm_prefix} +
-        make_invalid_size_section(1, size_t{section_contents.size() + 1}, section_contents);
+        make_invalid_size_section(1, size_t{section_contents.size() + 1}, section_contents) +
+        "00"_bytes;
```
gumb0: why this change?

chfast: We construct a section that claim that has size bigger that its actual content. But to keep the current error we need to add additional byte to the whole wasm binary, because the `expected_section_end` is now checked for out-of-bounds access up front.

PR #160, opened 2020-02-12
Raise nicer error for floating point instructions

axic:


lib/fizzy/types.hpp
```diff
@@ -162,9 +195,63 @@ enum class Instr : uint8_t
...
+    f32_convert_i64_s = 0xb4,
+    f32_convert_i64_u = 0xb5,
+    f32_demote_f64 = 0xb6,
+    f64_convert_i32_s = 0xb7,
+    f64_convert_i32_u = 0xb8,
+    f64_convert_i64_s = 0xb9,
+    f64_convert_i64_u = 0xba,
+    f64_promote_f32 = 0xbb,
+    i32_reinterpret_f32 = 0xbc,
+    i32_reinterpret_f64 = 0xbd,
```
```suggestion
    i64_reinterpret_f64 = 0xbd,
```

lib/fizzy/types.hpp
```diff
@@ -162,9 +195,63 @@ enum class Instr : uint8_t
...
+    i64_reinterpret_f32 = 0xbe,
```
```suggestion
    f32_reinterpret_i32 = 0xbe,
```

lib/fizzy/types.hpp
```diff
@@ -162,9 +195,63 @@ enum class Instr : uint8_t
...
+    i64_reinterpret_f64 = 0xbf,
```
```suggestion
    f64_reinterpret_i64 = 0xbf,
```

lib/fizzy/types.hpp
axic: You're correct:

Screenshot 2020-02-12 at 14 20 25

axic: Fixed.

PR #163, opened 2020-02-12
Parse expr bounds checking

chfast:
Requires #158

### DIFF
--- a/lib/fizzy/parser.hpp
+++ b/lib/fizzy/parser.hpp
@@ -14,7 +14,8 @@ template <typename T>
 using parser_result = std::tuple<T, const uint8_t*>;
 
 Module parse(bytes_view input);
-parser_result<Code> parse_expr(const uint8_t* input);
+
+parser_result<Code> parse_expr(const uint8_t* input, const uint8_t* end);
 
 template <typename T>
 parser_result<T> parse(const uint8_t* pos, const uint8_t* end);
--- a/lib/fizzy/parser_expr.cpp
+++ b/lib/fizzy/parser_expr.cpp
@@ -300,7 +300,7 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end)
         case Instr::call:
         {
             uint32_t imm;
-            std::tie(imm, pos) = leb128u_decode<uint32_t>(pos, pos + 4);  // FIXME: Bounds checking.
+            std::tie(imm, pos) = leb128u_decode<uint32_t>(pos, end);
             push(code.immediates, imm);
             break;
         }
@@ -324,7 +324,7 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end)
         case Instr::call_indirect:
         {
             uint32_t imm;
-            std::tie(imm, pos) = leb128u_decode<uint32_t>(pos, pos + 4);  // FIXME: Bounds checking.
+            std::tie(imm, pos) = leb128u_decode<uint32_t>(pos, end);
             push(code.immediates, imm);
 
             const uint8_t tableidx{*pos++};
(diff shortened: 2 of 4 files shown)


lib/fizzy/parser.hpp
```diff
@@ -14,7 +14,13 @@ template <typename T>
 using parser_result = std::tuple<T, const uint8_t*>;
 
 Module parse(bytes_view input);
-parser_result<Code> parse_expr(const uint8_t* input);
+
+parser_result<Code> parse_expr(const uint8_t* input, const uint8_t* end);
+
+inline auto parse_expr(const bytes& input)
```
chfast: This can be moved to tests.

lib/fizzy/parser_expr.cpp
```diff
@@ -394,8 +407,7 @@ parser_result<Code> parse_expr(const uint8_t* pos)
         }
         code.instructions.emplace_back(instr);
     }
-    if (!label_positions.empty())
-        throw parser_error{"code is not balanced (missing end statements)"};
+    assert(label_positions.empty());
```
gumb0: Why is this changed?

chfast: This is not reachable. The parsing ends when the final `end` is reached. Previously it was also not reachable, it is leftover from some earlier jumps implementation. I left assert as a documentation / target for fuzzing.

lib/fizzy/parser.hpp
chfast: Moved.

lib/fizzy/parser_expr.cpp
```diff
@@ -209,6 +209,9 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end)
 
         case Instr::block:
         {
+            if (pos == end)
```
axic: Why not move these to line 43?

axic: Only having it in cases where there's an immediate?

Actually, why are we not putting arity onto the immediate list?

lib/fizzy/parser_expr.cpp
```diff
@@ -209,6 +209,9 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end)
...
+                throw parser_error{"Unexpected EOF"};
+
             const uint8_t type{*pos};
```
axic: Perhaps it would be due having a helper for the block/valtype because this code is repeated three times: block, loop, if

lib/fizzy/parser_expr.cpp
```diff
@@ -349,15 +349,15 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end)
...
+            std::tie(imm, pos) = leb128s_decode<int32_t>(pos, end);
             push(code.immediates, static_cast<uint32_t>(imm));
             break;
         }
 
         case Instr::i64_const:
         {
             int64_t imm;
-            std::tie(imm, pos) = leb128s_decode<int64_t>(pos, pos + 10);  // FIXME: Bounds checking.
+            std::tie(imm, pos) = leb128s_decode<int64_t>(pos, end);
```
axic: Th commit log for this is wrong, as it is not only u32 immediates.

test/unittests/parser_expr_test.cpp
```diff
@@ -204,3 +204,14 @@ TEST(parser, control_instr_out_of_bounds)
     EXPECT_THROW_MESSAGE(parse_expr("03"_bytes), parser_error, "Unexpected EOF");
     EXPECT_THROW_MESSAGE(parse_expr("04"_bytes), parser_error, "Unexpected EOF");
 }
+
+TEST(parser, immediate_u32_out_of_bounds)
```
axic: This is not only about u32 immediates.

lib/fizzy/parser_expr.cpp
axic: Done in #174.

PR #164, opened 2020-02-13
Execution tests

chfast:


test/unittests/execute_test.cpp
```diff
@@ -1088,6 +1088,52 @@ TEST(execute, imported_function_call_with_arguments)
     EXPECT_EQ(ret[0], 42);
 }
 
+TEST(execute, imported_functions_call_indirect)
+{
+    /* wat2wasm
+    (module
+      (type $ft (func (param i32) (result i64)))
+      (func $sqr    (import "" "sqr") (param i32) (result i64))
+      (func $isqrt  (import "" "isqrt") (param i32) (result i64))
```
axic: Hmm, is empty namespace valid (according to spec)?

axic: I would suggest to use `env` which is the standard for C imports. Then we can discuss empty namespaces separately (note, wat2wasm can compile non-standard binaries)

PR #168, opened 2020-02-13
benchmarks: add memset benchmark

axic:


axic:
Turns out the stack passed in `call` is reversed.

test/benchmarks/memset.c
```diff
@@ -0,0 +1,15 @@
+#include <inttypes.h>
+#include <string.h>
+
+#define WASM_EXPORT __attribute__((visibility("default")))
+
+WASM_EXPORT unsigned memset_bench(unsigned fill, unsigned length)
+{
+    unsigned char input[length];
```
chfast: That's not even a portable C.

axic: It is in C99, isn't it? Anyway it doesn't matter in our case too much, since we only care about the wasm file.

PR #170, opened 2020-02-13
Support parsing floating points to raise unsupported_feature error

axic:


axic:
@gumb0 @chfast I think it may be time to review this whether we want to go ahead this direction.

lib/fizzy/parser.cpp
```diff
@@ -97,6 +97,12 @@ inline parser_result<ConstantExpression> parse_constant_expression(
             result.value.constant = static_cast<uint64_t>(value);
             break;
         }
+        case Instr::f32_const:
```
axic: @gumb0 these are silently unhandled. Not sure how to handle these nicely.

gumb0: From the point of view of running the spec tests it seems fine to ignore them here (I would zero out `result.value.constant` to make it deterministic)

The test accessing the floating point global would either be skipped if it uses some float opcode, or it will return the value of the global directly, then it will be skipped as the test expecting float result value.

But outside of tests this would mean that accessing such global directly or calling a function returning it will just silently return 0.

chfast:
> @gumb0 @chfast I think it may be time to review this whether we want to go ahead this direction.

Ok to me.

test/unittests/parser_test.cpp
```diff
@@ -1118,12 +1118,13 @@ TEST(parser, code_section_unsupported_fp_instructions)
...
+        const auto code_section = make_vec({code_bin});
+        const auto function_section = make_vec({"00"_bytes});
+        const auto bin =
+            bytes{wasm_prefix} + make_section(3, function_section) + make_section(10, code_section);
 
-        const auto expected_msg =
-            std::string{"unsupported floating point instruction "} + std::to_string(instr);
-        EXPECT_THROW_MESSAGE(parse(bin), parser_error, expected_msg.c_str());
+        const auto module = parse(bin);
+        EXPECT_EQ(module.codesec.size(), 1);
```
gumb0: Maybe rename this test now, because it's kind of not "unsupported" now

axic: It is still unsupported in terms of execution, but yeah, supported in terms of parsing.

axic:
> Looks good, maybe also add a unit test for execution throwing unsupported_feature

Added execute test. It is quite large, but I don't think I've missed anything.

test/unittests/execute_test.cpp
```diff
@@ -1048,3 +1048,221 @@ TEST(execute, memory_copy_32bytes)
...
+TEST(execute, fp_instructions)
+{
+    /* wat2wasm
+    (memory 1)
+
+    (func (param f32 f64)
+      ;; f32 ops
+      i32.const 0 ;; mem offset
+      f32.const 1
+      ;; get_local 0 -- this should work
```
axic: Anyone want to investigate? I think the local calculation may be wrong and this indexes out of bounds.

gumb0: You pass empty `args` vector to `execute`, so it's out of bounds

test/unittests/execute_test.cpp
```diff
@@ -1048,3 +1048,221 @@ TEST(execute, memory_copy_32bytes)
...
+        "0000000000f03fb01a44000000000000f03fb11a4101b21a4101b31a4201b41a4201b51a4101b71a4101b81a42"
+        "01b91a4201ba1a430000803fbc1a44000000000000f03fbd1a4101be1a4201bf1a0b");
+
+    const auto module = parse(bin);
+    auto instance = instantiate(module);
+
+    bool trap = false;
+    try
+    {
+        execute(instance, 0, {});
```
gumb0:
```suggestion
        execute(instance, 0, {0, 0});
```

test/unittests/execute_test.cpp
axic: Phew, rookie mistake :)

axic: But I think I'll just remove it because we can't pass floats properly yet.

test/unittests/execute_test.cpp
```diff
@@ -1048,3 +1048,221 @@ TEST(execute, memory_copy_32bytes)
...
+    }
+    catch (unsupported_feature const&)
+    {
+        trap = true;
+    }
+    ASSERT_TRUE(trap);
```
gumb0: I think you could just use `ASSERT_THROW` instead of `try/catch`?

axic: `EXPECT_THROW_MESSAGE`?

test/unittests/execute_test.cpp
```diff
@@ -1048,3 +1048,221 @@ TEST(execute, memory_copy_32bytes)
...
+
+    const auto module = parse(bin);
+    auto instance = instantiate(module);
+
+    // First function with floating point instructions.
+    EXPECT_THROW_MESSAGE(
+        execute(instance, 0, {}), unsupported_feature, "Floating point instruction.");
+
+    // Second function with floating point parameters.
+    const auto [trap, ret] = execute(instance, 1, {});
```
gumb0: could be just
```suggestion
    ASSERT_TRUE(execute(instance, 1, {}).trapped);
```

test/unittests/parser_test.cpp
axic: Done.

PR #171, opened 2020-02-13
Support linking exported-imported functions between two modules

gumb0:


test/unittests/execute_test.cpp
```diff
@@ -1135,6 +1135,50 @@ TEST(execute, imported_functions_call_indirect)
...
 }
 
+TEST(execute, imported_function_from_another_module)
+{
+    /* wat2wasm
+    (module
+      (func $add (param $lhs i32) (param $rhs i32) (result i32)
+        get_local $lhs
+        get_local $rhs
+        i32.add)
```
chfast: Please use noncommutative operation so we know the order of arguments is correct.

gumb0: Changed to sub

test/unittests/execute_test.cpp
```diff
@@ -1135,6 +1135,50 @@ TEST(execute, imported_functions_call_indirect)
...
+    const auto module2 = parse(bin2);
+
+    const auto func_idx = fizzy::find_exported_function(module1, "sub");
+    ASSERT_TRUE(func_idx.has_value());
+
+    auto add = [&instance1, func_idx](Instance&, std::vector<uint64_t> args) -> execution_result {
+        return fizzy::execute(instance1, *func_idx, std::move(args));
+    };
+
+    auto instance2 = instantiate(module2, {add});
```
axic: Ah I think this is confusing, the export is called `sub`, but the import and locals in this test case are called `add`.

test/unittests/execute_test.cpp
```diff
@@ -1135,6 +1135,50 @@ TEST(execute, imported_functions_call_indirect)
...
+    const auto bin2 = from_hex(
+        "0061736d0100000001070160027f7f017f020a01026d31036164640000030201000a0a0108002001200010000"
+        "b");
+    const auto module2 = parse(bin2);
+
+    const auto func_idx = fizzy::find_exported_function(module1, "sub");
+    ASSERT_TRUE(func_idx.has_value());
+
+    auto add = [&instance1, func_idx](Instance&, std::vector<uint64_t> args) -> execution_result {
+        return fizzy::execute(instance1, *func_idx, std::move(args));
```
axic: Eventually what would be nice is having a generic function of an export which can be passed here (and the interpreter could optimize it). But that is part of the external API work later.

test/unittests/execute_test.cpp
gumb0: Fixed.

PR #176, opened 2020-02-14
benchmarks: add blake2b

axic:


axic:
> What happened on ASAN build?

I have no idea, @chfast ?

chfast:
> > What happened on ASAN build?
> 
> I have no idea, @chfast ?

We are using libbenchmark incorrectly and hit one of its asserts.

chfast:
> > > What happened on ASAN build?
> > 
> > 
> > I have no idea, @chfast ?
> 
> We are using libbenchmark incorrectly and hit one of its asserts.

The assert is caused by `SkipWithError()`. Looks like not precise assert in libbenchmark, because docs say that `SkipWithError()` can be used anywhere in the benchmark function (before, after and inside benchmark loop).

PR #178, opened 2020-02-14
bench: Speed up wabt execution startup 

chfast:
WABT Executor/Thread will init both value and call stack vectors with zeros up front. To mitigate the overhead, we set sizes to much lower values than default ones.


test/utils/wabt_engine.cpp
```diff
@@ -71,7 +71,13 @@ std::optional<WasmEngine::FuncRef> WabtEngine::find_function(std::string_view na
 WasmEngine::Result WabtEngine::execute(
     WasmEngine::FuncRef func_ref, const std::vector<uint64_t>& args)
 {
-    wabt::interp::Executor executor{&m_env};
+    // WABT Executor/Thread will init both value and call stack vectors with zeros up front.
+    // This is not very smart design. To mitigate the overhead, we set sizes to much lower values
+    // than default ones.
```
axic: Can you mention the default sizes?

Also this is the size of the preallocated stack or the limit of the stack?

test/utils/wabt_engine.cpp
```diff
@@ -71,7 +71,13 @@ std::optional<WasmEngine::FuncRef> WabtEngine::find_function(std::string_view na
 WasmEngine::Result WabtEngine::execute(
     WasmEngine::FuncRef func_ref, const std::vector<uint64_t>& args)
 {
-    wabt::interp::Executor executor{&m_env};
+    // WABT Executor/Thread will init both value and call stack vectors with zeros up front.
+    // This is not very smart design. To mitigate the overhead, we set sizes to much lower values
```
gumb0: Is it necessary to insult wabt? :smirk: 
```suggestion
    // To mitigate the performance overhead, we set sizes to much lower values
```

test/utils/wabt_engine.cpp
chfast: > Also this is the size of the preallocated stack or the limit of the stack?

I'm guessing it's both.

PR #180, opened 2020-02-14
tests: add wasm_engine tests

axic:


test/unittests/wasm_engine_test.cpp
```diff
@@ -0,0 +1,34 @@
...
+      local.get $b
+      i64.mul
+      i32.wrap_i64
+    )
+    */
+    const auto wasm = from_hex(
+        "0061736d0100000001080160037f7e7f017f03020100070801047465737400000a0e010c00200020026bad2001"
+        "7ea70b");
+
+    auto fizzy = fizzy::test::create_wabt_engine();
```
axic: This example works on fizzy, traps on wabt and needs the stack fix for wasm3.

test/unittests/wasm_engine_test.cpp
```diff
@@ -0,0 +1,34 @@
+#include <gtest/gtest.h>
+#include <test/utils/hex.hpp>
+#include <test/utils/wasm_engine.hpp>
+
+using namespace fizzy;
+
+TEST(wasm_engine, multiple_args_ret_i32)
```
chfast: If the test is only for fizzy, maybe name the suite as `fizzy_engine`.

test/unittests/wasm_engine_test.cpp
```diff
@@ -0,0 +1,34 @@
...
+    const auto wasm = from_hex(
+        "0061736d0100000001080160037f7e7f017f03020100070801047465737400000a0e010c00200020026bad2001"
+        "7ea70b");
+
+    auto fizzy = fizzy::test::create_wabt_engine();
+    ASSERT_TRUE(fizzy->parse(wasm));
+    ASSERT_TRUE(fizzy->instantiate());
+    auto const func = fizzy->find_function("test");
+    ASSERT_TRUE(func.has_value());
+    // (52 - 21) * 5 => 155
```
chfast: The comment looks out-of-date.

test/unittests/wasm_engine_test.cpp
axic: Just pushed the fix.

test/unittests/wasm_engine_test.cpp
axic: wabt fails because the WabtEngine only supports i32 inputs.

test/unittests/wasm_engine_test.cpp
```diff
@@ -0,0 +1,69 @@
...
+    const auto wasm = from_hex(
+        "0061736d0100000001080160037f7f7f017f03020100070801047465737400000a0c010a00200020026b20016c"
+        "0b");
+
+    for (auto engine_create_fn : {create_fizzy_engine, create_wabt_engine, create_wasm3_engine})
+    {
+        auto engine = engine_create_fn();
+        ASSERT_TRUE(engine->parse(wasm));
+        ASSERT_TRUE(engine->instantiate());
+        auto const func = engine->find_function("test");
```
gumb0:
```suggestion
        const auto func = engine->find_function("test");
```
also in three cases below

axic: Fixed.

PR #181, opened 2020-02-17
Add APIs to find exported global/table/memory

gumb0:


gumb0:
I figured probably a more elegant approach to this API, working on it in https://github.com/wasmx/fizzy/pull/183

gumb0:
Closed in favor of #183

PR #182, opened 2020-02-17
spectest: support tests with imports

gumb0:
Requires https://github.com/wasmx/fizzy/pull/135


test/spectests/spectests.cpp
```diff
@@ -103,7 +121,17 @@ class test_runner
 
                 const auto registered_name = cmd.at("as").get<std::string>();
 
-                registered_names[registered_name] = module_name;
+                if (module_name == UnnamedModule)
+                {
+                    // Assign a name to unnamed module to avoid it being overwritten by another
+                    // unnamed one. Let the name be equal to registered name.
+                    assert(instances.find(registered_name) == instances.end());
```
axic: Should this be an assert or a fail? This requires spectest to be compiled in debug mode to be effective.

gumb0: Well it would be a failure of the assumption made by the testing code... I guess we can change it to failure for easier debugging.

gumb0: Changed to failure.

test/spectests/spectests.cpp
```diff
@@ -103,7 +121,25 @@ class test_runner
 
                 const auto registered_name = cmd.at("as").get<std::string>();
 
-                registered_names[registered_name] = module_name;
+                if (module_name == UnnamedModule)
```
axic: @gumb0 can you extend the smoke test with this case?

test/spectests/smoketest/smoketest.wast
```diff
@@ -22,7 +24,10 @@
 ;; expected i64 value
 (assert_return (invoke $Mod "foo.i64") (i64.const 4))
 
-(assert_trap (invoke "trap") "unreachable instruction")
+;; FIXME: invoke without module name (last module)
```
gumb0: this case is fixed in #201

test/spectests/spectests.cpp
gumb0: added

PR #183, opened 2020-02-17
Add APIs to find exports

gumb0:
Alternative to https://github.com/wasmx/fizzy/pull/181

Adds APIs for finding exported functions/globals/tables/memories in the form of `ExternalFunction`/`ExternalGlobal`/`ExternalTable`/`ExternalMemory` instances ready to be imported into another module.

~~The same should be done for functions, but this requires https://github.com/wasmx/fizzy/pull/171~~ Done.

Using this in spectests will look like this:
https://github.com/wasmx/fizzy/blob/b287cc004c3de064b469ed6e99f2ca262dd6dd92/test/spectests/spectests.cpp#L365-L410

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -2,6 +2,7 @@
 #include "limits.hpp"
 #include "stack.hpp"
 #include "types.hpp"
+#include <algorithm>
 #include <cassert>
 #include <cstring>
 
@@ -1439,4 +1440,31 @@ std::optional<FuncIdx> find_exported_function(const Module& module, std::string_
 
     return {};
 }
+
+std::optional<ExternalGlobal> find_exported_global(Instance& instance, std::string_view name)
+{
+    const auto it = std::find_if(instance.module.exportsec.begin(), instance.module.exportsec.end(),
+        [&name](const auto& export_) {
+            return export_.kind == ExternalKind::Global && export_.name == name;
+        });
+
+    if (it == instance.module.exportsec.end())
+        return std::nullopt;
+
+    const auto global_idx = it->index;
+    if (global_idx < instance.imported_globals.size())
+    {
+        // imported global is reexported
+        return ExternalGlobal{instance.imported_globals[global_idx].value,
+            instance.imported_globals[global_idx].is_mutable};
+    }
+    else
+    {
+        // global owned by instance
+        const auto module_global_idx = global_idx - instance.imported_globals.size();
+        return ExternalGlobal{&instance.globals[module_global_idx],
+            instance.module.globalsec[module_global_idx].is_mutable};
+    }
+}
+
 }  // namespace fizzy
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -1451,6 +1451,17 @@ std::optional<FuncIdx> find_exported_function(const Module& module, std::string_
     return {};
 }
 
+std::optional<ExternalFunction> find_exported_function(Instance& instance, std::string_view name)
+{
+    const auto opt_index = find_export(instance, ExternalKind::Function, name);
+    if (!opt_index.has_value())
+        return std::nullopt;
+
+    return [idx = *opt_index, &instance](fizzy::Instance&, std::vector<uint64_t> args) {
+        return execute(instance, idx, std::move(args));
+    };
+}
+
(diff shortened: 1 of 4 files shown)


lib/fizzy/execute.cpp
```diff
@@ -1437,4 +1448,74 @@ std::optional<FuncIdx> find_exported_function(const Module& module, std::string_
...
+    }
+    else
+    {
+        // imported table is reexported
+        const auto it_import =
+            std::find_if(instance.module.importsec.begin(), instance.module.importsec.end(),
+                [](const auto& import) { return import.kind == ExternalKind::Table; });
+        assert(it_import != instance.module.importsec.end());
+
+        return ExternalTable{instance.table.get(), it_import->desc.table.limits};
```
gumb0: Limits here are not exactly correct: table could have been imported with limits narrower than the ones defined in import's module, we don't save those during instantiate.

gumb0: Similar for memory, but there we save at least the provided upper limit (in `Instance::memory_max_pages`), I use it below in `find_exported_memory`

lib/fizzy/execute.hpp
```diff
@@ -20,23 +20,23 @@ struct execution_result
 
 struct Instance;
 
-using ImportedFunction = std::function<execution_result(Instance&, std::vector<uint64_t>)>;
+using ExternalFunction = std::function<execution_result(Instance&, std::vector<uint64_t>)>;
```
gumb0: maybe `external_function`?
There are `execution_result` and `Instance` in this file, so no consistency.

lib/fizzy/execute.cpp
```diff
@@ -1437,4 +1448,85 @@ std::optional<FuncIdx> find_exported_function(const Module& module, std::string_
...
+    {
+        // imported global is reexported
+        return ExternalGlobal{instance.imported_globals[global_idx].value,
+            instance.imported_globals[global_idx].is_mutable};
+    }
+    else
+    {
+        // global owned by instance
+        return ExternalGlobal{
+            &instance.globals[global_idx], instance.module.globalsec[global_idx].is_mutable};
```
gumb0: Index of the global is incorrect here, should be `global_idx - instance.imported_globals.size()`

gumb0: Fixed, also modified the test to contain both imported and non-imported globals.

gumb0:
Rebased.

lib/fizzy/execute.cpp
chfast: Can you add (possible disabled tests) for these?

lib/fizzy/execute.cpp
```diff
@@ -476,6 +477,16 @@ inline uint64_t popcnt64(uint64_t value) noexcept
...
     return static_cast<uint64_t>(__builtin_popcountll(value));
 }
+
+std::optional<uint32_t> find_export(
+    const Instance& instance, ExternalKind kind, std::string_view name)
+{
+    const auto it = std::find_if(instance.module.exportsec.begin(), instance.module.exportsec.end(),
+        [kind, name](const auto& export_) { return export_.kind == kind && export_.name == name; });
+
+    return (it == instance.module.exportsec.end() ? std::nullopt : std::make_optional(it->index));
```
chfast: My preference is to have the positive case first.
```suggestion
    return (it != instance.module.exportsec.end()) ? std::make_optional(it->index) : std::nullopt;
```

lib/fizzy/execute.cpp
gumb0: Added FIXME comments and disabled tests for table and memory cases of this.

lib/fizzy/execute.cpp
gumb0: Depends on the definition of positive... But I don't care, changed.

gumb0:
Rebased.

lib/fizzy/execute.cpp
```diff
@@ -479,6 +479,16 @@ inline uint64_t popcnt64(uint64_t value) noexcept
 {
     return static_cast<uint64_t>(__builtin_popcountll(value));
 }
+
+std::optional<uint32_t> find_export(
```
axic: Why not use this in `find_external_function` too?

PR #185, opened 2020-02-18
Enforce limit on number of local variables

axic:


test/unittests/parser_test.cpp
```diff
@@ -144,6 +144,22 @@ TEST(parser, code_locals_invalid_type)
     EXPECT_THROW_MESSAGE(parse(wasm), parser_error, "invalid valtype 123");
 }
 
+TEST(parser, code_locals_too_many)
+{
+    const auto large_num = "8080808008"_bytes;  // 0x80000000
+    for (auto const locals : {
```
gumb0:
```suggestion
    for (const auto& locals : {
```

axic: Fixed.

PR #186, opened 2020-02-18
Properly raise parser error on size mismatch of code expressions

axic:


lib/fizzy/parser.cpp
```diff
@@ -258,9 +262,9 @@ inline parser_result<Code> parse(const uint8_t* pos, const uint8_t* end)
         if (local_count > std::numeric_limits<uint32_t>::max())
             throw parser_error{"too many local variables"};
     }
-    std::get<0>(result).local_count = static_cast<uint32_t>(local_count);
+    result.local_count = static_cast<uint32_t>(local_count);
 
-    return result;
+    return {result, pos3};
```
axic: Should use `std::move(result)` here?

gumb0: Yes, I think it will be copied otherwise

lib/fizzy/parser.cpp
```diff
@@ -249,7 +249,11 @@ inline parser_result<Code> parse(const uint8_t* pos, const uint8_t* end)
 
     const auto [locals_vec, pos2] = parse_vec<Locals>(pos1, end);
 
-    auto result = parse_expr(pos2, end);
+    auto [result, pos3] = parse_expr(pos2, end);
```
gumb0: Could name it `[code, pos3]` not confuse with the whole tuple's type `parser_result`

axic: Yeah, I've used that initially, but realised most other functions use `result`. I can change it.

axic: Changed.

lib/fizzy/parser.cpp
axic: Done.

PR #188, opened 2020-02-18
Parse names of custom sections (as required by the spec)

axic:
image

### DIFF
--- a/lib/fizzy/parser.cpp
+++ b/lib/fizzy/parser.cpp
@@ -338,6 +338,8 @@ Module parse(bytes_view input)
             std::tie(module.datasec, it) = parse_vec<Data>(it, input.end());
             break;
         case SectionId::custom:
+            // NOTE: this section can be ignored, but the name must be parseable (and valid UTF-8)
+            parse_string(it, input.end());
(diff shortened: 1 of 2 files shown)


lib/fizzy/parser.cpp
```diff
@@ -338,6 +338,8 @@ Module parse(bytes_view input)
             std::tie(module.datasec, it) = parse_vec<Data>(it, input.end());
             break;
         case SectionId::custom:
+            // NOTE: this section can be ignored, but the name must be parseable (and valid UTF-8)
+            parse_string(it, input.end());
```
chfast: I think there is one more edge case here: when `parse_string` exceeds the `size` (and there are other sections after so `input.end()` is not exceeded). 
First add tests to confirm the issue. There are two possible fixes:
1. Use `it` returned by `parse_string()` and add additional checks.
2. `parse_string(it, expected_section_end)` (i.e. `parse_string(it, it + size)`).

The 2 solution was also proposed to other sections, but I believe this is not strictly required there. So I did not use `expected_section_end` mostly because "Unexpected EOF" error would be a bit confusing in this case.

PR #189, opened 2020-02-18
Do not allow wrongly ordered sections

axic:
Screenshot 2020-02-18 at 20 33 46

### DIFF
--- a/lib/fizzy/parser.cpp
+++ b/lib/fizzy/parser.cpp
@@ -332,9 +332,17 @@ Module parse(bytes_view input)
     input.remove_prefix(wasm_prefix.size());
 
     Module module;
+    SectionId last_id = SectionId::custom;
     for (auto it = input.begin(); it != input.end();)
     {
         const auto id = static_cast<SectionId>(*it++);
+        if (id != SectionId::custom)
+        {
+            if (id <= last_id)
+                throw parser_error{"Unexpected out-of-order section type"};
+            last_id = id;
+        }
+
         uint32_t size;
         std::tie(size, it) = leb128u_decode<uint32_t>(it, input.end());
 
--- a/test/unittests/parser_test.cpp
+++ b/test/unittests/parser_test.cpp
@@ -1193,6 +1193,27 @@ TEST(parser, interleaved_custom_section)
     EXPECT_EQ(module.codesec.size(), 1);
 }
 
+TEST(parser, wrongly_ordered_sections)
+{
+    auto create_empty_section = [](uint8_t id) -> bytes {
+        if (id == 8)  // start
+            return make_section(8, "00"_bytes);
+        else
+            return make_section(id, make_vec({}));
+    };
+
+    constexpr uint8_t first_section_id = 1;
+    constexpr uint8_t last_section_id = 11;
+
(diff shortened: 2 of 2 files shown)


lib/fizzy/parser.cpp
```diff
@@ -292,9 +292,14 @@ Module parse(bytes_view input)
     input.remove_prefix(wasm_prefix.size());
 
     Module module;
+    SectionId last_id = SectionId::custom;
     for (auto it = input.begin(); it != input.end();)
     {
         const auto id = static_cast<SectionId>(*it++);
+        if (id != SectionId::custom && id <= last_id)
+            throw parser_error{"Unexpected section type"};
+        last_id = id;
```
gumb0: I think you should do this assignment only if current section is not custom

axic: > I think you should do this assignment only if current section is not custom

Yes, good point.

test/unittests/parser_test.cpp
```diff
@@ -312,6 +264,13 @@ TEST(parser, type_section_functype_out_of_bounds)
     EXPECT_THROW_MESSAGE(parse(wasm), parser_error, "Unexpected EOF");
 }
 
+TEST(parser, import_section_empty)
+{
+    const auto bin = bytes{wasm_prefix} + make_section(2, make_vec({}));
+    const auto module = parse(bin);
+    EXPECT_EQ(module.typesec.size(), 0);
```
gumb0: I guess you intended to check `importsec` here.
The tests below check `typesec`, too

gumb0:
Also needs tests for incorrect order of sections.

test/unittests/parser_test.cpp
axic: Good catch.

axic:
> Also needs tests for incorrect order of sections.

Yes, didn't start adding the test yet.

test/unittests/parser_test.cpp
axic: Moved to #194

lib/fizzy/parser.cpp
```diff
@@ -331,9 +331,17 @@ Module parse(bytes_view input)
...
 
     Module module;
+    SectionId last_id = SectionId::custom;
     for (auto it = input.begin(); it != input.end();)
     {
         const auto id = static_cast<SectionId>(*it++);
+        if (id != SectionId::custom)
+        {
+            if (id <= last_id)
+                throw parser_error{"Unexpected out-of-order section type"};
```
chfast: Because this is part of (noncritical) validation there should be option to disable this check. It should also throw `validation_error` (new exception type).

axic: Well it is part of the binary format section and not the validation section, hence I believe `parser_error` is more appropriate.

chfast: Ok then.

axic:
> It looks with small modification is should be easy to also check if non-custom sections are duplicated.

It already checks for that, doesn't it?

gumb0:
You could add a test for duplicate sections.

axic:
Rebased. @chfast thanks for the extension. Are we OK merging this now?

PR #190, opened 2020-02-19
Fix handling of unused bits in LEB128 decoder

gumb0:
This adds the handling of the unused bits according to this note in the spec:
image
https://webassembly.github.io/spec/core/binary/values.html#integers

All of the `binary-leb128.wast` spectests pass with this fix.

### DIFF
--- a/lib/fizzy/leb128.hpp
+++ b/lib/fizzy/leb128.hpp
@@ -1,8 +1,12 @@
 #pragma once
 
 #include "exceptions.hpp"
-#include "types.hpp"
 #include <cstdint>
+#include <limits>
+#include <utility>
+
+static_assert((int8_t{-1} >> 1) == int8_t{-1},
+    "signed LEB128 decoder assumes arithmetic shift of negative values");
 
 namespace fizzy
 {
@@ -19,11 +23,14 @@ std::pair<T, const uint8_t*> leb128u_decode(const uint8_t* input, const uint8_t*
         if (input == end)
             throw parser_error("Unexpected EOF");
 
-        // TODO this ignores the bits in the last byte other than the least significant one
-        // So would not reject some invalid encoding with those bits set.
         result |= static_cast<T>((static_cast<T>(*input) & 0x7F) << result_shift);
         if ((*input & 0x80) == 0)
+        {
+            if (*input != (result >> result_shift))
+                throw parser_error("Invalid LEB128 encoding: unused bits set.");
+
             return {result, input + 1};
+        }
     }
 
     throw parser_error("Invalid LEB128 encoding: too many bytes.");
@@ -46,13 +53,29 @@ std::pair<T, const uint8_t*> leb128s_decode(const uint8_t* input, const uint8_t*
         result |= static_cast<T_unsigned>((static_cast<T_unsigned>(*input) & 0x7F) << result_shift);
         if ((*input & 0x80) == 0)
         {
-            // sign extend
-            if ((*input & 0x40) != 0 && (result_shift + 7 < sizeof(T_unsigned) * 8))
+            if (result_shift + 7 < sizeof(T_unsigned) * 8)
             {
-                constexpr auto all_ones = std::numeric_limits<T_unsigned>::max();
-                const auto mask = static_cast<T_unsigned>(all_ones << (result_shift + 7));
-                result |= mask;
+                // non-terminal byte of the encoding, extend the sign bit, if needed
+                if ((*input & 0x40) != 0)
+                {
+                    constexpr auto all_ones = std::numeric_limits<T_unsigned>::max();
+                    const auto mask = static_cast<T_unsigned>(all_ones << (result_shift + 7));
+                    result |= mask;
+                }
             }
+            else
+            {
+                // terminal byte of the encoding, need to check unused bits
+                const auto expected =
+                    (static_cast<uint8_t>(static_cast<T>(result) >> result_shift) & 0x7F);
+
+                if (*input != expected)
+                {
+                    throw parser_error(
+                        "Invalid LEB128 encoding: unused bits not equal to sign bit.");
+                }
+            }
+
             return {static_cast<T>(result), input + 1};
         }
     }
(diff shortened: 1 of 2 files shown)


gumb0:
Changed `last possible byte` to `terminal byte` in comments, to use the terms from spec.

axic:
In #193 I should have changed `#include "types.hpp"` to `#include <limits>` to reduce dependencies in this file. Can you do it here instead?

lib/fizzy/leb128.hpp
```diff
@@ -19,11 +19,16 @@ std::pair<T, const uint8_t*> leb128u_decode(const uint8_t* input, const uint8_t*
         if (input == end)
             throw parser_error("Unexpected EOF");
 
-        // TODO this ignores the bits in the last byte other than the least significant one
-        // So would not reject some invalid encoding with those bits set.
         result |= static_cast<T>((static_cast<T>(*input) & 0x7F) << result_shift);
         if ((*input & 0x80) == 0)
+        {
+            constexpr auto all_ones = std::numeric_limits<T>::max();
```
chfast:
```suggestion
            constexpr auto all_ones = ~T{0};
```

lib/fizzy/leb128.hpp
```diff
@@ -19,11 +19,16 @@ std::pair<T, const uint8_t*> leb128u_decode(const uint8_t* input, const uint8_t*
...
+            const auto unused_bits_mask = ~static_cast<uint8_t>(all_ones >> result_shift);
+            if (*input & unused_bits_mask)
```
chfast:
```suggestion
            if (*input != (result << result_shift))
```

lib/fizzy/leb128.hpp
gumb0: Hm, earlier you changed my `~T_unsigned{0}` to `std::numeric_limits<T_unsigned>()`
https://github.com/wasmx/fizzy/commit/28b0158c3c2655c64621029fa316f768f170f9a8

chfast: Yes, sorry. I think I wanted to see how it looks this way. Not convinced what is best.

lib/fizzy/leb128.hpp
gumb0: yeah this works, but with the right shift instead :+1:

chfast: Right

lib/fizzy/leb128.hpp
```diff
@@ -46,13 +49,31 @@ std::pair<T, const uint8_t*> leb128s_decode(const uint8_t* input, const uint8_t*
...
+                }
             }
+            else
+            {
+                // terminal byte of the encoding, need to check unused bits
+                const auto unused_bits_mask = ~static_cast<uint8_t>(all_ones >> result_shift);
+                // unused bits must be set if and only if sign bit is set
+                const auto unused_bits_expected =
+                    static_cast<T>(result) < 0 ? (unused_bits_mask & 0x7F) : 0;
+                if ((*input & unused_bits_mask) != unused_bits_expected)
```
chfast: You can also use shift to right here, but arithmetic one this time.
```suggestion
                if (*input == static_cast<uint8_t>(static_cast<T>(result) >> result_shift))
```

gumb0: But right-shifting negative values is implementation-dependent, that's why I avoided it here.

gumb0: I can add a static_assert to check the compiler.

chfast: You are right. I learned about this only recently and I keep forgetting. Anyway, UBSan seems to be happy.

lib/fizzy/leb128.hpp
```diff
@@ -46,13 +52,30 @@ std::pair<T, const uint8_t*> leb128s_decode(const uint8_t* input, const uint8_t*
         result |= static_cast<T_unsigned>((static_cast<T_unsigned>(*input) & 0x7F) << result_shift);
         if ((*input & 0x80) == 0)
         {
-            // sign extend
-            if ((*input & 0x40) != 0 && (result_shift + 7 < sizeof(T_unsigned) * 8))
+            constexpr auto all_ones = std::numeric_limits<T_unsigned>::max();
+
+            if (result_shift + 7 < sizeof(T_unsigned) * 8)
             {
-                constexpr auto all_ones = std::numeric_limits<T_unsigned>::max();
```
axic: This is still only used in this branch, however it should make no difference.

gumb0: Moved it back to the branch where it's only needed.

lib/fizzy/leb128.hpp
```diff
@@ -46,13 +52,29 @@ std::pair<T, const uint8_t*> leb128s_decode(const uint8_t* input, const uint8_t*
...
+                    constexpr auto all_ones = std::numeric_limits<T_unsigned>::max();
+                    const auto mask = static_cast<T_unsigned>(all_ones << (result_shift + 7));
+                    result |= mask;
+                }
             }
+            else
+            {
+                // terminal byte of the encoding, need to check unused bits
+                const auto expected =
+                    (static_cast<uint8_t>(static_cast<T>(result) >> result_shift) & 0x7f);
```
axic: Nitpick, in the rest of file you use `0x7F` (uppercase).

lib/fizzy/leb128.hpp
```diff
@@ -19,11 +22,14 @@ std::pair<T, const uint8_t*> leb128u_decode(const uint8_t* input, const uint8_t*
         if (input == end)
             throw parser_error("Unexpected EOF");
 
-        // TODO this ignores the bits in the last byte other than the least significant one
-        // So would not reject some invalid encoding with those bits set.
         result |= static_cast<T>((static_cast<T>(*input) & 0x7F) << result_shift);
         if ((*input & 0x80) == 0)
+        {
+            if (*input != (result >> result_shift))
```
axic: In the signed case the expected side is masked with `0x7f`. Is that needed here too?

gumb0: No because right-shifing unsigned value produces 0 in the most significant bit anyway.

lib/fizzy/leb128.hpp
gumb0: fixed

axic:
@gumb0 are any of these test cases not covered in the spec tests?

axic:
Rebased. Compilation failed due to missing include (`pair`) on macos.

gumb0:
> are any of these test cases not covered in the spec tests?

It's basically one case and there're many tests for it in `binary-leb128.wast` (search for "unused bits")
For example https://github.com/WebAssembly/spec/blob/0ff841dad1c42324d51166a8ded075f4af2fc202/test/core/binary-leb128.wast#L190

axic:
We didn't it trigger a failure though?

gumb0:
Actually the link I gave is not the right one, that's a valid encoding, which we parsed correctly.

But we did fail all of the `assert_malformed` tests checking for incorrect unused bits - that's all the tests in lines 525-955 (33 tests)

PR #191, opened 2020-02-19
Remove Instance from ImportedFunction signature

gumb0:
Originally `Instance` was added to the signature of imported function due to this comment https://github.com/wasmx/fizzy/pull/49#discussion_r365550587

But since `ImportedFunction` is `std::function` now, `instance` can be captured on the host side by the lambda implementing the function.

Also if it's a function from another wasm module being imported, its memory instructions are going to access that external module's memory.

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -558,7 +558,7 @@ Instance instantiate(Module module, std::vector<ImportedFunction> imported_funct
 execution_result execute(Instance& instance, FuncIdx func_idx, std::vector<uint64_t> args)
 {
     if (func_idx < instance.imported_functions.size())
-        return instance.imported_functions[func_idx](instance, std::move(args));
+        return instance.imported_functions[func_idx](std::move(args));
 
     const auto code_idx = func_idx - instance.imported_functions.size();
     assert(code_idx < instance.module.codesec.size());
--- a/lib/fizzy/execute.hpp
+++ b/lib/fizzy/execute.hpp
@@ -20,7 +20,7 @@ struct execution_result
 
 struct Instance;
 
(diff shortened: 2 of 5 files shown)


axic:
How can it capture the instance if the functions need to be passed to `instantiate`, e.g. before the instance exists?

axic:
First in a new PR a test case should be added with a host function using the wasm memory. Then we can go ahead to (re)design it.

gumb0:
> How can it capture the instance if the functions need to be passed to instantiate, e.g. before the instance exists?

Yeah that's true, I didn't think of this. However as `Instance` is move-assignable, this is possible:
```
    Instance instance;
    auto foo = [&instance](std::vector<uint64_t> args){
        return ...;
    };
    instance = instantiate(module, {fooo});
```
But it's not as straight-forward as I imagined, so I'll close this.


The motivation for this came from the confusion around several instances involved when you import function from one module to another:
```
    auto instance1 = instantiate(module1);

    auto foo = [&instance1, func_idx](Instance& /* instance2 */, std::vector<uint64_t> args) {
        return fizzy::execute(instance1, *func_idx, std::move(args));
    };

    auto instance2 = instantiate(module2, {foo});
```

axic:
> The motivation for this came from the confusion around several instances involved when you import function from one module to another:

In most engines this is achieved by having an abstraction over exports/imports such that you could assign an export to an import, without the need of what you did above. This also has the benefit that the engine can detect if the call goes to another instance and reduce overhead.

gumb0:
> In most engines this is achieved by having an abstraction over exports/imports such that you could assign an export to an import, without the need of what you did above. This also has the benefit that the engine can detect if the call goes to another instance and reduce overhead.

It's kind of what I introduce in simple form in https://github.com/wasmx/fizzy/pull/183

PR #194, opened 2020-02-20
tests: add parser tests for empty sections

axic:
Split off #189.


axic:
@gumb0 can you check this?

PR #195, opened 2020-02-20
Fix custom section name parsing

chfast:
Do not allow a custom section which name length exceeds the total size of the section.

### DIFF
--- a/test/unittests/parser_test.cpp
+++ b/test/unittests/parser_test.cpp
@@ -165,6 +165,13 @@ TEST(parser, custom_section_name_out_of_bounds)
     EXPECT_THROW_MESSAGE(parse(bin), parser_error, "Unexpected EOF");
 }
 
+TEST(parser, custom_section_name_exceeds_section_size)
+{
+    const auto bin = bytes{wasm_prefix} + make_invalid_size_section(0, 1, "01aa"_bytes) +
+                     make_section(0, "00"_bytes);
+    EXPECT_THROW_MESSAGE(parse(bin), parser_error, "Unexpected EOF");
+}
+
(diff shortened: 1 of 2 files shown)


test/unittests/parser_test.cpp
```diff
@@ -1054,9 +1061,9 @@ TEST(parser, interleaved_custom_section)
     const auto type_section = make_vec({functype_void_to_void});
     const auto func_section = make_vec({"00"_bytes});
     const auto code_section = make_vec({add_size_prefix("000b"_bytes)});
-    const auto bin = bytes{wasm_prefix} + make_section(0, bytes{4}) +
```
axic: These were actually valid.

chfast: They don't look valid to me. You promised a vector of 4 bytes without providing any actual bytes.

axic: I mean custom sections with `n` zero bytes is valid.

chfast: Oh, I see. This is one of the C++11 pitfalls.
The `bytes{4}` creates `"04"` because the constructor for `std::initializer_list` has priority if `{}` are used.
You meant `bytes(4)` to create `"00000000"`. To be fully precise such constructor does not exist (opposite to `std::vector`) - the zero must be explicit here: `bytes(4, 0)`.

chfast: I can convert it to what is was meant to be.

axic: It doesn't really matter for this test case. We should however add a test case which includes unparsed data (and not just a parsed name) in the custom section.

axic: Pushed this test case here.

PR #197, opened 2020-02-21
tests: use wat2wasm in end_to_end_test

axic:


test/unittests/end_to_end_test.cpp
```diff
@@ -140,43 +139,38 @@ TEST(end_to_end, nested_loops_in_c)
...
-        get_local $p1
-        i32.const -1
-        i32.add
-        tee_local $p1
-        br_if $L1
-      end
-    end)
+    /* wat2wasm
+    (module
+      (func (export "test") (param i32 i32)
```
axic: Can change this back to `(param $p0 i32) (param $p1 i32)`

PR #199, opened 2020-02-21
Add more WasmEngine tests

axic:


test/unittests/wasm_engine_test.cpp
```diff
@@ -5,6 +5,63 @@
...
+    const auto wasm = "0102"_bytes;
+
+    for (auto engine_create_fn : {create_fizzy_engine, create_wabt_engine, create_wasm3_engine})
+    {
+        auto engine = engine_create_fn();
+        ASSERT_FALSE(engine->parse(wasm));
+    }
+}
+
+TEST(wasm_engine, DISABLED_instantiate_error)
```
chfast: Why is this disabled?

axic: Need to fix #138 to make this pass on wasm3. And wabt had some other issue I can't remember.

axic: I can enable this test for fizzy only, if that helps.

chfast: I'm not sure we can have generically parse and instantiate separated. You can leave it for fizzy only.

PR #201, opened 2020-02-24
spectest: support "get" command

gumb0:


gumb0:
Added smoke test cases.

gumb0:
Rebased.

PR #203, opened 2020-02-25
Parser fuzzer

chfast:
Build with `-DFIZZY_TESTING=ON -DFIZZY_FUZZING=ON`.
Clang compiler is required to build this.

```
bin/fizzy-fuzz-parser corpus
```

```
bin/fizzy-fuzz-parser -help=1
```

### DIFF
--- a/lib/fizzy/parser.cpp
+++ b/lib/fizzy/parser.cpp
@@ -173,7 +173,8 @@ parser_result<std::string> parse_string(const uint8_t* pos, const uint8_t* end)
     if (!utf8_validate(pos, pos + size))
         throw parser_error{"Invalid UTF-8"};
 
-    auto ret = std::string(pos, pos + size);
+    const auto char_begin = reinterpret_cast<const char*>(pos);
+    auto ret = std::string(char_begin, char_begin + size);
     pos += size;
 
     return {std::move(ret), pos};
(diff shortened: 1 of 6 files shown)


axic:
Where is the corpus (or how did you generate it)? Taking all the wasm files from the spectests?

chfast:
> Where is the corpus (or how did you generate it)? Taking all the wasm files from the spectests?

Corpus can be an empty dir (you have to create it though).
The inputs are purely wasm binaries, so you can also dump all wasm files you have (I never tried it thought).

At current version, it quickly "crashes" because of huge memory allocations.

lib/fizzy/parser.cpp
```diff
@@ -152,7 +152,8 @@ parser_result<std::string> parse_string(const uint8_t* pos, const uint8_t* end)
     if (!utf8_validate(pos, pos + size))
         throw parser_error{"Invalid UTF-8"};
 
-    auto ret = std::string(pos, pos + size);
+    const auto char_begin = reinterpret_cast<const char*>(pos);
```
axic: Shouldn't this go into #204 ?

chfast: It is cherry picked from there. Needed to have it running.

chfast:
I'm going to merge this to start building corpus on CI.

chfast:
Pushed a fix for CI cache key: the save key must include the timestamp to "overwrite" the previous corpus.

PR #204, opened 2020-02-25
ci: Enable UBSan errors

chfast:
This makes CI to fail on UBSan warnings (previously silently ignored). And fixes issues we accumulated over time.

Requires #229, #225.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -88,8 +88,6 @@ commands:
           working_directory: ~/build
           command: |
             set +e
-            # TODO: Make UBSan warnings fatal.
-            # export UBSAN_OPTIONS=halt_on_error=1
             expected="  PASSED 4433, FAILED 42, SKIPPED 7338."
             result=$(bin/fizzy-spectests --skip-validation wasm-spec/test/core/json | tail -1)
             echo "Expected: $expected"
@@ -208,6 +206,7 @@ jobs:
     environment:
       BUILD_TYPE: RelWithDebInfo
       CMAKE_OPTIONS: -DENABLE_ASSERTIONS=ON -DSANITIZE=address,undefined,implicit-conversion,nullability
+      UBSAN_OPTIONS: halt_on_error=1
     steps:
       - build
       - test
--- a/circle.yml
+++ b/circle.yml
@@ -205,7 +205,7 @@ jobs:
     executor: linux-clang-9
     environment:
       BUILD_TYPE: RelWithDebInfo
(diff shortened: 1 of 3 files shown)


lib/fizzy/execute.cpp
```diff
@@ -566,7 +574,7 @@ execution_result execute(Instance& instance, FuncIdx func_idx, std::vector<uint6
     assert(code_idx < instance.module.codesec.size());
 
     const auto& code = instance.module.codesec[code_idx];
-    auto& memory = *instance.memory;
+    auto* const memory = instance.memory.get();
```
gumb0: Why this?

chfast: Why pointer in general? Because you cannot dereference null pointer.
Why `auto* const`? The `*` indicates it is a pointer. The `const` make it const. More verbose variant is `const auto`.

gumb0: > Why pointer in general? Because you cannot dereference null pointer.

I don't get what this fixes, you just dereference it many times instead of once now.

chfast: Let's move this discussion to: https://github.com/wasmx/fizzy/pull/225.

axic:
@chfast anything left to merge here?

chfast:
> @chfast anything left to merge here?

I moved ready-to-merge fixes to https://github.com/wasmx/fizzy/pull/229. This still requires #225 before we can enable UBSan errors on CI.

lib/fizzy/execute.cpp
gumb0: Ah I see now, in general module is allowed to have no memory, unless it has some memory instructions.

axic: So this PR as-is can be applied after #225 is merged?

circle.yml
```diff
@@ -208,6 +208,7 @@ jobs:
     environment:
       BUILD_TYPE: RelWithDebInfo
       CMAKE_OPTIONS: -DENABLE_ASSERTIONS=ON -DSANITIZE=address,undefined,implicit-conversion,nullability
+      UBSAN_OPTIONS: halt_on_error=1
```
axic: Will this also be triggered for the `spectest` job?

chfast: Sure.

axic: Can you remove the commented out stuff from spectest job then?

PR #205, opened 2020-02-25
Fix i32.{shr, div, rem}_s instructions

chfast:
If the binary_op() operates on i32 values, we want to zero-extend the final value to i64 so avoid int32_t -> uint64_t conversion which is the sign-extend one.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -98,6 +98,7 @@ jobs:
           ./wat2wasm4cpp.py test/unittests/end_to_end_test.cpp
           ./wat2wasm4cpp.py test/unittests/execute_call_test.cpp
           ./wat2wasm4cpp.py test/unittests/execute_control_test.cpp
+          ./wat2wasm4cpp.py test/unittests/execute_numeric_test.cpp
           ./wat2wasm4cpp.py test/unittests/execute_test.cpp
           ./wat2wasm4cpp.py test/unittests/parser_test.cpp
           ./wat2wasm4cpp.py test/unittests/wasm_engine_test.cpp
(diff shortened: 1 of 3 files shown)


axic:
These seem not to be covered by the spec tests, given we pass the i32/i64 spec tests.

chfast:
> These seem not to be covered by the spec tests, given we pass the i32/i64 spec tests.

That was my expectation.

axic:
This looks okay to me.

chfast:
> This looks okay to me.

Is this approved or what?

gumb0:
Looks good to me, just one suggestion to simplify the tests: I would just return the result of `i64.extend_u/i32`, it already would be seen what was incorrect.

chfast:
> Looks good to me, just one suggestion to simplify the tests: I would just return the result of `i64.extend_u/i32`, it already would be seen what was incorrect.

I changed that. The previous version comes from the fact JavaScript does not support `i64` numbers.

gumb0:
> I changed that. The previous version comes from the fact JavaScript does not support i64 numbers.

I meant also in `i32_div_s_stack_value` and `i32_rem_s_stack_value`

gumb0:
> These seem not to be covered by the spec tests, given we pass the i32/i64 spec tests.

It might be covered somewhere in `conversions.wast`, I didn't try those.

gumb0:
Yeah, `i64.extend_i32_u` tests were failing, `conversions.wast`, lines 38, 39, 41.

gumb0:
Ah no, those don't use binary operations, and they still fail after merge. It's a similar problem in fizzy-spectests passing i32 arguments to functions.

PR #196, opened 2020-02-21
Optimise parse_string and parse<Data>

axic:
On my computer it gives 15-7-0.5% speed up on the benchmarks as it highly depends on input. No measurable speed increase if there's no data section or substantial number of strings (exports/imports/custom sections).

PR #206, opened 2020-02-25
Use parse<uint8_t> in multiple places and rename it to parse_byte

axic:
Follow up of #196

### DIFF
--- a/lib/fizzy/parser.cpp
+++ b/lib/fizzy/parser.cpp
@@ -50,26 +50,32 @@ inline parser_result<uint8_t> parse(const uint8_t* pos, const uint8_t* end)
 }
 */
 
+ValType validate_valtype(uint8_t byte)
+{
+    switch (byte)
+    {
+    case 0x7F:
+        return ValType::i32;
+    case 0x7E:
+        return ValType::i64;
+    case 0x7D:  // f32
+        return ValType::f32;
+    case 0x7C:  // f64
+        return ValType::f64;
+    default:
+        throw parser_error{"invalid valtype " + std::to_string(byte)};
+    }
+}
+
 template <>
 parser_result<ValType> parse(const uint8_t* pos, const uint8_t* end)
 {
     if (pos == end)
         throw parser_error{"Unexpected EOF"};
 
-    const auto b = *pos++;
-    switch (b)
-    {
-    case 0x7F:
-        return {ValType::i32, pos};
-    case 0x7E:
-        return {ValType::i64, pos};
-    case 0x7D:  // f32
-        return {ValType::f32, pos};
-    case 0x7C:  // f64
-        return {ValType::f64, pos};
-    default:
-        throw parser_error{"invalid valtype " + std::to_string(b)};
-    }
+    const auto byte = *pos++;
+    const auto valtype = validate_valtype(byte);
+    return {valtype, pos};
 }
 
 const uint8_t* validate_valtype(const uint8_t* pos, const uint8_t* end)
@@ -99,7 +105,7 @@ inline parser_result<FuncType> parse(const uint8_t* pos, const uint8_t* end)
 
 inline std::tuple<bool, const uint8_t*> parse_global_type(const uint8_t* pos, const uint8_t* end)
 {
-    pos = validate_valtype(pos, end);
+    std::tie(std::ignore, pos) = parse<ValType>(pos, end);
 
     if (pos == end)
         throw parser_error{"Unexpected EOF"};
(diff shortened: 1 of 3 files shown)


axic:
@chfast here's the follow up. Haven't benchmarked it yet.

lib/fizzy/parser.cpp
```diff
@@ -65,6 +58,7 @@ inline parser_result<ConstantExpression> parse_constant_expression(
         if (pos == end)
             throw parser_error{"Unexpected EOF"};
 
+        // TODO: use parse_byte here?
```
chfast: Yes.

lib/fizzy/parser_expr.cpp
```diff
@@ -39,6 +39,7 @@ parser_result<uint8_t> parse_blocktype(const uint8_t* pos, const uint8_t* end)
     if (pos == end)
         throw parser_error{"Unexpected EOF"};
 
+    // TODO: use parse_byte here?
```
chfast: Yes.

lib/fizzy/parser_expr.cpp
```diff
@@ -64,6 +65,7 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end)
         if (pos == end)
             throw parser_error{"Unexpected EOF"};
 
+        // TODO: use parse_byte here?
```
chfast: Yes.

lib/fizzy/parser_expr.cpp
axic: I would say actually this one cannot be nicely changed because `validate_valtype` also wants to read it.

lib/fizzy/parser.cpp
axic: Done.

lib/fizzy/parser_expr.cpp
axic: Done.

lib/fizzy/parser_expr.cpp
chfast: Change to `void validate_valtype(uint8_t)`.

axic: `valida_valtype` is just a wrapper over `parse<ValType>`

lib/fizzy/parser.hpp
```diff
@@ -41,5 +41,8 @@ parser_result<std::string> parse_string(const uint8_t* pos, const uint8_t* end);
 /// This is used in parse_expr() (parser_expr.cpp).
 parser_result<std::vector<uint32_t>> parse_vec_i32(const uint8_t* pos, const uint8_t* end);
 
+/// Validates and converts the given byte to valtype.
+ValType validate_valtype(uint8_t byte);
+
 const uint8_t* validate_valtype(const uint8_t* pos, const uint8_t* end);
```
axic: This seems to be remnant of lost code?

lib/fizzy/parser.cpp
```diff
@@ -50,28 +50,34 @@ inline parser_result<uint8_t> parse(const uint8_t* pos, const uint8_t* end)
...
+{
+    if (pos == end)
+        throw parser_error{"Unexpected EOF"};
+
+    const auto byte = *pos++;
+    const auto valtype = validate_valtype(byte);
+    return {valtype, pos};
+}
+
 const uint8_t* validate_valtype(const uint8_t* pos, const uint8_t* end)
```
axic: Any point keeping both this and the template?

chfast: The template is not available in `parser_expr.cpp`.

axic: Yes, but do we need the template?

chfast: Template is required for `vec<ValType>`.

lib/fizzy/parser.cpp
```diff
@@ -39,52 +39,39 @@ inline parser_result<uint32_t> parse(const uint8_t* pos, const uint8_t* end)
...
-const uint8_t* validate_valtype(const uint8_t* pos, const uint8_t* end)
+template <>
+parser_result<ValType> parse(const uint8_t* pos, const uint8_t* end)
 {
-    const auto [_, next_pos] = parse<ValType>(pos, end);
-    return next_pos;
+    if (pos == end)
+        throw parser_error{"Unexpected EOF"};
+
+    const auto byte = *pos++;
```
axic, line 65: Why not use `parse_byte` here?

PR #208, opened 2020-02-26
tests: improve coverage of Stack

axic:


test/unittests/stack_test.cpp
```diff
@@ -48,3 +52,95 @@ TEST(stack, drop_and_peek)
...
+
+    auto bottom_item = stack.begin();
+    EXPECT_EQ(*bottom_item, 'a');
+
+    auto top_item = stack.end() - 1;
+    EXPECT_EQ(*top_item, 'c');
+}
+
+// This crashes.
+TEST(stack, DISABLED_peek_on_empty)
```
chfast: What's the point of these DISABLED tests?

axic: For us to argue here: should they pass? If yes, need to fix.

chfast: This should never pass.

test/unittests/stack_test.cpp
```diff
@@ -48,3 +52,95 @@ TEST(stack, drop_and_peek)
...
+{
+    Stack<char> stack;
+    stack.clear();
+}
+
+// This crashes.
+TEST(stack, DISABLED_drop_on_empty)
+{
+    Stack<char> stack;
+    stack.drop();
```
chfast: This should never pass.

test/unittests/stack_test.cpp
```diff
@@ -48,3 +52,95 @@ TEST(stack, drop_and_peek)
...
+}
+
+// This crashes on certain implementations.
+// Works on macOS.
+TEST(stack, DISABLED_pop_on_empty)
+{
+    Stack<char> stack;
+    stack.pop();
```
chfast: This should never pass.

test/unittests/stack_test.cpp
```diff
@@ -48,3 +52,95 @@ TEST(stack, drop_and_peek)
...
+
+    EXPECT_FALSE(stack.empty());
+    EXPECT_EQ(stack.size(), 2);
+
+    // grow stack
+    stack.resize(4);
+    EXPECT_FALSE(stack.empty());
+    EXPECT_EQ(stack.peek(), 0);
+    EXPECT_EQ(stack.peek(1), 0);
+    EXPECT_EQ(stack.size(), 4);
```
gumb0: could also check that initial elements are not changed

test/unittests/stack_test.cpp
```diff
@@ -48,3 +52,95 @@ TEST(stack, drop_and_peek)
...
+{
+    Stack<char> stack;
+    stack.push('a');
+    stack.push('b');
+    stack.push('c');
+
+    EXPECT_FALSE(stack.empty());
+    EXPECT_EQ(stack.size(), 3);
+
+    auto bottom_item = stack.begin();
```
gumb0: BTW `cend()` should be enough in the code currently using `end()`, maybe makes sense to change it.

Otherwise this test should check also that iterators allow to change elements.

test/unittests/stack_test.cpp
axic: Some of these result in segfault and not an exception.

gumb0: All standard containers produce UB on out of bounds access. It's a price for efficiency and avoiding extra checks.

test/unittests/stack_test.cpp
axic: `pop_back` is UB, so this is fine removing.

test/unittests/stack_test.cpp
axic: This uses `resize` internally, so it actually should just work, except if `size() - num_elements` is negative.

test/unittests/stack_test.cpp
axic: `peek` is our implementation, but by judging `pop_back` I guess we can agree to consider it UB.

test/unittests/stack_test.cpp
axic: > Otherwise this test should check also that iterators allow to change elements.

You mean something like what's in 61a86f41?

test/unittests/stack_test.cpp
axic: Added.

test/unittests/stack_test.cpp
gumb0: Yeah that would do.
But not sure if we should allow this if we create another implementation of stack.

axic: I think it is fine now, at least we have all the features used tested. If there's a follow up PR to remove the use of `end` and replace it with `cend` or to remove those use cases entirely, then we can remove this.

PR #209, opened 2020-02-26
New stack implementation

axic:
This is not meant to be merged, only for discussion.

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -609,13 +609,13 @@ execution_result execute(Instance& instance, FuncIdx func_idx, std::vector<uint6
             const auto target_imm = read<uint32_t>(immediates);
             LabelContext label{code.instructions.data() + target_pc,
                 code.immediates.data() + target_imm, arity, stack.size()};
-            labels.emplace_back(label);
+            labels.push(label);
             break;
         }
         case Instr::loop:
         {
             LabelContext label{pc - 1, immediates, 0, stack.size()};  // Target this instruction.
-            labels.push_back(label);
+            labels.push(label);
             break;
         }
         case Instr::if_:
@@ -630,7 +630,7 @@ execution_result execute(Instance& instance, FuncIdx func_idx, std::vector<uint6
 
                 LabelContext label{code.instructions.data() + target_pc,
                     code.immediates.data() + target_imm, arity, stack.size()};
-                labels.emplace_back(label);
+                labels.push(label);
             }
             else
             {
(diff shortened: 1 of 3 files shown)


lib/fizzy/execute.cpp
```diff
@@ -598,13 +598,13 @@ execution_result execute(Instance& instance, FuncIdx func_idx, std::vector<uint6
...
             LabelContext label{code.instructions.data() + target_pc,
                 code.immediates.data() + target_imm, arity, stack.size()};
-            labels.emplace_back(label);
+            labels.push(label);
             break;
         }
         case Instr::loop:
         {
             LabelContext label{pc - 1, immediates, 0, stack.size()};  // Target this instruction.
-            labels.push_back(label);
```
axic: Even currently this is inconsistent, in some place it uses emplace, in others push for labels.

lib/fizzy/execute.cpp
```diff
@@ -1420,7 +1420,10 @@ execution_result execute(Instance& instance, FuncIdx func_idx, std::vector<uint6
 end:
     assert(labels.empty() || trap);
     // move allows to return derived Stack<uint64_t> instance into base vector<uint64_t> value
-    return {trap, std::move(stack)};
+    if (stack.empty())
```
axic: I think we could merge this part.

lib/fizzy/parser_expr.cpp
```diff
@@ -239,7 +239,7 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end)
             code.immediates.push_back(arity);
 
             // Push label with immediates offset after arity.
-            label_positions.push_back({Instr::block, code.immediates.size()});
+            label_positions.push({Instr::block, code.immediates.size()});
```
axic: Wonder if the parser needs to use stack.

lib/fizzy/stack.hpp
```diff
@@ -6,31 +6,53 @@
 namespace fizzy
 {
 template <typename T>
-class Stack : public std::vector<T>
+class Stack
```
axic: This is based off evmone's because I was lazy to write the same for testing. Yes, it uses a fixed stack, just for conducting this simple test.

lib/fizzy/stack.hpp
```diff
@@ -6,31 +6,53 @@
...
 {
+    T storage[1024];
+
+    T* top_item;
```
chfast: Would be interesting to see the difference when fields order is reversed.

axic: Swapped.

axic:
Superseded by #247.

PR #210, opened 2020-02-26
Stack depth counting (and preallocation)

axic:


lib/fizzy/execute.cpp
```diff
@@ -571,8 +571,8 @@ execution_result execute(Instance& instance, FuncIdx func_idx, std::vector<uint6
     std::vector<uint64_t> locals = std::move(args);
     locals.resize(locals.size() + code.local_count);
 
-    // TODO: preallocate fixed stack depth properly
     Stack<uint64_t> stack;
+    stack.reserve(code.max_stack_depth);
```
chfast: To use full potential of this we need custom stack implementation.

lib/fizzy/parser_expr.cpp
```diff
@@ -408,8 +411,12 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end, bool have
         }
         }
         code.instructions.emplace_back(instr);
+        max_stack_depth += instr_stack_change(instr);
```
axic: I think we need to move this into/after a validation phase so that we have access to each of the sections (such as types) to know how many stack items are we popping/pushing (for example in `call`/`call_indirect`/etc).

axic:
Superseded by #244 and #247.

PR #211, opened 2020-02-26
Trap when call_indirect chooses uninitialized table element

gumb0:
From `call_indirect` [execution spec](https://webassembly.github.io/spec/core/exec/instructions.html#exec-call-indirect):

image

Spec test for this: https://github.com/WebAssembly/spec/blob/e3a3c12f51221daaa901691e674d7ff990029122/test/core/elem.wast#L353

This PR changes table's internal representation to `std::vector<std::optional<FuncIdx>>`.

It would be possible to represent it with `std::vector<std::function<execution_result(std::vector<uint64_t>)>>`, but `call_indirect` also checks the type of the called function against opcode's immediate, so we would need to store addditionally table function types or function indices. Vector of optionals seems at least simpler for now.

### DIFF
--- a/lib/fizzy/execute.hpp
+++ b/lib/fizzy/execute.hpp
@@ -22,11 +22,12 @@ struct Instance;
 
 using ExternalFunction = std::function<execution_result(Instance&, std::vector<uint64_t>)>;
 
-using table_ptr = std::unique_ptr<std::vector<FuncIdx>, void (*)(std::vector<FuncIdx>*)>;
+using table_elements = std::vector<std::optional<FuncIdx>>;
+using table_ptr = std::unique_ptr<table_elements, void (*)(table_elements*)>;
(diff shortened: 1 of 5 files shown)


lib/fizzy/execute.hpp
```diff
@@ -22,11 +22,12 @@ struct Instance;
 
 using ExternalFunction = std::function<execution_result(Instance&, std::vector<uint64_t>)>;
 
-using table_ptr = std::unique_ptr<std::vector<FuncIdx>, void (*)(std::vector<FuncIdx>*)>;
+using table_data = std::vector<std::optional<FuncIdx>>;
```
axic: `table_elem`?

gumb0: Renamed to `table_elements`

PR #212, opened 2020-02-26
parser: Do not reserve memory for vec elements

chfast:
Do not reserve memory for `size` number of elements. It can be a lie causing huge allocation without actual elements to follow.

### DIFF
--- a/lib/fizzy/parser.hpp
+++ b/lib/fizzy/parser.hpp
@@ -79,7 +79,13 @@ inline parser_result<std::vector<T>> parse_vec(const uint8_t* pos, const uint8_t
     std::tie(size, pos) = leb128u_decode<uint32_t>(pos, end);
 
     std::vector<T> result;
-    result.reserve(size);
+
+    // Reserve memory for vec elements if `size` value is reasonable.
+    // TODO: Adjust the limit constant by inspecting max vec size value
+    //       appearing in big set of example wasm binaries.
+    if (size < 128)
+        result.reserve(size);
+
     auto inserter = std::back_inserter(result);
     for (uint32_t i = 0; i < size; ++i)
         std::tie(inserter, pos) = parse<T>(pos, end);
--- a/test/bench_internal/CMakeLists.txt
+++ b/test/bench_internal/CMakeLists.txt
@@ -7,5 +7,5 @@ target_sources(fizzy-bench-internal PRIVATE
     parser_noinline.cpp
 )
 
-target_link_libraries(fizzy-bench-internal PRIVATE fizzy::fizzy benchmark::benchmark_main)
+target_link_libraries(fizzy-bench-internal PRIVATE fizzy::fizzy fizzy::test-utils benchmark::benchmark_main)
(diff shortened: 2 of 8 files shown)


chfast:
A unit test to be added.

lib/fizzy/parser.hpp
```diff
@@ -79,7 +79,9 @@ inline parser_result<std::vector<T>> parse_vec(const uint8_t* pos, const uint8_t
     std::tie(size, pos) = leb128u_decode<uint32_t>(pos, end);
 
     std::vector<T> result;
-    result.reserve(size);
+    // Do not reserve memory for `size` number of elements. It can be a lie
```
axic: How is this a lie?

chfast: _Can be_. Literally `ffffffff04`.

axic: Well that's going to fail anyway, won't it? Can't we just impose a rational limit here?

axic: > Can be. Literally ffffffff04.

If we do merge this (as it shouldn't mean any real difference at this point), please update the comment to be more clear.

chfast: It is going to report `parser_error` instead of crashing. A limit would require additional justification.

axic: >It is going to report parser_error instead of crashing. 

Yep, and that's what the spectests require probably (though they haven't failed -- so should report this upstream).

> A limit would require additional justification.

We have #8 for that, but I guess this specific case is not worth it.

chfast:
This cause up to 10% regression in parsing. So I will present a mitigation to this.

chfast:
This version preserves the performance. The top vectors size in all benchmarks is 2.

lib/fizzy/parser.hpp
```diff
@@ -79,7 +79,11 @@ inline parser_result<std::vector<T>> parse_vec(const uint8_t* pos, const uint8_t
     std::tie(size, pos) = leb128u_decode<uint32_t>(pos, end);
 
     std::vector<T> result;
-    result.reserve(size);
+
+    // Reserve memory for vec elements if `size` value is reasonable.
+    if (size < 128)
```
axic: I think a large number would be still reasonable, especially for the `Data` section. For everything else probably 256 would be more realistic.

chfast: 1. `Data` parsing does not use `parse_vec()` any more.
2. In all benchmarks max size is 2.
3. This still works correctly in else case.
4. I used `128` because `cmp` instruction is smaller then.

axic: > Data parsing does not use parse_vec() any more.

But the `Data` section itself is parsed into a vector.

> In all benchmarks max size is 2.

Yes, our benchmarks are tiny. I had another benchmark in the pipeline from ewasm-precompiles which is quite big, that should help us see more realistic numbers.

> I used 128 because cmp instruction is smaller then.

That's nice, but does it have a performance difference? In any case most of this conversation is pretty useless unless we benchmark all these options on more realistic data. **I suggest to keep a note to revisit this in the future.**

PR #213, opened 2020-02-27
spectest: fix i32 arguments/results handling

gumb0:
Fix to correctly parse and compare i32 signed values, similar to the fix in https://github.com/wasmx/fizzy/pull/205

Some tests were failing because of this https://github.com/wasmx/fizzy/pull/205#issuecomment-591490490

### DIFF
--- a/test/spectests/spectests.cpp
+++ b/test/spectests/spectests.cpp
@@ -5,6 +5,7 @@
 #include <fstream>
 #include <iostream>
 #include <sstream>
+#include <type_traits>
 #include <unordered_map>
 
 namespace fs = std::filesystem;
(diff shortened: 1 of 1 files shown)


test/spectests/spectests.cpp
```diff
@@ -122,25 +123,18 @@ class test_runner
...
-                    }
                     else
                     {
                         skip("Unsupported expected type '" + expected_type + "'.");
                         continue;
                     }
 
-                    if (expected_value != actual_value)
+                    const uint64_t actual_value = result->stack[0];
+                    if (expected_value != result->stack[0])
```
chfast:
```suggestion
                    if (expected_value != actual_value)
```

test/spectests/spectests.cpp
```diff
@@ -122,25 +123,18 @@ class test_runner
...
-                        actual_value = result->stack[0];
-                    }
                     else
                     {
                         skip("Unsupported expected type '" + expected_type + "'.");
                         continue;
                     }
 
-                    if (expected_value != actual_value)
+                    const uint64_t actual_value = result->stack[0];
```
gumb0: We don't cast actual value anymore, bacause if it's 32-bit, returned stack value must have 0s in the high 32 bits.

test/spectests/spectests.cpp
gumb0: fixed

PR #214, opened 2020-02-27
parser: Ensure there is a matching function section for each code section

axic:
Fixes 5 spectest failures.


axic:
@chfast @gumb0 single commit or have some separation of code change vs. test change?

chfast:
> @chfast @gumb0 single commit or have some separation of code change vs. test change?

Two commits sound fine.

PR #217, opened 2020-02-28
Parser and parser tests refactoring

chfast:
I want to hide as much of parser implementation as possible.


chfast:
> I'll have pain rebasing #170 and #214, though.

I can do the rebase.

PR #220, opened 2020-02-29
benchmarks: add ecpairing example

axic:


test/utils/wasm3_engine.cpp
```diff
@@ -83,7 +83,12 @@ void Wasm3Engine::set_memory(fizzy::bytes_view memory)
     uint32_t size;
     const auto data = m3_GetMemory(m_runtime, &size, 0);
     assert(data != nullptr);
-    std::memcpy(data, memory.data(), size);
+
+    // If memory grows during execution, there is no API to shrink it back to initial size.
```
axic: Can you rebase this PR to the latest API?

Then I can push the sources for ecpairing, shrink the wasm and we could merge this.

chfast: This fix is not needed any more as we don't want to patch the existing instance, we just create a new one.

axic:
@chfast @gumb0 added sources, let me know if this layout is good

axic:
My only concern is adding 200k of binary to the repo, so far the repo been tiny. But this benchmark can help us push for speed optim -- need to match/beat wabt after all.

PR #225, opened 2020-03-02
validation: Reject modules with memory instructions but without memory

chfast:
Requires: #238.

Issue: when there is no memory defined in a module any memory access instruction (store, load) causes a crash.

I noticed that wabt rejects such module on parsing. We should confirm it is valid solution because it is a nice one (no additional check is needed during execution).

### DIFF
--- a/lib/fizzy/exceptions.hpp
+++ b/lib/fizzy/exceptions.hpp
@@ -9,6 +9,11 @@ struct parser_error : public std::runtime_error
     using runtime_error::runtime_error;
 };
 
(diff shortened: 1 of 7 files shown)


gumb0:
Not sure about parsing, but validation spec certainly states that memory instructions require memory to be defined:

image

https://webassembly.github.io/spec/core/valid/instructions.html#memory-instructions

gumb0:
Not sure if we should add validation to floating point memory instructions, too, but it seems easy.

chfast:
I will add this for floating point.

PR #226, opened 2020-03-03
CI jobs optimization

chfast:


chfast:
@axic You need to adjust required CI job, because this PR changes their names.

PR #227, opened 2020-03-03
Parser refactor

chfast:


lib/fizzy/parser.cpp
```diff
@@ -438,4 +464,9 @@ Module parse(bytes_view input)
 
     return module;
 }
+
+parser_result<std::vector<uint32_t>> parse_i32s(const uint8_t* pos, const uint8_t* end)
```
axic: Why not `parse_i32_vec` or `parse_vec_i32`?

lib/fizzy/parser.hpp
```diff
@@ -51,23 +45,5 @@ inline parser_result<Limits> parse_limits(const uint8_t* pos, const uint8_t* end
...
-    //       appearing in big set of example wasm binaries.
-    if (size < 128)
-        result.reserve(size);
-
-    auto inserter = std::back_inserter(result);
-    for (uint32_t i = 0; i < size; ++i)
-        std::tie(inserter, pos) = parse<T>(pos, end);
-    return {result, pos};
-}
+parser_result<std::vector<uint32_t>> parse_i32s(const uint8_t* pos, const uint8_t* end);
```
axic: Perhaps a comment noting this is only used by `parse_expr.cpp` would be useful?

lib/fizzy/parser.cpp
```diff
@@ -64,6 +64,12 @@ parser_result<ValType> parse(const uint8_t* pos, const uint8_t* end)
     }
 }
 
+const uint8_t* validate_valtype(const uint8_t* pos, const uint8_t* end)
+{
+    const auto [_, next_pos] = parse<ValType>(pos, end);
```
axic: Why is this better than `std::tie(std::ignore, pos) = parse...` ?

lib/fizzy/parser.cpp
```diff
@@ -64,6 +64,12 @@ parser_result<ValType> parse(const uint8_t* pos, const uint8_t* end)
     }
 }
 
+const uint8_t* validate_valtype(const uint8_t* pos, const uint8_t* end)
```
axic: Shouldn't this be inline?

axic:
There seems to still be a few unnecessarily exported parser helpers. Anything to move in this PR?

lib/fizzy/parser.cpp
chfast: It cannot, because is used in other compilation unit.

lib/fizzy/parser.cpp
chfast: Can be, but it will weirdly update the input argument.

lib/fizzy/parser.cpp
axic: Oh thought this was the header. Nvm.

chfast:
> There seems to still be a few unnecessarily exported parser helpers. Anything to move in this PR?

Do you mean `parse_limits` and `parse_string`?

First is used in unittests. I can replace it with full wasm tests, but @gumb0 had some restrictions.

Second is used in microbenchmarking. Implementation is not exposed.

axic:
> Do you mean parse_limits and parse_string?

To be honest haven't checked the file before commenting. In the past we had *a lot* in there. Checked now, you are right, it is mostly `parse_limits` which would annoy me.

PR #228, opened 2020-03-03
test: Add make_code() helper for execute tests

chfast:


chfast:
After starting doing this, I realized there are only 2 usage of that left. I can also convert these to "wat2wasm" form.

axic:
> After starting doing this, I realized there are only 2 usage of that left. I can also convert these to "wat2wasm" form.

Please do.

test/unittests/execute_control_test.cpp
```diff
@@ -55,58 +55,49 @@ TEST(execute_control, block_br)
...
-
-    const auto [trap, ret] = execute(module, 0, {100});
-
-    ASSERT_FALSE(trap);
-    ASSERT_EQ(ret.size(), 1);
-    EXPECT_EQ(ret[0], 100);
+    /* wat2wasm --no-check
+    (func
+      (block
+        get_local 0  ;; Leaves an item on the stack what makes the loop invalid.
```
axic: Comment is wrong, this is not a loop.

PR #231, opened 2020-03-12
spectest: support assert_unlinkable and assert_uninstantiable

gumb0:


test/spectests/spectests.cpp
```diff
@@ -256,6 +256,49 @@ class test_runner
...
+                        std::move(imports->memories), std::move(imports->globals));
+                }
+                catch (const fizzy::parser_error& ex)
+                {
+                    fail(std::string{"Parsing failed with error: "} + ex.what());
+                    continue;
+                }
+                catch (const fizzy::instantiate_error&)
+                {
+                    pass();
```
axic: Should it check here for `instantiate` message ?

gumb0: Yes we can check.

test/spectests/spectests.cpp
```diff
@@ -256,6 +257,55 @@ class test_runner
...
+                        std::move(imports.globals));
+                }
+                catch (const fizzy::parser_error& ex)
+                {
+                    fail(std::string{"Parsing failed with error: "} + ex.what());
+                    continue;
+                }
+                catch (const fizzy::instantiate_error& ex)
+                {
+                    if (ex.what() == std::string{"Start function failed to execute"})
```
axic: Need to also check for `type == "assert_uninstantiable"`.

test/spectests/spectests.cpp
```diff
@@ -256,6 +257,55 @@ class test_runner
...
+                        pass();
+                    else
+                        fail(std::string{"Instantiation failed with error: "} + ex.what());
```
axic: Shouldn't this be pass if `type == "assert_unlinkable"`?

gumb0:
There're some UBSan warnings in spectest run, not sure if they're caused by this PR

axic:
>There're some UBSan warnings in spectest run, not sure if they're caused by this PR

Those been there before this PR. We need to fix them.

axic:
I'm merging this, we fix it later too.

PR #235, opened 2020-03-17
Check imported function type

gumb0:
1. Make `ExternalFunction` a struct containing `std::function` and `FuncType`.
2. In `match_imported_functions` check imported function types against import types declared in the module.
3. Refactor to get rid of `Instance::imported_function_types`, because this info is now included in `Instance::imported_functions`

This fixes failures in `imports.wast`

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -88,7 +88,7 @@ commands:
           working_directory: ~/build
           command: |
             set +e
-            expected="  PASSED 4433, FAILED 42, SKIPPED 7338."
+            expected="  PASSED 4451, FAILED 24, SKIPPED 7338."
             result=$(bin/fizzy-spectests --skip-validation wasm-spec/test/core/json | tail -1)
             echo "Expected: $expected"
             echo "Result: $result"
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -18,8 +18,7 @@ struct LabelContext
     size_t stack_height = 0;             ///< The stack height at the label instruction.
 };
 
-void match_imported_functions(const Module& module,
-    const std::vector<TypeIdx>& module_imported_types,
+void match_imported_functions(const std::vector<FuncType>& module_imported_types,
     const std::vector<ExternalFunction>& imported_functions)
 {
     if (module_imported_types.size() != imported_functions.size())
@@ -31,7 +30,7 @@ void match_imported_functions(const Module& module,
 
     for (size_t i = 0; i < imported_functions.size(); ++i)
     {
-        const auto& module_imported_type = module.typesec[module_imported_types[i]];
+        const auto& module_imported_type = module_imported_types[i];
         const auto& imported_type = imported_functions[i].type;
         if (module_imported_type.inputs != imported_type.inputs ||
             module_imported_type.outputs != imported_type.outputs)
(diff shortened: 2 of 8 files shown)


test/unittests/instantiate_test.cpp
```diff
@@ -56,6 +56,17 @@ TEST(instantiate, imported_functions_not_enough)
         "Module requires 1 imported functions, 0 provided");
 }
 
+TEST(instantiate, imported_function_wrong_type)
+{
+    Module module;
+    module.typesec.emplace_back(FuncType{{ValType::i32}, {ValType::i32}});
+    module.importsec.emplace_back(Import{"mod", "foo", ExternalKind::Function, {0}});
```
axic:
```suggestion
    module.importsec.emplace_back(Import{"mod", "foo", ExternalKind::Function, {FuncIdx{0}}});
```

gumb0: This repeats the pattern in other tests above, should I change it there, too?

axic: Yeah, sure.

gumb0: changed

lib/fizzy/execute.cpp
```diff
@@ -27,6 +27,18 @@ void match_imported_functions(const std::vector<TypeIdx>& module_imported_types,
                                 " imported functions, " +
                                 std::to_string(imported_functions.size()) + " provided");
     }
+
+    for (size_t i = 0; i < imported_functions.size(); ++i)
+    {
+        const auto& module_imported_type = module_imported_types[i];
+        const auto& imported_type = imported_functions[i].type;
+        if (module_imported_type.inputs != imported_type.inputs ||
```
chfast: It should be better to add `operator==` to `FuncType` to indicate that his type is comparable.

gumb0: Good idea, added

circle.yml
```diff
@@ -90,7 +90,7 @@ commands:
             set +e
             # TODO: Make UBSan warnings fatal.
             # export UBSAN_OPTIONS=halt_on_error=1
-            expected="  PASSED 4433, FAILED 42, SKIPPED 7338."
+            expected="  PASSED 4451, FAILED 24, SKIPPED 7338."
```
axic: Needs to be updated.

lib/fizzy/execute.cpp
```diff
@@ -27,6 +37,17 @@ void match_imported_functions(const std::vector<TypeIdx>& module_imported_types,
...
+
+    for (size_t i = 0; i < imported_functions.size(); ++i)
+    {
+        const auto& module_imported_type = module_imported_types[i];
+        const auto& imported_type = imported_functions[i].type;
+        if (module_imported_type != imported_type)
+        {
+            throw instantiate_error("Function " + std::to_string(i) +
+                                    " type doesn't match module's imported function type");
+        }
```
axic: Are parentheses still needed here?

lib/fizzy/execute.cpp
```diff
@@ -27,6 +37,17 @@ void match_imported_functions(const std::vector<TypeIdx>& module_imported_types,
                                 " imported functions, " +
                                 std::to_string(imported_functions.size()) + " provided");
     }
+
+    for (size_t i = 0; i < imported_functions.size(); ++i)
+    {
+        const auto& module_imported_type = module_imported_types[i];
+        const auto& imported_type = imported_functions[i].type;
+        if (module_imported_type != imported_type)
```
axic: How about `if (module_imported_types[i] != imported_functions[i].type)` ?

lib/fizzy/execute.cpp
gumb0: yes, because it's multi-line

gumb0:
UBSan error in `instantiate.imported_memory_invalid` seems to say there's an uninitialized `bool` inside `std::optional`...

any suggestions?

gumb0:
Fixed it.

PR #236, opened 2020-03-19
UB in shr instructions

gumb0:


lib/fizzy/execute.cpp
```diff
@@ -409,6 +409,7 @@ inline void comparison_op(Stack<uint64_t>& stack, Op<T> op) noexcept
 template <typename T>
 inline T shift_left(T lhs, T rhs) noexcept
 {
+    static_assert(!std::numeric_limits<T>::is_signed);
```
chfast: This may be a bit more precise.
```suggestion
    static_assert(std::is_unsigned_v<T>);
```

lib/fizzy/execute.cpp
```diff
@@ -418,13 +419,14 @@ template <typename T>
 inline T shift_right(T lhs, T rhs) noexcept
 {
     constexpr T num_bits{sizeof(T) * 8};
-    const auto k = rhs % num_bits;
+    const auto k = rhs & (num_bits - 1);
```
chfast: This is good change (still single CPU instruction). So maybe use the same pattern in all other cases?

chfast: Just an idea, but might be more readable.
```suggestion
    constexpr T mask{sizeof(T) * 8 - 1};
    const auto k = rhs & mask;
```

gumb0: Changed it in all shifts

gumb0: not sure it is

PR #237, opened 2020-03-23
call_indirect for functions from another modules

gumb0:
This fixes remaining failures in `elem.wast` and `linking.wast`. 

There are two scenarios:

1.
- module A has exported table
- module A adds via element section some of its functions into the table.
- module B imports this table
- `call_indirect` executed with this table in the instance of B will call a function in A.

Unit test demonstrating it: https://github.com/wasmx/fizzy/blob/0a51bc1ec4db349249113ed99a0c8371c1368c65/test/unittests/execute_call_test.cpp#L353

2. 
- module A has exported table
- module B imports this table and adds via element section some of its functions into this shared table.
- module B's start function traps, so it can't be instantiated.
- After this the change to shared table can't be rolled back, and so the table still must contain the functions of module B. They can be called from module A with `call_indirect`.
(Related discussion about this scenario:
https://github.com/WebAssembly/spec/issues/993
https://github.com/WebAssembly/spec/pull/994)

Unit test demonstrating it: https://github.com/wasmx/fizzy/blob/b5d12edf384710c0bb48f730cf0efedca2180c2a/test/unittests/execute_call_test.cpp#L396

Two problems are fixed in the same PR, because fixing only 1 leads to UB in the spectest for scenario 2 (calling a function from deleted instance)
https://github.com/WebAssembly/spec/blob/9f7b0ae427e5f078cf4951d1c974aa3bcda72bc2/test/core/linking.wast#L388

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -93,7 +93,7 @@ commands:
           working_directory: ~/build
           command: |
             set +e
-            expected="  PASSED 4451, FAILED 24, SKIPPED 7338."
+            expected="  PASSED 4468, FAILED 7, SKIPPED 7338."
             result=$(bin/fizzy-spectests --skip-validation wasm-spec/test/core/json | tail -1)
             echo "Expected: $expected"
             echo "Result: $result"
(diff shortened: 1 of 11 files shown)


gumb0:
Closing in favor of https://github.com/wasmx/fizzy/pull/246

PR #238, opened 2020-03-23
parser: Process code separately from code section

chfast:
When parsing code section (vec of codes), only record the wasm binary section containing the code (based on the code size available). The code are fully processed later, after information about all sections is collected. This decouples code parsing from parse_vec() and is preparation for passing additional information to the code parsing procedure.

### DIFF
--- a/lib/fizzy/types.hpp
+++ b/lib/fizzy/types.hpp
@@ -324,7 +324,8 @@ struct Element
     std::vector<FuncIdx> init;
 };
 
-// https://webassembly.github.io/spec/core/binary/modules.html#code-section
+/// The element of the code section.
+/// https://webassembly.github.io/spec/core/binary/modules.html#code-section
 struct Code
 {
     uint32_t local_count = 0;
@@ -336,6 +337,14 @@ struct Code
     // The decoded instructions' immediate values.
     // These are instruction-type dependent fixed size value in the order of instructions.
     bytes immediates;
+};
+
+/// The reference to the `code` in the wasm binary.
+///
+/// The distinct type is needed for parse_vec<> specialization.
+struct code_view : public bytes_view
+{
+    using bytes_view::bytes_view;
 };
 
 // https://webassembly.github.io/spec/core/binary/modules.html#data-section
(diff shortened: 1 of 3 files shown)


lib/fizzy/types.hpp
```diff
@@ -324,9 +324,16 @@ struct Element
     std::vector<FuncIdx> init;
 };
 
-// https://webassembly.github.io/spec/core/binary/modules.html#code-section
+/// The element of the code section.
+/// https://webassembly.github.io/spec/core/binary/modules.html#code-section
+///
+/// TODO: This type has currently 3 slightly different uses. Refactor uses to eliminate
```
gumb0: Well I'd prefer to split it in this PR, it doesn't seem too difficult. We'd have something like
```
struct code_view
{
    bytes_view code_binary;
};
parser_result<code_view> parse(const uint8_t* pos, const uint8_t* end);

Code parse_code(code_view code);
```

chfast: This is nice suggestion except it limits lazy code parsing.

lib/fizzy/types.hpp
```diff
@@ -324,7 +324,11 @@ struct Element
     std::vector<FuncIdx> init;
 };
 
-// https://webassembly.github.io/spec/core/binary/modules.html#code-section
+/// The element of the code section.
+/// https://webassembly.github.io/spec/core/binary/modules.html#code-section
+///
+/// TODO: This type has currently 2 slightly different uses. Refactor uses to eliminate
```
gumb0: What is the second use now besides parsing instructions?

lib/fizzy/types.hpp
```diff
@@ -336,6 +340,22 @@ struct Code
     // The decoded instructions' immediate values.
     // These are instruction-type dependent fixed size value in the order of instructions.
     bytes immediates;
+
+    Code() noexcept = default;
```
gumb0: I think you could get rid of default constructor by changing `parse_expr` to accumulate instructions and immediates in separate local variables and then constructing `Code` in the end with them.

Not sure it's worth the effort, but it seems that default-constructed `Code` is not used to represent any meaningful concept, but is just a convenience hack.

lib/fizzy/types.hpp
gumb0: Ah I guess the second one is storing locals.

lib/fizzy/types.hpp
chfast: I can actually remove these constructors.

PR #239, opened 2020-03-24
Avoid reinterpret_cast for converting bytes to string

chfast:


lib/fizzy/parser.cpp
```diff
@@ -6,6 +6,11 @@
 
 namespace fizzy
 {
+inline std::string to_string(bytes_view b)
```
axic: After #204 is merged do we still want this helper? I think it looks nice, but don't care too much.

chfast:
I have left only a tiny bit of it.

PR #241, opened 2020-03-25
Validate label index

chfast:


test/spectests/spectests.cpp
```diff
@@ -289,6 +296,11 @@ class test_runner
                     pass();
```
axic: I think actually we should check the `type` here. It is a pass in `assert_malformed`, but a fail in `assert_invalid`.

test/spectests/spectests.cpp
```diff
@@ -289,6 +296,11 @@ class test_runner
...
                     continue;
                 }
+                catch (fizzy::validation_error const&)
+                {
+                    pass();
```
axic: The inverse of https://github.com/wasmx/fizzy/pull/241/commits/32cc345a2b296e6625a1a3dcab3ec0a1bac84d24#r398071288 applies here.

chfast: This does not work.

test/spectests/spectests.cpp
```diff
@@ -289,6 +296,15 @@ class test_runner
...
                     continue;
                 }
+                catch (fizzy::validation_error const&)
+                {
+                    if (type == "assert_invalid")
+                    {
+                        pass();
+                        continue;
+                    }
+                    throw;
```
gumb0: it should fail instead

gumb0: `throw` here just stops executing current file and outputs error to stderr

gumb0:
Still need to hande `validation_error` in `assert_unlinkable`/`assert_uninstantiable`

test/spectests/spectests.cpp
```diff
@@ -289,6 +296,14 @@ class test_runner
                     pass();
```
gumb0:
```suggestion
                    if (type == "assert_malformed")
                        pass();
                    else
                        fail(std::string{"Unexpected parser error: "} + ex.what());
                    continue;
```

PR #242, opened 2020-03-25
CI: Manual benchmarking workflow

chfast:
1. New "benchmarking" workflow containing the "benchmark" job.
2. The "benchmarking" workflow does not run by default, but can be triggered via Circle CI API.
3. The script has been added that triggers the workflow.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,5 +1,10 @@
 version: 2.1
 
+parameters:
+  benchmark:
+    type: boolean
+    default: false
+
 executors:
   lint:
     docker:
@@ -279,7 +284,8 @@ jobs:
 workflows:
   version: 2
 
-  fizzy:
+  testing:
+    unless: <<pipeline.parameters.benchmark>>
     jobs:
       - lint
       - release-linux
@@ -292,6 +298,7 @@ workflows:
             - release-linux
 
   benchmarking:
+    when: <<pipeline.parameters.benchmark>>
     jobs:
       - release-linux
       - benchmark:
--- a/circle.yml
+++ b/circle.yml
@@ -278,6 +278,7 @@ jobs:
 
 workflows:
   version: 2
+
   fizzy:
     jobs:
       - lint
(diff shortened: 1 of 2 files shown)


axic:
What is the reason for this?

chfast:
> What is the reason for this?

Benchmarking will take a lot of time and is not worth to run it on every PR change.

It currently takes ~6 mins, but it will take much longer, because:
1. We will add more benchmark cases.
2. For comparison with master we need to build twice and benchmark twice.
3. For statistical analysis we also need more repetitions - at least 9, bump from 3.

I have not found any better trigger than using CircleCi's REST API.

axic:
> I have not found any better trigger than using CircleCi's REST API.

We could run it only on `master`, e.g. once a PR is merged. However this only makes sense if someone is actually checking the results.

chfast:
> We could run it only on `master`, e.g. once a PR is merged. However this only makes sense if someone is actually checking the results.

That is not good idea. We want to compare against master (not possible after merge). And have at least link to benchmarking report in PR check list.

axic:
Not sure I understand this PR then, isn't this just disabling benchmarking entirely on CI?

chfast:
> Not sure I understand this PR then, isn't this just disabling benchmarking entirely on CI?

There are 2 workflows now: testing and benchmarking. "Testing" runs for every PR change. "Benchmarking" must be manually triggered by Circle CI API (can be done with curl command of with the provided script).

I have triggered it for this PR, so you can see 

> ci/circleci: benchmark — Your tests passed on CircleCI! 

being a part of "Benchmarking" workflow: https://circleci.com/workflow-run/f2de36e5-4597-44bf-8f8b-21f45d4e50cf.

Later I will do following changes:
- eliminate dependency on `release-linux` job - we will build it in the benchmarking virtual machine.
- build PR branch base commit.
- compare benchmarking stats of these 2.
In final stage it is expected to take 30-60 mins.

On every day we will trigger benchmarking only when we are interested in performance impact.

axic:
> On every day we will trigger benchmarking only when we are interested in performance impact.

Who will trigger it? Can't we have a cronjob on circleci?

chfast:
> Who will trigger it? Can't we have a cronjob on circleci?

Anyone with CircleCI access can trigger it. No, cronjob is not useful because we need this for individual PRs.

PR #244, opened 2020-03-26
validation: Detect stack underflow

chfast:
Can have false positives for calls as callee type information is not available during parsing yet (see single DISABLED test).

Based on https://webassembly.github.io/spec/core/appendix/algorithm.html.

Requires #254.


lib/fizzy/instructions.hpp
```diff
@@ -0,0 +1,19 @@
...
+#include <cstdint>
+
+namespace fizzy
+{
+struct InstructionMetrics
+{
+    /// The minimum number of the stack items required for the instruction.
+    int8_t stack_height_required;
+
+    /// The EVM stack height change caused by the instruction execution,
```
gumb0:
```suggestion
    /// The stack height change caused by the instruction execution,
```

test/unittests/parser_expr_test.cpp
```diff
@@ -234,9 +234,9 @@ TEST(parser, load_store_immediates_out_of_bounds)
              Instr::i64_load32_u, Instr::i32_store, Instr::i64_store, Instr::i32_store8,
              Instr::i32_store16, Instr::i64_store8, Instr::i64_store16, Instr::i64_store32})
     {
-        const auto code_imm1 = bytes{uint8_t(instr), 0xa0};
+        const auto code_imm1 = bytes{0x41, 0x00, 0x41, 0x00, uint8_t(instr), 0xa0};
```
gumb0: To make tests more readable, maybe makes sense to define a constant like
```
constexpr i32_push = uint8_t(Instr::i32_push);
```
?

lib/fizzy/instructions.cpp
```diff
@@ -0,0 +1,231 @@
...
+    /* if_                 = 0x04 */ {1, -1},
+    /* else_               = 0x05 */ {0, 0},
+    /*                       0x06 */ {},
+    /*                       0x07 */ {},
+    /*                       0x08 */ {},
+    /*                       0x09 */ {},
+    /*                       0x0a */ {},
+    /* end                 = 0x0b */ {0, 0},
+    /* br                  = 0x0c */ {variadic, 0},
+    /* br_if               = 0x0d */ {variadic, 0},
```
gumb0: `br_if` an `br_table` pop one value from the stack

chfast: You are absolutely right. I confused myself that this depends on the target label. This should reduce complexity of this work by 10. Thanks.

lib/fizzy/parser_expr.cpp
```diff
@@ -57,13 +58,25 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end, bool have
...
 
-        const auto instr = static_cast<Instr>(*pos++);
+        const auto opcode = *pos++;
+        const auto& metrics = metrics_table[opcode];
+
+        if (stack_height < metrics.stack_height_required)
+            throw validation_error{"stack underflow"};
+
+        stack_height += metrics.stack_height_change;
+        code.max_stack_height = std::max(code.max_stack_height, stack_height);
```
gumb0: max height is used only in tests now?

chfast: It is used in the next PR: #247. I will move it there only.

lib/fizzy/types.hpp
```diff
@@ -337,6 +339,14 @@ struct Code
     // The decoded instructions' immediate values.
     // These are instruction-type dependent fixed size value in the order of instructions.
     bytes immediates;
+
+    Code() noexcept = default;
+
+    Code(uint32_t _local_count, std::vector<Instr> _instructions, bytes _immediates) noexcept
```
gumb0: Why are constructors needed again?

chfast: A new filed is added and `Code` is used in many tests. This makes them still compile, but they don't work anyway.

> The optimization depends on `max_stack_height`. This is trivial to compute, but there are a lot of tests constructing `Code` - these require to provide this information manually. It might be good moment to upgrade these to wat2wasm tests.

So we rather need update tests rather than a `Code` constructor.

chfast: The commit has been removed as well.

gumb0:
Please rebase, to get `validation_error` handling in spectests

chfast:
> Please rebase, to get `validation_error` handling in spectests

It's not merged yet: #241.

chfast:
Spectests failures are due to false positive stack underflow failures. We may want to fix the issue first.

lib/fizzy/parser_expr.cpp
```diff
@@ -24,6 +25,17 @@ struct LabelPosition
 {
     Instr instruction = Instr::unreachable;  ///< The instruction that created the label.
     size_t immediates_offset{0};             ///< The immediates offset for block instructions.
+
+    /// Whether the block has a result.
+    /// TODO: Cleanup arity vs has_result vs have_result.
```
gumb0, line 30: I guess it should better be `arity` if you use it in `+=`

test/unittests/validation_test.cpp
```diff
@@ -113,4 +113,202 @@ TEST(validation, br_table_default_invalid_label_index)
...
+      (block
+        drop
+      )
+    )
+    */
+    const auto wasm = from_hex("0061736d01000000010401600000030201000a0a010800410202401a0b0b");
+    EXPECT_THROW_MESSAGE(parse(wasm), validation_error, "stack underflow");
+}
+
+TEST(validation, stack_underflow_in_block_with_result)
```
gumb0, line 147: This one doesn't have underflow, maybe should be renamed?
```suggestion
TEST(validation, stack_with_block_result)
```

test/unittests/validation_test.cpp
```diff
@@ -113,4 +113,202 @@ TEST(validation, br_table_default_invalid_label_index)
...
+      ;; Call argument missing.
+      call $f
+    )
+    */
+    const auto wasm =
+        from_hex("0061736d01000000010a0260017f017f6000017f03030200010a0b02040020000b040010000b");
+    EXPECT_THROW_MESSAGE(parse(wasm), validation_error, "stack underflow");
+}
+
+TEST(validation, stack_underflow_for_unreachable)
```
gumb0, line 236:
```suggestion
TEST(validation, stack_after_unreachable)
```

lib/fizzy/parser_expr.cpp
```diff
@@ -255,6 +276,7 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end, bool have
             uint8_t arity;
             std::tie(arity, pos) = parse_blocktype(pos, end);
             code.immediates.push_back(arity);
+            frame.stack_height += arity;
```
gumb0, line 279: A bit confusing, maybe add a comment that it's for enclosing frame, something like "outer frame will have additional item on stack after exit from block with result"

chfast: Added more detailed comments in all instances.

lib/fizzy/parser_expr.cpp
```diff
@@ -294,6 +326,8 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end, bool have
         {
             if (frame.instruction != Instr::if_)
                 throw parser_error{"unexpected else instruction (if instruction missing)"};
+            frame.stack_height = 0;  // Reset stack height after if.
```
axic, line 329: Why is this reset?

gumb0: After passing `if` block, it reuses the same `frame` (`control_stack.top()`) for the `else` block - it has the same `instruction` and `immediates_offset`, but `stack_height` could have been changed in `if` block.

gumb0: Actually I think `unreachable` also has to be reset here?

chfast: I think it should. In the Validation Algorithm it pops the control frame and then pushes new one with the same types.

gumb0: It would be nice to add a test with `br` inside `if`

chfast: Fixed.

The `if-else` is poorly tested as there is big number of possible combinations.

chfast: > It would be nice to add a test with `br` inside `if`

Added one with `unreachable`, but I will add one with `br` too. Also `br_table` is followed by unreachable...

chfast:
@axic Final word here?

lib/fizzy/parser_expr.cpp
```diff
@@ -63,17 +71,27 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end, bool have
...
-        const auto& frame = control_stack.top();
+        auto& frame = control_stack.top();
+
+        const auto opcode = *pos++;
+        const auto& metrics = metrics_table[opcode];
+
+        if (frame.stack_height < metrics.stack_height_required && !frame.unreachable)
+            throw validation_error{"stack underflow"};
+
+        frame.stack_height += metrics.stack_height_change;
```
chfast, line 92: Should this be increased when `frame.unreachable`?

gumb0: I think it's simpler to always increase without condition, otherwise we'd need to also add condition when increasing it in `loop`, `block` and `if_`

test/unittests/CMakeLists.txt
```diff
@@ -23,6 +23,7 @@ target_sources(
     parser_test.cpp
     stack_test.cpp
     utf8_test.cpp
+    validation_stack_test.cpp
```
axic, line 26: Need to also add it to CI.

PR #246, opened 2020-03-30
call_indirect for functions from another modules

gumb0:
This fixes remaining failures in `elem.wast` and `linking.wast`. 

There are two scenarios:

1.
- module A has exported table
- module A adds via element section some of its functions into the table.
- module B imports this table
- `call_indirect` executed with this table in the instance of B will call a function in A.

Unit test demonstrating it: https://github.com/wasmx/fizzy/blob/0a51bc1ec4db349249113ed99a0c8371c1368c65/test/unittests/execute_call_test.cpp#L353

2. 
- module A has exported table
- module B imports this table and adds via element section some of its functions into this shared table.
- module B's start function traps, so it can't be instantiated.
- After this the change to shared table can't be rolled back, and so the table still must contain the functions of module B. They can be called from module A with `call_indirect`.
(Related discussion about this scenario:
https://github.com/WebAssembly/spec/issues/993
https://github.com/WebAssembly/spec/pull/994)

Unit test demonstrating it: https://github.com/wasmx/fizzy/blob/b5d12edf384710c0bb48f730cf0efedca2180c2a/test/unittests/execute_call_test.cpp#L396

Two problems are fixed in the same PR, because fixing only 1 leads to UB in the spectest for scenario 2 (calling a function from deleted instance)
https://github.com/WebAssembly/spec/blob/9f7b0ae427e5f078cf4951d1c974aa3bcda72bc2/test/core/linking.wast#L388

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -601,7 +601,35 @@ std::unique_ptr<Instance> instantiate(Module module,
         const auto funcidx = *instance->module.startfunc;
         assert(funcidx < instance->imported_functions.size() + instance->module.funcsec.size());
         if (execute(*instance, funcidx, {}).trapped)
+        {
+            // When element section modified imported table, and then start function trapped,
+            // modifications to the table are not rolled back.
+            // Instance in this case is not being returned to the user, so it needs to be kept alive
+            // as long as functions using it are alive in the table.
+            if (!imported_tables.empty() && !instance->module.elementsec.empty())
+            {
+                // Instance may be used by several functions added to the table,
+                // so we need a shared ownership here.
+                std::shared_ptr<Instance> shared_instance = std::move(instance);
+
+                for (size_t i = 0; i < shared_instance->module.elementsec.size(); ++i)
+                {
+                    auto it_table = shared_instance->table->begin() + elementsec_offsets[i];
+                    for ([[maybe_unused]] auto _ : shared_instance->module.elementsec[i].init)
+                    {
+                        // Wrap the function with the lambda capturing shared instance
+                        auto& table_function = (*it_table)->function;
+                        table_function = [shared_instance, func = std::move(table_function)](
+                                             fizzy::Instance& _instance, std::vector<uint64_t> args,
+                                             int depth) {
+                            return func(_instance, std::move(args), depth);
+                        };
+                        ++it_table;
+                    }
+                }
+            }
             throw instantiate_error("Start function failed to execute");
+        }
     }
 
     return instance;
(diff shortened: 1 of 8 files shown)


lib/fizzy/execute.cpp
```diff
@@ -636,7 +636,30 @@ std::unique_ptr<Instance> instantiate(Module module,
         const auto funcidx = *instance->module.startfunc;
         assert(funcidx < instance->imported_functions.size() + instance->module.funcsec.size());
         if (execute(*instance, funcidx, {}).trapped)
+        {
+            // In case some functions were added to the imported table,
```
chfast: Please give more detailed explanation in the comment.

test/unittests/execute_call_test.cpp
```diff
@@ -344,3 +344,46 @@ TEST(execute_call, imported_function_from_another_module)
...
+
+TEST(execute_call, imported_table_from_another_module)
+{
+    /* wat2wasm
+    (module
+      (func $sub (param $lhs i32) (param $rhs i32) (result i32)
+        get_local $lhs
+        get_local $rhs
+        i32.sub)
+      (table (export "tab") 3 funcref)
```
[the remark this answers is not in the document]
gumb0: Can be 1

gumb0:
I plan to extract to separate PR the switch to `unique_ptr<Instance>`, to make this one less huge

test/unittests/execute_call_test.cpp
```diff
@@ -410,3 +410,25 @@ TEST(execute, call_nonempty_stack)
...
+
+    const auto module = parse(wasm);
+    auto host_foo = [](Instance& instance, std::vector<uint64_t>) -> execution_result {
+        return execute(instance, 0, {});
+    };
+    const auto host_foo_type = module.typesec[0];
+
+    auto instance = instantiate(module, {{host_foo, host_foo_type}});
+
+    EXPECT_TRUE(execute(*instance, 0, {}).trapped);
```
axic, line 527:
```suggestion
    EXPECT_THAT(execute(*instance, 0, {}), Traps());
```

gumb0:
Rebased and changed some test assertions to using `Result()` and `Traps()`

lib/fizzy/execute.cpp
```diff
@@ -758,27 +768,43 @@ execution_result execute(
...
+            const auto& actual_type = called_func->type;
             const auto& expected_type = instance.module.typesec[expected_type_idx];
             if (expected_type != actual_type)
             {
                 trap = true;
                 goto end;
             }
 
-            if (!invoke_function(actual_type, *called_func_idx, instance, stack, depth))
+            const auto num_args = actual_type.inputs.size();
```
axic: Why did all this had to be move back here from within `invoke_function`?

gumb0: Duplication is resolved in the following commit

PR #247, opened 2020-03-30
Stack optimization

chfast:
Execution comparison
```
Comparing master-exec to opt-exec                                                                                                                                       
Benchmark                                                            Time             CPU      Time Old      Time New       CPU Old       CPU New                       
-------------------------------------------------------------------------------------------------------------------------------------------------
fizzy/execute/blake2b/512_bytes_rounds_1_mean                     -0.1219         -0.1218            98            86            98            86
fizzy/execute/blake2b/512_bytes_rounds_16_mean                    -0.1459         -0.1459          1472          1257          1472          1257
fizzy/execute/ecpairing/onepoint_mean                             -0.2044         -0.2044        606879        482851        606880        482854
fizzy/execute/keccak256/512_bytes_rounds_1_mean                   -0.0619         -0.0619           104            97           104            97
fizzy/execute/keccak256/512_bytes_rounds_16_mean                  -0.0729         -0.0729          1495          1386          1495          1386
fizzy/execute/memset/256_bytes_mean                               -0.1190         -0.1189             9             8             9             8
fizzy/execute/memset/60000_bytes_mean                             -0.1218         -0.1218          1809          1589          1809          1589
fizzy/execute/mul256_opt0/input0_mean                             -0.0909         -0.0909            30            27            30            27
fizzy/execute/mul256_opt0/input1_mean                             -0.0909         -0.0909            30            27            30            27
fizzy/execute/sha1/512_bytes_rounds_1_mean                        -0.1538         -0.1538           105            89           105            89
fizzy/execute/sha1/512_bytes_rounds_16_mean                       -0.1598         -0.1598          1456          1224          1456          1224
fizzy/execute/sha256/512_bytes_rounds_1_mean                      -0.1101         -0.1101            96            85            96            85
fizzy/execute/sha256/512_bytes_rounds_16_mean                     -0.1178         -0.1178          1301          1148          1301          1148
fizzy/execute/micro/factorial/10_mean                             -0.0329         -0.0321             2             2             2             2
fizzy/execute/micro/factorial/20_mean                             -0.0284         -0.0276             3             3             3             3
fizzy/execute/micro/fibonacci/24_mean                             -0.0893         -0.0893         16020         14588         16020         14589
fizzy/execute/micro/host_adler32/1_mean                           -0.0360         -0.0359             1             1             1             1
fizzy/execute/micro/host_adler32/100_mean                         -0.0509         -0.0504             7             7             7             7
fizzy/execute/micro/host_adler32/1000_mean                        -0.0462         -0.0462            68            65            68            65
fizzy/execute/micro/spinner/1_mean                                -0.0280         -0.0296             1             0             1             0
fizzy/execute/micro/spinner/1000_mean                             -0.1301         -0.1302            13            11            13            11
```

Parsing comparison

```
Comparing master-parse to opt-parse                                                                                                                                     
Benchmark                                               Time             CPU      Time Old      Time New       CPU Old       CPU New                                    
------------------------------------------------------------------------------------------------------------------------------------
fizzy/parse/blake2b_mean                             +0.0439         +0.0439            12            13            12            13
fizzy/parse/ecpairing_mean                           +0.0320         +0.0320           678           699           678           699
fizzy/parse/keccak256_mean                           +0.0297         +0.0297            20            21            20            21
fizzy/parse/memset_mean                              +0.0112         +0.0112             3             3             3             3
fizzy/parse/mul256_opt0_mean                         +0.0276         +0.0276             4             4             4             4
fizzy/parse/sha1_mean                                +0.0240         +0.0240            19            20            19            20
fizzy/parse/sha256_mean                              +0.0090         +0.0090            33            34            33            34
fizzy/parse/micro/factorial_mean                     +0.0057         +0.0057             1             1             1             1
fizzy/parse/micro/fibonacci_mean                     +0.0065         +0.0065             1             1             1             1
fizzy/parse/micro/host_adler32_mean                  -0.0044         -0.0044             1             1             1             1
fizzy/parse/micro/spinner_mean                       -0.0209         -0.0209             1             1             1             1
```

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -378,26 +378,25 @@ inline bool store_into_memory(bytes& memory, OperandStack& stack, const uint8_t*
 template <typename Op>
 inline void unary_op(OperandStack& stack, Op op) noexcept
 {
-    using T = decltype(op(stack.pop()));
-    const auto a = static_cast<T>(stack.pop());
-    stack.push(op(a));
+    using T = decltype(op(stack.top()));
+    stack.top() = op(static_cast<T>(stack.top()));
 }
 
 template <typename Op>
 inline void binary_op(OperandStack& stack, Op op) noexcept
 {
-    using T = decltype(op(stack.pop(), stack.pop()));
+    using T = decltype(op(stack.top(), stack.top()));
     const auto val2 = static_cast<T>(stack.pop());
-    const auto val1 = static_cast<T>(stack.pop());
-    stack.push(static_cast<std::make_unsigned_t<T>>(op(val1, val2)));
+    const auto val1 = static_cast<T>(stack.top());
+    stack.top() = static_cast<std::make_unsigned_t<T>>(op(val1, val2));
 }
 
 template <typename T, template <typename> class Op>
 inline void comparison_op(OperandStack& stack, Op<T> op) noexcept
 {
     const auto val2 = static_cast<T>(stack.pop());
-    const auto val1 = static_cast<T>(stack.pop());
-    stack.push(uint32_t{op(val1, val2)});
+    const auto val1 = static_cast<T>(stack.top());
+    stack.top() = uint32_t{op(val1, val2)};
 }
 
 template <typename T>
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -251,7 +251,7 @@ uint64_t eval_constant_expression(ConstantExpression expr,
         return globals[global_idx - imported_globals.size()];
 }
 
-void branch(uint32_t label_idx, LabelStack& labels, Stack<uint64_t>& stack, const Instr*& pc,
+void branch(uint32_t label_idx, LabelStack& labels, OperandStack& stack, const Instr*& pc,
     const uint8_t*& immediates) noexcept
 {
     assert(labels.size() > label_idx);
@@ -278,11 +278,11 @@ void branch(uint32_t label_idx, LabelStack& labels, Stack<uint64_t>& stack, cons
 
 template <class F>
 bool invoke_function(
-    const FuncType& func_type, const F& func, Instance& instance, Stack<uint64_t>& stack, int depth)
+    const FuncType& func_type, const F& func, Instance& instance, OperandStack& stack, int depth)
 {
     const auto num_args = func_type.inputs.size();
     assert(stack.size() >= num_args);
-    std::vector<uint64_t> call_args(stack.end() - static_cast<ptrdiff_t>(num_args), stack.end());
+    std::vector<uint64_t> call_args{stack.rend() - num_args, stack.rend()};
     stack.shrink(stack.size() - num_args);
 
     const auto ret = func(instance, std::move(call_args), depth + 1);
@@ -302,7 +302,7 @@ bool invoke_function(
 }
 
 inline bool invoke_function(const FuncType& func_type, uint32_t func_idx, Instance& instance,
-    Stack<uint64_t>& stack, int depth)
+    OperandStack& stack, int depth)
 {
     const auto func = [func_idx](Instance& _instance, std::vector<uint64_t> args, int _depth) {
         return execute(_instance, func_idx, std::move(args), _depth);
@@ -346,7 +346,7 @@ inline DstT extend(SrcT in) noexcept
 }
 
 template <typename DstT, typename SrcT = DstT>
-inline bool load_from_memory(bytes_view memory, Stack<uint64_t>& stack, const uint8_t*& immediates)
+inline bool load_from_memory(bytes_view memory, OperandStack& stack, const uint8_t*& immediates)
 {
     const auto address = static_cast<uint32_t>(stack.pop());
     // NOTE: alignment is dropped by the parser
@@ -361,7 +361,7 @@ inline bool load_from_memory(bytes_view memory, Stack<uint64_t>& stack, const ui
 }
 
 template <typename DstT>
-inline bool store_into_memory(bytes& memory, Stack<uint64_t>& stack, const uint8_t*& immediates)
+inline bool store_into_memory(bytes& memory, OperandStack& stack, const uint8_t*& immediates)
 {
     const auto value = static_cast<DstT>(stack.pop());
     const auto address = static_cast<uint32_t>(stack.pop());
@@ -376,7 +376,7 @@ inline bool store_into_memory(bytes& memory, Stack<uint64_t>& stack, const uint8
 }
 
 template <typename Op>
-inline void unary_op(Stack<uint64_t>& stack, Op op) noexcept
+inline void unary_op(OperandStack& stack, Op op) noexcept
 {
     using T = decltype(op(stack.pop()));
     const auto a = static_cast<T>(stack.pop());
@@ -384,7 +384,7 @@ inline void unary_op(Stack<uint64_t>& stack, Op op) noexcept
 }
 
 template <typename Op>
-inline void binary_op(Stack<uint64_t>& stack, Op op) noexcept
+inline void binary_op(OperandStack& stack, Op op) noexcept
 {
     using T = decltype(op(stack.pop(), stack.pop()));
     const auto val2 = static_cast<T>(stack.pop());
@@ -393,7 +393,7 @@ inline void binary_op(Stack<uint64_t>& stack, Op op) noexcept
 }
 
 template <typename T, template <typename> class Op>
-inline void comparison_op(Stack<uint64_t>& stack, Op<T> op) noexcept
+inline void comparison_op(OperandStack& stack, Op<T> op) noexcept
 {
     const auto val2 = static_cast<T>(stack.pop());
     const auto val1 = static_cast<T>(stack.pop());
@@ -645,8 +645,7 @@ execution_result execute(
     std::vector<uint64_t> locals = std::move(args);
     locals.resize(locals.size() + code.local_count);
 
-    // TODO: preallocate fixed stack depth properly
-    Stack<uint64_t> stack;
+    OperandStack stack(static_cast<size_t>(code.max_stack_height));
 
     LabelStack labels;
     // The function's implicit block.
@@ -1540,8 +1539,7 @@ execution_result execute(
 
 end:
     assert(labels.empty() || trap);
-    // move allows to return derived Stack<uint64_t> instance into base vector<uint64_t> value
-    return {trap, std::move(stack)};
+    return {trap, {stack.rbegin(), stack.rend()}};
 }
 
 execution_result execute(const Module& module, FuncIdx func_idx, std::vector<uint64_t> args)
--- a/lib/fizzy/parser_expr.cpp
+++ b/lib/fizzy/parser_expr.cpp
@@ -137,9 +137,19 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end, const Mod
         auto& frame = control_stack.top();
         const auto& metrics = metrics_table[opcode];
 
-        if (!frame.unreachable &&
-            (frame.stack_height - frame.parent_stack_height) < metrics.stack_height_required)
-            throw validation_error{"stack underflow"};
+        if (!frame.unreachable)
+        {
+            if ((frame.stack_height - frame.parent_stack_height) < metrics.stack_height_required)
+                throw validation_error{"stack underflow"};
+
+            // Update code's max_stack_height using frame.stack_height of the previous instruction.
+            // At this point frame.stack_height includes additional changes to the stack height
+            // if the previous instruction is a call/call_indirect.
+            // This way the update is skipped for end/else instructions (because their frame is
+            // already popped/reset), but it does not matter, as these instructions do not modify
+            // stack height anyway.
+            code.max_stack_height = std::max(code.max_stack_height, frame.stack_height);
+        }
 
         frame.stack_height += metrics.stack_height_change;
 
--- a/lib/fizzy/stack.hpp
+++ b/lib/fizzy/stack.hpp
@@ -6,6 +6,7 @@
 
 #include <cassert>
 #include <cstdint>
+#include <memory>
 #include <vector>
 
 namespace fizzy
@@ -47,5 +48,80 @@ public:
         assert(new_size <= size());
         resize(new_size);
     }
+};
+
+class OperandStack
+{
+    /// The pointer to the top item, or below the stack bottom if stack is empty.
+    ///
+    /// This pointer always alias m_storage, but it is kept as the first field
+    /// because it is accessed the most. Therefore, it must be initialized
+    /// in the constructor after the m_storage.
+    uint64_t* m_top;
+
+    /// The storage for items.
+    std::unique_ptr<uint64_t[]> m_storage;
+
+public:
+    /// Default constructor. Sets the top item pointer to below the stack bottom.
+    explicit OperandStack(size_t max_stack_height)
+      : m_storage{std::make_unique<uint64_t[]>(max_stack_height)}
+    {
+        m_top = m_storage.get() - 1;
+    }
+
+    OperandStack(const OperandStack&) = delete;
+    OperandStack& operator=(const OperandStack&) = delete;
+
+    /// The current number of items on the stack (aka stack height).
+    [[nodiscard]] size_t size() const noexcept
+    {
+        return static_cast<size_t>(m_top + 1 - m_storage.get());
+    }
+
+    /// Returns the reference to the top item.
+    /// Requires non-empty stack.
+    [[nodiscard]] auto& top() noexcept
+    {
+        assert(size() != 0);
+        return *m_top;
+    }
+
+    /// Returns the reference to the stack item on given position from the stack top.
+    /// Requires index < size().
+    [[nodiscard]] auto& operator[](size_t index) noexcept
+    {
+        assert(index < size());
+        return *(m_top - index);
+    }
+
+    /// Pushes an item on the stack.
+    /// The stack max height limit is not checked.
+    void push(uint64_t item) noexcept { *++m_top = item; }
+
+    /// Returns an item popped from the top of the stack.
+    /// Requires non-empty stack.
+    auto pop() noexcept
+    {
+        assert(size() != 0);
+        return *m_top--;
+    }
+
+    /// Shrinks the stack to the given new size by dropping items from the top.
+    ///
+    /// Requires new_size <= size().
+    /// shrink(0) clears entire stack and moves the top pointer below the stack base.
+    void shrink(size_t new_size) noexcept
+    {
+        assert(new_size <= size());
+        // For new_size == 0, the m_top will point below the storage.
+        m_top = m_storage.get() + new_size - 1;
+    }
+
+    /// Returns iterator to the bottom of the stack.
+    [[nodiscard]] const uint64_t* rbegin() const noexcept { return m_storage.get(); }
+
+    /// Returns end iterator counting from the bottom of the stack.
+    [[nodiscard]] const uint64_t* rend() const noexcept { return m_top + 1; }
 };
 }  // namespace fizzy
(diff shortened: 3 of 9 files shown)


lib/fizzy/execute.cpp
```diff
@@ -323,15 +323,20 @@ const FuncType& function_type(const Instance& instance, FuncIdx idx)
...
+    OperandStack& stack, int depth)
 {
     if (depth == CallStackLimit)
         return false;
 
     const auto num_args = func_type.inputs.size();
     assert(stack.size() >= num_args);
-    std::vector<uint64_t> call_args(stack.end() - static_cast<ptrdiff_t>(num_args), stack.end());
-    stack.resize(stack.size() - num_args);
+    std::vector<uint64_t> call_args;
```
gumb0, line 333: It still can be done with a constructor, for example
```suggestion
    const auto args_begin = &stack[0] - num_args + 1;
    std::vector<uint64_t> call_args(args_begin, args_begin + num_args);
```

lib/fizzy/execute.cpp
```diff
@@ -323,15 +323,20 @@ const FuncType& function_type(const Instance& instance, FuncIdx idx)
...
+    if (num_args != 0)
+    {
+        call_args.reserve(num_args);
+        std::copy_n(&stack[num_args - 1], num_args, std::back_inserter(call_args));
+    }
+    stack.shrink(stack.size() - num_args);
```
gumb0, line 339: if you stay with `if (num_args != 0)`, this line can be inside

gumb0:
`unary_op`, `binary_op`, `comparison_op` could also assign to `stack.top()` instead of pop then push

lib/fizzy/execute.cpp
chfast: The `&stack[0]` is invalid for empty stack.

chfast:
> `unary_op`, `binary_op`, `comparison_op` could also assign to `stack.top()` instead of pop then push

Not easy with of the casts present there.

lib/fizzy/execute.cpp
gumb0: Can it?

chfast: Yes. I unsuccessfully pushed this change last week.

lib/fizzy/stack.hpp
```diff
@@ -1,38 +1,109 @@
...
+    /// Requires size <= size().
+    /// shrink(0) clears entire stack and moves the top pointer below the stack base
+    /// (as in constructor) causing pointer arithmetic overflow - this is intended
+    /// and sanitizer checks for such behavior are disabled.
+    [[gnu::no_sanitize("pointer-overflow"), clang::no_sanitize("pointer-overflow")]] void shrink(
+        size_t size) noexcept
+    {
+        assert(size <= this->size());
+        // For size == 0, the m_top will point below the storage.
+        m_top = m_storage.get() + (size - 1);
```
gumb0, line 106: Are parentheses required here?
Wouldn't `m_storage.get() + size - 1`  or `m_storage.get() - 1 + size` avoid arithmetic overflow?

lib/fizzy/stack.hpp
```diff
@@ -1,38 +1,109 @@
...
+    }
+
+    /// Shrinks the stack to the given size by dropping item from the top.
+    ///
+    /// Requires size <= size().
+    /// shrink(0) clears entire stack and moves the top pointer below the stack base
+    /// (as in constructor) causing pointer arithmetic overflow - this is intended
+    /// and sanitizer checks for such behavior are disabled.
+    [[gnu::no_sanitize("pointer-overflow"), clang::no_sanitize("pointer-overflow")]] void shrink(
+        size_t size) noexcept
```
gumb0, line 102: maybe
```suggestion
        size_t new_size) noexcept
```

lib/fizzy/stack.hpp
```diff
@@ -1,38 +1,109 @@
...
+
+    /// Returns an item popped from the top of the stack.
+    /// Requires non-empty stack.
+    auto pop() noexcept
+    {
+        assert(size() != 0);
+        return *m_top--;
+    }
+
+    /// Shrinks the stack to the given size by dropping item from the top.
```
gumb0, line 95:
```suggestion
    /// Shrinks the stack to the given size by dropping items from the top.
```

test/unittests/stack_test.cpp
```diff
@@ -1,4 +1,5 @@
 #include "stack.hpp"
+#include <gmock/gmock.h>
 #include <gtest/gtest.h>
```
gumb0, line 3: I think it's enough to include only gmock?

lib/fizzy/stack.hpp
```diff
@@ -1,38 +1,109 @@
...
     using std::vector<T>::size;
 
-    // Also used: size(), resize(), clear(), empty(), end()
+    /// Push an item to the stack.
+    void push(T val) { std::vector<T>::emplace_back(val); }
 
-    void push(T val) { emplace_back(val); }
+    /// Returns the reference to the top item on the stack.
+    ///
+    /// This is not currently in use (except unit tests), but left as is classic stack method.
```
gumb0, line 23: It's used only as const in tests, should we make it const?

lib/fizzy/execute.cpp
gumb0: Hm right, because `operator[]` dereferences it.
What do you think about adding `cbegin()`/`cend()` to `OperandStack`?

(Here it just seems so many additional calls + `if` appeared, but maybe my intuition is wrong and it's not much of a difference performance-wise. Complexity-wise it's ok I guess)

lib/fizzy/stack.hpp
```diff
@@ -1,38 +1,109 @@
...
+        m_top = m_storage.get() - 1;
+    }
+
+    OperandStack(const OperandStack&) = delete;
+    OperandStack& operator=(const OperandStack&) = delete;
+
+    /// The current number of items on the stack (aka stack height).
+    [[nodiscard]] size_t size() const noexcept
+    {
+        return static_cast<size_t>(m_top + 1 - &m_storage[0]);
```
gumb0, line 64: Is `&m_storage[0]` fine when you call this with 0 max height size in tests?

lib/fizzy/parser_expr.cpp
```diff
@@ -91,6 +93,10 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end, bool have
 
         frame.stack_height += metrics.stack_height_change;
 
+        if (!frame.unreachable)
+            func_stack_height += metrics.stack_height_change;
```
gumb0, line 97: What about `block`, `loop`, `if_`? I think non-0-arity block should affect `func_stack_height`, too?

lib/fizzy/execute.cpp
```diff
@@ -1553,8 +1557,10 @@ execution_result execute(
 
 end:
     assert(labels.empty() || trap);
-    // move allows to return derived Stack<uint64_t> instance into base vector<uint64_t> value
-    return {trap, std::move(stack)};
+    // move allows to return derived OperandStack instance into base vector<uint64_t> value
```
gumb0, line 1560: remove comment

lib/fizzy/execute.cpp
```diff
@@ -1553,8 +1557,10 @@ execution_result execute(
...
+    return {trap, stack.size() != 0 ?
+                      std::vector<uint64_t>{&stack[stack.size() - 1], &stack[0] + 1} :
```
gumb0, line 1562: here as well would be useful to have
```
const uint64_t* OperandStack::cbegin() const;
const uint64_t* OperandStack::cend() const;
```

lib/fizzy/execute.cpp
```diff
@@ -415,15 +420,15 @@ inline bool store_into_memory(bytes& memory, Stack<uint64_t>& stack, const uint8
 }
 
 template <typename Op>
-inline void unary_op(Stack<uint64_t>& stack, Op op) noexcept
+inline void unary_op(OperandStack& stack, Op op) noexcept
 {
     using T = decltype(op(stack.pop()));
     const auto a = static_cast<T>(stack.pop());
```
gumb0, line 426: > Not easy with of the casts present there.

Not sure why, at least for `unary_op` here this seems possible:
```suggestion
    stack.top() = op(static_cast<T>(stack.top()))
```
or 
```suggestion
    auto& a = stack.top();
    a = op(static_cast<T>(a))
```

lib/fizzy/execute.cpp
chfast: The `&stack[0]` should be fine on it's own, UBSan should pass it. The problem is with the `assert()` inside `operator[]`.
One option is to drop the assert and move on.
I wasn't keen to add `begin()`/`end()` because the `std::vector call_args` is to be removed Soon™. But I think I will add it now.

lib/fizzy/parser_expr.cpp
chfast: No, they don't (but please validate that claim, because I was thinking the same as you at first).
They affect validation. E.g.  for `(block (result i32))` the validation must make sure that there is exactly one item on the stack when reaching `end` of this block. So here we are safe to only follow every other instruction stack change.

lib/fizzy/stack.hpp
chfast: It looks fine, but I can use `m_storage.get()`.

lib/fizzy/execute.cpp
chfast: I used `rbegin()`, `rend()` to be consistent with the indexing order, but I'm not confident that is the best. Please recheck.

lib/fizzy/stack.hpp
chfast: That helped with sanitizer, thanks.

lib/fizzy/parser_expr.cpp
gumb0: Yeah I see, sounds right. Block start/end doesn't pop/push anything by itself, only the instructions inside do.

lib/fizzy/execute.cpp
```diff
@@ -415,24 +415,22 @@ inline bool store_into_memory(bytes& memory, Stack<uint64_t>& stack, const uint8
...
 template <typename Op>
-inline void binary_op(Stack<uint64_t>& stack, Op op) noexcept
+inline void binary_op(OperandStack& stack, Op op) noexcept
 {
-    using T = decltype(op(stack.pop(), stack.pop()));
+    using T = decltype(op(stack.top(), stack.top()));
     const auto val2 = static_cast<T>(stack.pop());
-    const auto val1 = static_cast<T>(stack.pop());
-    stack.push(static_cast<std::make_unsigned_t<T>>(op(val1, val2)));
+    stack.top() = static_cast<std::make_unsigned_t<T>>(op(static_cast<T>(stack.top()), val2));
```
gumb0, line 429: I'd keep `val1`, this line looks too complicated (should be easy for compiler to optimize it away)
```suggestion
    const auto val1 = static_cast<T>(stack.top());
    stack.top() = static_cast<std::make_unsigned_t<T>>(op(val1, val2));
```
Or as alternative you could at least declare `T_unsigned` on a separate line

lib/fizzy/execute.cpp
```diff
@@ -415,24 +415,22 @@ inline bool store_into_memory(bytes& memory, Stack<uint64_t>& stack, const uint8
...
 }
 
 template <typename T, template <typename> class Op>
-inline void comparison_op(Stack<uint64_t>& stack, Op<T> op) noexcept
+inline void comparison_op(OperandStack& stack, Op<T> op) noexcept
 {
     const auto val2 = static_cast<T>(stack.pop());
     const auto val1 = static_cast<T>(stack.pop());
```
gumb0, line 436: this can also be optimized
```suggestion
    const auto val1 = static_cast<T>(stack.top());
    stack.top() = uint32_t{op(val1, val2)};
```

lib/fizzy/execute.cpp
gumb0: I think it's good.

test/unittests/stack_test.cpp
```diff
@@ -71,62 +68,99 @@ TEST(stack, clear)
...
 }
 
-TEST(stack, iterator)
+TEST(operand_stack, construct)
 {
-    Stack<char> stack;
-    stack.push('a');
-    stack.push('b');
-    stack.push('c');
+    OperandStack stack{0};
```
gumb0, line 79: Nit: as with other containers I'd prefer parenthesis here not to confuse with initializer list.
In execute it looks ok because it uses `max_stack_height`, but here it looks like stack containing 0.
```suggestion
    OperandStack stack(0);
```

gumb0:
Looks good, please also add a test for stack's `rbegin` / `rend`

test/unittests/stack_test.cpp
chfast: Done.

chfast:
> please also add a test for stack's `rbegin` / `rend`

Addded.

test/unittests/stack_test.cpp
```diff
@@ -3,130 +3,118 @@
...
-        EXPECT_EQ(stack.peek(item), 'x');
+TEST(operand_stack, rbegin_rend)
+{
+    OperandStack stack(3);
+    EXPECT_EQ(stack.rbegin(), stack.rend());
+
+    stack.push(1);
+    stack.push(2);
+    stack.push(3);
+    EXPECT_NE(stack.rbegin(), stack.rend());
```
gumb0, line 105: I guess `EXPECT_LT` should work, too

test/unittests/stack_test.cpp
```diff
@@ -3,130 +3,118 @@
...
+    EXPECT_EQ(*(stack.rend() - 1), 3);
 }
 
-TEST(stack, clear_on_empty)
+TEST(operand_stack, to_vector)
 {
-    Stack<char> stack;
-    stack.clear();
+    OperandStack stack(3);
+    EXPECT_TRUE(std::vector(stack.rbegin(), stack.rend()).empty());
```
gumb0, line 113: Maybe `EXPECT_THAT` would be nice here
```suggestion
    EXPECT_THAT(std::vector(stack.rbegin(), stack.rend()), IsEmpty());
```

test/unittests/stack_test.cpp
```diff
@@ -3,130 +3,118 @@
...
+
+    stack.push(1);
+    stack.push(2);
+    stack.push(3);
+
+    EXPECT_EQ(std::vector(stack.rbegin(), stack.rend()), (std::vector<uint64_t>{1, 2, 3}));
```
gumb0, line 119: Maybe `EXPECT_THAT` would be nice here
```suggestion
    EXPECT_THAT(std::vector(stack.rbegin(), stack.rend()), ElementsAre(1, 2, 3));
```

lib/fizzy/parser_expr.cpp
```diff
@@ -136,9 +136,16 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end, const Mod
...
-            (frame.stack_height - frame.parent_stack_height) < metrics.stack_height_required)
-            throw validation_error{"stack underflow"};
+        if (!frame.unreachable)
+        {
+            if ((frame.stack_height - frame.parent_stack_height) < metrics.stack_height_required)
+                throw validation_error{"stack underflow"};
+
+            // Update code's max_stack_height using frame.stack_height of the previous instruction.
+            // The frame.stack_height may have been updated by call instruction and it is fine
+            // to omit value for end instruction.
```
gumb0, line 146: I'm not sure what do you mean by "it is fine to omit value for end instruction."? Is it about final end instruction?

lib/fizzy/stack.hpp
```diff
@@ -4,39 +4,85 @@
...
 
+#include <cassert>
 #include <cstdint>
+#include <memory>
 #include <vector>
 
 namespace fizzy
 {
-template <typename T>
-class Stack : public std::vector<T>
```
gumb0, line 13: Can we keep `Stack`, because it is now used in #312?
Or what can be the other way forward?

test/unittests/validation_stack_test.cpp
```diff
@@ -548,8 +548,8 @@ TEST(validation_stack, if_invalid_end_stack_height)
     */
     const auto wasm = from_hex(
         "0061736d01000000010401600000030201000a1701150042014102047e4201420205420342041a0b1a1a0b");
-    parse(wasm);
-    // TODO: Add max stack height check.
+    const auto module = parse(wasm);
+    EXPECT_EQ(module.codesec[0].max_stack_height, 3);
```
gumb0, line 552: is the comment on line 537 outdated?

lib/fizzy/stack.hpp
chfast: I will drop the commit removing it, but I also changed the interface. Is it fine, or you would like to keep the original one?

lib/fizzy/parser_expr.cpp
chfast: The each `end` ending a frame. I will update the comment to

> The frame.stack_height may have been updated by call instruction. This also omits every `end` and `else` instructions as they pop/reset the frame object.

test/unittests/validation_stack_test.cpp
chfast: No, it explains `--no-check` and why this is invalid wasm. But we don't have "too many results" validation check yet (there is a placeholder for it added in #319). We will need to modify this test's expectations at some point later.

lib/fizzy/stack.hpp
gumb0: I think it's fine, I'll rebase on new interface.

lib/fizzy/parser_expr.cpp
gumb0: Still not sure I understand the second part. Do you mean that there's no need to update `max_stack_height` on `end` and `else`? But it's kinda not related to why we update `max_stack_height` according to previous instruction...

I would say we don't update `max_stack_height` on `end` and `else`, because `frame.stack_heigh` doesn't grow there anyway. Maybe here I'd just remove the second sentence.

But also wouldn't it be more straightforward to update here after `frame.stack_height` growth (for current instruction) + in one other place - `update_caller_frame`. No need for complicated comment here then. But you would need to pass `code` there...

chfast: I try to explain why the last update `frame.stack_height += metrics.stack_height_change` is ignored for `code.max_stack_height`.

I was considering changing `update_caller_frame` that way, but decided to try having the update in single place (and this is result of it).

So removing last sentence is good enough resolution? Or maybe move it to `frame.stack_height += metrics.stack_height_change`?

gumb0: If we keep the update here, I'd change the comment to something like
```
// Update code's max_stack_height using frame.stack_height of the previous instruction.
// At this point `frame.stack_height` includes additional changes to the stack if previous instruction was a call.
// This way this update is skipped for the end/else instruction (becaue their frame is already popped/reset), but it doesn't matter because they don't grow stack anyway.
```

chfast: Comment updated.

lib/fizzy/stack.hpp
chfast: Done.

lib/fizzy/parser_expr.cpp
```diff
@@ -137,9 +137,19 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end, const Mod
...
+        if (!frame.unreachable)
+        {
+            if ((frame.stack_height - frame.parent_stack_height) < metrics.stack_height_required)
+                throw validation_error{"stack underflow"};
+
+            // Update code's max_stack_height using frame.stack_height of the previous instruction.
+            // At this point frame.stack_height includes additional changes to the stack height
+            // if the previous instruction is a call/call_indirect.
+            // This way the update is skipped for end/else instructions (because their frame is
+            // already popped/reset), but it does not matter at these instructions do not modify
```
gumb0, line 149:
```suggestion
            // already popped/reset), but it does not matter, as these instructions do not modify
```

PR #248, opened 2020-03-31
test: Upgrade execution tests to wat2wasm (1)

chfast:


test/unittests/execute_call_test.cpp
```diff
@@ -210,35 +203,43 @@ TEST(execute_call, call_indirect_uninited_table)
...
+      call 0
+    )
+    */
+    const auto wasm = from_hex(
+        "0061736d010000000105016000017f020b01036d6f6403666f6f0000030201000a0601040010000b");
+
+    const auto module = parse(wasm);
 
-    auto host_foo = [](Instance&, std::vector<uint64_t>) -> execution_result {
+    constexpr auto host_foo = [](Instance&, std::vector<uint64_t>) -> execution_result {
```
axic, line 217: Does the `constexpr` matters here?

chfast: Not very much. This follows the fashion to `const` everything possible. Here it also means that this lambda is compile-time constant and does not depend on variables around it.

PR #249, opened 2020-04-01
validation: Check correctness of type indicies in function section

chfast:


test/unittests/parser_test.cpp
```diff
@@ -440,39 +440,44 @@ TEST(parser, function_section_empty)
...
+                     make_section(3, function_section) + make_section(10, code_section);
     const auto module = parse(bin);
     ASSERT_EQ(module.funcsec.size(), 1);
     EXPECT_EQ(module.funcsec[0], 0);
 }
 
 TEST(parser, function_section_with_multiple_functions)
 {
-    const auto function_section = "04000142ff01"_bytes;
+    const auto type_section = make_vec({make_functype({}, {}), make_functype({}, {}),
```
gumb0, line 455: this could have been just 2 types

chfast: The original test had unique type_idx. I kept them distinct at least.

PR #250, opened 2020-04-01
Dynamically allocate Instance

gumb0:
Required for https://github.com/wasmx/fizzy/pull/246

### DIFF
--- a/lib/fizzy/execute.hpp
+++ b/lib/fizzy/execute.hpp
@@ -64,10 +64,23 @@ struct Instance
     std::vector<uint64_t> globals;
     std::vector<ExternalFunction> imported_functions;
     std::vector<ExternalGlobal> imported_globals;
+
+    Instance(Module _module, bytes_ptr _memory, size_t _memory_max_pages, table_ptr _table,
+        std::vector<uint64_t> _globals, std::vector<ExternalFunction> _imported_functions,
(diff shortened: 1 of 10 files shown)


gumb0:
clang-tidy false positive seems to disappear, so I removed `NOLINT`

gumb0:
clang-tidy warning still happens in #246, so I won't remove `NOLINT` it here.

gumb0:
Rebased.

gumb0:
Rebased.

gumb0:
Rebased.

lib/fizzy/execute.hpp
```diff
@@ -64,10 +64,23 @@ struct Instance
     std::vector<uint64_t> globals;
     std::vector<ExternalFunction> imported_functions;
     std::vector<ExternalGlobal> imported_globals;
+
+    Instance(Module _module, bytes_ptr _memory, size_t _memory_max_pages, table_ptr _table,
```
axic, line 68: `explicit`?

gumb0: `explicit` is useful only for single parameter constructors (or the ones where parameters after the first one have default values), others can only be called explicitly anyway.

axic: Ok, C++ noob here.

PR #251, opened 2020-04-02
CI tweaks

chfast:


circle.yml
```diff
@@ -230,7 +233,8 @@ jobs:
       - checkout
       - build
       - test
-      - benchmark
+      - benchmark:
+          min_time: "0.01"
```
axic, line 237: What is this doing?

codecov.yml
```diff
@@ -1,5 +1,5 @@
 codecov:
-  require_ci_to_pass: yes
+  require_ci_to_pass: no
```
axic, line 2: Why?

circle.yml
chfast: Normally benchmarks run for at least 0.5 s. But in sanitizers run we only want to find bugs in benchmarking code. With this they run only for 10 ms and whole job finishes faster. Except for the ecpairing where single iteration takes ~30 s.

codecov.yml
chfast: Nvm, I thought it did something else.

PR #254, opened 2020-04-02
test: Update tests to avoid stack underflow 

chfast:
This is going to be reported as validation failure so fix it before #244.

### DIFF
--- a/test/unittests/parser_expr_test.cpp
+++ b/test/unittests/parser_expr_test.cpp
@@ -202,11 +202,11 @@ TEST(parser, unexpected_else)
 
 TEST(parser, call_indirect_table_index)
 {
-    const auto code1_bin = "1122000b"_bytes;
+    const auto code1_bin = i32_const(0) + "1122000b"_bytes;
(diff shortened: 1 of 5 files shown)


test/unittests/parser_expr_test.cpp
```diff
@@ -213,7 +215,7 @@ TEST(parser, control_instr_out_of_bounds)
 {
     EXPECT_THROW_MESSAGE(parse_expr("02"_bytes), parser_error, "Unexpected EOF");
     EXPECT_THROW_MESSAGE(parse_expr("03"_bytes), parser_error, "Unexpected EOF");
-    EXPECT_THROW_MESSAGE(parse_expr("04"_bytes), parser_error, "Unexpected EOF");
+    EXPECT_THROW_MESSAGE(parse_expr("410004"_bytes), parser_error, "Unexpected EOF");
```
gumb0, line 218:
```suggestion
    EXPECT_THROW_MESSAGE(parse_expr(i32_const(0) + "04"_bytes), parser_error, "Unexpected EOF");
```

PR #255, opened 2020-04-03
test: Upgrade execution tests to wat2wasm (2)

chfast:

PR #257, opened 2020-04-03
test: add EXPECT_EMPTY_RESULT

axic:
Merge after #255 and #259.

### DIFF
--- a/test/unittests/execute_test.cpp
+++ b/test/unittests/execute_test.cpp
@@ -14,6 +14,36 @@ inline bool operator==(const execution_result& lhs, const execution_result& rhs)
 {
     return lhs.trapped == rhs.trapped && lhs.stack == rhs.stack;
 }
+
+
+struct Trapped
+{
+    bool operator==(const execution_result& lhs) { return lhs.trapped == true; }
+};
+
+struct Result
+{
+    explicit Result() : m_stack{} {}
+
+    // TODO: support both i32/i64
+    explicit Result(uint64_t value) : m_stack{value} {}
+
+    // TODO: remove this once the API changes
+    explicit Result(std::vector<uint64_t> stack) : m_stack{std::move(stack)} {}
+
+    std::vector<uint64_t> m_stack;
+
+//    bool operator==(const execution_result& lhs)
+//    {
+//        return lhs.trapped == false && lhs.stack == m_stack;
+//    }
+};
+
+inline bool operator==(const execution_result& lhs, const Result& rhs)
+{
+    return lhs.trapped == false && lhs.stack == rhs.m_stack;
+}
+
 }  // namespace fizzy
 
 TEST(execute, end)
@@ -22,7 +52,7 @@ TEST(execute, end)
     (func)
     */
     const auto wasm = from_hex("0061736d01000000010401600000030201000a040102000b");
-    EXPECT_THAT(execute(parse(wasm), 0, {}), (execution_result{false, {}}));
+    EXPECT_THAT(execute(parse(wasm), 0, {}), Result());
 }
 
 TEST(execute, drop)
--- a/test/unittests/execute_test.cpp
+++ b/test/unittests/execute_test.cpp
@@ -13,9 +13,7 @@ TEST(execute, end)
     (func)
     */
     const auto wasm = from_hex("0061736d01000000010401600000030201000a040102000b");
-    const auto [trap, ret] = execute(parse(wasm), 0, {});
-    ASSERT_FALSE(trap);
-    EXPECT_EQ(ret.size(), 0);
+    EXPECT_EMPTY_RESULT(execute(parse(wasm), 0, {}));
 }
 
 TEST(execute, drop)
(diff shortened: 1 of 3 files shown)


test/unittests/execute_test.cpp
```diff
@@ -13,9 +13,7 @@ TEST(execute, end)
     (func)
     */
     const auto wasm = from_hex("0061736d01000000010401600000030201000a040102000b");
-    const auto [trap, ret] = execute(parse(wasm), 0, {});
-    ASSERT_FALSE(trap);
-    EXPECT_EQ(ret.size(), 0);
+    EXPECT_EMPTY_RESULT(execute(parse(wasm), 0, {}));
```
axic, line 16: How about merging `EXPECT_*_RESULT` to expect an array and deduce the size/values from it?

```
EXPECT_RESULT(execute(parse(wasm), 0, {}), {})
EXPECT_RESULT(execute(parse(wasm), 0, {}), {1})
EXPECT_RESULT(execute(parse(wasm), 0, {}), {1, 2})
```

@chfast @gumb0 what od you think?

test/utils/asserts.hpp
```diff
@@ -11,6 +11,14 @@
         EXPECT_EQ(r.stack[0], (expected)); \
     } while (false)
 
+#define EXPECT_EMPTY_RESULT(result)   \
+    do                                \
+    {                                 \
+        const auto r = (result);      \
+        ASSERT_FALSE(r.trapped);      \
```
gumb0, line 18: It's called `EXPECT_` but in fact does `ASSERT`.
I think I would change both `ASSERT`s to `EXPECT`s

axic: Yeah assert vs expect is plagued all over the test suite. if we want to be fully consistent then trapped should still be assert, but the size can be changed to expect. (See `EXPECT_RESULT`).

test/unittests/execute_test.cpp
chfast: Macros don't go along with `{}`: `EXPECT_RESULT(execute(parse(wasm), 0, {}), {1, 2})` will be parsing error. You need `EXPECT_RESULT(execute(parse(wasm), 0, {}), ({1, 2}))`.
I also don't expect returning all remaining stack will stay with us for long. You may try to get it work with `std::optional<uint64_t>`.

chfast:
If we are here, I think it's better to migrate/use `EXPECT_THAT()` from gmock. They work if you include `#include <gmock/gmock-matchers.h>`.

- https://github.com/google/googletest/blob/master/googlemock/docs/cook_book.md#using-matchers
- https://github.com/google/googletest/blob/master/googlemock/docs/cook_book.md#writing-new-matchers-quickly-newmatchers

Then we can implement `Result()` matcher instead of fighting with macros and use it as
```
EXPECT_THAT(execute(instance, 0, {}), Result());
EXPECT_THAT(execute(instance, 1, {}), Result(123));
```

axic:
> If we are here, I think it's better to migrate/use EXPECT_THAT() from gmock. 

Sounds good to me.

test/unittests/execute_test.cpp
```diff
@@ -1,21 +1,37 @@
...
 #include "limits.hpp"
 #include "parser.hpp"
+#include <gmock/gmock.h>
 #include <gtest/gtest.h>
 #include <test/utils/asserts.hpp>
 #include <test/utils/hex.hpp>
 
 using namespace fizzy;
 
+MATCHER(Trapped, "")
```
axic, line 11: Are we adding these to `asserts.hpp`?

chfast: Yes.

chfast, line 11: I expect that `""` is used in failure messages. Please check how it looks add descriptions.

axic: Yes, the manual said that. Will need to come up with something.

axic:
```
/Users/alex/Projects/fizzy/test/unittests/execute_test.cpp:34: Failure
Value of: execute(parse(wasm), 0, {})
Expected: trapped
  Actual: 32-byte object <00-CA 7C-0F 01-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00> (of type fizzy::execution_result)
```

It seems matchers are just lowercased. However need to have some kind of formatter for `execution_result`.

test/unittests/execute_test.cpp
```diff
@@ -1,21 +1,37 @@
 #include "execute.hpp"
 #include "limits.hpp"
 #include "parser.hpp"
+#include <gmock/gmock.h>
 #include <gtest/gtest.h>
```
gumb0, line 5: Is including just `gmock.h` enough?

test/unittests/execute_test.cpp
gumb0, line 11: following other matchers in gmock
```suggestion
MATCHER(Traps, "traps")
```

test/unittests/execute_test.cpp
```diff
@@ -1,21 +1,37 @@
...
+}
+MATCHER(Result, "")
+{
+    return arg.trapped == false && arg.stack.size() == 0;
+}
+MATCHER_P(Result, value, "")
+{
+    return arg.trapped == false && arg.stack.size() == 1 && arg.stack[0] == uint64_t(value);
+}
+MATCHER_P2(Result, value1, value2, "")
```
gumb0, line 23: Where do we need this?

axic: There were some tests doing it, but couldn't find them yesterday anymore.

PR #258, opened 2020-04-06
benchmarks: add fibonacci and factorial

axic:


axic:
The failure in wabt is due to "argument type mismatch", e.g. we are passing i32, while the code requires i64.

We have three options:
a) change the factorial code to use i32 (and that will truncate at some point)
b) like a) but also use a lower input so that it doesn't truncate
c) try to "fix" wabt_engine

@chfast @gumb0 which one should we go with?

chfast:
> We have three options:
> a) change the factorial code to use i32 (and that will truncate at some point)
> b) like a) but also use a lower input so that it doesn't truncate
> c) try to "fix" wabt_engine
> 
> @chfast @gumb0 which one should we go with?

a)

axic:
Ok then I invite you to fix it, hehe :wink:

chfast:
> Ok then I invite you to fix it, hehe wink

Fixed.

test/utils/wabt_engine.cpp
```diff
@@ -82,12 +82,18 @@ WasmEngine::Result WabtEngine::execute(
 
     const auto* e = reinterpret_cast<const wabt::interp::Export*>(func_ref);
 
+    const auto func_sig = m_env.GetFuncSignature(e->index);
```
axic, line 85: Ah, clever!

test/utils/wabt_engine.cpp
```diff
@@ -82,12 +82,18 @@ WasmEngine::Result WabtEngine::execute(
...
+
     wabt::interp::TypedValues typed_args;
-    for (const auto arg : args)
+    for (size_t i = 0; i < args.size(); ++i)
```
axic, line 88: Can you add `assert(args.size() == func_sig.param_types.size())`?

chfast: Good suggestion. Added.

axic:
```
fizzy/execute/ecpairing/multipoint 32703759095 ns   32702482549 ns            1
fizzy-bench: /home/builder/project/test/utils/wabt_engine.cpp:86: virtual WasmEngine::Result fizzy::test::WabtEngine::execute(WasmEngine::FuncRef, const std::vector<uint64_t> &): Assertion `func_sig->param_types.size() == args.size()' failed.
```

If this error is reported for `ecpairing`, then something is wrong. According to ecpairing.wat the function takes 0 inputs.

PR #260, opened 2020-04-07
Add OperandStack class

chfast:


axic:
This is the proper version from #247? Do you want to merge this standalone, before 0.1.0 ?

chfast:
I was just bored waiting for other PRs to be merged. This is cleaned up ValueStack class from #247. No relation to 0.1.0.

test/unittests/stack_test.cpp
```diff
@@ -130,3 +131,94 @@ TEST(stack, clear_on_empty)
     Stack<char> stack;
     stack.clear();
 }
+
+TEST(operand_stack, construct)
+{
+    OperandStack stack{0};
+    EXPECT_THAT(stack.size(), 0);
```
axic, line 138: Why use `EXPECT_THAT` in these simple comparisons?

chfast: I though you want to migrate to `EXPECT_THAT()` so I started using it.

axic: I don't mind personally. Not sure about impact/benefit/difference. For the more complex return types it looks like a good solution as you've suggested.

lib/fizzy/stack.hpp
```diff
@@ -35,4 +37,65 @@ class Stack : public std::vector<T>
     /// Drops @a num_elements elements from the top of the stack.
     void drop(size_t num_elements = 1) noexcept { resize(size() - num_elements); }
 };
+
+class OperandStack
+{
+    /// The pointer to the top item, or below the stack bottom if stack is empty.
+    uint64_t* m_top;
```
gumb0, line 44:
```suggestion
    uint64_t* m_top = nullptr;
```

lib/fizzy/stack.hpp
```diff
@@ -35,4 +37,65 @@ class Stack : public std::vector<T>
...
+
+    /// The storage for items.
+    std::unique_ptr<uint64_t[]> m_storage;
+
+public:
+    /// Default constructor. Sets the top item pointer to below the stack bottom.
+    explicit OperandStack(size_t max_stack_height)
+      : m_top{nullptr} /* temporarily */, m_storage{new uint64_t[max_stack_height]}
```
gumb0, line 52:
```suggestion
      : m_storage{std::make_unique<uint64_t[]>(max_stack_height)}
```

lib/fizzy/stack.hpp
```diff
@@ -35,4 +37,65 @@ class Stack : public std::vector<T>
...
+        assert(size() != 0);
+        return *m_top--;
+    }
+
+    /// Shrinks the stack to the given size by dropping item from the top.
+    /// Requires size <= size();
+    [[gnu::no_sanitize("pointer-overflow"), clang::no_sanitize("pointer-overflow")]] void shrink(
+        size_t size) noexcept
+    {
+        assert(size <= this->size());
```
gumb0, line 96: What does `size()` return here when `shrink` is called from constructor?
Maybe it's better to restrict this method to `size < this->size()` and initialize `m_top` in constructor explicitly?

lib/fizzy/stack.hpp
```diff
@@ -35,4 +37,65 @@ class Stack : public std::vector<T>
...
+      : m_top{nullptr} /* temporarily */, m_storage{new uint64_t[max_stack_height]}
+    {
+        shrink(0);
+    }
+
+    OperandStack(const OperandStack&) = delete;
+    OperandStack& operator=(const OperandStack&) = delete;
+
+    /// The current number of items on the stack (aka stack height).
+    size_t size() noexcept { return static_cast<size_t>(m_top + 1 - &m_storage[0]); }
```
gumb0, line 61:
```suggestion
    size_t size() const noexcept { return static_cast<size_t>(m_top + 1 - &m_storage[0]); }
```

lib/fizzy/stack.hpp
```diff
@@ -35,4 +37,65 @@ class Stack : public std::vector<T>
...
+    /// Returns the reference to the stack item on given position from the stack top.
+    /// Requires index < size().
+    auto& operator[](size_t index) noexcept
+    {
+        assert(index < size());
+        return *(m_top - index);
+    }
+
+    /// Pushes an item on the stack.
+    /// The stack max height limit is not checked.
```
gumb0, line 80: You could add assert to check this.

lib/fizzy/stack.hpp
chfast: I didn't what to do that because `m_top` is never `nullptr` actually but always alias `m_storage`.
This would be nicer if order of fields reverted, but we want to keep `m_top` as first one because it is accessed the most (then `&m_top === this`).

lib/fizzy/stack.hpp
chfast: Hard to do without keeping `max_stack_height`.

lib/fizzy/stack.hpp
chfast: This is bug, fixed in #247.

test/unittests/stack_test.cpp
gumb0, line 138: Explicitly using `Eq` matcher arguably is more readable (you could read it like a sentence "expect that stack size equals 0")
```suggestion
    EXPECT_THAT(stack.size(), Eq(0));
```
but I think it's ok without it.

lib/fizzy/stack.hpp
gumb0: still not reslolved?

lib/fizzy/stack.hpp
gumb0: why not include it here?

lib/fizzy/stack.hpp
gumb0: It would be useful to add a comment why we keep it the first.

lib/fizzy/stack.hpp
gumb0: ah no, it is, was looking at the wrong commit

PR #261, opened 2020-04-07
Add call depth parameter to ExternalFunction

gumb0:
This is required for `call_indirect` in #246 to check the call depth limit.
But also allows calls to functions imported from another module to start not from 0 depth, but from current value.

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -285,9 +285,6 @@ const FuncType& function_type(const Instance& instance, FuncIdx idx)
 bool invoke_function(const FuncType& func_type, uint32_t func_idx, Instance& instance,
     Stack<uint64_t>& stack, int depth)
 {
-    if (depth == CallStackLimit)
-        return false;
-
     const auto num_args = func_type.inputs.size();
     assert(stack.size() >= num_args);
     std::vector<uint64_t> call_args(stack.end() - static_cast<ptrdiff_t>(num_args), stack.end());
@@ -593,10 +590,12 @@ std::unique_ptr<Instance> instantiate(Module module,
 execution_result execute(
     Instance& instance, FuncIdx func_idx, std::vector<uint64_t> args, int depth)
 {
-    assert(depth >= 0 && depth <= CallStackLimit);
+    assert(depth >= 0);
+    if (depth > CallStackLimit)
+        return {true, {}};
 
     if (func_idx < instance.imported_functions.size())
-        return instance.imported_functions[func_idx].function(instance, std::move(args));
+        return instance.imported_functions[func_idx].function(instance, std::move(args), depth);
 
     const auto code_idx = func_idx - instance.imported_functions.size();
     assert(code_idx < instance.module.codesec.size());
@@ -1544,8 +1543,8 @@ std::optional<ExternalFunction> find_exported_function(Instance& instance, std::
         return std::nullopt;
 
     const auto idx = *opt_index;
-    auto func = [idx, &instance](fizzy::Instance&, std::vector<uint64_t> args) {
-        return execute(instance, idx, std::move(args));
+    auto func = [idx, &instance](fizzy::Instance&, std::vector<uint64_t> args, int depth) {
+        return execute(instance, idx, std::move(args), depth);
     };
 
     return ExternalFunction{std::move(func), function_type(instance, idx)};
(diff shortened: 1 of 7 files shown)


lib/fizzy/execute.cpp
```diff
@@ -325,9 +325,6 @@ const FuncType& function_type(const Instance& instance, FuncIdx idx)
 bool invoke_function(const FuncType& func_type, uint32_t func_idx, Instance& instance,
     Stack<uint64_t>& stack, int depth)
 {
-    if (depth == CallStackLimit)
```
gumb0, line 328: I decided that it's better to move this check into `execute`, because the host calling exported functions via `ExternalFunction` now can call `execute` with arbitrary `depth` value.

lib/fizzy/execute.hpp
```diff
@@ -22,7 +22,7 @@ struct Instance;
 
 struct ExternalFunction
 {
-    std::function<execution_result(Instance&, std::vector<uint64_t>)> function;
+    std::function<execution_result(Instance&, std::vector<uint64_t>, int depth)> function;
```
axic, line 25: I think this is not nice -- any way to avoid it?

gumb0: Another way I see is create another type alongside `ExternalFunction` which will be used only in the table and will have `depth`

gumb0: I mean this PR then will not be needed (`ExternalFunction` doesn't change), but then the solution in #246 will change from https://github.com/wasmx/fizzy/pull/246/files#diff-5292aa18dd21e020324623b99674f3da to something like
```
struct TableFunction
{
    std::function<execution_result(std::vector<uint64_t>, int depth)> function;
    FuncType type;
};
using table_elements = std::vector<std::optional<TableFunction>>;
```

gumb0: Here's this alternative https://github.com/wasmx/fizzy/pull/266
But I'm not sure if it's better.

chfast:
> This is required for `call_indirect` in #246 to check the call depth limit.

Why this is required for `call_indirect` / table but not for functions imported from other modules?

gumb0:
> Why this is required for `call_indirect` / table but not for functions imported from other modules?

https://github.com/wasmx/fizzy/blob/607deee85bad9f45297f27bb7ff4f12cb905269d/test/unittests/execute_call_test.cpp#L392 
Well this test case shows that imported host function can recurse back to the caller, and to handle this case it's actually required to have `depth` in `ExternalFunction` (but also required for host function to pass correct `depth` value to callback)

For wasm functions imported from another module it's probably not possible to create infinite recursion (without `call_indirect`)
But then without `depth` in `ExternalFunction` each call to imported function resets depth to 0, so theoretically you could go over implementation's depth limit and get native stack overflow using imported functions.
(There's no spec test for this, maybe because it would be dependent on implementation call limit)

lib/fizzy/execute.cpp
```diff
@@ -631,10 +628,11 @@ std::unique_ptr<Instance> instantiate(Module module,
 execution_result execute(
     Instance& instance, FuncIdx func_idx, std::vector<uint64_t> args, int depth)
 {
-    assert(depth >= 0 && depth <= CallStackLimit);
+    if (depth < 0 || depth > CallStackLimit)
```
chfast, line 631: Too defensive IMHO. Move `depth < 0` back to assert.
Note: I don't consider `execute()` a public API.

gumb0: I changed it to if, because now via `ExternalFunction` returned from `find_exported_function` it can be called with arbitrary `depth` value. Somehow this feels more like a value like can come from outside of the library.
But I guess it's still not public API, I can change it.

gumb0:
> Did you think if the call depth could be bound with instance?

What do you mean? Having it as a member of `Instance`?

chfast:
> > Did you think if the call depth could be bound with instance?
> 
> What do you mean? Having it as a member of `Instance`?

Yes. I'm guessing this is not good idea, but I'd like to see explicit examples where this fails.

gumb0:
> Yes. I'm guessing this is not good idea, but I'd like to see explicit examples where this fails.

Well one failure example is similar to "no depth passed to `ExternalFunction`": if you chain N instances calling imported function from the next one, you can get call depth `N * CallStackLimit`, so it can crash.

And a general objection is that it would be not possible to have several executions in parallel on the same instance, and probably general purpose wasm engine shouldn't disallow that (parallel calculation not using the same memory/globals should be safe)

gumb0:
Rebased.

gumb0:
Included in https://github.com/wasmx/fizzy/pull/246

PR #264, opened 2020-04-09
Modify Stack type interface in preparation for new implementation

chfast:


lib/fizzy/stack.hpp
```diff
@@ -30,9 +30,13 @@ class Stack : public std::vector<T>
         return res;
     }
 
-    T peek(size_t depth = 0) const noexcept { return (*this)[size() - depth - 1]; }
+    T& operator[](size_t index) noexcept { return std::vector<T>::operator[](size() - index - 1); }
```
gumb0, line 33: What about `const` versions of `operator[]` and `top`?

chfast: Not needed (Stack is never used as const).

gumb0: What if it is at some point in some functios, will then `vector`'s version with opposite semantics be called or a confusing compiler message output?

chfast: This is excerpt from #265 to make the other PR smaller. There `OperandStack` does not extend `std::vector`. Can be merge as part of #265 if that's preferred.

See also #260

gumb0: But it still doesn't delete `Stack` class.
Can you `=delete` const versions, if this class doesn't allow them?

PR #267, opened 2020-04-14
Call instructions stack validation

gumb0:
Depends on #281 and #305.

PR #268, opened 2020-04-15
refactor: move the loop splitting imports from instantiate to parser

gumb0:
Also add validation of imported function type index.

Required for #267

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -481,13 +481,12 @@ inline uint64_t popcnt64(uint64_t value) noexcept
     return static_cast<uint64_t>(__builtin_popcountll(value));
 }
 
-std::optional<uint32_t> find_export(
-    const Instance& instance, ExternalKind kind, std::string_view name)
+std::optional<uint32_t> find_export(const Module& module, ExternalKind kind, std::string_view name)
 {
-    const auto it = std::find_if(instance.module.exportsec.begin(), instance.module.exportsec.end(),
+    const auto it = std::find_if(module.exportsec.begin(), module.exportsec.end(),
         [kind, name](const auto& export_) { return export_.kind == kind && export_.name == name; });
 
-    return (it != instance.module.exportsec.end() ? std::make_optional(it->index) : std::nullopt);
+    return (it != module.exportsec.end() ? std::make_optional(it->index) : std::nullopt);
 }
 
 }  // namespace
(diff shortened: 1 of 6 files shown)


lib/fizzy/execute.cpp
```diff
@@ -162,36 +162,10 @@ void match_imports(const Module& module, const std::vector<ExternalFunction>& im
...
-        default:
-            assert(false);
-        }
-    }
-
-    match_imported_functions(imported_function_types, imported_functions);
-    match_imported_tables(imported_table_types, imported_tables);
-    match_imported_memories(imported_memory_types, imported_memories);
-    match_imported_globals(imported_globals_mutability, imported_globals);
+    match_imported_functions(module.imported_function_types, imported_functions);
```
chfast, line 165: In case `match_imports` has single user, I would remove it (adding TODO comment for now).

lib/fizzy/parser.cpp
```diff
@@ -489,6 +489,8 @@ Module parse(bytes_view input)
         switch (import.kind)
         {
         case ExternalKind::Function:
+            if (import.desc.function_type_index >= module.typesec.size())
+                throw validation_error{"unknown type"};
```
chfast, line 493:
```suggestion
                throw validation_error{"invalid type index of an imported function"};
```

lib/fizzy/parser.cpp
```diff
@@ -513,44 +512,38 @@ Module parse(bytes_view input)
...
-        [](const auto& import) noexcept { return import.kind == ExternalKind::Function; });
-    const auto total_func_count = static_cast<size_t>(imported_func_count) + module.funcsec.size();
+    const auto total_func_count = module.imported_function_types.size() + module.funcsec.size();
 
     if (module.startfunc && *module.startfunc >= total_func_count)
         throw parser_error{"invalid start function index"};
 
-    const auto have_memory = !module.memorysec.empty() || imported_mem_count != 0;
+    const auto have_memory =
+        !module.memorysec.empty() || !module.imported_memory_types.empty() != 0;
```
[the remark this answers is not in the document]
gumb0: hm I remember fixing these... got lost after rebase or something

lib/fizzy/execute.cpp
gumb0: I will remove it

lib/fizzy/parser.cpp
gumb0: This was a string from spec tests, but I agree more specific message is better

axic:
Can you merge "code review" commit into others?

gumb0:
> Can you merge "code review" commit into others?

done

PR #270, opened 2020-04-16
CHANGELOG and bumpversion

chfast:


chfast:
Still no way to update CHANGELOG via bumpversion...

PR #271, opened 2020-04-16
validation: table and memory number

gumb0:
1. Throw `validation_error` instead of `parser_error` in table/memory number checks.
2. Move related tests to validation_test.cpp
3. assert in `instantiate` for the same checks instead of throwing `instantiate_error`
3.1 remove related cases from instantiation tests

3.1 is required for converting instantiation tests to wat2wasm https://github.com/wasmx/fizzy/blob/a534d42636da97700db634e2640b8947f44a4735/test/unittests/instantiate_test.cpp#L172-L184

### DIFF
--- a/test/utils/wasm_binary.hpp
+++ b/test/utils/wasm_binary.hpp
@@ -5,6 +5,24 @@
 
 namespace fizzy::test
 {
+inline bytes add_size_prefix(const bytes& content)
+{
+    return leb128u_encode(content.size()) + content;
+}
+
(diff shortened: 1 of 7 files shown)


test/utils/wasm_binary.hpp
```diff
@@ -5,6 +5,24 @@
 
 namespace fizzy::test
 {
+inline bytes add_size_prefix(const bytes& content)
+{
+    return test::leb128u_encode(content.size()) + content;
```
axic, line 10: No need for these explicit namespaces here.

PR #280, opened 2020-04-23
Add a short license header to the sources

axic:
@gumb0 @chfast do we want something like this?


lib/fizzy/bytes.hpp
```diff
@@ -1,3 +1,7 @@
+// Fizzy: A fast WebAssembly interpreter
+// Copyright 2019-2020 The Fizzy Authors.
+// Licensed under the Apache License, Version 2.0.
```
axic, line 3: @chfast suggested https://spdx.org/ids-how

So would we do
```
```
or

```
// Licensed under the Apache License, Version 2.0.
```

chfast: The first one.

axic:
Are we placing this into cmake files and shell scripts too?

PR #281, opened 2020-04-23
refactor: move Module and function_type helper to module.hpp

gumb0:
Needed for https://github.com/wasmx/fizzy/pull/267

### DIFF
--- a/lib/fizzy/module.hpp
+++ b/lib/fizzy/module.hpp
@@ -44,18 +44,18 @@ struct Module
     std::vector<Memory> imported_memory_types;
     // Mutability of globals defined in import section
     std::vector<bool> imported_globals_mutability;
+
+    const FuncType& get_function_type(FuncIdx idx) const noexcept
+    {
+        assert(idx < imported_function_types.size() + funcsec.size());
+
+        if (idx < imported_function_types.size())
+            return imported_function_types[idx];
+
+        const auto type_idx = funcsec[idx - imported_function_types.size()];
+        assert(type_idx < typesec.size());
+
+        return typesec[type_idx];
+    }
 };
-
-inline const FuncType& function_type(const Module& module, FuncIdx idx) noexcept
-{
-    assert(idx < module.imported_function_types.size() + module.funcsec.size());
-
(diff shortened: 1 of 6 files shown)


lib/fizzy/module.hpp
```diff
@@ -0,0 +1,20 @@
+#pragma once
```
chfast, line 1: Now you need the header comment.

lib/fizzy/module.hpp
```diff
@@ -0,0 +1,20 @@
...
+
+#include "types.hpp"
+#include <cassert>
+
+namespace fizzy
+{
+inline const FuncType& function_type(const Module& module, FuncIdx idx)
```
chfast, line 8:
```suggestion
inline const FuncType& get_function_type(const Module& module, FuncIdx idx) noexcept
```

lib/fizzy/module.hpp
```diff
@@ -0,0 +1,24 @@
+// Fizzy: A fast WebAssembly interpreter
+// Copyright 2019-2020 The Fizzy Authors.
```
chfast, line 2:
```suggestion
// Copyright 2020 The Fizzy Authors.
```

lib/fizzy/module.hpp
```diff
@@ -9,7 +9,7 @@
 
 namespace fizzy
 {
-inline const FuncType& function_type(const Module& module, FuncIdx idx)
+inline const FuncType& get_function_type(const Module& module, FuncIdx idx)
```
chfast, line 49:
```suggestion
inline const FuncType& get_function_type(const Module& module, FuncIdx idx) noexcept
```

chfast: You can also make it a `Module`'s method now, but I'm not pushing this. It's rather blurry line what should be a method and what a free function.

gumb0: I'll make it a method, should it still be called `get_function_type` ?

chfast: Yes.

PR #285, opened 2020-04-24
test: add two more tests for wasm_engines

axic:


test/utils/wasm_engine.hpp
```diff
@@ -43,6 +43,7 @@ class WasmEngine
 
     /// Initializes the beginning of the instance's memory.
     /// The `memory` must not be empty.
+    /// Returns false if no memory is available (exported).
```
gumb0, line 46:
```suggestion
    /// Returns false if no memory is available (exported) or if `memory` doesn't fit into instance's memory.
```

axic: Fixed.

axic:
@gumb0 @chfast added test case for `init_memory` with oversized input

PR #286, opened 2020-04-24
benchmarks: add empty and single point example to ecpairing

axic:


test/benchmarks/ecpairing.inputs
```diff
@@ -1,5 +1,23 @@
...
+
+twopoint
+ecpairing_twopoint
+
+
+
+
 multipoint
-ecpairing
+ecpairing_multipoint
```
axic, line 20: Lets temporarily remove twopoint and multipoint cases from here as they take quite a bit of time.

PR #288, opened 2020-04-24
ci: fix shebang in ci-benchmark.sh

gumb0:
My shell (fish) requires shebang to be the first bytes in the file.

Fixes #280.


test/ci-benchmark.sh
```diff
@@ -1,9 +1,9 @@
+#!/usr/bin/env sh
```
axic, line 1: Ah this was a typo from me, even though I looked at this file, forgot it when copying the headers.

PR #289, opened 2020-04-24
Increase spectests.cpp coverage

gumb0:
- Add remaining normal cases to `smoketest.wast`
- Add cases that will be skipped to `smoketest.wast`
- Add cases that will fail to `smoketest-failures.wast`
- Run three smoke tests on CI:
  - `smoketest.wast `
  - `smoketest.wast` with `--skip-validation`
  - `smoketest-failures.wast`

This covers most of the remaining `spectests.cpp`, the lines left not covered would mostly be hit for invalid input json / json not generated with `wast2json`.


circle.yml
```diff
@@ -69,7 +69,7 @@ commands:
     - run:
         name: "Run smoketest with fizzy-spectests"
         working_directory: ~/build
-        command: bin/fizzy-spectests --skip-validation ~/project/test/spectests/smoketest
+        command: bin/fizzy-spectests ~/project/test/spectests/smoketest
```
axic, line 72: Based on the coverage you may want to run it with both options.

gumb0: Perhaps even three times, one more with failures ignored

test/spectests/smoketest-failures/failures.wast
```diff
@@ -0,0 +1,24 @@
+;; invalid
+(module (func (drop)))
+
+;; unlinkable
+(module (import "Mod-unnknown" "foo" (func)))
```
axic, line 5: Is this intentionally using `unnknown` and not `unknown` ?

gumb0: No, thanks :)

PR #291, opened 2020-04-24
benchmarks: fix sha1 benchmarks to use the correct logic (endianness)

axic:
Does it make sense merging this? It slows down the benchmark, but should not really have too much of an effect on comparing the engines.


test/benchmarks/sha1.c
```diff
@@ -28,8 +28,29 @@
 
 #define IS_ALIGNED_32(p) (0 == (3 & ((const char*)(p) - (const char*)0)))
 
-#define be2me_32(x) (x)
-#define be32_copy(to, index, from, length) memcpy((to) + (index), (from), (length))
```
axic, line 32: Alternative solution: leave a comment here.

PR #293, opened 2020-04-24
bench: Rework parse step

chfast:


test/utils/wasm_engine.hpp
```diff
@@ -26,16 +26,13 @@ class WasmEngine
 
     virtual ~WasmEngine() noexcept = default;
 
-    /// Parses input bytes and creates internal wasm module out of it.
+    /// Parses input wasm binary. The created module is discarded.
     /// Returns false on parsing error.
-    /// Consecutive invocations replace the internal module and invalidate the internal instance.
-    virtual bool parse(bytes_view input) = 0;
+    virtual bool parse(bytes_view input) const = 0;
```
gumb0, line 31: Maybe this should be `try_parse` ?

test/utils/fizzy_engine.cpp
```diff
@@ -29,28 +29,24 @@ std::unique_ptr<WasmEngine> create_fizzy_engine()
...
 
-bool FizzyEngine::instantiate()
+bool FizzyEngine::instantiate(bytes_view wasm_binary)
 {
     try
     {
-        m_instance = fizzy::instantiate(std::move(m_instance->module));
+        m_instance = fizzy::instantiate(fizzy::parse(wasm_binary));
     }
     catch (const fizzy::instantiate_error&)
```
gumb0, line 51: What about parse and vlidation errors? I think they should be caught here, too.

test/utils/wabt_engine.cpp
```diff
@@ -76,14 +84,6 @@ std::optional<WasmEngine::FuncRef> WabtEngine::find_function(std::string_view na
 WasmEngine::Result WabtEngine::execute(
     WasmEngine::FuncRef func_ref, const std::vector<uint64_t>& args)
 {
-    // WABT Executor/Thread will init both value and call stack vectors with zeros up front.
-    // To mitigate the performance overhead, we set sizes to much lower values
```
gumb0, line 80: Why do we not do this anymore?

chfast: Because we create single executor per `WabtEngine` instance and the initialization cost of it is not measured any more. So we can use default values as it does not matter any more.

test/utils/wasm_engine.hpp
chfast: Can be, but this API is unstable enough and not keen to do esthetic changes.

test/utils/fizzy_engine.cpp
```diff
@@ -46,13 +46,13 @@ bool FizzyEngine::parse(bytes_view input)
...
-bool FizzyEngine::instantiate()
+bool FizzyEngine::instantiate(bytes_view wasm_binary)
 {
     try
     {
-        m_instance = fizzy::instantiate(std::move(m_instance->module));
+        m_instance = fizzy::instantiate(fizzy::parse(wasm_binary));
     }
-    catch (const fizzy::instantiate_error&)
+    catch (...)
```
axic, line 51: Why not catch parser_error/instantiate_error? Do we want any other exception to be caught here? If so, would also do the same for `parse()`?

test/utils/wabt_engine.cpp
```diff
@@ -31,15 +31,20 @@ std::unique_ptr<WasmEngine> create_wabt_engine()
 
 bool WabtEngine::parse(bytes_view input)
 {
+    wabt::interp::Environment env;
+    wabt::interp::DefinedModule* module{nullptr};
```
axic, line 38: Whose's the owner of this pointer? Is it a smart pointer?

test/utils/wasm_engine.hpp
```diff
@@ -31,11 +31,9 @@ class WasmEngine
     /// Consecutive invocations replace the internal module and invalidate the internal instance.
     virtual bool parse(bytes_view input) = 0;
 
-    /// Instantiates the internal module creating an internal instance.
+    /// Instantiates the internal module from the wasm binary input (parsing included).
     /// Returns false on instantiation error.
```
axic, line 34: This is somewhat incorrect now.

test/utils/fizzy_engine.cpp
```diff
@@ -46,13 +46,13 @@ bool FizzyEngine::parse(bytes_view input)
     return true;
```
axic, line 42: Can't comment there but why do we assign to `m_instance` in `parse` after this change?

test/utils/wasm3_engine.cpp
```diff
@@ -64,9 +64,29 @@ bool Wasm3Engine::parse(bytes_view input)
     return true;
```
axic: I believe we don't need `LoadModule` and hence the runtime here, just `ParseModule`?

test/utils/fizzy_engine.cpp
chfast: We don't any more.

test/utils/fizzy_engine.cpp
chfast: This is resistant to changes in exceptions in fizzy. Still requires some debugging when this happens. This is also done for parse().

test/utils/wabt_engine.cpp
chfast: This rather works as out param and env owns it.

test/utils/wasm_engine.hpp
```diff
@@ -31,11 +31,9 @@ class WasmEngine
     /// Consecutive invocations replace the internal module and invalidate the internal instance.
```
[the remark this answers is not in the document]
chfast: No.

test/utils/fizzy_engine.cpp
axic: I saw it was done for parse in a subsequent commit. It is fine as long as it is consistent.

test/utils/wabt_engine.cpp
axic: Ok.

test/utils/fizzy_engine.cpp
axic: Yeah it was a subsequent commit.

test/utils/wasm3_engine.cpp
chfast: This is done later.

test/utils/wasm_engine.hpp
axic: Looking at the final diff, this is the only nit left.

test/utils/wasm_engine.hpp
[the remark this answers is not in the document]
axic: Fixing this is I think the only bit left then.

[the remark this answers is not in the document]
chfast: This particular comment is remove in the commit that simplifies parse().

test/utils/wasm_engine.hpp
chfast: Fixed.

axic:
@chfast cannot this be simplified (a single loop for all engines) after these changes: https://github.com/wasmx/fizzy/blob/master/test/unittests/wasm_engine_test.cpp#L36-L51

chfast:
> @chfast cannot this be simplified (a single loop for all engines) after these changes: https://github.com/wasmx/fizzy/blob/master/test/unittests/wasm_engine_test.cpp#L36-L51

I simplified it, but 2 issues remain.

PR #294, opened 2020-04-24
cmake: update wasm3 to 0.4.7 and apply patches locally

axic:
~~Need to compare benchmarks pre and post this change.~~

Should rebase my fork off wasm3 0.4.7 so we have a stable release we are benchmarking against.

### DIFF
--- a/cmake/ProjectWasm3.cmake
+++ b/cmake/ProjectWasm3.cmake
@@ -20,6 +20,7 @@ ExternalProject_Add(wasm3
     BINARY_DIR ${binary_dir}
     URL https://github.com/wasm3/wasm3/archive/v0.4.7.tar.gz
     URL_HASH SHA256=11e863a643f605d62a5276e342abb01a65d33d138d01ea0070622a3f78fa1bd5
+    PATCH_COMMAND ${CMAKE_CURRENT_LIST_DIR}/apply_patches.sh
     CMAKE_ARGS
     -DCMAKE_INSTALL_PREFIX=<INSTALL_DIR>
     -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
--- a/cmake/ProjectWasm3.cmake
+++ b/cmake/ProjectWasm3.cmake
@@ -11,22 +11,21 @@ set(binary_dir ${prefix}/src/wasm3-build)
 set(include_dir ${source_dir})
 set(wasm3_library ${binary_dir}/source/${CMAKE_STATIC_LIBRARY_PREFIX}m3${CMAKE_STATIC_LIBRARY_SUFFIX})
 
-# Should use a release, like https://github.com/wasm3/wasm3/archive/v0.4.6.tar.gz, once changes are upstreamed.
-
 ExternalProject_Add(wasm3
     EXCLUDE_FROM_ALL 1
     PREFIX ${prefix}
-    DOWNLOAD_NAME wasm3-ext-api-v2.tar.gz
+    DOWNLOAD_NAME wasm3-v0.4.7.tar.gz
     DOWNLOAD_DIR ${prefix}/downloads
     SOURCE_DIR ${source_dir}
     BINARY_DIR ${binary_dir}
-    URL https://github.com/axic/wasm3/archive/ext-api-v2.tar.gz
-    URL_HASH SHA256=6163efd2f9c5088a6c0a7e5c8f815871d8f70af5de3240c52be6c9473411fbdf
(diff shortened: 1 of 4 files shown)


axic:
If https://github.com/wasm3/wasm3/pull/129 and https://github.com/wasm3/wasm3/pull/130 are merged we could abandon our wasm3 fork for testing.

chfast:
The `ext-api-v3` seems to be broken. Is this one any better?

axic:
Yes, this one works.

cmake/ProjectWasm3.cmake
```diff
@@ -16,17 +16,18 @@ set(wasm3_library ${binary_dir}/source/${CMAKE_STATIC_LIBRARY_PREFIX}m3${CMAKE_S
...
-    URL_HASH SHA256=6163efd2f9c5088a6c0a7e5c8f815871d8f70af5de3240c52be6c9473411fbdf
+    URL https://github.com/axic/wasm3/archive/ext-api-v4.tar.gz
+    URL_HASH SHA256=5ceb543ef5f44d563cef639701d35616c323663a3b7e41996b4bd56ab59c941c
     CMAKE_ARGS
     -DCMAKE_INSTALL_PREFIX=<INSTALL_DIR>
     -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
     -DCMAKE_BUILD_TYPE=Release
-    -DBUILD_WASI_SUPPORT=FALSE
+    -DBUILD_WASI=none
+    -DBUILD_PORTABLE=ON
```
chfast, line 30: Why do you insist on disabling `-march=native`?

axic: It was a necessity because otherwise bench was crashing (binary was built on a different VM than run). I asked you a while ago whether that is still the case but you didn't respond.

However, I think it would be fair to either build all engines with `-march=native` or none, to be on a level playing field. Otherwise we are chasing invalid targets.

chfast: It is not an issue any more.

We can build with native for benchmarks, but you are overestimating the impact of it. Trying to modify other project's compiler flags seems to be dead end.

axic: That change is not an issue, the `m3_CallProper` is.

axic: To clarify, my main personal worry was to release fizzy 0.1.0 with my wasm3 fork. I want to delete my wasm3 fork, but if we release and I delete that, that means fizzy 0.1.0 is "broken".

Couldn't we alternatively just keep the changes as a patch in `ProjectWasm3.cmake`?

cmake/apply_patches.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env sh
+
+source_dir=$(pwd)
+echo $source_dir
+
+patches_dir=$(dirname $0)
```
axic, line 6: Ah, nice trick. I had a shell script too, but forgot this idea.

chfast:
The "patches" approach seems to be bad idea.
1. The `PATCH_COMMAND` should work also if called again. I cannot make it work because the `CallProper` chunk is always applied.
2. Also `patch` tool is not installed on CI images.
3. I tried `git apply` instead - it should work if used outside repo, but will not if the build dir is ignored inside Fizzy repo.

Downloading a complete zip archive much more reliable.

axic:
> The PATCH_COMMAND should work also if called again. I cannot make it work because the CallProper chunk is always applied.

Not sure I understand what this means.

> Downloading a complete zip archive much more reliable.

This is what we have right now. Question is where would it download it from?

cmake/ProjectWasm3.cmake
```diff
@@ -11,22 +11,24 @@ set(binary_dir ${prefix}/src/wasm3-build)
 set(include_dir ${source_dir})
 set(wasm3_library ${binary_dir}/source/${CMAKE_STATIC_LIBRARY_PREFIX}m3${CMAKE_STATIC_LIBRARY_SUFFIX})
 
-# Should use a release, like https://github.com/wasm3/wasm3/archive/v0.4.6.tar.gz, once changes are upstreamed.
+# Should use a release, like https://github.com/wasm3/wasm3/archive/v0.4.7.tar.gz, once changes are upstreamed.
```
chfast, line 14: This is actually used.

axic: I guess it is not relevant since we have the patches here. Removed.

axic:
@chfast do you want to trigger the benchmarking?

drumenbassik (commented):
A

PR #297, opened 2020-04-25
Resolve jumps

chfast:
Goal is to move label stack for `block` and `loop` opcodes from execution to parser and result jumps targets there. Finally, for every `br`, `br_if`, `br_table`, `return`, `if` and `else` we will have the destination (as relative or absolute code offset) available as an immediate value during execution.

Resolving `loop` is simple. For every `br` inside `loop` the destination is the `loop` offset (available as the top of the label stack).

Resolving `block` is not simple because `br` destination is at the `end` after the `br` itself. So in the `block` label we need an array of `br` to be resolved when `end` is reached.


axic:
Closing as replaced by #312.

PR #298, opened 2020-04-25
benchmarks: Remove execute/ecparing/empty

chfast:
It does not do anything.

### DIFF
--- a/test/benchmarks/ecpairing.inputs
+++ b/test/benchmarks/ecpairing.inputs
@@ -1,9 +1,3 @@
-empty
-ecpairing_empty
-
-
-
-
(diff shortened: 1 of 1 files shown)


axic:
Hm, you are correct it indeed doesn't perform any pairing, just the interface overhead:
- https://github.com/wasmx/fizzy/blob/master/test/benchmarks/ecpairing/src/lib.rs#L13
- https://github.com/ewasm/ethereum-bn128.rs/blob/master/src/lib.rs#L121

I remembered it may be doing something.

Preferably this would also remove the `_empty` method from the source and rebuild the wasm, but oh well.

PR #300, opened 2020-04-29
Fix disabled tests around reexports of imported table/memory

gumb0:
Imported table and memory are allowed to have narrower limits than the ones defined in the module. If it is then reexported, the limits must be original (narrower) ones, otherwise it might not fit as an import to another module. The disabled tests demonstrate it https://github.com/wasmx/fizzy/blob/7dabb5255a2f1b9fe7fb253a617851184951e696/test/unittests/api_test.cpp#L278

PR adds new members to `Instance` to save the limits of either local or imported table/memory. These limits are then returned from `find_exported_table`/`find_exported_memory`.

I also tried to use `std::variant` to refactor table in https://github.com/wasmx/fizzy/pull/302 but the approach in this PR seems now much simpler.

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -181,7 +181,7 @@ std::tuple<table_ptr, Limits> allocate_table(
         return {table_ptr{nullptr, null_delete}, Limits{}};
 }
 
-std::tuple<bytes_ptr, size_t> allocate_memory(const std::vector<Memory>& module_memories,
+std::tuple<bytes_ptr, Limits> allocate_memory(const std::vector<Memory>& module_memories,
     const std::vector<ExternalMemory>& imported_memories)
 {
     static const auto bytes_delete = [](bytes* b) noexcept { delete b; };
@@ -191,12 +191,12 @@ std::tuple<bytes_ptr, size_t> allocate_memory(const std::vector<Memory>& module_
 
     if (module_memories.size() == 1)
     {
-        const size_t memory_min = module_memories[0].limits.min;
-        const size_t memory_max =
-            (module_memories[0].limits.max.has_value() ? *module_memories[0].limits.max :
-                                                         MemoryPagesLimit);
+        const auto memory_min = module_memories[0].limits.min;
+        const auto memory_max = module_memories[0].limits.max;
+
         // FIXME: better error handling
-        if ((memory_min > MemoryPagesLimit) || (memory_max > MemoryPagesLimit))
+        if ((memory_min > MemoryPagesLimit) ||
+            (memory_max.has_value() && *memory_max > MemoryPagesLimit))
         {
             throw instantiate_error("Cannot exceed hard memory limit of " +
                                     std::to_string(MemoryPagesLimit * PageSize) + " bytes.");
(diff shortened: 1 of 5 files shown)


lib/fizzy/execute.cpp
```diff
@@ -187,42 +190,41 @@ std::tuple<bytes_ptr, size_t> allocate_memory(const std::vector<Memory>& module_
 
     if (module_memories.size() == 1)
     {
-        const size_t memory_min = module_memories[0].limits.min;
-        const size_t memory_max =
-            (module_memories[0].limits.max.has_value() ? *module_memories[0].limits.max :
-                                                         MemoryPagesLimit);
+        const auto memory_min = module_memories[0].limits.min;
+        const auto memory_max = module_memories[0].limits.max;
```
gumb0, line 194: The execution logic of `memory.grow` should stay the same, but for unlimited memory we don't save `MemoryPagesLimit` in `Instance` (we save only original module's or impored memory limits). For unlimited memory `MemoryPagesLimit` is enforced in `memory.grow` implemetation.

lib/fizzy/execute.cpp
```diff
@@ -187,42 +190,41 @@ std::tuple<bytes_ptr, size_t> allocate_memory(const std::vector<Memory>& module_
...
 
         bytes_ptr memory{imported_memories[0].data, null_delete};
-        return {std::move(memory), memory_max};
+        return {std::move(memory), imported_memories[0].limits};
     }
     else
     {
         bytes_ptr memory{nullptr, null_delete};
-        return {std::move(memory), MemoryPagesLimit};
+        return {std::move(memory), Limits{}};
```
gumb0, line 227: No need to set any limit when memory is not present, because `memory.grow` would be invalid.

lib/fizzy/execute.cpp
```diff
@@ -578,9 +579,9 @@ std::unique_ptr<Instance> instantiate(Module module,
     // because table functions will capture the pointer to instance.
     // FIXME: clang-tidy warns about potential memory leak for moving memory (which is in fact
     // safe), but also erroneously points this warning to std::move(table)
-    auto instance = std::make_unique<Instance>(std::move(module), std::move(memory), memory_max,
+    auto instance = std::make_unique<Instance>(std::move(module), std::move(memory), memory_limits,
```
gumb0, line 582: No `std::move` for limits, because clang-tidy complains that it's a move of trivial type.

axic:
Rebased.

gumb0:
> The code looks good, but can you explain that is the core issue it solves?

Added a paragraph to description.

test/unittests/instantiate_test.cpp
```diff
@@ -419,7 +426,8 @@ TEST(instantiate, memory_single_unspecified_maximum)
     auto instance = instantiate(module);
 
     ASSERT_EQ(instance->memory->size(), PageSize);
-    EXPECT_EQ(instance->memory_max_pages * PageSize, 256 * 1024 * 1024);
+    EXPECT_EQ(instance->memory_limits.min, 1);
+    EXPECT_FALSE(instance->memory_limits.max.has_value());
```
axic, line 430: This is a bit misleading though, because we do have a hard limit, albeit I believe this is more in line with the standard, that that hard limit is enforced only during `memory.grow`.

PR #302, opened 2020-04-30
Represent table and memory as std::variant of local/external/none

gumb0:
This is alternative to https://github.com/wasmx/fizzy/pull/300, should fix disabled tests.


axic:
Merged #300. Please reopen if this is still useful.

PR #303, opened 2020-05-06
cmake: Upgrade cable.cmake to 0.5.0

chfast:


axic:
Is it possible to retroactively add the "git subtree" annotation here?

chfast:
> Is it possible to retroactively add the "git subtree" annotation here?

I'm not using git subtree in Fizzy because I only want these 2 files.

PR #304, opened 2020-05-07
refactor: Unify br and return execution implementation

gumb0:


lib/fizzy/execute.cpp
```diff
@@ -744,24 +752,24 @@ execution_result execute(
         }
         case Instr::end:
         {
+            // When we reach end instruction normally, labels have one implicit function block.
```
gumb0, line 755: Still a bit of a special case remains here - in one case we pop implicit block and go to end, in another just go to end.

lib/fizzy/parser_expr.cpp
```diff
@@ -361,6 +360,15 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end, bool have
             break;
         }
 
+        case Instr::return_:
```
chfast, line 363: You should push `Instr::br` to `code.instructions` instead.

gumb0: Why? I think it might help with debugging to keep it as `return`, and in execution loop it's just  one handler for `case`s of `br`/`br_if`/`return` .

chfast: Ok.

axic:
Rebase wasn't enough as #281 changed interfaces.

I also suggest we move this to 0.2.0 as it could affect speed.

axic:
@gumb0 can you rebase and post benchmarking difference between master and this change?

gumb0:
> @gumb0 can you rebase and post benchmarking difference between master and this change?

https://app.circleci.com/pipelines/github/wasmx/fizzy/2512/workflows/8278e3e5-08fe-4afa-bcfe-e220cb32729c/jobs/15126/parallel-runs/0/steps/0-119

PR #306, opened 2020-05-10
benchmarks: add input_len argument to blake2b, keccak256, sha1, and sha256

axic:
Will squash it.


test/benchmarks/blake2b.c
```diff
@@ -216,9 +216,9 @@ int blake2b(void *out, size_t outlen,
     return 0;
 }
 
-WASM_EXPORT unsigned blake2b_bench(unsigned fill, unsigned rounds)
+WASM_EXPORT unsigned blake2b_bench(unsigned input_len, unsigned fill, unsigned rounds)
 {
-    unsigned char input[512];
+    unsigned char input[input_len];
```
chfast, line 221: The VLA is know to bloat code and affect performance negatively.
See https://godbolt.org/z/No483N.
- You can see instruction counts for both version in Clang Opt Viewer.
- GCC even eliminates the 64-byte memset.

axic: The `memset.c` benchmark uses VLA already.

The whole point here is not to optimise the benchmarking code, but make it configurable for different workloads: (assuming the memset is not significant) we can test for 512 byte long hashing, 2048 byte long hashing, etc.

chfast: Ok, somehow I thought this is for output len...

PR #307, opened 2020-05-10
Validate memory page limits

axic:


lib/fizzy/parser.cpp
```diff
@@ -213,6 +213,8 @@ inline parser_result<Memory> parse(const uint8_t* pos, const uint8_t* end)
 {
     Limits limits;
     std::tie(limits, pos) = parse_limits(pos, end);
+    if (limits.min > 65536 || (limits.max.has_value() && *limits.max > 65536))
```
axic, line 216: Should I add a constant into `limits.h`?

lib/fizzy/parser.cpp
```diff
@@ -213,6 +213,8 @@ inline parser_result<Memory> parse(const uint8_t* pos, const uint8_t* end)
...
+        throw validation_error{"maximum memory limit (4GB) exceeded"};
```
axic, line 217: Any message change?

lib/fizzy/parser.cpp
axic, line 216: Need to add unit test for this.

gumb0: There is already `PageSize`

gumb0: Ah, it's not a page size, then yes, I think it's better to add a new const

axic: What should it be named?

gumb0: can be `MemoryPagesValidationLimit`

axic: Added.

lib/fizzy/parser.cpp
axic: Changed it.

lib/fizzy/parser.cpp
axic: Added unit test.

axic:
Need to squash some commits before merging.

lib/fizzy/limits.hpp
```diff
@@ -6,9 +6,17 @@
 
 namespace fizzy
 {
+// The page size as defined by the WebAssembly 1.0 specification.
 constexpr unsigned PageSize = 65536;
+
+// The maximum memory page limit as defined by the specification. It is only possible to address 4
+// GB (32-bit) of memory.
```
axic, line 13: clang-format splits this up strangely.

PR #309, opened 2020-05-11
Change some FIXMEs to TODOs

axic:
We use more TODOs more often and actually these cases were not bugs, just missing features.

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -194,7 +194,7 @@ std::tuple<bytes_ptr, Limits> allocate_memory(const std::vector<Memory>& module_
         const auto memory_min = module_memories[0].limits.min;
         const auto memory_max = module_memories[0].limits.max;
 
-        // FIXME: better error handling
+        // TODO: better error handling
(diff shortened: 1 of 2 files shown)


lib/fizzy/execute.cpp
```diff
@@ -565,7 +565,8 @@ std::unique_ptr<Instance> instantiate(Module module,
 
     // We need to create instance before filling table,
     // because table functions will capture the pointer to instance.
-    // FIXME: clang-tidy warns about potential memory leak for moving memory (which is in fact
+    //
+    // NOTE: clang-tidy warns about potential memory leak for moving memory (which is in fact
```
gumb0, line 569: @chfast wanted to deal with it at some point, that's why it was `FIXME`

axic: So change it to `TODO`? I don't think it is a bug in our code.

gumb0: Yes, I'd make it `TODO`

axic: Fixed.

PR #310, opened 2020-05-11
Validate i32/i64.load/store alignment

axic:
Based on section 3.3.4: https://webassembly.github.io/spec/core/_download/WebAssembly.pdf


circle.yml
```diff
@@ -354,12 +354,12 @@ jobs:
           at: ~/build
       - spectest:
           skip_validation: true
-          expected_passed: 4482
-          expected_failed: 8
+          expected_passed: 4481
+          expected_failed: 9
```
axic, line 358: Hmm.

lib/fizzy/parser_expr.cpp
```diff
@@ -489,8 +489,46 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end, const Mod
         case Instr::i64_store16:
         case Instr::i64_store32:
         {
+            size_t width = 0;
```
axic, line 492: This clearly not nice, should move it into `instructions.cpp`.

lib/fizzy/parser_expr.cpp
```diff
@@ -489,8 +489,46 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end, const Mod
...
+            default:
+                break;
+            }
+
             // alignment
-            std::tie(std::ignore, pos) = leb128u_decode<uint32_t>(pos, end);
+            uint32_t align;
+            std::tie(align, pos) = leb128u_decode<uint32_t>(pos, end);
+            // NOTE: align > 4 is the hard limit (64 / 8) >> 1, and checking it to avoid overflows
+            if ((align > 4) || ((1 << align) > (width / 8)))
```
axic, line 530: Could be simplified once having instruction metrics properly.

lib/fizzy/parser_expr.cpp
axic: @chfast what do you think, can we add a `bit_width`/`byte_width`/`memory_width` entry `InstructionMetrics`?

lib/fizzy/parser_expr.cpp
```diff
@@ -490,7 +490,11 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end, const Mod
         case Instr::i64_store32:
         {
             // alignment
-            std::tie(std::ignore, pos) = leb128u_decode<uint32_t>(pos, end);
+            uint32_t align;
+            std::tie(align, pos) = leb128u_decode<uint32_t>(pos, end);
+            // NOTE: align > 4 is the hard limit (64 / 8) >> 1, and checking it to avoid overflows
+            if ((align > 4) || ((1 << align) > metrics.memory_width))
```
axic, line 496: Could also just use `uint64_t` to avoid the need for the "overflow" check?

lib/fizzy/instructions.hpp
```diff
@@ -16,6 +16,9 @@ struct InstructionMetrics
     /// The stack height change caused by the instruction execution,
     /// i.e. stack height _after_ execution - stack height _before_ execution.
     int8_t stack_height_change;
+
+    /// The memory width in bytes the instruction operates on or 0 in case not using the memory.
+    uint8_t memory_width;
```
axic, line 21: I guess this structure was aligned at 16-bits and we're breaking that now?

chfast: This is not necessary the best option to add a field to `InstructionMetrics` for sparse values (most are not relevant). This significantly increases the metrics table. But is good enough for now.

It had size 2, alignment 1. Now is size 3 alignment 1.

axic: Can go back to the extra switch statement too :wink:

circle.yml
axic: This is due to another shortcoming in the spectest, will submit an issue.

lib/fizzy/instructions.hpp
chfast: I'm for the extra switch statement. For what I checked, it will be implemented as a tiny jump table.

axic: Bruh, the instruction metrics route is so much nicer.

chfast: I benchmarked this and it does not make any difference in parsing so this can stay this way.

chfast, line 21: You may add some constructors, so no need to repeat `0` for instructions where memory alignment is not relevant.

lib/fizzy/parser_expr.cpp
```diff
@@ -546,7 +546,10 @@ parser_result<Code> parse_expr(
         case Instr::i64_store32:
         {
             // alignment
-            std::tie(std::ignore, pos) = leb128u_decode<uint32_t>(pos, end);
+            uint32_t align;
+            std::tie(align, pos) = leb128u_decode<uint32_t>(pos, end);
+            if ((1 << static_cast<uint64_t>(align)) > metrics.memory_width)
```
gumb0, line 551: why cast is needed?

axic: My compiler was complaining that shifting `1 << uint32_t` can't be represented in `uint32_t`, for some reason it picked up the type from `align`. Setting the type on `1` would be nicer I guess.

However, the compiler is not complaining anymore so can remove.

chfast: Nicer way is doing it as `uint64_t{1} << align` (or `uint32_t{1} << align`, depending on needs).

lib/fizzy/parser_expr.cpp
```diff
@@ -619,7 +619,10 @@ parser_result<Code> parse_expr(
         case Instr::i64_store32:
         {
             // alignment
-            std::tie(std::ignore, pos) = leb128u_decode<uint32_t>(pos, end);
+            uint32_t align;
+            std::tie(align, pos) = leb128u_decode<uint32_t>(pos, end);
+            if ((uint64_t{1} << align) > metrics.memory_width)
```
gumb0, line 624: You could store `log(max_align)` in the table, then you wouldn't need to shift here.

Otherwise I think you need first also to check `align < 64`, or shift would be undefined behavior.

axic: You know what I'll just go back to the original commit 975fe8a677.

```
            // NOTE: align > 4 is the hard limit (64 / 8) >> 1, and checking it to avoid overflows
            if ((align > 4) || ((1 << align) > metrics.memory_width))
```

axic: Added back the correct version (align <= 3).

lib/fizzy/parser_expr.cpp
```diff
@@ -618,8 +618,47 @@ parser_result<Code> parse_expr(
         case Instr::i64_store16:
         case Instr::i64_store32:
         {
+            size_t width = 0;
+            switch (instr)
```
chfast: I don't think it make sense to keep this commit.

lib/fizzy/parser_expr.cpp
```diff
@@ -619,7 +619,12 @@ parser_result<Code> parse_expr(
         case Instr::i64_store32:
         {
             // alignment
-            std::tie(std::ignore, pos) = leb128u_decode<uint32_t>(pos, end);
+            uint32_t align;
```
chfast, line 622:
```suggestion
            uint32_t alignment;
```
and drop `// alignment` comment?

test/unittests/validation_test.cpp
```diff
@@ -129,6 +129,149 @@ TEST(validation, memory_size_no_memory)
         parse(wasm), validation_error, "memory instructions require imported or defined memory");
 }
 
+TEST(validation, store8_invalid_alignment)
```
chfast, line 132: I see some issues with tests:
- we don't have error location information so it is unclear which `align=` is causing the exception,
- positive tests are missing, we should add ones for max alignment, e.g `i32.store8 align=1`, `i32.store16 align=2`, etc.
- Is `align=0` ok?

lib/fizzy/parser_expr.cpp
axic: It is called `align` in the spec.

chfast: Yes, drop the comment.

axic: Fixed.

test/unittests/validation_test.cpp
gumb0: Yes, test should better be split to valid cases / one invalid instruction per case.

axic: I tried to avoid having >100 test cases. Any suggestions?

axic: In section 4.4.4:
> Note: The alignment memarg.align in load and store instructions does not affect the semantics. It is an indication that the offset ea at which the memory is accessed is intended to satisfy the property ea mod 2**memarg.align = 0. A WebAssembly implementation can use this hint to optimize for the intended use. Unaligned access violating that property is still allowed and must succeed regardless of the annotation. However, it may be substantially slower on some hardware.

So `align=0` would mean `address % 1 == 0` must be satisfied for aligned access. To me it suggests align must be in the range of [0, 3] (as seen in the comment in the parser).

axic: @gumb0 @chfast happy with these kind of tests (check `store8_alignment`)? I can go through the pain of doing it for every option... will be probably a few hundred lines.

test/unittests/validation_test.cpp
```diff
@@ -129,6 +129,170 @@ TEST(validation, memory_size_no_memory)
         parse(wasm), validation_error, "memory instructions require imported or defined memory");
 }
 
+TEST(validation, store8_alignment)
```
chfast, line 132: I think it is fine for `i32.store8` (and similarly for other instructions) to have biggest passing (1) and smallest failing (2) aligns. So 2 cases per instruction.

test/unittests/validation_test.cpp
```diff
@@ -129,6 +129,170 @@ TEST(validation, memory_size_no_memory)
...
+}
+
+TEST(validation, store16_invalid_alignment)
+{
+    /* wat2wasm --no-check
+    (memory 1)
+    (func (param i32)
+      get_local 0
+      i32.const 0
+      i32.store16 align=1
```
chfast, line 178: But I also mean to drop these lines to only have single failing instruction in a test case. Otherwise it is unknown which instruction the failure is coming from.

axic:
> Plus a test for align=0?

`align=1` in text is what turns into `align=0` in binary, remember it is the 2's exponent which is stored in the binary, while the text has the actual value (`2^e`).

axic:
> Plus a test for align=0?

The text format represents the actual value, while the binary contains the exponent,  e.g. `align=1` in the text equals to `0` encoded in the binary (`2^0 => 1`)

test/unittests/validation_test.cpp
axic: I know, I only changed `store8` so far to check if everyone is happy before I spend 2 hrs changing the rest of the test cases.

chfast:
> The text format represents the actual value, while the binary contains the exponent, e.g. `align=1` in the text equals to `0` encoded in the binary (`2^0 => 1`)

Then no.

test/unittests/validation_test.cpp
chfast: Ok. I suggested reducing number of cases...

test/unittests/validation_test.cpp
```diff
@@ -129,6 +129,170 @@ TEST(validation, memory_size_no_memory)
         parse(wasm), validation_error, "memory instructions require imported or defined memory");
 }
 
+TEST(validation, store8_alignment)
+{
+    /* wat2wasm --no-check
```
gumb0, line 134: should be valid
```suggestion
    /* wat2wasm
```

axic: Added.

lib/fizzy/instructions.hpp
```diff
@@ -16,6 +16,18 @@ struct InstructionMetrics
...
```
chfast, line 21: I can see now that it would be simpler to keep this as `max_align` of `{1,2,3,4}`, as in your tests.

axic: Maybe, but `memory_width` is also fitting and describes what it does.

chfast: I mean `{0,1,2,3}` so you don't need to pre-process it during validation:
```
            if (align > metrics.max_align)
                throw validation_error{"alignment cannot exceed operand size"};
```

axic: I see. Even if we change that I still wouldn't use `instruction_metrics` in the tests, otherwise the tests aren't that prone to mistakes :)

chfast: Agreed.

axic:
```
    /// The largest acceptable alignment value satisfying `2 ** max_align < memory_width` where memory_width
    /// is the number of bytes the instruction operates on. max_align is 0 if it does not operate on memory.
    uint8_t max_align;
```

This is actually a bit weird because `max_align = 0` is a valid value and with the current way `memory_width = 0` would be invalid which is the default on non-memory instructions. It is only a nitpick. The difference in the parser is a single shift less.

lib/fizzy/parser_expr.cpp
```diff
@@ -678,8 +678,12 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end, uint32_t
         case Instr::i64_store16:
         case Instr::i64_store32:
         {
-            // alignment
-            std::tie(std::ignore, pos) = leb128u_decode<uint32_t>(pos, end);
+            uint32_t align;
+            std::tie(align, pos) = leb128u_decode<uint32_t>(pos, end);
+            // NOTE: [0, 3] is the correct range (the hard limit is log2(64 / 8)) and checking it to
+            // avoid overflows
+            if ((align > 3) || (align > metrics.max_align))
```
chfast, line 685:
```suggestion
            if (align > metrics.max_align)
```

lib/fizzy/instructions.hpp
chfast: Not an issue to me.

lib/fizzy/parser_expr.cpp
axic: I would still prefer the `memory_width` version of the code.

lib/fizzy/instructions.hpp
```diff
@@ -16,6 +17,24 @@ struct InstructionMetrics
     /// The stack height change caused by the instruction execution,
     /// i.e. stack height _after_ execution - stack height _before_ execution.
     int8_t stack_height_change;
+
+    /// The largest acceptable alignment value satisfying `2 ** max_align < memory_width` where
+    /// memory_width is the number of bytes the instruction operates on.
+    ///
+    /// This field may contain invalid value for instructions not needing it.
```
gumb0, line 24: It's default-initialized to 0, I wouldn't call it invalid

lib/fizzy/parser_expr.cpp
chfast: Why? To have nice looking table? It is mistake to keep in table data which you later need to modify by a constant expression.

lib/fizzy/instructions.hpp
axic: 0 is invalid on non-memory ops, because the change preferred by @chfast makes 0 a valid value (1<<0 = 1).

test/unittests/validation_test.cpp
```diff
@@ -129,6 +129,99 @@ TEST(validation, memory_size_no_memory)
         parse(wasm), validation_error, "memory instructions require imported or defined memory");
 }
 
+TEST(validation, store_alignment)
+{
+    // NOTE: could use instruction_metrics here, but better to have two sources of truth for testing
+    std::map<Instr, int> max_align{
```
gumb0, line 135:
```suggestion
    const std::map<Instr, int> max_align{
```

test/unittests/validation_test.cpp
```diff
@@ -129,6 +129,99 @@ TEST(validation, memory_size_no_memory)
...
+        {Instr::i32_store8, 0}, {Instr::i32_store16, 1}, {Instr::i32_store, 2},
+        // TODO: split these out once we have type validation
+        {Instr::i64_store8, 0}, {Instr::i64_store16, 1}, {Instr::i64_store32, 2},
+        {Instr::i64_store, 3},
+        // TODO: include floating point
+    };
+
+    for (auto instr : {Instr::i32_store8, Instr::i32_store16, Instr::i32_store, Instr::i64_store8,
```
gumb0, line 143: Could iterate here over `max_align` map

lib/fizzy/instructions.hpp
gumb0: 0 is valid value for memory instructions, and nonrelevant for non-memory instructions. To me it never contains "invalid value" like garbage.

axic: It does contain invalid value for non-memory ops.

test/unittests/validation_test.cpp
axic: Done.

test/unittests/validation_test.cpp
axic: Done.

test/unittests/validation_test.cpp
```diff
@@ -129,6 +129,99 @@ TEST(validation, memory_size_no_memory)
...
+        {Instr::i32_store8, 0}, {Instr::i32_store16, 1}, {Instr::i32_store, 2},
+        // TODO: split these out once we have type validation
+        {Instr::i64_store8, 0}, {Instr::i64_store16, 1}, {Instr::i64_store32, 2},
+        {Instr::i64_store, 3},
+        // TODO: include floating point
+    };
+
+    for (const auto test_case : test_cases)
+    {
+        auto instr = test_case.first;
```
gumb0, line 145:
```suggestion
        const auto instr = test_case.first;
```

test/unittests/validation_test.cpp
```diff
@@ -129,6 +129,99 @@ TEST(validation, memory_size_no_memory)
...
+        auto max_align = test_case.second;
```
gumb0, line 146:
```suggestion
        const auto max_align = test_case.second;
```

axic: Done.

test/unittests/validation_test.cpp
axic: Done.

PR #315, opened 2020-05-16
Change exception messages to lowercase + no-dot

axic:


lib/fizzy/parser.cpp
```diff
@@ -175,7 +175,7 @@ inline parser_result<Global> parse(const uint8_t* pos, const uint8_t* end)
 inline parser_result<Limits> parse_limits(const uint8_t* pos, const uint8_t* end)
 {
     if (pos == end)
-        throw parser_error{"Unexpected EOF"};
+        throw parser_error{"unexpected EOF"};
```
chfast, line 178: I would do something more about this one. It is repeated many times and the error message is confusing as it can mean the unexpected end of a section.

axic: I'd reword them in a different PR, if it needs to be done.

PR #316, opened 2020-05-16
fizzy-bench: show more useful error messages

axic:


test/bench/bench.cpp
```diff
@@ -112,13 +113,14 @@ void validate_benchmark_case(benchmark::State& state, fizzy::test::WasmEngine& e
...
         if (!result.value)
-            return state.SkipWithError("Missing result value");
+            return state.SkipWithError(("Missing result value (expected: " + expected.data() + ")"));
         else if (*result.value != *benchmark_case.expected_result)
-            return state.SkipWithError("Incorrect result");
+            return state.SkipWithError("Incorrect result value");
     }
     else if (result.value)
-        return state.SkipWithError("Unexpected result");
+            return state.SkipWithError(("Unexpected result value: " + std::to_string(*result.value)).data());
```
axic, line 123: When I was adding the benchmark it was always frustrating getting the right values out. Not sure what is the nicest way to add error messages here.

test/benchmarks/micro/factorial.inputs
```diff
@@ -2,7 +2,7 @@
 factorial
 10
 
-3628800
```
axic, line 5: This was the correct one, just changed it to trigger an error message.

PR #317, opened 2020-05-16
Validate missing memory

axic:


axic:
Need to add a unit test before merging.

test/unittests/parser_test.cpp
```diff
@@ -1241,10 +1243,23 @@ TEST(parser, data_section_memidx_nonzero)
 {
     const auto section_contents = make_vec({"0141010b0100"_bytes});
     const auto bin = bytes{wasm_prefix} + make_section(11, section_contents);
-
     EXPECT_THROW_MESSAGE(parse(bin), parser_error, "unexpected memidx value 1");
 }
 
+TEST(parser, DISABLED_empty_data_section_without_memory)
+{
+    const auto bin = bytes{wasm_prefix} + make_section(11, make_vec({}));
```
axic, line 1251: Since empty data sections end up being `datasec.empty()` in the parser, this case is not detected.

@gumb0 did you say the spec states an empty data section can or cannot be present if no memory section is present?

gumb0: It cannot. Any data section is invalid if no memory is present, according to validation spec.

image

axic: Any suggestion how to clearly support that in our code? Since section kind's are strictly ordered (and we enforce that already), the best option seems to be moving the check into the switch statement right before `parse_vec<Data>`. What do you think?

gumb0: It requires info about imports, so to move data and element section checks into the switch, you would need to move "split imports" loop inside `case SectionId::import`, too.

I'm fine in this PR to leave like it is now.

gumb0: The current check covers empty data section, too, doesn't it? So this test can be just enabled?

axic: > The current check covers empty data section, too, doesn't it? So this test can be just enabled?

It does not, if it is enabled it fails. You can try yourself and perhaps come up with a clever solution.

gumb0: Ah, I misunderstood what is '"empty data section", I thought it's "data section with empty data vector" (that would be invalid), but you mean "section consisting of empty vector". I can't find any restriction for a data section as a whole, I guess any section with empty contents is allowed.

Also, in binary spec it says:
image

gumb0, line 1251: So I think this test could be changed to having empty data vector
```suggestion
    const auto bin = bytes{wasm_prefix} + make_section(11, make_vec({0041010b00}));
```

axic: Good find, will take your change then.

PR #318, opened 2020-05-18
API for resolving function imports

gumb0:


lib/fizzy/execute.cpp
```diff
@@ -1567,6 +1567,37 @@ execution_result execute(const Module& module, FuncIdx func_idx, std::vector<uin
...
+            throw instantiate_error(
+                "imported function " + import.module + "." + import.name + " is required");
+        }
+
+        assert(import.desc.function_type_index < module.typesec.size());
+        external_functions.emplace_back(ExternalFunction{
+            std::move(it->function), module.typesec[import.desc.function_type_index]});
+    }
+
+    if (external_functions.size() != imported_functions.size())
```
axic, line 1595: Please remove this.

axic:
Can you add some unit tests?

chfast:
Maybe to so much functional, but I was imaging the instantiation workflow to look something like:

```
auto module = parse(...);
auto builder = InstanceBuilder{module};
builder.add_imported_function("host", "name", func_ptr);
builder.add_imported_function("host", "xxx", func_ptr_xxx);
auto instance = builder.instantiate();
```

axic:
I think that is an option too, but shouldn't we design that as part of the API discussion topic?

Here my goal would be something quick we can add to test the overhead of host calls.

gumb0:
> Maybe to so much functional, but I was imaging the instantiation workflow to look something like:
> 
> ```
> auto module = parse(...);
> auto builder = InstanceBuilder{module};
> builder.add_imported_function("host", "name", func_ptr);
> builder.add_imported_function("host", "xxx", func_ptr_xxx);
> auto instance = builder.instantiate();
> ```

Looks good for C++ API, but maybe this should be a part of higher-level C++ bindings... I'd leave it simpler in this PR.

lib/fizzy/execute.hpp
```diff
@@ -100,6 +100,20 @@ execution_result execute(
 // TODO: remove this helper
 execution_result execute(const Module& module, FuncIdx func_idx, std::vector<uint64_t> args);
 
+// Function that should be used by instantiate as imports, identified by module and function name.
+struct ImportedFunction
```
gumb0, line 104: I'm not very happy with the name, maybe `NamedFunctionImport` or `NamedFunction`?

lib/fizzy/execute.cpp
```diff
@@ -1567,6 +1567,34 @@ execution_result execute(const Module& module, FuncIdx func_idx, std::vector<uin
...
+            continue;
+
+        const auto it = std::find_if(
+            imported_functions.begin(), imported_functions.end(), [&import](const auto& func) {
+                return import.module == func.module && import.name == func.name;
+            });
+
+        if (it == imported_functions.end())
+        {
+            throw instantiate_error(
```
gumb0, line 1586: Decided it's not worth making separate `resolve_error`

axic:
Changelog entry perhaps?

lib/fizzy/execute.hpp
```diff
@@ -100,6 +100,20 @@ execution_result execute(
...
+{
+    std::string module;
+    std::string name;
+    std::function<execution_result(Instance&, std::vector<uint64_t>, int depth)> function;
```
axic, line 108: I think `ImportedFunction` should have two more fields:
- `std::vector<ValType> inputs;`
- `std::optional<ValType> result;` (where empty means `void`)

And check that against the type section. Abort on mismatch.

gumb0: It can be then
```
struct NamedExternalFunction
{
    std::string module;
    std::string name;
    ExternalFunction external_function;
};
```

axic: Yeah I guess it could, but that's quite a few different structs to address when declaring these. (From a user perspective.)

axic: Alternative is to use a string (similar to what wasm3 and emscripten is doing): parameters followed by colon followed by return types.

e.g. `ii:v` would mean it takes two 32-bit inputs and returns void.

The same in wasm3 is `v(ii)` and `iiv` in emscripten.

test/unittests/api_test.cpp
```diff
@@ -11,6 +11,146 @@
...
+    EXPECT_THAT(execute(*instance_reordered, 1, {}), Result(1));
+    EXPECT_THAT(execute(*instance_reordered, 2, {}), Result(2));
+    EXPECT_THAT(execute(*instance_reordered, 3, {}), Result());
+
+
+    std::vector<ImportedFunction> imported_functions_extra = {
+        {"mod1", "foo1", {}, ValType::i32, function_returning_value(0)},
+        {"mod1", "foo2", {ValType::i32}, ValType::i32, function_returning_value(1)},
+        {"mod2", "foo1", {ValType::i32}, ValType::i32, function_returning_value(2)},
+        {"mod2", "foo2", {ValType::i64}, std::nullopt, function_returning_void},
```
axic, line 90: Can you add one test case somewhere which takes 2 inputs?

gumb0: Made this void function take 2 parameters

axic:
Looks good to me, but as discussed in chat lets do the changelog in a single PR at the release time instead, so please remove the commit :)

gumb0:
Looks good to me, but as discussed in chat lets do the changelog in a single PR at the release time instead, so please remove the commit :)

> Looks good to me, but as discussed in chat lets do the changelog in a single PR at the release time instead, so please remove the commit :)

@chfast though wanted to have it in each PR, as I understood.

PR #312, opened 2020-05-12
Resolve br/br_if/br_table in parser

gumb0:
Replaces #297 

I tried to make the change history to gradually introduce resolving in parser, at first keeping control stack in execute and old instruction immediates, then deleting it in the very end. So this can be reviewed by commit.

For `br_table` it currently puts arity into immedaites for each label. This should be further optimized to have one arity immediate for entire `br_table` instruction.

PR #319, opened 2020-05-19
More stack height validation

chfast:
This validates if frames leave enough results they promised, however this check is missing for function frames (function type information is missing, but this is fixed in #312 - see a TODO comment). The situation where there is more results left is also not reported, but is easy to add - see TODO comment. This cannot be applied now because a lot of false positives when function result arity is so far assumed 0.

The frame `stack_height` now uses absolute values what will be useful for jump resolving in #312 and makes computing code's max stack height straight forward (as it is max value of any `frame.stack_height`).

### DIFF
--- a/lib/fizzy/parser_expr.cpp
+++ b/lib/fizzy/parser_expr.cpp
@@ -93,6 +93,23 @@ void update_caller_frame(ControlFrame& frame, const FuncType& func_type)
     frame.stack_height += stack_height_change;
 }
 
+void validate_result_count(const ControlFrame& frame)
+{
+    if (frame.unreachable)
+        return;
+
+    // This is checked by "stack underflow".
+    assert(frame.stack_height >= frame.parent_stack_height);
+
+    if (frame.stack_height < frame.parent_stack_height + frame.arity)
+        throw validation_error{"missing result"};
+
+    // TODO: Enable "too many results" check when having information about number of function
+    //       results.
+    // if (frame.stack_height > frame.parent_stack_height + frame.arity)
+    //     throw validation_error{"too many results"};
+}
+
 }  // namespace
 
 parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end, const Module& module)
@@ -329,6 +346,8 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end, const Mod
             if (frame.instruction != Instr::if_)
                 throw parser_error{"unexpected else instruction (if instruction missing)"};
 
+            validate_result_count(frame);  // else is the end of if.
+
             // Reset frame after if. The if result type validation not implemented yet.
             frame.stack_height = frame.parent_stack_height;
             frame.unreachable = false;
@@ -348,6 +367,8 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end, const Mod
 
         case Instr::end:
         {
+            validate_result_count(frame);
+
             if (control_stack.size() > 1)
             {
                 if (frame.instruction != Instr::loop)  // If end of block/if/else instruction.
--- a/lib/fizzy/parser_expr.cpp
+++ b/lib/fizzy/parser_expr.cpp
@@ -33,6 +33,9 @@ struct ControlFrame
     /// The instruction that created the label.
     const Instr instruction{Instr::unreachable};
 
+    /// Number of results returned from the frame.
+    const uint8_t arity{0};
+
     /// The immediates offset for block instructions.
     const size_t immediates_offset{0};
 
@@ -48,9 +51,10 @@ struct ControlFrame
     /// after branches).
     bool unreachable{false};
 
-    ControlFrame(
-        Instr _instruction, int _parent_stack_height, size_t _immediates_offset = 0) noexcept
+    ControlFrame(Instr _instruction, uint8_t _arity, int _parent_stack_height,
+        size_t _immediates_offset = 0) noexcept
       : instruction{_instruction},
+        arity{_arity},
         immediates_offset{_immediates_offset},
         parent_stack_height{_parent_stack_height},
         stack_height{_parent_stack_height}
@@ -100,7 +104,9 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end, const Mod
     // For a block/if/else instruction the value is the block/if/else's immediate offset.
     std::stack<ControlFrame> control_stack;
 
-    control_stack.push({Instr::block, 0});  // The function's implicit block.
+    // TODO: Get function result arity from the function type in Module.
+    const uint8_t func_result_arity = 0;
+    control_stack.emplace(Instr::block, func_result_arity, 0);  // The function's implicit block.
 
     const auto metrics_table = get_instruction_metrics_table();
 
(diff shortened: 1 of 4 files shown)


lib/fizzy/parser_expr.cpp
```diff
@@ -76,6 +91,23 @@ void update_caller_frame(ControlFrame& frame, const FuncType& func_type)
     frame.stack_height += stack_height_change;
 }
 
+void validate_result_count(const ControlFrame& frame)
```
axic, line 94: Why isn't this a helper or a constructor on `ControlFrame`?

lib/fizzy/parser_expr.cpp
```diff
@@ -100,10 +134,16 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end, const Mod
...
         const auto& metrics = metrics_table[opcode];
 
-        if (frame.stack_height < metrics.stack_height_required && !frame.unreachable)
-            throw validation_error{"stack underflow"};
+        if (!frame.unreachable)
+        {
+            if (frame.stack_height - metrics.stack_height_required < frame.parent_stack_height)
+                throw validation_error{"stack underflow"};
+
+            frame.stack_height += metrics.stack_height_change;
```
gumb0, line 142: I understand that you do it only for reachable instructions, so that unreachable code doesnt't influence `code.max_stack_height`.

But it doesn't look correct to update `frame.stack_height` here only when not unreachable, und update it always in `end` instruction (line 325, parent frame stack shouldn't be increased after passing unreachable block then) and also in `update_caller_frame` (line 91, unreachable `call` shouldn't update stack)

lib/fizzy/parser_expr.cpp
```diff
@@ -76,6 +91,23 @@ void update_caller_frame(ControlFrame& frame, const FuncType& func_type)
...
 }
 
+void validate_result_count(const ControlFrame& frame)
+{
+    if (!frame.unreachable)
+    {
+        // This is checked by "stack underflow".
+        assert(frame.stack_height >= frame.parent_stack_height);
+
+        if (frame.stack_height - frame.arity < frame.parent_stack_height)
```
gumb0, line 101: just a minor nitpick, but to me it would look more straightforward as
```suggestion
        if (frame.stack_height < frame.parent_stack_height + frame.arity)
```
or
```suggestion
        if (frame.stack_height - frame.parent_stack_height < frame.arity)
```
first one preferable

lib/fizzy/parser_expr.cpp
```diff
@@ -76,6 +91,23 @@ void update_caller_frame(ControlFrame& frame, const FuncType& func_type)
...
+            throw validation_error{"missing result"};
+
+        // TODO: Enable "too many results" check when having information about number of function
+        //       results.
+        // if (frame.stack_height - frame.arity > frame.parent_stack_height)
```
gumb0, line 106: This I'd express as
```suggestion
        if (frame.stack_height > frame.parent_stack_height + frame.arity)
```

lib/fizzy/parser_expr.cpp
gumb0: Maybe it would be better to update `code.max_stack_height` only when reachable and `frame.stack_height` always?

gumb0:
This commit I think should be enough to get function arity information https://github.com/wasmx/fizzy/pull/312/commits/5e86830ffd874274783204a7e0c12e17fefb077a

lib/fizzy/parser_expr.cpp
chfast: Conceptually: because `ControlFrame` is pure data type, and it does not know how to do validation.
Also: there is `update_caller_frame()`.

lib/fizzy/parser_expr.cpp
chfast: Make sense, I will use `frame.parent_stack_height + frame.arity`.

lib/fizzy/parser_expr.cpp
chfast: :+1:

lib/fizzy/parser_expr.cpp
chfast: It was like that before, so I will revert the change.

chfast:
> This commit I think should be enough to get function arity information [5e86830](https://github.com/wasmx/fizzy/commit/5e86830ffd874274783204a7e0c12e17fefb077a)

I'm leaving this for later, so we can parallelize validation and stack optimization PRs.

lib/fizzy/parser_expr.cpp
```diff
@@ -295,12 +305,12 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end, const Mod
             std::tie(arity, pos) = parse_blocktype(pos, end);
             code.immediates.push_back(arity);
 
+            // Push label with immediates offset after arity.
+            control_stack.push({Instr::block, frame.stack_height, code.immediates.size()});
```
gumb0: This `push` can theoretically invalidate `frame` reference, so `frame` should be updated before.

gumb0: ah this is changed in later commit

chfast: Oops, I have not see this. In this case this is fine as we use `std::deque`, but still I should have think about this before :) I guess I need some tests that causes stack memory grow here and then ASan should find the bug, if it ever happen.

chfast: https://github.com/wasmx/fizzy/issues/321

gumb0: Yeah I think I saw this happening in #312 (which uses `Stack` based on `vector`), that's why I noticed. Kind of subtle confusing error.

chfast: Let me know if you have any repro tests. I will try to handle that on the current master.

lib/fizzy/parser_expr.cpp
```diff
@@ -87,7 +121,9 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end, const Mod
     // For a block/if/else instruction the value is the block/if/else's immediate offset.
     std::stack<ControlFrame> control_stack;
 
-    control_stack.push({Instr::block});  // The function's implicit block.
+    // TODO: Get function result arity from the function type in Module.
```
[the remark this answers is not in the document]
chfast: This is done in #312. We can extract relevant part from #312, as a new PR later.

lib/fizzy/parser_expr.cpp
```diff
@@ -87,7 +121,9 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end, const Mod
...
+    const uint8_t func_result_arity = 0;
+    control_stack.push({Instr::block, func_result_arity, 0});  // The function's implicit block.
```
gumb0, line 126: `push`es can be `emplace`

lib/fizzy/parser_expr.cpp
```diff
@@ -76,6 +93,23 @@ void update_caller_frame(ControlFrame& frame, const FuncType& func_type)
     frame.stack_height += stack_height_change;
 }
 
+void validate_result_count(const ControlFrame& frame)
+{
+    if (!frame.unreachable)
```
gumb0, line 98: perhaps early return
```suggestion
    if (frame.unreachable)
        return;
```

lib/fizzy/parser_expr.cpp

lib/fizzy/parser_expr.cpp

PR #320, opened 2020-05-19
Add a test case for using host functions

axic:
Depends on #318.


test/unittests/wasm_engine_test.cpp
```diff
@@ -230,3 +230,32 @@ TEST(wasm_engine, memory)
         EXPECT_FALSE(engine->init_memory(bytes(PageSize + 4, 0)));
     }
 }
+
+TEST(wasm_engine, host_function)
+{
+    /* wat2wasm
+    (func $extfunc (import "env" "crc32") (param i32) (result i32))
```
axic, line 237: Should this use memory so we test that too?

test/unittests/wasm_engine_test.cpp
```diff
@@ -230,3 +230,32 @@ TEST(wasm_engine, memory)
...
+    */
+    const auto wasm = from_hex(
+        "0061736d0100000001060160017f017f020d0103656e760563726333320000030201000503010001071102066d"
+        "656d6f72790200047465737400010a08010600200010000b");
+
+    for (auto engine_create_fn : all_engines)
+    {
+        auto engine = engine_create_fn();
+        ASSERT_TRUE(engine->parse(wasm));
+        ASSERT_TRUE(engine->instantiate(wasm));
```
axic, line 252: Ideally this test case would add the imported function here, but right now I would refrain from that because that introduces a call overhead in our abstraction. Maybe I'm wrong here, @gumb0 @chfast?

test/utils/wabt_engine.cpp
```diff
@@ -35,6 +36,17 @@ std::unique_ptr<WasmEngine> create_wabt_engine()
 bool WabtEngine::parse(bytes_view input) const
 {
     wabt::interp::Environment env;
+
+    wabt::interp::HostModule* hostModule = env.AppendHostModule("env");
+    assert(hostModule != nullptr);
```
axic, line 41: Should return false.

axic:
Will also add a micro-benchmark which calls `env::crc32()` N-times and adds the results together (`v += crc32(v)`.

test/benchmarks/micro/host_adler32.wast
```diff
@@ -0,0 +1,24 @@
+(module
```
axic, line 1: Will remove this.

axic:
> Big maintenance burden. Is it number of host function times number of engines?

Only plan to have this single host function for unit testing and benchmarking.

test/utils/fizzy_engine.cpp
```diff
@@ -24,6 +25,19 @@ class FizzyEngine : public WasmEngine
...
 
+namespace
+{
+// fizzy::execution_result host_mul256(fizzy::Instance&, std::vector<uint64_t>, int)
+//{
+//    return {false, {}};
+//}
+fizzy::execution_result host_crc32(fizzy::Instance& instance, std::vector<uint64_t> args, int)
+{
+    auto ret = fizzy::adler32(instance.memory->substr(args[0], args[1]));
```
chfast, line 36: Doesn't `substr()` do a copy?

axic: Given `memory` is a `bytes_view`: https://en.cppreference.com/w/cpp/string/basic_string_view/substr

test/benchmarks/micro/host_adler32.inputs
```diff
@@ -0,0 +1,18 @@
+1
+adler32
```
axic, line 2: Need to change this name due to issue in wasm3: https://github.com/wasm3/wasm3/issues/143

test/unittests/wasm_engine_test.cpp
axic: Using memory now.

test/benchmarks/micro/host_adler32.wast
```diff
@@ -0,0 +1,24 @@
...
+  (import "env" "adler32" (func (param i32 i32) (result i32)))
+  (memory (export "memory") 1)
+  (data (i32.const 0) "11223344")
```
axic, line 4: Should fix this to use hexadecimal notation, so we actually run it on 0x11223344 and not "1122".

test/utils/wabt_engine.cpp
axic: I think this is fine as it is. Would fail on memory allocation only.

test/utils/adler32.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+#pragma once
+
+#include "bytes.hpp"
+
+namespace fizzy
+{
+// This calculates the Adler-32 checksum of the data.
+// It is used in benchmarking.
+inline uint32_t adler32(bytes_view data)
```
gumb0, line 13:
```suggestion
inline uint32_t adler32(bytes_view data) noexcept
```

test/utils/wasm3_engine.cpp
```diff
@@ -34,6 +35,18 @@ class Wasm3Engine : public WasmEngine
...
 };
 
+namespace
+{
+const void* env_adler32(IM3Runtime /*runtime*/, uint64_t* stack, void* mem)
+{
+    uint64_t* ret = stack;
+    uint32_t offset = *(uint32_t*)stack++;
+    uint32_t length = *(uint32_t*)stack++;
+    *ret = fizzy::adler32({&reinterpret_cast<uint8_t*>(mem)[offset], length});
```
gumb0, line 45:
```suggestion
    *ret = fizzy::adler32({reinterpret_cast<uint8_t*>(mem) + offset, length});
```

test/utils/wasm3_engine.cpp
```diff
@@ -34,6 +35,18 @@ class Wasm3Engine : public WasmEngine
     Result execute(FuncRef func_ref, const std::vector<uint64_t>& args) final;
 };
 
+namespace
+{
+const void* env_adler32(IM3Runtime /*runtime*/, uint64_t* stack, void* mem)
+{
+    uint64_t* ret = stack;
+    uint32_t offset = *(uint32_t*)stack++;
```
gumb0, line 43:
```suggestion
    const auto offset = *static_cast<uint32_t*>(stack++);
```

test/utils/fizzy_engine.cpp
```diff
@@ -46,12 +58,19 @@ bool FizzyEngine::instantiate(bytes_view wasm_binary)
 {
     try
     {
-        m_instance = fizzy::instantiate(fizzy::parse(wasm_binary));
+        auto module = fizzy::parse(wasm_binary);
```
chfast, line 61:
```suggestion
        const auto module = fizzy::parse(wasm_binary);
```

test/utils/fizzy_engine.cpp
```diff
@@ -24,6 +25,17 @@ class FizzyEngine : public WasmEngine
     Result execute(FuncRef func_ref, const std::vector<uint64_t>& args) final;
 };
 
+namespace
+{
+fizzy::execution_result env_adler32(fizzy::Instance& instance, std::vector<uint64_t> args, int)
+{
+    assert(instance.memory != nullptr);
+    // Since memory is a bytes_view, substr produces another bytes_view and no copy is performed.
+    const auto ret = fizzy::adler32(instance.memory->substr(args[0], args[1]));
```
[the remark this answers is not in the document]
axic: Pribably mistaken it with something. It produces the correct result though. Do you suggest `bytes_view{memory.begin() + args[0]..` or something like that as the solution?

test/utils/wasm3_engine.cpp
axic: Will do, it was copy paste from wasm3 :)

test/utils/wabt_engine.cpp
```diff
@@ -44,6 +57,26 @@ bool WabtEngine::parse(bytes_view input) const
 
 bool WabtEngine::instantiate(bytes_view wasm_binary)
 {
+    wabt::interp::HostModule* hostModule = m_env.AppendHostModule("env");
+    assert(hostModule != nullptr);
+
+    hostModule->AppendFuncExport("adler32", {{wabt::Type::I32, wabt::Type::I32}, {wabt::Type::I32}},
+        [=](const wabt::interp::HostFunc*, const wabt::interp::FuncSignature*,
```
gumb0, line 64: Quite big for a lambda, can you make the code inside a separate private method?

Also why not use the same function in `parse`?

test/utils/wasm3_engine.cpp
chfast, line 43: Is this `((uint32_t*)stack)++` or `(uint32_t*)(stack++)`?

test/utils/wabt_engine.cpp
axic: Parse doesn't need it, it only wants the signature.

I had a different version using a `uniqu_ptr` but something wasnt working. The problem is `GetMemory` will only return correctly after instantiation, so it can't be captured prior. Also wabt has no context argument, so we'll need a lambda in any case, but could extarct out into a helper.

I would prefer however not tod any of that here, but someone else could optimize these in subsequent PRs.

test/utils/adler32.hpp
```diff
@@ -0,0 +1,28 @@
+// Fizzy: A fast WebAssembly interpreter
+// Copyright 2019-2020 The Fizzy Authors.
```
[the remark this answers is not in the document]
axic: Done.

test/utils/wasm3_engine.cpp
axic: Done.

test/utils/wasm3_engine.cpp
axic: I believe the intention is `*(uint32_t*)(stack++)`. It is a copy paste from wasm3.

test/utils/fizzy_engine.cpp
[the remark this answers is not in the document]
axic: Fixed.

test/utils/fizzy_engine.cpp
```diff
@@ -24,6 +25,17 @@ class FizzyEngine : public WasmEngine
     Result execute(FuncRef func_ref, const std::vector<uint64_t>& args) final;
 };
 
+namespace
+{
+fizzy::execution_result env_adler32(fizzy::Instance& instance, std::vector<uint64_t> args, int)
+{
+    assert(instance.memory != nullptr);
+    // Since memory is a bytes_view, substr produces another bytes_view and no copy is performed.
```
gumb0, line 33: comment is kinda outdated now :)

test/utils/wabt_engine.cpp
```diff
@@ -44,6 +57,26 @@ bool WabtEngine::parse(bytes_view input) const
...
+            const wabt::interp::TypedValues& args, wabt::interp::TypedValues& results) {
+            const auto offset = args[0].value.i32;
+            const auto length = args[1].value.i32;
+            // TODO: move the memory lookup outside in order to reduce overhead
+            assert(m_env.GetMemoryCount() > 0);
+            auto memory = m_env.GetMemory(0);
+            assert(memory != nullptr);
+            auto memory_data = memory->data;
```
gumb0, line 72: what's the type of `memory_data`?

gumb0: it's `vector<char>`

axic: Some wabt memory type. Apparently `std::vector<char> data`.

test/utils/fizzy_engine.cpp
axic: Will remove before merging.

test/unittests/wasm_engine_test.cpp
chfast: I'm not sure this will help in the case were we have only single host function.

test/utils/fizzy_engine.cpp
chfast: :+1:

test/unittests/wasm_engine_test.cpp
axic: For this PR it would not make any difference, however if the interface would be flexible, we could add more functions in fizzy-bench or have external tools use it, such as ewasm-bench, with a multitude of functions.

PR #323, opened 2020-05-20
Simplify find_exported_function(Module, name) helper

axic:


axic:
Could be also removed :)

I'd defer API changes after we've merged the benchmark for host functions and the optimisation PRs. Should do a large overhaul then.

PR #324, opened 2020-05-20
refactor: Group control instructions together in expression parser loop

gumb0:
Preparation for #312. 

Just reorders cases in instruction parser loop to have control instructions together and ordered by binary encoding.


gumb0:
Rebased.

PR #325, opened 2020-05-20
wasm3: add some assertions

axic:


axic:
The reason here is we know if an update breaks our assumptions. The wasm3 documentation is unclear, hence the asserts.

PR #327, opened 2020-05-21
Add wamr to benchmarking

axic:


chfast:
I'm against adding any more wasm runtimes to fizzy. Babysitting poorly developed other's projects build scripts is too much work will little benefit for every day development.

axic:
This won't be finished anytime soon.

PR #328, opened 2020-05-22
Compute functions' max stack height

chfast:
This aggregates the information about stack height in each frame in a function to compute the max function's stack height. This value will be used to preallocate storage for operand stack for execution.


axic:
Merged as part of #247.

PR #329, opened 2020-05-22
Add fizzy-wasi tool

axic:
TODO:
- [x] fix ProjectUVWASI because it doesn't properly link in libuv (on macos at least)
- [x] add FIZZY_WASI cmake option
- [x] implement `fd_write` (add memory translation helpers)
- [x] maybe restructure to make error handling nicer
- [x] change from `wasi_unstable` to `wasi_snapshot_preview1`
- [x] add CI task running `helloworld.wasm` and expect "hello world\n" on cout

Optional: support helloworld from rust
- [x] wasi_snapshot_preview1::fd_prestat_dir_name (dummy)
- [x] wasi_snapshot_preview1::fd_prestat_get
- [x] wasi_snapshot_preview1::environ_sizes_get
- [x] wasi_snapshot_preview1::environ_get (dummy)

Use hello world from https://github.com/bytecodealliance/wasmtime/blob/master/docs/WASI-tutorial.md#web-assembly-text-example as test.

Also supports running hello world generated by Rust (using wasilibc).

Spec explanation: https://github.com/WebAssembly/WASI/blob/master/phases/README.md


test/wasi/wasi.cpp
```diff
@@ -18,9 +18,29 @@ namespace
...
+                  << std::endl;
+        return false;
+    }
+
+    // TODO: load wasm file
     (void)cmd;
-    (void)argc;
-    (void)argv;
+
+    uvwasi_destroy(&state);
```
axic, line 100: Maybe we should not care about this as it is a single lifetime tool.

test/wasi/wasi.cpp
```diff
@@ -0,0 +1,124 @@
...
+    try
+    {
+        if (argc < 2)
+        {
+            std::cerr << "Missing executable argument" << std::endl;
+            return -1;
+        }
+
+        std::string cmd{argv[1]};
+        const bool res = run(cmd, argc -= 2, argv += 2);
```
axic, line 116: Should just use `-2` and `+2` here.

axic:
@chfast @gumb0 so should we create a `wasi` directory or a `tools/wasi` directory? (Anticipating there may be another runner tool in the future.)

axic:
There seems to be a linking error:
```
/usr/bin/ld: cannot find -luv
```

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,150 @@
...
+    uint32_t iovptr = static_cast<uint32_t>(args[1]);
+    uint32_t iovcnt = static_cast<uint32_t>(args[2]);
+    uint32_t nwritten = static_cast<uint32_t>(args[3]);
+
+    std::vector<uvwasi_ciovec_t> iovs(iovcnt);
+    // TODO: not sure what to pass as end, passing memory size...
+    uvwasi_errno_t ret = uvwasi_serdes_readv_ciovec_t(instance.memory->data(),
+        instance.memory->size(), iovptr, iovs.data(), static_cast<uvwasi_size_t>(iovs.size()));
+    if (ret != UVWASI_ESUCCESS)
+        return {false, {ret}};
```
axic, line 44: Need to double check the spec whether we return error code here or trpa.

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,150 @@
...
+        return false;
+    }
+
+    // TODO: implement nicer error handling
+    auto wasm_binary = load_wasm_file(cmd);
+
+    auto module = fizzy::parse(wasm_binary);
+    auto imports = fizzy::resolve_imported_functions(module,
+        {
+            // TODO: support both wasi_snapshot_preview1 and wasi_unstable
```
axic, line 86: New decision: only support `wasi_snapshot_preview1` for now.

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,150 @@
...
+    uvwasi_errno_t err = uvwasi_init(&state, &options);
+    if (err != UVWASI_ESUCCESS)
+    {
+        std::cerr << "Failed to initialise UVWASI: " << uvwasi_embedder_err_code_to_string(err)
+                  << std::endl;
+        return false;
+    }
+
+    // TODO: implement nicer error handling
+    auto wasm_binary = load_wasm_file(cmd);
```
axic, line 81: Make this const.

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+    auto module = fizzy::parse(wasm_binary);
```
axic, line 83: Make this const?

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,150 @@
...
+    auto imports = fizzy::resolve_imported_functions(module,
+        {
+            // TODO: support both wasi_snapshot_preview1 and wasi_unstable
+            {"wasi_unstable", "proc_exit", {fizzy::ValType::i32}, std::nullopt, wasi_proc_exit},
+            {"wasi_unstable", "fd_write",
+                {fizzy::ValType::i32, fizzy::ValType::i32, fizzy::ValType::i32,
+                    fizzy::ValType::i32},
+                fizzy::ValType::i32, wasi_fd_write},
+        });
+    auto instance = fizzy::instantiate(module, imports);
```
axic, line 93: Probably should move module and imports here.

README.md
```diff
@@ -40,6 +40,20 @@ $ cmake --build .
 This will build Fizzy as a library and since there is no public API
 (the so called *embedder API* in WebAssembly) yet, this is not very useful.
 
+## WASI
+
+Building with the `FIZZY_WASI` option will output a `fizzy-wasi` binary implementing
+the [WASI] API (`wasi_snapshot_preview1` to be precise). It uses [uvwasi] under the hood.
```
axic, line 46: Should mention this is in incomplete.

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,150 @@
...
+#include "execute.hpp"
+#include "parser.hpp"
+#include <uvwasi.h>
+#include <cassert>
+#include <filesystem>
+#include <fstream>
+#include <iostream>
+#include <sstream>
+#include <type_traits>
+#include <unordered_map>
```
gumb0, line 14: not used?

axic: Removed.

cmake/ProjectUVWASI.cmake
```diff
@@ -0,0 +1,54 @@
...
+    CMAKE_ARGS
+    -DCMAKE_INSTALL_PREFIX=<INSTALL_DIR>
+    -DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}
+    -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
+    -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
+    -DUVWASI_DEBUG_LOG=OFF
+    -DCODE_COVERAGE=OFF
+    -DASAN=OFF
+    -DUVWASI_BUILD_TESTS=OFF
+//    INSTALL_COMMAND ""
```
axic, line 37: Should remove.

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,146 @@
...
+namespace fs = std::filesystem;
+
+namespace
+{
+// Global state.
+// NOTE: we do not care about uvwasi_destroy(), because it only clears file descriptors (currently)
+// and we are a single-run tool. This may change in the future and should reevaluate.
+uvwasi_t state;
+
+fizzy::execution_result wasi_proc_exit(fizzy::Instance&, std::vector<uint64_t> args, int)
```
gumb0, line 22: can do this and remove `return` statement
```suggestion
[[noreturn]] fizzy::execution_result wasi_proc_exit(fizzy::Instance&, std::vector<uint64_t> args, int)
```

gumb0: Ah no, doesn't work because of `-Winvalid-noreturn`

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,185 @@
...
+{
+    uint32_t environc = static_cast<uint32_t>(args[0]);
+    // TODO: implement properly (only returns 0 now)
+    uvwasi_serdes_write_uint32_t(instance.memory->data(), environc, 0);
+    return {false, {UVWASI_ESUCCESS}};
+}
+
+fizzy::bytes load_wasm_file(const fs::path& path)
+{
+    std::ifstream wasm_file{path};
```
axic, line 81: What is the best way to check if the file is not found? Should use something from `std::filesystem`?

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,185 @@
...
+    return fizzy::bytes(
+        std::istreambuf_iterator<char>{wasm_file}, std::istreambuf_iterator<char>{});
+}
+
+bool run(const fs::path& cmd, int argc, char** argv)
```
axic, line 86: If we are not using `std::filesystem` for anything, could as well just use `string` here and above.

tools/wasi/wasi.cpp
gumb0:
```
  if (!wasm_file)
    // open failed
```

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,185 @@
...
+                  << std::endl;
+        return false;
+    }
+
+    // TODO: implement nicer error handling
+    const auto wasm_binary = load_wasm_file(cmd);
+
+    auto module = fizzy::parse(wasm_binary);
+    auto imports = fizzy::resolve_imported_functions(module,
+        {
```
gumb0, line 110: Move it to a constant at namespace level, perhaps it would be better formatted

README.md
```diff
@@ -40,6 +40,20 @@ $ cmake --build .
...
+Building with the `FIZZY_WASI` option will output a `fizzy-wasi` binary implementing
+the [WASI] API (a very limited subset of `wasi_snapshot_preview1`, to be precise).
+It uses [uvwasi] under the hood. It can be used to execute WASI-compatible binaries on the command line.
+
+```sh
+$ mkdir build && cd build
+$ cmake -DFIZZY_WASI=ON ..
+$ cmake --build .
+```
+
```
gumb0, line 54: Perhaps we could add an example of `fizzy-wasi` usage like
```
$ bin/fizzy-wasi ../tools/wasi/smoketest/helloworld.wasm 
Hello, world!
```

axic: Done.

tools/wasi/wasi.cpp
axic: This works when trying to open a directory for example, but not with missing files.

circle.yml
```diff
@@ -92,6 +92,10 @@ commands:
         environment:
           LLVM_PROFILE_FILE: spectests-failures.profraw
         command: bin/fizzy-spectests ~/project/test/spectests/smoketest-failures || true
+    - run:
+        name: "Run smoketest with fizzy-wasi"
+        working_directory: ~/build
+        command: bin/fizzy-wasi ~/project/tools/wasi/smoketest/helloworld.wasm
```
axic, line 98: Will wait for #364 and move these to the new directory.

tools/wasi/wasi.cpp
axic: Done. Not much better formatted though, but will be nicer to keep the run code more readable.

tools/wasi/wasi.cpp
axic: @gumb0 @chfast any suggestions?

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,232 @@
...
+    {"wasi_unstable", "fd_prestat_dir_name",
+        {fizzy::ValType::i32, fizzy::ValType::i32, fizzy::ValType::i32}, fizzy::ValType::i32,
+        wasi_return_enosys},
+    {"wasi_unstable", "environ_sizes_get", {fizzy::ValType::i32, fizzy::ValType::i32},
+        fizzy::ValType::i32, wasi_environ_sizes_get},
+    {"wasi_unstable", "environ_get", {fizzy::ValType::i32, fizzy::ValType::i32},
+        fizzy::ValType::i32, wasi_return_enosys},
+};
+
+bool run(const fs::path& cmd, int argc, char** argv)
```
axic, line 145: Perhaps rename `cmd` to `wasm_path` ?

tools/wasi/wasi.cpp
gumb0: Works for me (returns `false`) for non-existing file

README.md
```diff
@@ -40,6 +40,26 @@ $ cmake --build .
...
+Try it with a Hello World example:
+```sh
+$ bin/fizzy-wasi ../tools/wasi/smoketest/helloworld.wasm
```
axic, line 57: Need to update to new path.

circle.yml
axic: This can be replaced/removed after #364.

tools/wasi/wasi.cpp
gumb0: You could leave just `int argc, char** argv` and use `argv[0]` instead of `cmd`

axic: The reason I did it this way so that I can just pass `argc/argv` to wasi as is.

gumb0: I mean you would do it anyway. Currently `cmd` and `argv[0]` in this function are the same string.

tools/wasi/wasi.cpp
axic: > This works when trying to open a directory for example, but not with missing files.

Sorry, I meant the reverse:

```
$ bin/fizzy-wasi a
Failed to open file: a
$ bin/fizzy-wasi _deps
Exception: invalid wasm module prefix
```

axic:
@chfast @gumb0 can I ask for a review?

cmake/ProjectUVWASI.cmake
```diff
@@ -0,0 +1,53 @@
...
+set(ProjectUVWASIIncluded TRUE)
+
+include(ExternalProject)
+
+set(prefix ${CMAKE_BINARY_DIR}/_deps)
+set(source_dir ${prefix}/src/uvwasi)
+set(binary_dir ${prefix}/src/uvwasi-build)
+set(include_dir ${source_dir}/include)
+set(uvwasi_library ${binary_dir}/${CMAKE_STATIC_LIBRARY_PREFIX}uvwasi_a${CMAKE_STATIC_LIBRARY_SUFFIX})
+set(uv_library ${binary_dir}/_deps/libuv-build/${CMAKE_STATIC_LIBRARY_PREFIX}uv_a${CMAKE_STATIC_LIBRARY_SUFFIX})
```
chfast, line 13: How is this library built?

axic:
> Can you at least clean up commits?

I keep doing that, but then changes are requested, I do it in a new commit for review, but no review happens. I'll clean it up in the end.

cmake/ProjectUVWASI.cmake
axic: You added this line :wink:

It is needed because uvwasi builds libuv but doesn't link it into the library (could upstream it). We don't want to build another libuv ourselves due to potential version incompatibility.

chfast: Both are static libraries, so you cannot link uv.a to uvwasi.a.

Answering my question: uvwasi downloads source code of uv using git and git tag, then builds it as a part of uvwasi CMake project.

axic: > Both are static libraries, so you cannot link uv.a to uvwasi.a.

If you only consider "linking" as creating a shared library, yes. However `.a` is just an archive so one can certainly "link" aka merge two archives.

tools/wasi/wasi.cpp
gumb0: To distinguish real files from the rest it's best to use `std::filesystem`.

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,233 @@
...
+    if (!start_function.has_value())
+    {
+        std::cerr << "File is not WASI compatible (_start not found)" << std::endl;
+        return false;
+    }
+
+    // Manually validate type signature here
+    // TODO: do this in find_exported_function
+    const auto func_type = instance->module.get_function_type(*start_function);
+    if (func_type.inputs.size() != 0 || func_type.outputs.size() != 0)
```
gumb0, line 189: could be also
```suggestion
    if (func_type != FuncType{})
```

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,233 @@
...
+    {
+        std::cerr << "File is not WASI compatible (_start has invalid signature)" << std::endl;
+        return false;
+    }
+
+    if (instance->memory == nullptr)
+    {
+        std::cerr << "File is not WASI compatible (no memory exported)" << std::endl;
```
gumb0, line 197: Does it require to export the memory? Then you probably need to call `find_exported_memory` to really check if it's exported.

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,233 @@
...
+        return false;
+    }
+
+    const auto [trapped, result_stack] = fizzy::execute(*instance, *start_function, {});
+    // TODO: remove assert and check for signature above
```
gumb0, line 202: Looks to be done

tools/wasi/wasi.cpp
axic: Good catch, since we added the comparison operator :)

tools/wasi/wasi.cpp
axic: Good point, this may not be the best check, but `find_export_memory` requires an export name. All we can there is an exported memory.

tools/wasi/wasi.cpp
axic: You can try opening an issue on `uvwasi`.

README.md
```diff
@@ -42,6 +42,26 @@ $ cmake --build .
 This will build Fizzy as a library and since there is no public API
 (the so called *embedder API* in WebAssembly) yet, this is not very useful.
 
+## WASI
+
+Building with the `FIZZY_WASI` option will output a `fizzy-wasi` binary implementing
+the [WASI] API (a very limited subset of `wasi_snapshot_preview1`, to be precise).
```
gumb0, line 48: It's not straigtforward to find this `wasi_snapshot_preview1` in WASI repo, I think this is the right link https://github.com/WebAssembly/WASI/blob/master/phases/snapshot/witx/wasi_snapshot_preview1.witx perhaps we should include it.

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,235 @@
...
+        wasi_return_enosys},
+    {"wasi_unstable", "environ_sizes_get", {fizzy::ValType::i32, fizzy::ValType::i32},
+        fizzy::ValType::i32, wasi_environ_sizes_get},
+    {"wasi_unstable", "environ_get", {fizzy::ValType::i32, fizzy::ValType::i32},
+        fizzy::ValType::i32, wasi_return_enosys},
+};
+
+bool run(int argc, char** argv)
+{
+    uvwasi_options_t options = {
```
gumb0, line 146:
```suggestion
    const uvwasi_options_t options = {
```

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,235 @@
...
+        3,           // sizeof fd_table
+        0, nullptr,  // NOTE: no remappings
+        static_cast<uvwasi_size_t>(argc), const_cast<const char**>(argv),
```
gumb0, line 149: does it work without `const_cast`?
```suggestion
        static_cast<uvwasi_size_t>(argc), argv,
```

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,235 @@
...
+        nullptr,  // TODO: support env
+        0, 1, 2,
+        nullptr,  // NOTE: no special allocator
+    };
+
+    uvwasi_errno_t err = uvwasi_init(&state, &options);
```
gumb0, line 155:
```suggestion
    const uvwasi_errno_t err = uvwasi_init(&state, &options);
```

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,235 @@
...
+    if (err != UVWASI_ESUCCESS)
+    {
+        std::cerr << "Failed to initialise UVWASI: " << uvwasi_embedder_err_code_to_string(err)
+                  << std::endl;
```
gumb0, line 159:
```suggestion
                  << "\n";
```

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,235 @@
...
+        return false;
+    }
+
+    if (!std::filesystem::is_regular_file(argv[0]))
+    {
+        std::cerr << "File does not exists or is irregular: " << argv[0] << std::endl;
```
gumb0, line 165:
```suggestion
        std::cerr << "File does not exists or is irregular: " << argv[0] << "\n";
```

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,235 @@
...
+        return false;
+    }
+
+    std::ifstream wasm_file{argv[0]};
+    if (!wasm_file)
+    {
+        std::cerr << "Failed to open file: " << argv[0] << std::endl;
```
gumb0, line 172:
```suggestion
        std::cerr << "Failed to open file: " << argv[0] << "\n";
```

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,235 @@
...
+    // TODO: do this in find_exported_function
+    if (instance->module.get_function_type(*start_function) != fizzy::FuncType{})
+    {
+        std::cerr << "File is not WASI compatible (_start has invalid signature)" << std::endl;
+        return false;
+    }
+
+    if (!fizzy::find_exported_memory(*instance, "memory").has_value())
+    {
+        std::cerr << "File is not WASI compatible (no memory exported)" << std::endl;
```
gumb0, line 200:
```suggestion
        std::cerr << "File is not WASI compatible (no memory exported)\n";
```

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,235 @@
...
+        return false;
+    }
+
+    const auto [trapped, result_stack] = fizzy::execute(*instance, *start_function, {});
+    assert(result_stack.empty());
```
gumb0, line 205: I think we return stack of trapped function? Then this will fail, should assert only if not trapped.

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,235 @@
...
+    if (trapped)
+    {
+        std::cerr << "Execution aborted with WebAssembly trap" << std::endl;
```
gumb0, line 208:
```suggestion
        std::cerr << "Execution aborted with WebAssembly trap\n";
```

README.md
gumb0: Or this one is more readable https://github.com/WebAssembly/WASI/blob/master/phases/snapshot/docs.md

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,235 @@
...
+fizzy::execution_result wasi_environ_sizes_get(
+    fizzy::Instance& instance, std::vector<uint64_t> args, int)
+{
+    uint32_t environc = static_cast<uint32_t>(args[0]);
+    // TODO: implement properly (only returns 0 now)
+    uvwasi_serdes_write_uint32_t(instance.memory->data(), environc, 0);
+    return {false, {UVWASI_ESUCCESS}};
+}
+
+// TODO: support both wasi_snapshot_preview1 and wasi_unstable
```
gumb0, line 99: Where is `wasi_unstable` defined?

README.md
gumb0, line 48: Should `wasi_unstable` be mentioned here?

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,235 @@
...
+const std::vector<fizzy::ImportedFunction> wasi_functions = {
+    {"wasi_snapshot_preview1", "proc_exit", {fizzy::ValType::i32}, std::nullopt, wasi_proc_exit},
+    // clock_time_get
```
gumb0, line 102: Is this TODO or what does it mean?

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,235 @@
...
+    // Should not reach this.
+    return {false, {}};
+}
+
+fizzy::execution_result wasi_fd_write(fizzy::Instance& instance, std::vector<uint64_t> args, int)
+{
+    uint32_t fd = static_cast<uint32_t>(args[0]);
+    uint32_t iovptr = static_cast<uint32_t>(args[1]);
+    uint32_t iovcnt = static_cast<uint32_t>(args[2]);
+    uint32_t nwritten = static_cast<uint32_t>(args[3]);
```
gumb0, line 37:
```suggestion
    const auto fd = static_cast<uint32_t>(args[0]);
    const auto iovptr = static_cast<uint32_t>(args[1]);
    const auto iovcnt = static_cast<uint32_t>(args[2]);
    const auto nwritten = static_cast<uint32_t>(args[3]);
```

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+    std::vector<uvwasi_ciovec_t> iovs(iovcnt);
+    // TODO: not sure what to pass as end, passing memory size...
+    uvwasi_errno_t ret = uvwasi_serdes_readv_ciovec_t(instance.memory->data(),
+        instance.memory->size(), iovptr, iovs.data(), static_cast<uvwasi_size_t>(iovs.size()));
```
gumb0, line 42: Maybe just using `iovcnt` without cast is simpler
```suggestion
        instance.memory->size(), iovptr, iovs.data(), iovcnt);
```

tools/wasi/wasi.cpp
gumb0, line 37: Perhaps this is `nwrittenptr` and `nwritten_` is just `nwritten`
```suggestion
    uint32_t nwrittenptr = static_cast<uint32_t>(args[3]);
```

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,235 @@
...
+    uvwasi_serdes_write_prestat_t(instance.memory->data(), prestat_ptr, &buf);
+
+    return {false, {ret}};
+}
+
+fizzy::execution_result wasi_environ_sizes_get(
+    fizzy::Instance& instance, std::vector<uint64_t> args, int)
+{
+    uint32_t environc = static_cast<uint32_t>(args[0]);
+    // TODO: implement properly (only returns 0 now)
```
gumb0, line 94: This function should be returning two size values, if I'm looking at correct spec
https://github.com/WebAssembly/WASI/blob/4fa4f4a82844f1dd5e45cbcb7c714e602f6962de/phases/snapshot/witx/wasi_snapshot_preview1.witx#L39

README.md
axic: Added the witx because it links to the readme too.

tools/wasi/wasi.cpp
axic: In `master`, but I want to clean up to only support preview1. Some examples I used were compiled for `unstable` but no point trying to support that.

README.md
axic: See above, will remove it.

tools/wasi/wasi.cpp
axic: Changed this, based on https://github.com/WebAssembly/WASI/blob/master/phases/snapshot/witx/wasi_snapshot_preview1.witx

tools/wasi/wasi.cpp
axic: `uvwasi_init` is not const: `uvwasi_errno_t uvwasi_init(uvwasi_t* uvwasi, uvwasi_options_t* options);`

tools/wasi/wasi.cpp
axic: For some reason this doesn't work for me unless I change the function signature of `run`.

tools/wasi/wasi.cpp
axic: I think in these cases the flushing `std::endl` does is beneficial/ok. Any other reason not to use it?

tools/wasi/wasi.cpp
axic: Hmm, good point. Will just move the assert.

tools/wasi/wasi.cpp
axic:
```
/Users/alex/Projects/fizzy/tools/wasi/wasi.cpp:42:60: error: implicit conversion loses integer precision: 'std::__1::vector<uvwasi_ciovec_s, std::__1::allocator<uvwasi_ciovec_s>
      >::size_type' (aka 'unsigned long') to 'uvwasi_size_t' (aka 'unsigned int') [-Werror,-Wshorten-64-to-32]
        instance.memory->size(), iovptr, iovs.data(), iovs.size());
                                                      ~~~~~^~~~~~
```

tools/wasi/wasi.cpp
axic: I copied these from uvwasi, but perhaps we should consider following the naming of `wasi.witx`. That does not use `ptr` anywhere, we could decide to add `_ptr` to pointers.

What do you think?

test/smoketests/CMakeLists.txt
```diff
@@ -4,3 +4,4 @@
 
 add_subdirectory(benchmarks)
 add_subdirectory(spectests)
+add_subdirectory(wasi)
```
chfast, line 7: This should be conditionally added?

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,235 @@
...
+#include <filesystem>
+#include <fstream>
+#include <iostream>
+
+namespace
+{
+// Global state.
+// NOTE: we do not care about uvwasi_destroy(), because it only clears file descriptors (currently)
+// and we are a single-run tool. This may change in the future and should reevaluate.
+uvwasi_t state;
```
chfast, line 18: You can wrap it with `std::unique_ptr`.

test/smoketests/CMakeLists.txt
axic: For the tests? Yeah I forgot that one.

tools/wasi/wasi.cpp
axic: How would that help?

chfast: It will call `uvwasi_destroy()` at the program exit.

axic: How? Is there a way to override the destructor of `unique_ptr`? Is the "second arg" that?

axic: Yeah, don't see too much benefit, or at least I couldn't make it nice.

gumb0: This looks to work
```
std::unique_ptr<uvwasi_t, void(*)(uvwasi_t*)> state{new uvwasi_t, uvwasi_destroy};
```

axic: I used `decltype(&uvwasi_destroy)` for the signature, but disliked using the `new` keyword and the indirection this adds without any benefit.

tools/wasi/wasi.cpp
gumb0: I think adding `_ptr` would be good.

tools/wasi/wasi.cpp
gumb0: Yes, but `iovcnt` has the right type and equals to this size.

tools/wasi/wasi.cpp
gumb0: Only for consistency. stderr is unbuffered anyway.

tools/wasi/wasi.cpp
axic: Removed.

test/smoketests/CMakeLists.txt
axic: Done.

tools/wasi/wasi.cpp
axic: You are correct, fixed.

tools/wasi/wasi.cpp
axic: @gumb0 any comment?

gumb0: Yeah, looks like there's no way around cast here

https://isocpp.org/wiki/faq/const-correctness#constptrptr-conversion
http://c-faq.com/ansi/constmismatch.html

chfast: Did you try declaring `main()` as `int main(int c, const char**v)`?

tools/wasi/wasi.cpp
chfast: The benefit is that `uvwasi_destroy` is used and we can detect issues with it.

There is a memory leak in @gumb0's suggestions.

tools/wasi/wasi.cpp
axic: Sure, I can do that. In a previous PR @gumb0 mentioned that `main` cannot be declared noexcept and perhaps const could pose such a problem too: https://en.cppreference.com/w/cpp/language/main_function

But changing both run and main to const avoids the cast.

chfast: I'm fine with `char**`.

tools/wasi/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+# Fizzy: A fast WebAssembly interpreter
+# Copyright 2019-2020 The Fizzy Authors.
+# SPDX-License-Identifier: Apache-2.0
+
+include(ProjectUVWASI)
+
+set(fizzy_include_dir ${PROJECT_SOURCE_DIR}/lib/fizzy)
+
+add_executable(fizzy-wasi wasi.cpp)
+target_compile_features(fizzy-wasi PRIVATE cxx_std_17)
```
chfast, line 10: Not needed, as `fizzy::fizzy` has it.

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+    uvwasi_errno_t err = uvwasi_init(&state, &options);
+    if (err != UVWASI_ESUCCESS)
+    {
+        std::cerr << "Failed to initialise UVWASI: " << uvwasi_embedder_err_code_to_string(err)
+                  << "\n";
+        return false;
+    }
+
+    if (!std::filesystem::is_regular_file(argv[0]))
```
chfast, line 141: It maybe better to just relay on failing `wasm_file{argv[0]}` instead of using `<filesystem>` (to avoid building issues with GCC8 and on macOS).

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,213 @@
...
+    const auto wasm_binary =
+        fizzy::bytes(std::istreambuf_iterator<char>{wasm_file}, std::istreambuf_iterator<char>{});
+
+    auto module = fizzy::parse(wasm_binary);
+    auto imports = fizzy::resolve_imported_functions(module, wasi_functions);
+    auto instance = fizzy::instantiate(std::move(module), std::move(imports));
+    assert(instance != nullptr);
+
+    const auto start_function = fizzy::find_exported_function(instance->module, "_start");
+    if (!start_function.has_value())
```
chfast, line 162:
```suggestion
    if (!start_function)
```
This is also fine.

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,213 @@
...
+uvwasi_t state;
+
+fizzy::ExecutionResult wasi_return_enosys(fizzy::Instance&, fizzy::span<const uint64_t>, int)
+{
+    return UVWASI_ENOSYS;
+}
+
+fizzy::ExecutionResult wasi_proc_exit(fizzy::Instance&, fizzy::span<const uint64_t> args, int)
+{
+    uvwasi_proc_exit(&state, static_cast<uvwasi_exitcode_t>(args[0]));
```
chfast, line 27: This can fail, so maybe trap if not successful?

tools/wasi/wasi.cpp
axic: Ah didn't see you proposed `iovcnt`.

tools/wasi/wasi.cpp
axic: Fixed.

tools/wasi/wasi.cpp
axic: It traps already?

tools/wasi/wasi.cpp
axic: Just learned that today, but I found it more confusing in Solidity at least. This one I guess can be changed.

tools/wasi/wasi.cpp
axic: Discussed most of that in https://github.com/wasmx/fizzy/pull/329#discussion_r430463086.

`ifstream` is terrible at error reporting, just returns an empty bytestring in case of errors.

tools/wasi/CMakeLists.txt
axic: This was a copy of the cmake file from spectests, all of the tools have this line too. Should be removed then.

tools/wasi/wasi.cpp
chfast: It returns `uvwasi_errno_t` whatever it means in this context.

tools/wasi/wasi.cpp
chfast: It is not great, but you can check `wasm_file.fail()` and print that this file cannot be opened (without the reason why).

axic:
The new `-Weverything` warnings:
```
31/76] Building CXX object tools/wasi/CMakeFiles/fizzy-wasi.dir/wasi.cpp.o
FAILED: tools/wasi/CMakeFiles/fizzy-wasi.dir/wasi.cpp.o 
/usr/bin/cmake -E __run_co_compile --tidy="clang-tidy;--extra-arg-before=--driver-mode=g++" --source=/home/builder/project/tools/wasi/wasi.cpp -- /usr/bin/clang++  -I/home/builder/project/lib/fizzy -isystem _deps/src/uvwasi/include -isystem _deps/src/uvwasi-build -g -Wpedantic -Werror -Wall -Wextra -Wshadow -Wconversion -Wsign-conversion -Wno-unknown-pragmas -fstack-protector-strong -Wimplicit-fallthrough -Wcast-qual -Wcast-align -Wextra-semi -Wmissing-declarations -Wold-style-cast -Wfinal-dtor-non-final-class -Wnewline-eof -Wunreachable-code-break -Wduplicate-enum -Weverything -Wno-c++98-compat -Wno-c++98-compat-pedantic -Wno-covered-switch-default -Wno-double-promotion -Wno-float-equal -Wno-padded -Wno-switch-enum -std=c++17 -MD -MT tools/wasi/CMakeFiles/fizzy-wasi.dir/wasi.cpp.o -MF tools/wasi/CMakeFiles/fizzy-wasi.dir/wasi.cpp.o.d -o tools/wasi/CMakeFiles/fizzy-wasi.dir/wasi.cpp.o -c /home/builder/project/tools/wasi/wasi.cpp
/home/builder/project/tools/wasi/wasi.cpp:105:44: error: declaration requires an exit-time destructor [clang-diagnostic-exit-time-destructors,-warnings-as-errors]
const std::vector<fizzy::ImportedFunction> wasi_functions = {
                                           ^
/home/builder/project/tools/wasi/wasi.cpp:105:44: error: declaration requires a global destructor [clang-diagnostic-global-constructors,-warnings-as-errors]
1119 warnings generated.
Suppressed 1117 warnings (1117 in non-user code).
Use -header-filter=.* to display errors from all non-system headers. Use -system-headers to display errors from system headers as well.
2 warnings treated as errors
```

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,216 @@
...
+    const auto environc = args[0].as<uint32_t>();
+    const auto environ_buf_size = args[1].as<uint32_t>();
+    // TODO: implement properly (only returns 0 now)
+    uvwasi_serdes_write_uint32_t(instance.memory->data(), environc, 0);
+    uvwasi_serdes_write_uint32_t(instance.memory->data(), environ_buf_size, 0);
+    return fizzy::Value{uint32_t{UVWASI_ESUCCESS}};
+}
+
+// TODO: support both wasi_snapshot_preview1 and wasi_unstable?
+const std::vector<fizzy::ImportedFunction> wasi_functions = {
```
chfast, line 105:
```suggestion
constexpr fizzy::ImportedFunction wasi_functions[] = {
```

chfast (approved):

tools/wasi/wasi.cpp
axic:
```
/Users/alex/Projects/fizzy/tools/wasi/wasi.cpp:105:35: error: constexpr variable cannot have non-literal type 'fizzy::ImportedFunction const[]'
constexpr fizzy::ImportedFunction wasi_functions[] = {
                                  ^
/Users/alex/Projects/fizzy/lib/fizzy/instantiate.hpp:101:17: note: 'ImportedFunction' is not literal because it has data member 'module' of non-literal type 'std::string'
      (aka 'basic_string<char, char_traits<char>, allocator<char> >')
    std::string module;
                ^
```

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,216 @@
...
+{
+    const auto fd = args[0].as<uint32_t>();
+    const auto iov_ptr = args[1].as<uint32_t>();
+    const auto iov_cnt = args[2].as<uint32_t>();
+    const auto nwritten_ptr = args[3].as<uint32_t>();
+
+    std::vector<uvwasi_ciovec_t> iovs(iov_cnt);
+    // TODO: not sure what to pass as end, passing memory size...
+    uvwasi_errno_t ret = uvwasi_serdes_readv_ciovec_t(instance.memory->data(),
+        instance.memory->size(), iov_ptr, iovs.data(), static_cast<uvwasi_size_t>(iovs.size()));
```
gumb0, line 44: I suggested this before already
```suggestion
        instance.memory->size(), iov_ptr, iovs.data(), iov_cnt);
```

tools/wasi/wasi.cpp
gumb0: Can be const

tools/wasi/wasi.cpp
axic: The only easy solution seems to be moving it back into `main`.

tools/wasi/CMakeLists.txt
chfast: Our CMake scripts are not perfect, but I still vote for removing the line.

cmake/ProjectUVWASI.cmake
```diff
@@ -0,0 +1,53 @@
...
+endif()
+
+ExternalProject_Add(uvwasi
+    EXCLUDE_FROM_ALL 1
+    PREFIX ${prefix}
+    DOWNLOAD_NAME uvwasi-0.0.10.tar.gz
+    DOWNLOAD_DIR ${prefix}/downloads
+    SOURCE_DIR ${source_dir}
+    BINARY_DIR ${binary_dir}
+    URL https://github.com/cjihrig/uvwasi/archive/v0.0.10.tar.gz
```
chfast, line 26: The version `0.0.11` has been released 20 days ago.

cmake/ProjectUVWASI.cmake
```diff
@@ -0,0 +1,53 @@
...
+    URL_HASH SHA256=39135f4dd4a44013399ceed7166391ffc5c09655e4bfbf851da2be039e6985df
+    CMAKE_ARGS
+    -DCMAKE_INSTALL_PREFIX=<INSTALL_DIR>
+    -DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}
```
chfast, line 30:
```suggestion
```
The uvwasi is C project, so this is not needed.

cmake/ProjectUVWASI.cmake
```diff
@@ -0,0 +1,53 @@
...
+)
+
+add_library(uvwasi::uvwasi STATIC IMPORTED)
+
+file(MAKE_DIRECTORY ${include_dir})
+
+set_target_properties(
+    uvwasi::uvwasi
+    PROPERTIES
+    IMPORTED_CONFIGURATIONS Release
```
chfast, line 47:
```suggestion
    IMPORTED_CONFIGURATIONS ${CMAKE_BUILD_TYPE}
```
This should rather match the build type. Other way of fixing this is always build Release. This can be change later if someone needs to debug uvwasi.

cmake/ProjectUVWASI.cmake
axic: I know, but it does not compile, so thought delaying it.

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,217 @@
...
+    const auto environ_buf_size = args[1].as<uint32_t>();
+    // TODO: implement properly (only returns 0 now)
+    uvwasi_serdes_write_uint32_t(instance.memory->data(), environc, 0);
+    uvwasi_serdes_write_uint32_t(instance.memory->data(), environ_buf_size, 0);
+    return fizzy::Value{uint32_t{UVWASI_ESUCCESS}};
+}
+
+bool run(int argc, const char** argv)
+{
+    // TODO: support both wasi_snapshot_preview1 and wasi_unstable?
```
chfast, line 106: For me the `wasi_snapshot_preview1` is the only answer, not worth a TODO.

tools/wasi/wasi.cpp
chfast: Add comment then.

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,217 @@
...
+    const std::vector<fizzy::ImportedFunction> wasi_functions = {
```
chfast, line 107:
```suggestion
    const fizzy::ImportedFunction wasi_functions[] = {
```

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,217 @@
...
+    auto module = fizzy::parse(wasm_binary);
+    auto imports = fizzy::resolve_imported_functions(module, wasi_functions);
+    auto instance = fizzy::instantiate(
+        std::move(module), std::move(imports), {}, {}, {}, fizzy::MemoryPagesValidationLimit);
+    assert(instance != nullptr);
+
+    const auto start_function = fizzy::find_exported_function(instance->module, "_start");
+    if (!start_function.has_value())
+    {
+        std::cerr << "File is not WASI compatible (_start not found)\n";
```
chfast, line 168: Not a job for now, but the wasi support should be designed as a library, otherwise it is difficult to test it.

chfast: This case is not tested.

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,217 @@
...
+        return false;
+    }
+
+    // Manually validate type signature here
+    // TODO: do this in find_exported_function
+    if (instance->module.get_function_type(*start_function) != fizzy::FuncType{})
+    {
+        std::cerr << "File is not WASI compatible (_start has invalid signature)\n";
```
chfast, line 176: This is not tested.

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,217 @@
...
+        return false;
+    }
+
+    if (!fizzy::find_exported_memory(*instance, "memory").has_value())
+    {
+        std::cerr << "File is not WASI compatible (no memory exported)\n";
```
chfast, line 182: This is not tested.

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,217 @@
...
+        return false;
+    }
+
+    const auto result = fizzy::execute(*instance, *start_function, {});
+    if (result.trapped)
+    {
+        std::cerr << "Execution aborted with WebAssembly trap\n";
```
chfast, line 189: This is not tested.

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    std::ifstream wasm_file{argv[0]};
+    if (!wasm_file)
+    {
+        std::cerr << "Failed to open file: " << argv[0] << "\n";
+        return false;
+    }
+    const auto wasm_binary =
+        fizzy::bytes(std::istreambuf_iterator<char>{wasm_file}, std::istreambuf_iterator<char>{});
```
chfast, line 157: This is not tested.

tools/wasi/wasi.cpp
```diff
@@ -0,0 +1,217 @@
...
+        static_cast<uvwasi_size_t>(argc), argv,
+        nullptr,  // TODO: support env
+        0, 1, 2,
+        nullptr,  // NOTE: no special allocator
+    };
+
+    uvwasi_errno_t err = uvwasi_init(&state, &options);
+    if (err != UVWASI_ESUCCESS)
+    {
+        std::cerr << "Failed to initialise UVWASI: " << uvwasi_embedder_err_code_to_string(err)
```
chfast, line 139: How are you going to test this?

axic: Checking uvwasi this is triggered in two realistic cases (in version 0.10.0): 1) out of memory; 2) invalid paths supplied in `options->preopen`. We do not make use of that feature yet.

tools/wasi/wasi.cpp
axic:
```
/home/builder/project/tools/wasi/wasi.cpp: In function 'bool {anonymous}::run(int, const char**)':
/home/builder/project/tools/wasi/wasi.cpp:160:62: error: could not convert '(const fizzy::ImportedFunction*)(& wasi_functions)' from 'const fizzy::ImportedFunction*' to 'std::vector<fizzy::ImportedFunction>'
  160 |     auto imports = fizzy::resolve_imported_functions(module, wasi_functions);
      |                                                              ^~~~~~~~~~~~~~
      |                                                              |
      |                                                              const fizzy::ImportedFunction*
```

`resolve` requires a vector currently, I'd leave this do be fixed after this is merged.

cmake/ProjectUVWASI.cmake
axic: This is actually a typo in wabt/wasm3 too.

cmake/ProjectUVWASI.cmake
chfast: What's up with this?

tools/wasi/wasi.cpp
chfast:
```suggestion
uvwasi_t state;
std::unique_ptr<uvwasi_t, decltype(&uvwasi_destroy)> state_destroyer{&state, &uvwasi_destroy};
```

chfast (changes requested):

chfast:
The idea how to test at least some aspects of WASI.
In [`uvwasi_options_t`](https://github.com/cjihrig/uvwasi#uvwasi_options_t) you can set file descriptions to previously created memory buffers (probably requires POSIX/OS-specific API). This allows providing input and checkout output in unittests.

Although, I'm still ok with merging this in the current from, and leave more extensive testing to some other day.

axic:
@chfast your magic does not compile:
```
FAILED: tools/wasi/CMakeFiles/fizzy-wasi.dir/wasi.cpp.o 
/usr/bin/cmake -E __run_co_compile --tidy="clang-tidy;--extra-arg-before=--driver-mode=g++" --source=/home/builder/project/tools/wasi/wasi.cpp -- /usr/bin/clang++  -I/home/builder/project/lib/fizzy -I/home/builder/project/include -isystem _deps/src/uvwasi/include -isystem _deps/src/uvwasi-build -g -Wpedantic -Werror -Wall -Wextra -Wshadow -Wconversion -Wsign-conversion -Wno-unknown-pragmas -fstack-protector-strong -Wimplicit-fallthrough -Wcast-qual -Wcast-align -Wmissing-declarations -Wextra-semi -Wold-style-cast -Wfinal-dtor-non-final-class -Wnewline-eof -Wunreachable-code-break -Wduplicate-enum -Weverything -Wno-c++98-compat -Wno-c++98-compat-pedantic -Wno-covered-switch-default -Wno-double-promotion -Wno-float-equal -Wno-padded -Wno-switch-enum -std=c++17 -MD -MT tools/wasi/CMakeFiles/fizzy-wasi.dir/wasi.cpp.o -MF tools/wasi/CMakeFiles/fizzy-wasi.dir/wasi.cpp.o.d -o tools/wasi/CMakeFiles/fizzy-wasi.dir/wasi.cpp.o -c /home/builder/project/tools/wasi/wasi.cpp
/home/builder/project/tools/wasi/wasi.cpp:18:54: error: declaration requires an exit-time destructor [clang-diagnostic-exit-time-destructors,-warnings-as-errors]
std::unique_ptr<uvwasi_t, decltype(&uvwasi_destroy)> state_destroyer{&state, &uvwasi_destroy};
                                                     ^
/home/builder/project/tools/wasi/wasi.cpp:18:54: error: declaration requires a global destructor [clang-diagnostic-global-constructors,-warnings-as-errors]
1135 warnings generated.
Suppressed 1133 warnings (1133 in non-user code).
Use -header-filter=.* to display errors from all non-system headers. Use -system-headers to display errors from system headers as well.
2 warnings treated as errors
```

tools/wasi/wasi.cpp
axic: Created a PR upstream: https://github.com/cjihrig/uvwasi/pull/155

chfast (approved):

gumb0 (approved):

tools/wasi/wasi.cpp
```diff
@@ -132,7 +132,7 @@ bool run(int argc, const char** argv)
         nullptr,  // NOTE: no special allocator
     };
 
-    uvwasi_errno_t err = uvwasi_init(&state, &options);
+    const uvwasi_errno_t err = uvwasi_init(&state, &options);
```
axic, line 134: Why is this const stuff so important on this?

chfast: The `const` should be the default.

axic: Should protest C++ to follow Rust with making const-by-default and have a mutable specifier on others :)

chfast: We both know this is not possible.

PR #330, opened 2020-05-22
ci: Run sanitizers job on macos / ugrade to XCode 11.5.0

chfast:
This is valuable because this tests usage of `libc++` standard library.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -17,7 +17,7 @@ executors:
       - image: ethereum/cpp-build-env:14-clang-10
   macos:
     macos:
-      xcode: 11.3.0
+      xcode: 11.5.0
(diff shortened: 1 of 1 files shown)


circle.yml
```diff
@@ -20,6 +20,13 @@ executors:
       xcode: 11.3.0
 
 commands:
+  description: "Install macos system dependencies"
```
axic, line 23: Nit: macOS is the official spelling

gumb0:
Why is it not "Required"? Should it be?

axic:
> Why is it not "Required"? Should it be?

You mean on github? I'll add it once merged.

PR #331, opened 2020-05-22
utils: add signature to WasmEngine::find_function

axic:
This is required for wamr support. Also it can help us avoid mistakes if we validate signatures in the engines.

Should also add signature type validation to `find_function`. That helps with both this PR and #329.


axic:
This is required for wamr support. Also it can help us avoid mistakes if we validate signatures in the engines.

test/bench/bench.cpp
```diff
@@ -234,6 +236,17 @@ void load_benchmark(const fs::path& path, const std::string& name_prefix)
...
             case InputsReadingState::FuncName:
                 benchmark_case->func_name = std::move(l);
+                st = InputsReadingState::FuncSignature;
+                break;
+
+            case InputsReadingState::FuncSignature:
+                std::cout << l << std::endl;
+                if (l.find_first_of(":") != l.find_last_of(":"))
+                    throw std::runtime_error{"Multiple occurances of ':' found in signature"};
+                // Only allow i (i32) I (i64) v (void) as types
```
axic, line 246: Need to decide on a design here, the main question is about return types and no arguments cases:
a) Require colon and void type, e.g. `i:v`, `i:i`
b) Allow empty return type, no need for void, e.g. `i:`
c) Omit colon if no return type, e.g. `i`
d) Require void on input side if there no inputs, but there's an output, e.g. `v:i`
e) Allow empty arguments, e.g. `:i`

I'm not sure what is the best, but leaning towards c) and e).

cc @chfast @gumb0

test/bench/bench.cpp
```diff
@@ -234,6 +236,17 @@ void load_benchmark(const fs::path& path, const std::string& name_prefix)
 
             case InputsReadingState::FuncName:
                 benchmark_case->func_name = std::move(l);
+                st = InputsReadingState::FuncSignature;
+                break;
+
+            case InputsReadingState::FuncSignature:
```
axic, line 242: @chfast suggested to conflate the signature into the name line, such as `keccak256_bnech iii:i`

test/bench/bench.cpp
gumb0: I like a) because everything is explicit and easier to parse

axic: In the case of a), what happens when there are no inputs? `:v` and `:i` ?

chfast: 1. If this is based on WAMR I would use what they have.
2. I would not use `v` at all to be consistent with wasm types, where it is `[]`.
3. I would keep `:` always for parsing. So `:` is also valid.

axic: It is not based on WAMR. Mentioned it here before: https://github.com/wasmx/fizzy/pull/318#discussion_r427904518

- wasm3 uses a version of this, but has a reverse notation where return value is first (e.g. `v(ii)`)
- emscripten uses a version of this: it has no separation of input vs output (e.g. `iiv`), but rather assumes the last type is either a type or void
- wamr uses a version of this, but unclear what they do for the return value

axic: @chfast are you saying these rules then:
- `:` no input, no output
- `i:` single i32 input, no output
- `:i` single i32 output, no input
- etc.

chfast: Yes. But if this is something different than anything else, I don't care so much.

axic: We may want to improve/change/use it later on in the external API, then we can come back. All we need right now is just to make the benchmark tool more convenient to use and to support wamr.

gumb0: > In the case of a), what happens when there are no inputs? :v and :i ?

`v:v`, `v:i`
I guess it's d) or a) + d)

test/utils/fizzy_engine.cpp
```diff
@@ -72,9 +73,48 @@ bytes_view FizzyEngine::get_memory() const
     return {m_instance->memory->data(), m_instance->memory->size()};
 }
 
-std::optional<WasmEngine::FuncRef> FizzyEngine::find_function(std::string_view name) const
+namespace
 {
-    return fizzy::find_exported_function(m_instance->module, name);
+bool matching_signature(const fizzy::FuncType& func_type, std::string_view signature)
```
axic, line 78: Probably would be nicer to have simple translator instead (sig -> func_type) and equal operator for `func_type`.

test/bench/bench.cpp
```diff
@@ -234,6 +236,19 @@ void load_benchmark(const fs::path& path, const std::string& name_prefix)
...
+                std::cout << l << std::endl;
+                if (l.find_first_of(":") == std::string::npos)
```
axic, line 244: Should this validation be moved as a helper on `wasm_engine` instead?

test/utils/fizzy_engine.cpp
```diff
@@ -72,9 +73,54 @@ bytes_view FizzyEngine::get_memory() const
...
 }
 
-std::optional<WasmEngine::FuncRef> FizzyEngine::find_function(std::string_view name) const
+namespace
 {
-    return fizzy::find_exported_function(m_instance->module, name);
+FuncType translate_signature(std::string_view signature)
+{
+    std::string_view inputs(signature.begin(), signature.find(":"));
+    std::string_view outputs(inputs.end() + 1);  // Is this a terrible hack?
```
gumb0, line 81:
```suggestion
    const auto delimiter_pos = signature.find(":");
    assert(delimiter_pos != std::string_view::npos);
    std::string_view inputs(signature.begin(), delimiter_pos);
    std::string_view outputs(delimiter_pos + 1, signature.end());
```

test/utils/fizzy_engine.cpp
```diff
@@ -72,9 +73,54 @@ bytes_view FizzyEngine::get_memory() const
...
+
+    FuncType func_type;
+    for (auto input : inputs)
+    {
+        if (input == 'i')
```
gumb0, line 86: don't repeat these `if`s, make a helper `fizzy::ValType charToValType(char)`

test/utils/fizzy_engine.cpp
```diff
@@ -72,9 +73,54 @@ bytes_view FizzyEngine::get_memory() const
...
+namespace
 {
-    return fizzy::find_exported_function(m_instance->module, name);
+FuncType translate_signature(std::string_view signature)
+{
+    std::string_view inputs(signature.begin(), signature.find(":"));
+    std::string_view outputs(inputs.end() + 1);  // Is this a terrible hack?
+
+    FuncType func_type;
+    for (auto input : inputs)
```
gumb0, line 84:
```suggestion
    for (const auto input : inputs)
```

test/utils/fizzy_engine.cpp
```diff
@@ -72,9 +73,54 @@ bytes_view FizzyEngine::get_memory() const
...
+    {
+        if (input == 'i')
+            func_type.inputs.push_back(fizzy::ValType::i32);
+        else if (input == 'I')
+            func_type.inputs.push_back(fizzy::ValType::i64);
+        else
+            throw std::runtime_error{"invalid type"};
+    }
+    for (auto output : outputs)
```
gumb0, line 93:
```suggestion
    for (const auto output : outputs)
```

test/utils/fizzy_engine.cpp
```diff
@@ -72,9 +73,54 @@ bytes_view FizzyEngine::get_memory() const
...
+        else if (output == 'I')
+            func_type.outputs.push_back(fizzy::ValType::i64);
+        else
+            throw std::runtime_error{"invalid type"};
+    }
+
+    return func_type;
+}
+
+inline bool operator!=(const FuncType& lhs, const FuncType& rhs)
```
gumb0, line 106: This should go to `types.hpp` or other common header.
Porbably with `operator==`, too, and unit tests.

test/utils/wasm_engine.hpp
```diff
@@ -55,6 +56,17 @@ class WasmEngine
     /// Executes the function of the given index.
     /// Requires instantiate().
     virtual Result execute(FuncRef func_ref, const std::vector<uint64_t>& args) = 0;
+
+    static void validate_function_signature(std::string_view signature)
```
gumb0, line 60: I would prefer to make it a free function in `fizzy::test`, but not so important

gumb0:
Do you plan to add this validation to all of the engines?

axic:
>Do you plan to add this validation to all of the engines?

Initially I thought so, but I think it is fine to only do it in fizzy (it detects problems in all our test cases) and only do the conversion for a given engine when there's need (such as wamr would depend on it for implementation).

test/utils/fizzy_engine.cpp
axic: Feels like an overkill to be honest.

axic: Did it anyway.

test/utils/fizzy_engine.cpp
axic:
```
/Users/alex/Projects/fizzy/test/utils/fizzy_engine.cpp:111:22: error: no matching constructor for initialization of 'std::string_view' (aka 'basic_string_view<char>')
    std::string_view outputs(delimiter_pos + 1, signature.end());
                     ^       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

test/utils/fizzy_engine.cpp
axic: Done in #360.

test/utils/wasm_engine.hpp
axic: Would that mean to make inline in the header or creating yet another file, `wasm_engine.cpp`, to place it?

test/utils/fizzy_engine.cpp
axic: @gumb0 any suggestions?

axic:
Squashed this down quite a bit.

test/utils/fizzy_engine.cpp
chfast: `&signature[delimiter_pos] + 1`.

chfast: But probably should be easier to do `signature.substr(0, delim)` and `signature.substr(delim+1)`.

axic: Yeah that's nicer, pushing that.

test/utils/fizzy_engine.cpp
```diff
@@ -27,6 +28,31 @@ class FizzyEngine : public WasmEngine
...
+    else
+        throw std::runtime_error{"invalid type"};
+}
+
+FuncType translate_signature(std::string_view signature)
+{
+    const auto delimiter_pos = signature.find(":");
+    assert(delimiter_pos != std::string_view::npos);
+    std::string_view inputs = signature.substr(0, delimiter_pos);
+    std::string_view outputs = signature.substr(delimiter_pos + 1);
```
axic, line 46: Could also make these `auto`?

test/bench/bench.cpp
chfast: That would be nicer for reading input files, but I don't insist.

test/utils/fizzy_engine.cpp
```diff
@@ -27,6 +28,31 @@ class FizzyEngine : public WasmEngine
...
+
+    FuncType func_type;
+    for (const auto input : inputs)
```
chfast, line 49: This can be done as `std::transform(std::begin(input), std::end(input), std::back_inserter(func_type.inputs), translate_valtype)` or something...

test/utils/fizzy_engine.cpp
chfast: Can be even `const auto`.

axic: Doesn't `string_view` imply const?

test/utils/fizzy_engine.cpp
axic: Sure, though I'm not convinced it is easier to read.

axic: Changed.

test/utils/fizzy_engine.cpp
axic: Changed.

chfast: `string_view` references immutable data, but reference can be changed (reassigned or move).

test/utils/wasm_engine.hpp
axic: @chfast any preference?

a) as it is now
b) as a free function in the header
c) as a free function in a new `.cpp` file

chfast: For sanity, it is preferable to move implementation to .cpp file.
For static method / free function I think both options are fine.

test/utils/wasm_engine.hpp
```diff
@@ -37,7 +37,8 @@ class WasmEngine
 
     /// Finds an exported function in the internal instance.
     /// Requires instantiate().
-    virtual std::optional<FuncRef> find_function(std::string_view name) const = 0;
+    virtual std::optional<FuncRef> find_function(
+        std::string_view name, std::string_view signature) const = 0;
```
axic, line 41: Should this be `signature` or `sig`?

chfast: `signature`.

test/utils/wasm_engine.cpp
```diff
@@ -0,0 +1,20 @@
+// Fizzy: A fast WebAssembly interpreter
+// Copyright 2019-2020 The Fizzy Authors.
```
chfast, line 2:
```suggestion
// Copyright 2020 The Fizzy Authors.
```

PR #333, opened 2020-05-24
ci: Store fuzzing corpus as artifact

chfast:


chfast:
> Assuming the goal here is to store the corpus so it can be inspected for reproduction, if needed.

Yes. I wanted to check why it had not found the bug in #334. Current fuzzer is pretty week.
You can also download the corpus from CI and merge with a one you have locally.

PR #334, opened 2020-05-24
Fix function section validation

chfast:
The validation of type indices in function section must be fully
finished before processing any code. This is because there may be a call
in code[i] referencing function[i + n].

### DIFF
--- a/test/unittests/parser_expr_test.cpp
+++ b/test/unittests/parser_expr_test.cpp
@@ -350,3 +350,15 @@ TEST(parser_expr, call_1arg_1result)
     const auto module = parse(wasm);
     ASSERT_EQ(module.codesec.size(), 2);
 }
+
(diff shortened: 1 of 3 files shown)


test/unittests/parser_expr_test.cpp
```diff
@@ -349,4 +349,16 @@ TEST(parser_expr, call_1arg_1result)
 
     const auto module = parse(wasm);
     ASSERT_EQ(module.codesec.size(), 2);
+}
+
+TEST(parser_expr, call_unexisted_typeid)
```
chfast, line 354:
```suggestion
TEST(parser_expr, call_nonexisting_typeidx)
```

axic: Not merging this part?

chfast: oops...

PR #219, opened 2020-02-28
New version of execution_result

axic:
This re-implements the execution_result struct. Now it has three valid
states: value, void and trap. The struct is simplified, close to POD
type, std::vector is not used any more.

In benchmarks for certain cases (such as ecpairing) deleting an `std::vector` took quite a big chunk of time. I thought perhaps the moved stack for the return value could be the culprit, hence did this change. However my local benchmarks are inconclusive.

PR #340, opened 2020-05-26
bench: Fix execution pre-validation

axic:
This adds `fizzy-bench` integration tests and then fixes crashes.

> Maybe we should merge this after 0.2.0 in case it would make comparison harder to 0.1.0?
> Or could release a patch (0.1.1) with this commit.

The changes are fine to land as they don't modify any benchmarking loop.


axic:
> All of this is checked in validate_benchmark_case where a test run is performed. I don't see point of repeating this.

I may have missed it but, I do get a crash on inputs in `execution` which return a failure in the `instantiate` case.

chfast:
> I may have missed it but, I do get a crash on inputs in `execution` which return a failure in the `instantiate` case.

Please investigate where exactly the problem is and why `validate_benchmark_case` is not able to catch it.

axic:
@chfast pushed your fixes here, but feel free to close.

test/integration_tests/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+# Fizzy: A fast WebAssembly interpreter
+# Copyright 2020 The Fizzy Authors.
+# SPDX-License-Identifier: Apache-2.0
+
+add_test(
```
axic, line 5: Hmm, couldn't we run spectest smoketests the same way?

chfast: Sure. I have not connected the dots..

test/utils/fizzy_engine.cpp
```diff
@@ -100,7 +100,6 @@ WasmEngine::Result FizzyEngine::execute(
 {
     const auto [trapped, result_stack] =
         fizzy::execute(*m_instance, static_cast<uint32_t>(func_ref), args);
-    assert(result_stack.size() <= 1);
```
axic, line 103: This just shows fizzy's "public api" is broken right now :)

Hopefully after the span things are going in, we can merge a version of #219 can go in and fix this.

axic:
Looks good to me, but a bit worried if this goes in first before #331, then #331 needs to be adjusted further to not cause a drop in coverage.

test/utils/fizzy_engine.cpp
chfast: I was confused at first, but truly we dump whole stack also in the case when returning `trapped`.
I can add disabled test for such case, but I don't think this really helps.

chfast:
> Looks good to me, but a bit worried if this goes in first before #331, then #331 needs to be adjusted further to not cause a drop in coverage.

No problem, this can go after #331.

test/bench/bench.cpp
```diff
@@ -88,72 +88,92 @@ struct ExecutionBenchmarkCase
...
 
-    validate_benchmark_case(state, *engine, benchmark_case);
+    const auto engine = create_fn();
+    const bool ok = validate_benchmark_case(state, *engine, benchmark_case);
 
-    const auto has_memory = !benchmark_case.memory.empty();
-    engine->instantiate(*benchmark_case.wasm_binary);
-    const auto func_ref = engine->find_function(benchmark_case.func_name, benchmark_case.func_sig);
+    std::optional<fizzy::test::WasmEngine::FuncRef> func_ref;
+    if (ok)
```
gumb0, line 172: Can you explain the changes here in this function?
It looks like when validation is not ok, you would still go and dereference `func_ref` at line 185

gumb0: Ah I guess it works because `validate_benchmark_case` would calll `state.SkipWithError` and the loop below  will not be executed.

But why then not just early return
```
  if (!validate_benchmark_case(...))
    return;
```

chfast: Yes. The loop will not run, but we need to call `auto _ : state` anyway (this is a libbenchmark limitation in the version we use).

axic:
Looks good to me, @gumb0 okay to merge?

test/bench/bench.cpp
gumb0: Perhaps it's woth to add a comment that we need to call it even when validation failed.

PR #343, opened 2020-05-26
validation: use correct function arity in stack height validation

gumb0:
Required for #312 

Resolves one TODO in `parse_expr` and 45 validation spec tests (must be the tests checking if function stack has enough results at the end of execution)

### DIFF
--- a/lib/fizzy/parser.hpp
+++ b/lib/fizzy/parser.hpp
@@ -40,8 +40,10 @@ inline parser_result<uint8_t> parse_byte(const uint8_t* pos, const uint8_t* end)
 ///
 /// @param input    The beginning of the expr binary input.
 /// @param end      The end of the binary input.
+/// @param func_idx Index of the function being parsed.
 /// @param module   Module that this code is part of.
-parser_result<Code> parse_expr(const uint8_t* input, const uint8_t* end, const Module& module);
+parser_result<Code> parse_expr(
+    const uint8_t* input, const uint8_t* end, FuncIdx func_idx, const Module& module);
 
 parser_result<std::string> parse_string(const uint8_t* pos, const uint8_t* end);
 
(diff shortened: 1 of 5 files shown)


lib/fizzy/parser.hpp
```diff
@@ -40,8 +40,10 @@ inline parser_result<uint8_t> parse_byte(const uint8_t* pos, const uint8_t* end)
 ///
 /// @param input    The beginning of the expr binary input.
 /// @param end      The end of the binary input.
+/// @param type_idx Index of the type of the function being parsed.
```
chfast, line 43:
```suggestion
/// @param type_idx Valid index of the type of the function being parsed.
```

Why not just pass `func_idx` instead?

gumb0: I guess it was handy when the loop in calling `parse_expr` was also validating `type_idx`
It doesn't anymore, and it seems `func_idx` might be more future-proof, so I'll change it.

lib/fizzy/parser_expr.cpp
```diff
@@ -122,9 +123,11 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end, const Mod
     // For a block/if/else instruction the value is the block/if/else's immediate offset.
     std::stack<ControlFrame> control_stack;
 
-    // TODO: Get function result arity from the function type in Module.
-    const uint8_t func_result_arity = 0;
-    control_stack.emplace(Instr::block, func_result_arity, 0);  // The function's implicit block.
+    const auto func_type_idx = module.funcsec[func_idx];
```
axic, line 126: Why not pass in `func_type_idx`?

gumb0: It was like that originally, I changed it just because `func_idx` seems to be a more natural parameter to "parse instuctions of a function", might be more useful in the future for something.

PR #346, opened 2020-05-27
test: Add a unit test reusing args as locals

chfast:


axic:
Should we also add a bench for this?

chfast:
> Should we also add a bench for this?

No, but I wander if LLVM backend is utilizing this...

PR #350, opened 2020-05-27
Raise validation error on element provided but missing table section

axic:
This was previously a parser error.


axic:
Pre: PASSED 4852, FAILED 580, SKIPPED 6381.
Post: PASSED 4855, FAILED 577, SKIPPED 6381.

Will update it once #317 is merged.

PR #351, opened 2020-05-27
Raise validation error if referring to unknown global or setting immutable global

axic:


lib/fizzy/parser_expr.cpp
```diff
@@ -505,6 +505,19 @@ parser_result<Code> parse_expr(
         {
             uint32_t imm;
             std::tie(imm, pos) = leb128u_decode<uint32_t>(pos, end);
+
+            if (instr == Instr::global_get || instr == Instr::global_set)
+            {
+                if (module.globalsec.size() <= imm)
```
gumb0, line 511:
```suggestion
                if (module.get_global_count() <= imm)
```

lib/fizzy/types.hpp
```diff
@@ -51,6 +51,9 @@ using TableIdx = uint32_t;
 // https://webassembly.github.io/spec/core/syntax/modules.html#syntax-memidx
```
gumb0, line 51: BTW these links now go to 1.1 spec

gumb0:
Init expressions of globals need to be validated, too, but I would do that in another PR, as it seems a bit more complicated.

lib/fizzy/types.hpp
axic: Do you want to open an issue/PR for us to direct-link to 1.0?

lib/fizzy/module.hpp
```diff
@@ -74,5 +74,13 @@ struct Module
     {
         return !memorysec.empty() || !imported_memory_types.empty();
     }
+
+    bool is_global_mutable(GlobalIdx idx) const noexcept
```
axic, line 78: Hmm, shouldn't we have a unit test for this?

axic: For the sake of progress I'm merging this, but perhaps we should add unit tests for each module helper in a new PR>

lib/fizzy/types.hpp
gumb0: https://github.com/wasmx/fizzy/issues/366

gumb0:
Globals validation is poorly covered by spec tests, I added some cases to "to upstream" list

PR #353, opened 2020-05-27
Raise validation error on invalid start function index

axic:


test/unittests/parser_test.cpp
```diff
@@ -781,15 +781,15 @@ TEST(parser, start_invalid_index)
                      make_section(3, func_section) + make_section(8, start_section) +
                      make_section(10, code_section);
 
-    EXPECT_THROW_MESSAGE(parse(bin), parser_error, "invalid start function index");
```
gumb0, line 784: these tests should better be moved to `validation_test.cpp` now

axic: Hm, I did that quite a few times. Perhaps we should merge these small changes as is and then do a sweep of moving things across?

axic:
Also fixed stat function type checking -- should that be in a new PR or extend this one?

test/unittests/parser_test.cpp
gumb0: Fine

chfast: Let's coordinate with moving bigger code pieces.

PR #356, opened 2020-05-27
Enable "too many results" stack height validation

chfast:
Requires #355.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -252,8 +252,8 @@ jobs:
       - benchmark:
           min_time: "0.01"
       - spectest:
-          expected_passed: 4903
-          expected_failed: 529
+          expected_passed: 4997
+          expected_failed: 435
           expected_skipped: 6381
 
   sanitizers-macos:
@@ -270,8 +270,8 @@ jobs:
       - benchmark:
           min_time: "0.01"
       - spectest:
-          expected_passed: 4903
-          expected_failed: 529
+          expected_passed: 4997
+          expected_failed: 435
           expected_skipped: 6381
 
   benchmark:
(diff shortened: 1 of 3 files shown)


test/unittests/validation_stack_test.cpp
```diff
@@ -160,7 +180,7 @@ TEST(validation_stack, loop_with_result_stack_underflow)
     EXPECT_THROW_MESSAGE(parse(wasm), validation_error, "stack underflow");
 }
 
-TEST(validation_stack, DISABLED_loop_too_many_results)
+TEST(validation_stack, loop_too_many_results)
```
gumb0, line 183: Another test suggestion, not sure if already exists: unreachable insrtuction after `br` that would push extra result, but should be valid, as it's unreachable.

chfast: This is actually invalid. You cannot have additional items in polymorphic stack.

circle.yml
```diff
@@ -252,8 +252,8 @@ jobs:
       - benchmark:
           min_time: "0.01"
       - spectest:
-          expected_passed: 4903
-          expected_failed: 529
+          expected_passed: 4997
```
gumb0, line 255: wow nice increase

My PRs are getting like 2 new tests passed :confounded:

test/unittests/validation_stack_test.cpp
```diff
@@ -482,6 +502,33 @@ TEST(validation_stack, unreachable_call_indirect)
     parse(wasm);
 }
 
+TEST(validation_stack, unreachable_too_many_results)
+{
+    // TODO: These are actually examples of invalid wasm.
```
gumb0, line 507: Then maybe better to make it disabled tests expecting exception?

circle.yml
chfast: This is part of type validation, so I expect it has a lot of test cases.

PR #357, opened 2020-05-28
Validate exports

gumb0:


lib/fizzy/parser.cpp
```diff
@@ -534,6 +535,39 @@ Module parse(bytes_view input)
...
+            break;
+        case ExternalKind::Global:
+            if (export_.index >= total_global_count)
+                throw validation_error{"invalid index of an exported global"};
+            break;
+        default:
+            assert(false);
+        }
+        if (const auto [_, inserted] = export_names.emplace(export_.name); !inserted)
+            throw validation_error("duplicate export name " + export_.name);
```
chfast, line 569: This is not tested.

gumb0: Ah right, forgot this

gumb0: Added a test.

lib/fizzy/parser.cpp
```diff
@@ -534,6 +535,37 @@ Module parse(bytes_view input)
...
+    for (const auto& export_ : module.exportsec)
+    {
+        switch (export_.kind)
+        {
+        case ExternalKind::Function:
+            if (export_.index >= total_func_count)
+                throw validation_error{"invalid index of an exported function"};
+            break;
+        case ExternalKind::Table:
+            if (export_.index != 0 || !module.has_table())
```
gumb0, line 552: Changed to using `module` class helpers

lib/fizzy/parser.cpp
```diff
@@ -534,6 +535,37 @@ Module parse(bytes_view input)
         throw parser_error("malformed binary: number of function and code entries must match");
 
     const auto total_func_count = module.imported_function_types.size() + module.funcsec.size();
+    const auto total_global_count =
+        module.imported_globals_mutability.size() + module.globalsec.size();
+
+    // Validate exports.
+    std::unordered_set<std::string_view> export_names;
```
axic, line 542: Is this much faster than `vector` or what is the reason?

lib/fizzy/parser.cpp
```diff
@@ -534,6 +535,37 @@ Module parse(bytes_view input)
...
+                throw validation_error{"invalid index of an exported memory"};
+            break;
+        case ExternalKind::Global:
+            if (export_.index >= total_global_count)
+                throw validation_error{"invalid index of an exported global"};
+            break;
+        default:
+            assert(false);
+        }
+        if (const auto [_, inserted] = export_names.emplace(export_.name); !inserted)
```
axic, line 566: I so much dislike this syntax of C++17 🤷

axic: `if (!std::get<1>(export_names.emplace(export_.name))` isn't nicer.

But this entire thing is in a for loop, so why not just keep these as two statements (like the rest of codebase)?

test/unittests/parser_test.cpp
```diff
@@ -697,37 +697,48 @@ TEST(parser, export_section_empty)
...
-    EXPECT_EQ(module.exportsec[1].index, 0x43);
+    EXPECT_EQ(module.exportsec[1].index, 0);
     EXPECT_EQ(module.exportsec[2].name, "bar");
     EXPECT_EQ(module.exportsec[2].kind, ExternalKind::Memory);
-    EXPECT_EQ(module.exportsec[2].index, 0x44);
+    EXPECT_EQ(module.exportsec[2].index, 0);
     EXPECT_EQ(module.exportsec[3].name, "xyz");
     EXPECT_EQ(module.exportsec[3].kind, ExternalKind::Global);
-    EXPECT_EQ(module.exportsec[3].index, 0x45);
+    EXPECT_EQ(module.exportsec[3].index, 0);
```
axic, line 741: Too bad the old test was invalid, but nicer to see what we're testing.

test/unittests/validation_test.cpp
```diff
@@ -231,3 +231,68 @@ TEST(validation, call_indirect_no_table)
     const auto wasm = from_hex("0061736d0100000001050160017f00030201000a0901070020001101000b");
     EXPECT_THROW_MESSAGE(parse(wasm), validation_error, "call_indirect without defined table");
 }
+
+TEST(validation, export_invald_index)
```
axic, line 235: invald -> invalid

test/unittests/validation_test.cpp
```diff
@@ -231,3 +231,68 @@ TEST(validation, call_indirect_no_table)
...
+    const auto wasm_no_mem = from_hex("0061736d01000000070501016d0200");
+    EXPECT_THROW_MESSAGE(
+        parse(wasm_no_mem), validation_error, "invalid index of an exported memory");
+
+    /* wat2wasm --no-check
+    (memory 1)
+    (export "m" (memory 1))
+    */
+    const auto wasm_mem = from_hex("0061736d010000000503010001070501016d0201");
+    EXPECT_THROW_MESSAGE(parse(wasm_mem), validation_error, "invalid index of an exported memory");
```
axic, line 276: Should we have different message whether memory/table section is present or not?

lib/fizzy/parser.cpp
chfast: With set it is straight-forward to check if element exists.

lib/fizzy/parser.cpp
chfast: `export_names.emplace(export_.name).second` will work.

test/unittests/validation_test.cpp
chfast: I think this is fine - 0 is invalid index in some context.

lib/fizzy/parser.cpp
gumb0: changed it to @chfast's suggestion

test/unittests/validation_test.cpp
gumb0: fixed

gumb0:
Cases with exporting table/memory 0 when none is defined are missing from the spectests, added to the "to upstream" list.

PR #361, opened 2020-05-29
clang-tidy: Enable readability checks

chfast:


axic:
@chfast you want to land this?

test/utils/asserts.hpp
```diff
@@ -8,7 +8,7 @@
 #include <gmock/gmock.h>
 #include <iosfwd>
 
-MATCHER(Traps, "")
+MATCHER(Traps, "")  // NOLINT(readability-redundant-string-init)
```
axic, line 11: Btw what is this error?

test/utils/wabt_engine.cpp
```diff
@@ -87,10 +87,7 @@ bool WabtEngine::instantiate(bytes_view wasm_binary)
 
     // Run start function (this will return ok if no start function is present)
     const wabt::interp::ExecResult r = m_executor.RunStartFunction(m_module);
-    if (r.result != wabt::interp::Result::Ok)
-        return false;
-
-    return true;
+    return r.result == wabt::interp::Result::Ok;
```
axic, line 90: I still hoped to prove both branches are covered with #381, but will add back the inverse of this commit for a test.

test/utils/asserts.hpp
chfast: It generates code like `std::string s = ""`.

PR #362, opened 2020-05-29
Add module helpers

gumb0:
This will also help #351


axic:
Merging this as it seems harmless to benchmarking speed.

PR #363, opened 2020-05-29
Validate local count in local.set/get/tee

axic:


lib/fizzy/parser_expr.cpp
```diff
@@ -147,6 +147,8 @@ parser_result<Code> parse_expr(
     const auto function_arity = static_cast<uint8_t>(module.typesec[func_type_idx].outputs.size());
     // The function's implicit block.
     control_stack.emplace(Instr::block, function_arity, 0);
+    
+    local_count += module.typesec[func_type_idx].inputs.size();
```
axic, line 151: I guess the suggestion here is to keep `local_count` const and have another local variable here.

lib/fizzy/parser.cpp
```diff
@@ -346,19 +346,20 @@ inline Code parse_code(code_view code_binary, FuncIdx func_idx, const Module& mo
...
-
     uint64_t local_count = 0;
     for (const auto& l : locals_vec)
     {
         local_count += l.count;
         if (local_count > std::numeric_limits<uint32_t>::max())
             throw parser_error{"too many local variables"};
     }
+
+    auto [code, pos2] = parse_expr(pos1, end, static_cast<uint32_t>(local_count), func_idx, module);
```
axic, line 357: First I tried to pass on `locals_vec` but realised it would need to do this counting again. We can refactor this when type checking is introduced, because then we'll need the types of locals.

lib/fizzy/parser_expr.cpp
```diff
@@ -564,6 +566,13 @@ parser_result<Code> parse_expr(
         {
             uint32_t imm;
             std::tie(imm, pos) = leb128u_decode<uint32_t>(pos, end);
+
+            if (instr == Instr::local_get || instr == Instr::local_set)
```
axic, line 570: Need to add `local_tee` here. However it seems that doesn't change the number of successful validation tests, which suggests they are not covered  upstream.

lib/fizzy/parser_expr.cpp
```diff
@@ -133,8 +133,8 @@ void push_branch_immediates(const ControlFrame& frame, bytes& immediates)
 
 }  // namespace
 
-parser_result<Code> parse_expr(
-    const uint8_t* pos, const uint8_t* end, FuncIdx func_idx, const Module& module)
+parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end, uint32_t local_count,
+    FuncIdx func_idx, const Module& module)
 {
     Code code;
```
axic, line 139: Could move `code.local_count` next to this or even set it in the constructor?

lib/fizzy/parser_expr.cpp
```diff
@@ -148,6 +148,11 @@ parser_result<Code> parse_expr(
     // The function's implicit block.
     control_stack.emplace(Instr::block, function_arity, 0);
 
+    code.local_count = local_count;
+
+    // This is needed for aiding validation.
+    local_count += static_cast<uint32_t>(module.typesec[func_type_idx].inputs.size());
```
axic, line 154: Okay to increment the argument or should have a new temporary?

lib/fizzy/parser_expr.cpp
```diff
@@ -562,6 +567,13 @@ parser_result<Code> parse_expr(
         {
             uint32_t imm;
             std::tie(imm, pos) = leb128u_decode<uint32_t>(pos, end);
+
+            if (instr == Instr::local_get || instr == Instr::local_set || instr == Instr::local_tee)
+            {
+                if (local_count <= imm)
```
axic, line 573: Should swap operands for readability?

lib/fizzy/parser.hpp
```diff
@@ -42,8 +42,8 @@ inline parser_result<uint8_t> parse_byte(const uint8_t* pos, const uint8_t* end)
 /// @param end      The end of the binary input.
 /// @param func_idx Index of the function being parsed.
 /// @param module   Module that this code is part of.
-parser_result<Code> parse_expr(
-    const uint8_t* input, const uint8_t* end, FuncIdx func_idx, const Module& module);
+parser_result<Code> parse_expr(const uint8_t* input, const uint8_t* end, uint32_t local_count,
```
chfast, line 45: Remember to update docs comment.

lib/fizzy/parser_expr.cpp
chfast: Setting in constructor sounds good. You can also make the field const.

There is also option to have `code` object with `local_count` pre-created in `module`. It may be good to leave a TODO comment about this.

lib/fizzy/parser_expr.cpp
chfast: It is better to have separate variable (I find modifying args values confusing). But I have not suggestions how to name the new variable.

lib/fizzy/parser_expr.cpp
chfast: Yes. The "constant" on right is better.

lib/fizzy/parser_expr.cpp
```diff
@@ -562,6 +567,13 @@ parser_result<Code> parse_expr(
...
+                    throw validation_error{"unknown local index"};
```
chfast, line 574:
```suggestion
                    throw validation_error{"invalid local index"};
```

lib/fizzy/parser_expr.cpp
axic: `max_local_index` ?

lib/fizzy/parser_expr.cpp
axic: Done.

lib/fizzy/parser_expr.cpp
axic: Done.

lib/fizzy/parser_expr.cpp
axic: Done.

lib/fizzy/parser_expr.cpp
axic: Added comment for now.

lib/fizzy/parser.hpp
axic: Done.

lib/fizzy/parser_expr.cpp
```diff
@@ -148,6 +150,12 @@ parser_result<Code> parse_expr(
     // The function's implicit block.
     control_stack.emplace(Instr::block, function_arity, 0);
 
+    // This is needed for aiding validation.
+    assert((uint64_t{local_count} + module.typesec[func_type_idx].inputs.size()) <=
```
axic, line 154: Probably should make this a proper parser error...

lib/fizzy/parser_expr.cpp
```diff
@@ -560,9 +568,16 @@ parser_result<Code> parse_expr(
         case Instr::local_set:
         case Instr::local_tee:
         {
-            uint32_t imm;
-            std::tie(imm, pos) = leb128u_decode<uint32_t>(pos, end);
-            push(code.immediates, imm);
+            uint32_t idx;
+            std::tie(idx, pos) = leb128u_decode<uint32_t>(pos, end);
+
+            if (instr == Instr::local_get || instr == Instr::local_set || instr == Instr::local_tee)
```
chfast, line 574: This is always true.

axic: Right, will remove.

It is a remnant from before #351 was meged.

axic: Fixed.

lib/fizzy/parser.hpp
```diff
@@ -38,12 +38,13 @@ inline parser_result<uint8_t> parse_byte(const uint8_t* pos, const uint8_t* end)
...
 ///
-/// @param input    The beginning of the expr binary input.
-/// @param end      The end of the binary input.
-/// @param func_idx Index of the function being parsed.
-/// @param module   Module that this code is part of.
-parser_result<Code> parse_expr(
-    const uint8_t* input, const uint8_t* end, FuncIdx func_idx, const Module& module);
+/// @param input       The beginning of the expr binary input.
+/// @param end         The end of the binary input.
+/// @param local_count The number of locals defined.
```
gumb0, line 43: We are going to need types for type validation, maybe makes sense to pass `const Locals&` right away?

axic: This was discussed in the an earlier comment. I started out with `std::vector<Locals>` but then local counting and validation of it has to be moved into `parse_expr`, which I think isn't the nicest.

I'd defer this work once we implement type checking.

lib/fizzy/parser.hpp
```diff
@@ -38,12 +38,13 @@ inline parser_result<uint8_t> parse_byte(const uint8_t* pos, const uint8_t* end)
...
+/// @param func_idx    Index of the function being parsed.
+/// @param module      Module that this code is part of.
+parser_result<Code> parse_expr(const uint8_t* input, const uint8_t* end, const uint32_t local_count,
```
gumb0, line 46: We don't declare arguments passed by value `const` in other places
```suggestion
parser_result<Code> parse_expr(const uint8_t* input, const uint8_t* end, uint32_t local_count,
```

lib/fizzy/parser_expr.cpp
gumb0, line 154: Would be nice to declare `const auto& func_type = module.typesec[func_type_idx];`
(you would need to rename a variable in `call` and `call_indirect` cases)

lib/fizzy/parser_expr.cpp
```diff
@@ -148,6 +150,12 @@ parser_result<Code> parse_expr(
     // The function's implicit block.
     control_stack.emplace(Instr::block, function_arity, 0);
 
+    // This is needed for aiding validation.
+    assert((uint64_t{local_count} + module.typesec[func_type_idx].inputs.size()) <=
+           std::numeric_limits<uint32_t>::max());
+    const uint32_t max_local_index =
```
gumb0, line 156: Well it's not max index, maybe `total_local_arg_count` or `local_and_arg_count`

lib/fizzy/parser.hpp
axic: Fixed.

lib/fizzy/parser_expr.cpp
axic: Added.

lib/fizzy/parser_expr.cpp
axic: Why isn't it max index? The immediate it is compared against is called "local index".

lib/fizzy/parser.hpp
chfast: Maybe create `Code` object in `Module` before `parse_expr`?

lib/fizzy/parser_expr.cpp
gumb0: Because it's max index + 1

lib/fizzy/parser.hpp
gumb0: > Maybe create `Code` object in `Module` before `parse_expr`?

I would advice against this, because it would make `code` confusingly both input and output parameter

gumb0:
Looks ok, needs a test fro validation failure.

lib/fizzy/parser.hpp
chfast: It does not mean that `parse_expr` should return `Code`. Maybe it should return `Expr`. But is also bad to ask `parse_expr` to move the information about locals (here `local_count`, maybe type information later) from one place to another. That's unneeded and trivial responsibility.

axic:
@gumb0 added tests.

lib/fizzy/parser_expr.cpp
```diff
@@ -138,21 +138,30 @@ void push_branch_immediates(const ControlFrame& frame, bytes& immediates)
...
-    const auto function_arity = static_cast<uint8_t>(module.typesec[func_type_idx].outputs.size());
+    const auto& func_type = module.typesec[func_type_idx];
+
+    const auto function_arity = static_cast<uint8_t>(func_type.outputs.size());
     // The function's implicit block.
     control_stack.emplace(Instr::block, function_arity, 0);
 
+    // This is needed for aiding validation.
+    assert(
+        (uint64_t{local_count} + func_type.inputs.size()) <= std::numeric_limits<uint32_t>::max());
```
axic, line 162: I think this case is also missing from the spectests.

PR #364, opened 2020-05-29
Add smoketests to ctest

chfast:


axic:
Need to also add it to ctest.

PR #354, opened 2020-05-27
Single arity immediate per br_table instruction

gumb0:
Get rid of duplicated arity in `br_table` immediates.

PR #368, opened 2020-06-02
Validate arity of labels in br_table

gumb0:
Required for #354.

There's a slight complication around loop's arity: turns out if there's loop in `br_table` labels, it's arity is considered always 0 (and then `br_table` is valid only if all labels have arity 0). 
It's difficult to find clear confirmation of this in the spec, but spec tests cover it in this case https://github.com/WebAssembly/spec/blob/636b862b9c8a25ad65fb240fefd673e7f23bcdd0/test/core/br_table.wast#L1229-L1244

### DIFF
--- a/lib/fizzy/parser_expr.cpp
+++ b/lib/fizzy/parser_expr.cpp
@@ -119,6 +119,13 @@ void validate_result_count(const ControlFrame& frame)
     //     throw validation_error{"too many results"};
 }
 
+inline uint8_t get_branch_arity(const ControlFrame& frame) noexcept
+{
+    // For loops arity is considered always 0, because br executed in loop jumps to the top,
+    // resetting frame stack to 0, so it should not keep top stack value even if loop has a result.
+    return frame.instruction == Instr::loop ? 0 : frame.arity;
+}
+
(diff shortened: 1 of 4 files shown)


gumb0:
I will squash test commits before merging.

lib/fizzy/parser_expr.cpp
```diff
@@ -481,19 +486,23 @@ parser_result<Code> parse_expr(
...
+
             // Remember immediates offset for all br items to fill them at end instruction.
             push(code.immediates, static_cast<uint32_t>(label_indices.size()));
             for (const auto idx : label_indices)
             {
                 auto& branch_frame = control_stack[idx];
-                branch_frame.br_immediate_offsets.push_back(code.immediates.size());
 
+                if (get_branch_arity(branch_frame) != default_branch_arity)
+                    throw validation_error{"br_table labels have inconsistent block result types"};
```
chfast, line 499:
```suggestion
                    throw validation_error{"br_table labels have inconsistent types"};
```
... as wabt is reporting this...

gumb0:
Changed error message and squashed test commits.

gumb0:
Rebased.

PR #369, opened 2020-06-03
refactor: use proper variable names instead of generic imm in parser_expr

axic:
Inspired by #310 and #363.

### DIFF
--- a/lib/fizzy/parser_expr.cpp
+++ b/lib/fizzy/parser_expr.cpp
@@ -560,41 +560,41 @@ parser_result<Code> parse_expr(
         case Instr::local_set:
         case Instr::local_tee:
         {
-            uint32_t imm;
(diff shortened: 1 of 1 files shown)


lib/fizzy/parser_expr.cpp
```diff
@@ -560,9 +560,9 @@ parser_result<Code> parse_expr(
         case Instr::local_set:
         case Instr::local_tee:
         {
-            uint32_t imm;
-            std::tie(imm, pos) = leb128u_decode<uint32_t>(pos, end);
-            push(code.immediates, imm);
+            uint32_t idx;
```
axic, line 563: Should this be `local_idx`?

chfast: `local_idx` sounds nice.

PR #370, opened 2020-06-03
test: move all helpers under the fizzy::test namespace

axic:


test/utils/asserts.cpp
```diff
@@ -5,7 +5,7 @@
 #include "asserts.hpp"
 #include <ostream>
 
-namespace fizzy
+namespace fizzy::test
```
chfast, line 8: This will break [ADL](https://en.cppreference.com/w/cpp/language/adl).

axic: Not sure I understand, a this notation was already used in some files.

chfast: I mean putting `operator<<` for class `fizzy::execution_result` in some other namespace breaks ADL.

test/utils/hex.hpp
```diff
@@ -37,7 +37,7 @@ inline namespace literals
 /// Operator for "" literals, e.g. "0a0b0c0d"_bytes.
 inline fizzy::bytes operator""_bytes(const char* literal, size_t /*length*/)
 {
-    return fizzy::from_hex(literal);
+    return fizzy::test::from_hex(literal);
```
gumb0, line 40: looks like it could be just
```suggestion
    return from_hex(literal);
```

test/utils/asserts.cpp
axic: But if the PR compiles then nothing broke or am I missing something?

gumb0: Did you try output of failed `EXPECT_THAT(..., Result(x))` ? It might be broken.

With this change any `cout << result;` would not work unles it's inside `fizzy::test` or has `using namespace fizzy::test`.

I woul leave it in `fizzy`

axic: > Did you try output of failed EXPECT_THAT(..., Result(x)) ? It might be broken.
>
>With this change any cout << result; would not work unless it's inside fizzy::test or has using namespace fizzy::test.

All tests are like that.

gumb0: The code outputting `expected_result` in case of failure is not in the tests, it's in google.test.
This is what happens after this change:
```
Value of: execute(module, 0, {3, 6, 0})
Expected: result 5
  Actual: 32-byte object <00-6E 47-91 FC-7F 00-01 80-8C 81-02 00-00 00-00 88-8C 81-02 00-00 00-00 88-8C 81-02 00-00 00-00> (of type fizzy::execution_result)
```

axic: I don't really get how is it better for this reason to be in `fizzy` and not `fizzy::test`.

gumb0: When compiler sees `cout << result` it will include namespace of `execution_result` in the search for `operator<<` definition (so it will searchi in `fizzy`, but not in `fizzy::test`)

test/utils/hex.hpp
axic: Actually no, because this is in a difference namespace (`literals`).

test/utils/asserts.cpp
axic: How about this version of the PR then. Removed moving the `<<` operator.

test/utils/hex.hpp
chfast: But `literals` is in `fizzy::test`.

axic:
@chfast @gumb0 should we merge?

PR #371, opened 2020-06-03
Add cmake-format and cmake-lint

axic:


axic:
The rules required are just too strict and have too many stylistic differences to current. We may be able to relax it however, but closing for now.

PR #372, opened 2020-06-04
Validate constant expressions for global/offsets initialization

gumb0:
More validation for number of instructions in const expr will be added in #377 

I will squash test commits before merging.

### DIFF
--- a/lib/fizzy/parser.cpp
+++ b/lib/fizzy/parser.cpp
@@ -550,6 +550,9 @@ Module parse(bytes_view input)
     if (!module.elementsec.empty() && !module.has_table())
         throw validation_error("element section encountered without a table section");
 
+    for (const auto& element : module.elementsec)
+        validate_constant_expression(element.offset, module);
+
     const auto total_global_count = module.get_global_count();
     for (const auto& global : module.globalsec)
     {
--- a/lib/fizzy/parser.cpp
+++ b/lib/fizzy/parser.cpp
@@ -535,6 +535,9 @@ Module parse(bytes_view input)
     if (!module.datasec.empty() && !module.has_memory())
         throw validation_error("data section encountered without a memory section");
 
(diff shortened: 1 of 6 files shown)


test/unittests/instantiate_test.cpp
```diff
@@ -852,7 +852,7 @@ TEST(instantiate, globals_initialized_from_imported)
     const auto bin_invalid2 = from_hex("0061736d01000000060b027f00412a0b7f0123000b");
     const auto module_invalid2 = parse(bin_invalid2);
 
-    EXPECT_THROW_MESSAGE(instantiate(module_invalid2, {}, {}), instantiate_error,
+    EXPECT_THROW_MESSAGE(instantiate(module_invalid2, {}, {}), validation_error,
```
gumb0, line 855: should be moved to validation tests

gumb0:
> 1. Once you move a test to validation_test.cpp in the same commit where functional change is. Other time this is done in two separate commits.

I made separate commits with the moves, now there're more test commits.

> 2. I'd rather use `global.get`, not `global_get` in comments and error messages.

Changed.

lib/fizzy/parser.cpp
```diff
@@ -12,6 +12,22 @@
 
 namespace fizzy
 {
+namespace
+{
+void validate_constant_expression(const ConstantExpression& const_expr, Module const& module)
```
gumb0, line 17:
```suggestion
void validate_constant_expression(const ConstantExpression& const_expr, const Module& module)
```

axic:
Will review this later today.

gumb0:
Rebased.

gumb0:
Rebased.

gumb0:
Reordered commits.

axic:
Replaced by individual PRs.

PR #377, opened 2020-06-08
Validate number of instructions in constant expression

gumb0:
~~Based on #372~~
Based on #397

According to [validation spec](https://webassembly.github.io/spec/core/valid/modules.html#globals), result of constant expression has type
- `[t]` for global initializer
- `[i32]` for offset initializers

This implies that constant expression can have only single `*.const` or `global.get` instruction.

Spec tests check that with expected error "type mismatch": https://github.com/WebAssembly/spec/blob/60facfbb74cbb9f818258bbb07d02c581352e592/test/core/global.wast#L281-L289

### DIFF
--- a/lib/fizzy/parser.cpp
+++ b/lib/fizzy/parser.cpp
@@ -123,60 +123,62 @@ inline parser_result<ConstantExpression> parse_constant_expression(
 {
     ConstantExpression result;
 
-    Instr instr;
-    do
+    uint8_t opcode;
+    std::tie(opcode, pos) = parse_byte(pos, end);
+
+    const auto instr = static_cast<Instr>(opcode);
+    switch (instr)
     {
-        uint8_t opcode;
-        std::tie(opcode, pos) = parse_byte(pos, end);
+    default:
+        throw validation_error{
+            "unexpected instruction in the constant expression: " + std::to_string(*(pos - 1))};
 
-        instr = static_cast<Instr>(opcode);
-        switch (instr)
-        {
-        default:
-            throw validation_error{
-                "unexpected instruction in the constant expression: " + std::to_string(*(pos - 1))};
+    case Instr::end:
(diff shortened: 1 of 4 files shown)


test/unittests/validation_test.cpp
```diff
@@ -52,6 +52,19 @@ TEST(validation, globals_invalid_initializer)
     const auto bin_invalid3 = from_hex("0061736d01000000060b027f00412a0b7f0123000b");
     EXPECT_THROW_MESSAGE(parse(bin_invalid3), validation_error,
         "global can be initialized by another const global only if it's imported");
+
+    /* wat2wasm --no-check
+      (global (mut i64) (i32.const 16) (i64.const -1))
```
chfast, line 57:
```suggestion
      (global (mut i64) (i64.const 16) (i64.const -1))
```
To make sure it fails on instructions count, not individual type mismatch.

lib/fizzy/parser.cpp
```diff
@@ -121,60 +121,62 @@ inline parser_result<ConstantExpression> parse_constant_expression(
 {
     ConstantExpression result;
 
-    Instr instr;
-    do
+    uint8_t opcode;
+    std::tie(opcode, pos) = parse_byte(pos, end);
+
+    const Instr instr = static_cast<Instr>(opcode);
```
chfast, line 127:
```suggestion
    const auto instr = static_cast<Instr>(opcode);
```

lib/fizzy/parser.cpp
```diff
@@ -12,6 +12,22 @@
 
 namespace fizzy
 {
+namespace
+{
+void validate_constant_expression(const ConstantExpression& const_expr, Module const& module)
```
axic, line 17: I'm still confused by C++'s random placement of `const`.

PR #381, opened 2020-06-08
test: add test case for failing start function in wasm_engine_test

axic:


axic:
Not sure this is a fault of our binding or upstream, but it definitely isn't the fault of the patch we have as I have tried it without the patch.

test/utils/wasm3_engine.cpp
```diff
@@ -84,14 +84,13 @@ bool Wasm3Engine::instantiate(bytes_view wasm_binary)
     // Transfers ownership to runtime.
     if (m3_LoadModule(m_runtime, module) != m3Err_none)
     {
-        m3_FreeModule(module);
+        // NOTE: apparently m3_FreeModule isn't needed in neither the failure nor the success case
         return false;
     }
 
     auto ret = m3_LinkRawFunction(module, "env", "adler32", "i(ii)", env_adler32);
     if (ret != m3Err_none && ret != m3Err_functionLookupFailed)
```
gumb0, line 92:
```suggestion
    return (ret == m3Err_none || ret == m3Err_functionLookupFailed);
```

axic: I really dislike this setting of clang-tidy.

test/utils/wasm3_engine.cpp
```diff
@@ -84,14 +84,13 @@ bool Wasm3Engine::instantiate(bytes_view wasm_binary)
     // Transfers ownership to runtime.
     if (m3_LoadModule(m_runtime, module) != m3Err_none)
     {
-        m3_FreeModule(module);
+        // NOTE: apparently m3_FreeModule isn't needed in neither the failure nor the success case
```
chfast, line 87: How this is known?

test/utils/wasm3_engine.cpp
chfast: You can disable it.

test/utils/wasm3_engine.cpp
axic: By going through the wasm3 code. Basically it seem once a module is parsed you have two choices: `LoadModule` or `FreeModule`, and if `LoadModule` was called then it is the liability of that to deal with it. Started opening in issue on their repo to clarify this, but then I guess forgot to submit.

axic: Asked the question: https://github.com/wasm3/wasm3/issues/160

axic:
@chfast @gumb0 suggest we merge this.

test/utils/wasm3_engine.cpp
```diff
@@ -84,18 +84,12 @@ bool Wasm3Engine::instantiate(bytes_view wasm_binary)
...
         return false;
     }
 
     auto ret = m3_LinkRawFunction(module, "env", "adler32", "i(ii)", env_adler32);
-    if (ret != m3Err_none && ret != m3Err_functionLookupFailed)
-    {
-        m3_FreeRuntime(m_runtime);
```
axic, line 94: This is a second bugfix, since we didn't set `m_runtime` to null, it may have attempted to free it twice (in the destructor).

PR #384, opened 2020-06-10
Validate stack height after unreachable

gumb0:
This applies the logic of [Validation algorithm](https://webassembly.github.io/spec/core/appendix/algorithm.html) regarding unreachable to stack height checks.

Specifically
- After unreachable any underflowing pop is ignored, but pushes add to stack growth.
image

- When current frame is marked unreachable, stack height is reset.
image

"Simplify" commit is not squashed yet to show the different nested organization of nested ifs. I find the final one better.

### DIFF
--- a/lib/fizzy/parser_expr.cpp
+++ b/lib/fizzy/parser_expr.cpp
@@ -93,28 +93,39 @@ void update_caller_frame(ControlFrame& frame, const FuncType& func_type)
 {
     const auto stack_height_required = static_cast<int>(func_type.inputs.size());
 
-    if (!frame.unreachable &&
-        (frame.stack_height - frame.parent_stack_height) < stack_height_required)
-        throw validation_error{"call/call_indirect instruction stack underflow"};
-
-    const auto stack_height_change =
-        static_cast<int>(func_type.outputs.size()) - stack_height_required;
-    frame.stack_height += stack_height_change;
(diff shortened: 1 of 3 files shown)


lib/fizzy/parser_expr.cpp
```diff
@@ -162,21 +179,32 @@ parser_result<Code> parse_expr(
...
+        {
             // Update code's max_stack_height using frame.stack_height of the previous instruction.
             // At this point frame.stack_height includes additional changes to the stack height
             // if the previous instruction is a call/call_indirect.
             // This way the update is skipped for end/else instructions (because their frame is
             // already popped/reset), but it does not matter, as these instructions do not modify
             // stack height anyway.
-            code.max_stack_height = std::max(code.max_stack_height, frame.stack_height);
-        }
+            if (!frame.unreachable)
```
gumb0, line 203: If we pull this `max_stack_height` out of outer `if`, it looks like we can unify stack height update/checks for instructions and for calls, not sure if it's worth it. The error message for stack underflow would not be specific for calls then...

chfast: I think this can be done in separate PR to confirm coverage of this version is full.

gumb0:
Squashed "simplify" commit.

PR #385, opened 2020-06-11
Validate types - keeping both stack height and type validation

gumb0:


lib/fizzy/stack.hpp
```diff
@@ -44,8 +44,15 @@ class Stack : public std::vector<T>
...
     T& operator[](size_t index) noexcept { return std::vector<T>::operator[](size() - index - 1); }
 
+    T operator[](size_t index) const noexcept
+    {
+        return std::vector<T>::operator[](size() - index - 1);
+    }
+
     T& top() noexcept { return (*this)[0]; }
 
+    T top() const noexcept { return (*this)[0]; }
```
gumb0, line 54: To avoid copying complex types, it could be
```suggestion
    std::conditional_t<std::is_fundamental_v<T>, T, const T&>
    T top() const noexcept { return (*this)[0]; }
```
but I think it would be unneeded complexity, as currently it's used only for `Stack<ValType>`

(Also, const `operator[]` is added only for completeness, it's not used in the code)

axic:
Please rebase.

gumb0:
Replaced by #402

PR #386, opened 2020-06-15
Validate stack height at branch instructions

gumb0:


lib/fizzy/parser_expr.cpp
```diff
@@ -124,6 +124,17 @@ inline uint8_t get_branch_arity(const ControlFrame& frame) noexcept
     return frame.instruction == Instr::loop ? 0 : frame.arity;
 }
 
+inline void validate_branch_stack_height(
+    const ControlFrame& current_frame, const ControlFrame& branch_frame)
+{
+    // assert(current_frame.stack_height >= current_frame.parent_stack_height);
```
gumb0, line 130: This can be enabled after #384 is merged

lib/fizzy/parser_expr.cpp
```diff
@@ -124,6 +124,17 @@ inline uint8_t get_branch_arity(const ControlFrame& frame) noexcept
...
+
+    if (!current_frame.unreachable && get_branch_arity(branch_frame) == 1 &&
```
chfast, line 132: I think this can be expressed without assuming `branch_airty ∈{0,1}`. E.g. like 
```cpp
current_frame.stack_height - get_branch_arity(branch_frame) < current_frame.parent_stack_height
```
Just make sure this uses signed integer types.

gumb0: Nice suggestion, I changed it to be similar to checks in `validate_result_count`

PR #390, opened 2020-06-23
parser: Save global value type during parsing

gumb0:
Required for type validation, pulled out from #385

### DIFF
--- a/test/unittests/instantiate_test.cpp
+++ b/test/unittests/instantiate_test.cpp
@@ -382,6 +382,27 @@ TEST(instantiate, imported_globals_mismatched_mutability)
         "global 0 mutability doesn't match module's global mutability");
 }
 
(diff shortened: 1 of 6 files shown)


test/unittests/instantiate_test.cpp
```diff
@@ -382,6 +382,27 @@ TEST(instantiate, imported_globals_mismatched_mutability)
         "global 0 mutability doesn't match module's global mutability");
 }
 
+TEST(instantiate, DISABLED_imported_globals_mismatched_type)
```
gumb0, line 385: To ebable this, we'll need to change `ExternalGlobal` type to include value type.

Also, it looks to be not covered in spec tests.

axic: Did you add this to the list?

gumb0: Added.

PR #391, opened 2020-06-25
Keep number of stack items to drop for branches

chfast:
Instead of keeping information about the absolute height to rewind the
stack to, keep the number of stack items to drop (without the number of
results - arity of the result type).

This makes OperandStack simpler (no need to track the bottom of the stack). No speed change.

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -264,11 +264,11 @@ void branch(
     {
         assert(arity == 1);
         const auto result = stack.top();
-        stack.shrink(stack.size() - stack_drop);
+        stack.drop(stack_drop);
         stack.top() = result;
     }
     else
-        stack.shrink(stack.size() - stack_drop);
+        stack.drop(stack_drop);
 }
 
 template <class F>
@@ -278,7 +278,7 @@ bool invoke_function(
     const auto num_args = func_type.inputs.size();
     assert(stack.size() >= num_args);
     std::vector<uint64_t> call_args{stack.rend() - num_args, stack.rend()};
-    stack.shrink(stack.size() - num_args);
+    stack.drop(num_args);
 
     const auto ret = func(instance, std::move(call_args), depth + 1);
     // Bubble up traps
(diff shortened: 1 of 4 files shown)


lib/fizzy/parser_expr.cpp
```diff
@@ -153,14 +153,20 @@ inline void validate_branch_stack_height(
 }
 
 
-void push_branch_immediates(const ControlFrame& frame, bytes& immediates)
+void push_branch_immediates(const ControlFrame& frame, int branch_stack_height, bytes& immediates)
```
gumb0, line 156: Better rename parameters to be similar to variables on the caller side
```suggestion
void push_branch_immediates(const ControlFrame& branch_frame, int stack_height, bytes& immediates)
```

axic:
@chfast did you benchmark the effect of this?

chfast:
> @chfast did you benchmark the effect of this?

> No speed change.

lib/fizzy/execute.cpp
```diff
@@ -248,23 +248,24 @@ void branch(
...
-    const auto stack_height = static_cast<size_t>(read<uint32_t>(immediates));
+    const auto stack_drop = read<uint32_t>(immediates);
     const auto arity = read<uint8_t>(immediates);
 
     pc = code.instructions.data() + code_offset;
     immediates = code.immediates.data() + imm_offset;
 
     // When branch is taken, additional stack items must be dropped.
-    assert(stack.size() >= stack_height + arity);
+    assert(static_cast<int32_t>(stack_drop) >= 0);
```
gumb0, line 258: It was `int` on the parser side
```suggestion
    assert(static_cast<int>(stack_drop) >= 0);
```

lib/fizzy/parser_expr.cpp
```diff
@@ -153,14 +153,21 @@ inline void validate_branch_stack_height(
 }
 
 
-void push_branch_immediates(const ControlFrame& frame, bytes& immediates)
+void push_branch_immediates(
+    const ControlFrame& branch_frame, int stack_height, bytes& immediates) noexcept
 {
+    const auto arity = get_branch_arity(branch_frame);
+
+    // How much stack items to drop when taking the branch.
```
gumb0, line 161:
```suggestion
    // How many stack items to drop when taking the branch.
```

gumb0:
Looks good, two minor nitpicks.

PR #393, opened 2020-06-25
Validate constant expressions for data offset initialization

gumb0:
Second part of https://github.com/wasmx/fizzy/pull/372


gumb0:
> So these are not covered by the spectests? If so, can you add this "our list"?

Yes, I'll add it.

PR #394, opened 2020-06-26
Validate constant expressions in globals initializers

gumb0:
First part of https://github.com/wasmx/fizzy/pull/372

### DIFF
--- a/lib/fizzy/parser.cpp
+++ b/lib/fizzy/parser.cpp
@@ -12,6 +12,22 @@
 
 namespace fizzy
 {
+namespace
+{
+void validate_constant_expression(const ConstantExpression& const_expr, const Module& module)
+{
+    if (const_expr.kind == ConstantExpression::Kind::Constant)
+        return;
+
+    const auto global_idx = const_expr.value.global_index;
+    if (global_idx >= module.get_global_count())
+        throw validation_error{"invalid global index in constant expression"};
+
+    if (module.is_global_mutable(global_idx))
+        throw validation_error{"constant expression can use global.get only for const globals"};
+}
+}  // namespace
+
 template <typename T>
 parser_result<T> parse(const uint8_t* pos, const uint8_t* end);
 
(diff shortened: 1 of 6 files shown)


test/unittests/parser_test.cpp
```diff
@@ -594,10 +594,15 @@ TEST(parser, global_single_mutable_const_inited)
 
 TEST(parser, global_single_const_global_inited)
```
axic, line 595: Test name doesn't reflect the test anymore.

gumb0: Renamed

lib/fizzy/parser.cpp
```diff
@@ -12,6 +12,22 @@
...
+{
+    if (const_expr.kind == ConstantExpression::Kind::Constant)
+        return;
+
+    const auto global_idx = const_expr.value.global_index;
+    if (global_idx >= module.get_global_count())
+        throw validation_error{"invalid global index in constant expression"};
+
+    if (module.is_global_mutable(global_idx))
+        throw validation_error("constant expression can use global.get only for const globals");
```
axic, line 27: Why is this using `validation_error()` while the other `validation_error{}` ?

lib/fizzy/parser.cpp
```diff
@@ -531,11 +547,26 @@ Module parse(bytes_view input)
...
+    for (const auto& global : module.globalsec)
+    {
+        validate_constant_expression(global.expression, module);
+
+        // Wasm spec section 3.3.7 constrains initialization by another global to const imports only
+        // https://webassembly.github.io/spec/core/valid/instructions.html#expressions
+        if (global.expression.kind == ConstantExpression::Kind::GlobalGet &&
+            global.expression.value.global_index >= module.imported_globals_mutability.size())
+        {
+            throw validation_error(
```
axic, line 560: Also uses parenthesis. It seems we are inconsistently using parentheses and curly brackets.

lib/fizzy/parser.cpp
gumb0: Changed it to always curly braces in my changes.

gumb0:
Rebased.

axic:
Rebased. @gumb0 please check.

PR #403, opened 2020-07-01
Minimum viable type validation

gumb0:


lib/fizzy/instructions.cpp
```diff
@@ -43,20 +43,20 @@ constexpr InstructionMetrics instruction_metrics_table[256] = {
     /*                       0x19 */ {},
 
     // 5.4.2 Parametric instructions
-    /* drop                = 0x1a */ {1, -1},
-    /* select              = 0x1b */ {3, -2},
+    /* drop                = 0x1a */ {0, 0},
```
gumb0, line 46: These are changed only for consistency with number of types in `instruction_type_table` below.

In fact these values are not used anymore at all, and I plan to remove `instruction_metrics_table` in another PR, or rather transform it into the table with only memory alignment values.

lib/fizzy/parser.cpp
```diff
@@ -373,13 +373,18 @@ inline Code parse_code(code_view code_binary, FuncIdx func_idx, const Module& mo
             throw parser_error{"too many local variables"};
     }
 
-    const auto [code, pos2] =
-        parse_expr(pos1, end, static_cast<uint32_t>(local_count), func_idx, module);
+    // TODO: Clarify in spec what happens if count of locals and arguments exceed uint32_t::max()
```
gumb0, line 376: I'm moving this assertion from `parse_expr` to `parse_code`, because now `parser_expr` doesn't get `local_count` and iterates `locals_vec` anyway on each local access, to find the type.

lib/fizzy/parser_expr.cpp
```diff
@@ -88,50 +83,55 @@ parser_result<std::optional<ValType>> parse_blocktype(const uint8_t* pos, const
...
+    const auto frame_stack_height = static_cast<int>(operand_stack.size());
+    const auto inputs_size = static_cast<int>(inputs.size());
+
+    // Stack is polymorphic after unreachable instruction: underflow is ignored,
+    // but we need to count stack growth to detect extra values at the end of the block.
+    if (!frame.unreachable && frame_stack_height < frame.parent_stack_height + inputs_size)
+        throw validation_error{"stack underflow"};
+
+    // Update operand_stack.
+    for (auto it = inputs.begin() + inputs.size(); it != inputs.begin(); --it)
```
gumb0, line 98: Would be nice to have reverse iterator for `span` instead of this, but I would leave it for another PR.

lib/fizzy/instructions.hpp
```diff
@@ -39,4 +41,12 @@ struct InstructionMetrics
 
 const InstructionMetrics* get_instruction_metrics_table() noexcept;
 
+struct InstructionType
+{
+    std::initializer_list<ValType> inputs;
```
gumb0, line 46: I'm still not sure whether it's valid to use `initializer_list` like this.

Declaring `constexpr` `initalizer_list` object works only if it's static local or global, but doesn't compile if it's auto https://godbolt.org/z/iWUg6B

The discussions that I found are not very clear to me, but seem to come to the conclusion that standard doesn't guarantee it to be valid.
https://stackoverflow.com/questions/16063123/is-it-legal-to-declare-a-constexpr-initializer-list-object
https://stackoverflow.com/questions/27496004/why-isnt-stdinitializer-list-defined-as-a-literal-type

gumb0: Just learned that in C++20 `std::vector` can be `constexpr`

gumb0: Found relatively simple solution with a custom struct instead of `initializer_list`, pushing it as a separate commit for now.

lib/fizzy/instructions.hpp
```diff
@@ -41,10 +41,26 @@ struct InstructionMetrics
 
 const InstructionMetrics* get_instruction_metrics_table() noexcept;
 
+template <typename T, size_t MaxSize>
+struct constexpr_array
```
gumb0, line 45: not sure about the name

lib/fizzy/parser_expr.cpp
axic: If we go ahead with this PR I'd prefer span to be merged separately anyway and in that PR this could be implemented too.

lib/fizzy/instructions.hpp
chfast: > Just learned that in C++20 `std::vector` can be `constexpr`, that would help a lot.

I don't think it would help. It is `constexpr` because you can use `new` and `delete` in constexpr evaluation (`consteval`), but for initialization dynamic memory allocation would be needed, so compilation would probably fail. See also `constinit`.

gumb0:
Rebased and squashed `span`-related and `constexpr_array` commits.

lib/fizzy/module.hpp
```diff
@@ -68,19 +68,19 @@ struct Module
         return imported_global_types.size() + globalsec.size();
     }
 
+    GlobalType get_global_type(GlobalIdx idx) const noexcept
```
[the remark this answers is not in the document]
gumb0: https://github.com/wasmx/fizzy/pull/411

gumb0:
Renamed `constexpr_array` -> `constexpr_vector` and put type table into separate first commit.

gumb0:
Rebased.

lib/fizzy/instructions.hpp
```diff
@@ -4,6 +4,8 @@
 
 #pragma once
 
+#include "span.hpp"
```
gumb0, line 7: not needed

lib/fizzy/instructions.hpp
```diff
@@ -39,4 +41,32 @@ struct InstructionMetrics
 
 const InstructionMetrics* get_instruction_metrics_table() noexcept;
 
+template <typename T, size_t MaxSize>
+struct constexpr_vector
```
axic, line 45: Why have this in `instructions.hpp` and not in `utils.hpp` or something like that? Also should have specific tests for it.

gumb0: I put it into separate file, now I'll make another PR for it.

gumb0: https://github.com/wasmx/fizzy/pull/415

lib/fizzy/instructions.hpp
```diff
@@ -39,4 +41,13 @@ struct InstructionMetrics
 
 const InstructionMetrics* get_instruction_metrics_table() noexcept;
 
+
+struct InstructionType
```
axic, line 45: Can you add a comment here that wasm 1.0 spec only has instructions which take at most 2 items and return at most 1 item?

lib/fizzy/parser_expr.cpp
```diff
@@ -216,7 +251,8 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end, uint32_t
         std::tie(opcode, pos) = parse_byte(pos, end);
 
         auto& frame = control_stack.top();
-        const auto& metrics = metrics_table[opcode];
+        const auto metrics = metrics_table[opcode];
+        const auto types = type_table[opcode];
```
chfast, line 255:
```suggestion
        const auto type = type_table[opcode];
```

lib/fizzy/parser_expr.cpp
```diff
@@ -88,50 +83,55 @@ parser_result<std::optional<ValType>> parse_blocktype(const uint8_t* pos, const
     return {validate_valtype(type), pos};
 }
 
-void update_stack_height(ControlFrame& frame, int stack_height_required, int stack_height_change)
+void update_operand_stack(const ControlFrame& frame, Stack<ValType>& operand_stack,
+    span<const ValType> inputs, span<const ValType> outputs)
```
chfast, line 87:
```suggestion
    const InstructionType& type)
```

lib/fizzy/constexpr_vector.hpp
```diff
@@ -0,0 +1,31 @@
...
+public:
+    template <typename... U>
+    constexpr constexpr_vector(U&&... u) noexcept
+      : m_array{std::forward<U>(u)...}, m_size(sizeof...(U))
+    {}
+
+    constexpr const T* data() const noexcept { return m_array; }
+    constexpr std::size_t size() const noexcept { return m_size; }
+
+    constexpr const T* begin() const noexcept { return m_array; }
```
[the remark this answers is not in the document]
gumb0: Actually with reverse iterator for span these are used only in one test https://github.com/wasmx/fizzy/blob/3c7324f551b25c860b6c578b846dfabd717619be/test/unittests/parser_test.cpp#L1143

But looks convenient there, I would leave it.

lib/fizzy/constexpr_vector.hpp
```diff
@@ -0,0 +1,31 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <utility>
+
+namespace fizzy
+{
+template <typename T, std::size_t MaxSize>
+struct constexpr_vector
```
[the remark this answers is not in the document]
gumb0: I would agree if we kept it in `instructions.hpp`, but with moving it into separate utility header, it seems appropriate to make it generic.

lib/fizzy/parser_expr.cpp
gumb0: This won't work, because this function is used for both `InstructionType` and `FuncType`, that's the whole reason of using `span` here.

lib/fizzy/parser_expr.cpp
gumb0: Ok, but I'll have to rename 4 variables for the block types in the switch below.

lib/fizzy/instructions.cpp
chfast: How are these change not breaking anything?

gumb0: These values are not used anymore, only types table is used for checks including stack height checks.

`drop` just drops an operand from the stack.

`select` pops one `i32` from the stack according to `instruction_type_table`, then drops one operand without type checking in this PR (to be added in #413)
I guess we don't have a unit test checking required number of operands for `select`, and spec tests for it also look not perfect, see description in #413.

gumb0: I've added a disabled test in this PR.

chfast: I would leave the table unmodified with a comment it is obsolete.

Do I understand it right that now you can create a program that will crash because of the wrong max stack height? And that will be fixed with validation of all instructions?

gumb0: Yes, executing the disabled `select` test will crash. I believe it's only for `select`.

(I will roll back the changes to metrics.)

lib/fizzy/parser_expr.cpp
```diff
@@ -173,41 +175,74 @@ void push_branch_immediates(
...
+
+ValType find_local_type(
+    const std::vector<ValType>& params, const std::vector<Locals>& locals, LocalIdx idx)
+{
+    if (idx < params.size())
+        return params[idx];
+
+    const auto local_idx = idx - params.size();
+    uint64_t local_count = 0;
+    for (const auto& l : locals)
```
chfast, line 212: This algorithm will have to be replaced. There exist log2 algorithm without needing to expand to vector of individual types.

lib/fizzy/instructions.cpp
chfast: I'm not saying there is a need to rollback anything. But we should crasher reproducers in tests. I can work on these as a part of the review.

lib/fizzy/parser_expr.cpp
```diff
@@ -489,13 +530,14 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end, uint32_t
                     // stack drop and arity were already stored in br handler
                 }
             }
-            const auto type = frame.type;
+            const auto frame_type = frame.type;
```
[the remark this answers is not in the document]
gumb0: No, it's invalidated by `control_stack.pop()`

lib/fizzy/instructions.cpp
gumb0: > I'm not saying there is a need to rollback anything.

I meant just the changes that don't influence anything.

gumb0: This is the test that will crash in execution https://github.com/wasmx/fizzy/blob/d039bd454a3aa902a3cf3ff532e7f1b5ebb5b920/test/unittests/validation_stack_test.cpp#L13-L25

chfast: I came up with very similar test, pushed to the same commit.

lib/fizzy/parser_expr.cpp
```diff
@@ -173,41 +175,74 @@ void push_branch_immediates(
...
 
-inline void mark_frame_unreachable(ControlFrame& frame) noexcept
+inline void mark_frame_unreachable(ControlFrame& frame, Stack<ValType>& operand_stack) noexcept
 {
     frame.unreachable = true;
-    frame.stack_height = frame.parent_stack_height;
+    operand_stack.shrink(static_cast<size_t>(frame.parent_stack_height));
 }
 
+inline void drop_operand(
```
chfast, line 184: In the "Validation Algorithm" this is a `pop_opd()`. Is the returned type ever needed?

lib/fizzy/parser_expr.cpp
```diff
@@ -622,20 +668,51 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end, uint32_t
         }
 
         case Instr::local_get:
+        {
+            uint32_t local_idx;
```
chfast, line 672:
```suggestion
            LocalIdx local_idx;
```

lib/fizzy/parser_expr.cpp
```diff
@@ -88,50 +83,55 @@ parser_result<std::optional<ValType>> parse_blocktype(const uint8_t* pos, const
...
+    // Stack is polymorphic after unreachable instruction: underflow is ignored,
+    // but we need to count stack growth to detect extra values at the end of the block.
+    if (!frame.unreachable && frame_stack_height < frame.parent_stack_height + inputs_size)
+        throw validation_error{"stack underflow"};
+
+    // Update operand_stack.
+    for (auto it = inputs.rbegin(); it != inputs.rend(); ++it)
     {
-        frame.stack_height += stack_height_change;
+        if (frame.unreachable &&
```
chfast, line 100: Add comment explaining what it does.

lib/fizzy/parser_expr.cpp
```diff
@@ -88,50 +83,55 @@ parser_result<std::optional<ValType>> parse_blocktype(const uint8_t* pos, const
...
+        const auto actual_type = operand_stack.pop();
+        if (actual_type != expected_type)
+            throw validation_error{"type mismatch"};
     }
+    for (const auto output_type : outputs)
+        operand_stack.push(output_type);
 }
 
-inline void update_caller_frame(ControlFrame& frame, const FuncType& func_type)
+inline void update_caller_frame(
```
chfast, line 113: Maybe drop it then?

lib/fizzy/parser_expr.cpp
gumb0: I guess you mean something like iterating over it once to create an array of cumulative sums, then using binary search on this array.

I can add a TODO.

chfast: Yes, this is what I mean. And I'm also for doing it later.

lib/fizzy/parser_expr.cpp
gumb0: Yeah can be dropped

lib/fizzy/parser_expr.cpp
gumb0: I can rename it to `pop_operand`.

Though it's not exactly the same thing as `pop_opd()` in Validation Algorithm, where it's a low-level primitive used to build all other checks. There return value is needed to check against the expected value and also return value can be `Unknown`. 
We never push unknown values into stack, so here return value would always be equal to `expected_type` input value.

chfast:
```
Comparing master to typev                                                                                                                                                                                                                                                       
Benchmark                                                  Time             CPU      Time Old      Time New       CPU Old       CPU New                                                                                                                                         
---------------------------------------------------------------------------------------------------------------------------------------
fizzy/parse/blake2b_mean                                +0.5400         +0.5400            14            22            14            22
fizzy/parse/ecpairing_mean                              +0.7163         +0.7163           774          1329           774          1329
fizzy/parse/keccak256_mean                              +0.5849         +0.5849            26            41            26            41
fizzy/parse/memset_mean                                 +0.3773         +0.3773             4             6             4             6
fizzy/parse/mul256_opt0_mean                            +0.5480         +0.5480             5             7             5             7
fizzy/parse/sha1_mean                                   +0.5750         +0.5766            23            37            23            37
fizzy/parse/sha256_mean                                 +0.7632         +0.7718            36            63            35            63
fizzy/parse/micro/eli_interpreter_mean                  +0.1986         +0.2042             3             4             3             4
fizzy/parse/micro/factorial_mean                        +0.1988         +0.2043             1             1             1             1
fizzy/parse/micro/fibonacci_mean                        +0.2253         +0.2310             1             1             1             1
fizzy/parse/micro/host_adler32_mean                     +0.1422         +0.1470             1             2             1             2
fizzy/parse/micro/spinner_mean                          +0.1556         +0.1556             1             1             1             1
```

lib/fizzy/parser_expr.cpp
chfast: You don't have to rename it. I was just curious.

lib/fizzy/parser_expr.cpp
gumb0: fixed

lib/fizzy/parser_expr.cpp
gumb0: Added comment.

lib/fizzy/parser_expr.cpp
gumb0: Removed it.

gumb0:
I'm getting somewhat less of regression on my laptop
```
fizzy/parse/blake2b_mean                                +0.2133         +0.2153            41            50            41            50
fizzy/parse/ecpairing_mean                              +0.3124         +0.3131          2191          2876          2190          2876
fizzy/parse/keccak256_mean                              +0.2555         +0.2565            73            92            73            92
fizzy/parse/memset_mean                                 +0.1513         +0.1517            10            12            10            12
fizzy/parse/mul256_opt0_mean                            +0.2082         +0.2086            13            15            13            15
fizzy/parse/sha1_mean                                   +0.2586         +0.2587            65            82            65            82
fizzy/parse/sha256_mean                                 +0.2389         +0.2391           112           138           112           138
fizzy/parse/micro/eli_interpreter_mean                  +0.1198         +0.1205             7             8             7             8
fizzy/parse/micro/factorial_mean                        +0.1289         +0.1289             2             2             2             2
fizzy/parse/micro/fibonacci_mean                        +0.1435         +0.1435             2             2             2             2
fizzy/parse/micro/host_adler32_mean                     +0.0668         +0.0669             3             3             3             3
fizzy/parse/micro/spinner_mean                          +0.0921         +0.0922             2             2             2             2
```

lib/fizzy/instructions.cpp
```diff
@@ -227,11 +227,237 @@ constexpr InstructionMetrics instruction_metrics_table[256] = {
...
+// Order of input parameters is the order they are popped from stack,
+// which is consistent with the order in FuncType::inputs.
+constexpr InstructionType instruction_type_table[256] = {
+    // 5.4.1 Control instructions
+    /* unreachable         = 0x00 */ {{}, {}},
+    /* nop                 = 0x01 */ {{}, {}},
+    /* block               = 0x02 */ {{}, {}},
+    /* loop                = 0x03 */ {{}, {}},
+    /* if_                 = 0x04 */ {{ValType::i32}, {}},
+    /* else_               = 0x05 */ {{}, {}},
```
axic, line 240: Why have these initialisers `{{}, {}},` when empty opcodes can go with `{}` ?

gumb0: It can, I just found it somewhat more explicit to show both input and output empty. Would you prefer to make it `{}`?

axic: It doesn't really make any difference.

lib/fizzy/parser_expr.cpp
```diff
@@ -389,32 +404,32 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end, uint32_t
 
         case Instr::block:
         {
-            std::optional<ValType> type;
-            std::tie(type, pos) = parse_blocktype(pos, end);
+            std::optional<ValType> block_type;
```
axic, line 431: Why do you need this refactoring btw?

gumb0: `type` would shadow the new variable in the outer scope (which I called `types` originally, but @chfast suggested to rename)

lib/fizzy/parser_expr.cpp
```diff
@@ -215,7 +229,8 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end, uint32_t
         std::tie(opcode, pos) = parse_byte(pos, end);
 
         auto& frame = control_stack.top();
-        const auto& metrics = metrics_table[opcode];
+        const auto metrics = metrics_table[opcode];
+        const auto type = type_table[opcode];
```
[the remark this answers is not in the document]
gumb0: This is a mistake I think, will make them references.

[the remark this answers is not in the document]
gumb0: Fixed.

PR #359, opened 2020-05-28
Pass function args by reference using span<>

chfast:
This passes function arguments by reference (instead of by value with `std::vector`). This eliminates one allocation per call because - previously one allocation was done to create the vector, then another to create locals.

Changing `std::vector` to `const std::vector&` may have been enough, but `span` is more generic and easier to refactor with.

The `snap` does not have unit tests as it is trivial and not all methods are currently used after full transition.

```
fizzy/execute/blake2b/512_bytes_rounds_1_mean                     -0.0834         -0.0835            85            78            85            78
fizzy/execute/blake2b/512_bytes_rounds_16_mean                    -0.0758         -0.0758          1262          1166          1262          1166
fizzy/execute/ecpairing/onepoint_mean                             -0.1312         -0.1312        482229        418964        482231        418967
fizzy/execute/keccak256/512_bytes_rounds_1_mean                   -0.0796         -0.0796            97            90            97            90
fizzy/execute/keccak256/512_bytes_rounds_16_mean                  -0.0627         -0.0627          1391          1304          1391          1304
fizzy/execute/memset/256_bytes_mean                               -0.0782         -0.0786             8             7             8             7
fizzy/execute/memset/60000_bytes_mean                             -0.0921         -0.0921          1582          1436          1582          1436
fizzy/execute/mul256_opt0/input0_mean                             -0.1303         -0.1302            28            24            28            24
fizzy/execute/mul256_opt0/input1_mean                             -0.1234         -0.1234            28            24            28            25
fizzy/execute/sha1/512_bytes_rounds_1_mean                        -0.1152         -0.1152            89            79            89            79
fizzy/execute/sha1/512_bytes_rounds_16_mean                       -0.1149         -0.1149          1220          1080          1220          1080
fizzy/execute/sha256/512_bytes_rounds_1_mean                      -0.1615         -0.1615            86            72            86            72
fizzy/execute/sha256/512_bytes_rounds_16_mean                     -0.1501         -0.1501          1152           979          1152           979
fizzy/execute/micro/factorial/10_mean                             -0.0534         -0.0531             1             1             1             1
fizzy/execute/micro/factorial/20_mean                             -0.0798         -0.0801             2             2             2             2
fizzy/execute/micro/fibonacci/24_mean                             -0.0374         -0.0374         14579         14033         14579         14033
fizzy/execute/micro/host_adler32/1_mean                           -0.0425         -0.0435             1             1             1             1
[...]

PR #404, opened 2020-07-06
Span utility

chfast:
Implements some parts of `std::span<>` from C++20.

Taken from #359 and extended for needs in #403.

### DIFF
--- a/lib/fizzy/span.hpp
+++ b/lib/fizzy/span.hpp
@@ -4,6 +4,18 @@
 
 #pragma once
 
+#if __cplusplus > 201703L
+
+#include <span>
+
+namespace fizzy
+{
+template <typename T>
+using span = std::span<T>;
+}
+
+#else
+
 #include <iterator>
 #include <type_traits>
 
@@ -49,3 +61,5 @@ public:
     constexpr reverse_iterator rend() const noexcept { return reverse_iterator{begin()}; }
 };
 }  // namespace fizzy
+
+#endif /* __cplusplus > 201703L */
--- a/lib/fizzy/span.hpp
+++ b/lib/fizzy/span.hpp
@@ -1,0 +1,51 @@
+// Fizzy: A fast WebAssembly interpreter
+// Copyright 2019-2020 The Fizzy Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <iterator>
+#include <type_traits>
+
+namespace fizzy
+{
+/// The span describes an object that can refer to a contiguous sequence of objects with the first
+/// element of the sequence at position zero.
+///
+/// This is minimal implementation of C++20's std::span:
+/// https://en.cppreference.com/w/cpp/container/span
+/// Only `const T` is supported.
+template <typename T, typename = typename std::enable_if_t<std::is_const_v<T>>>
+class span
+{
+    T* const m_begin = nullptr;
+    const std::size_t m_size = 0;
+
+public:
+    using value_type = std::remove_cv_t<T>;
+    using iterator = T*;
+    using reverse_iterator = std::reverse_iterator<iterator>;
+
+    constexpr span() = default;
+    constexpr span(const span&) = default;
+
+    constexpr span(T* begin, std::size_t size) noexcept : m_begin{begin}, m_size{size} {}
+
+    /// Constructor from a container type. Requires the Container to implement std::data() and
+    /// std::size().
+    template <typename Container>
+    constexpr span(const Container& container) noexcept
+      : m_begin{std::data(container)}, m_size{std::size(container)}
+    {}
+
+    constexpr T& operator[](std::size_t index) const noexcept { return m_begin[index]; }
+
+    [[nodiscard]] constexpr std::size_t size() const noexcept { return m_size; }
+
+    constexpr iterator begin() const noexcept { return m_begin; }
+    constexpr iterator end() const noexcept { return m_begin + m_size; }
+
+    constexpr reverse_iterator rbegin() const noexcept { return reverse_iterator{end()}; }
+    constexpr reverse_iterator rend() const noexcept { return reverse_iterator{begin()}; }
+};
+}  // namespace fizzy
--- a/test/unittests/span_test.cpp
+++ b/test/unittests/span_test.cpp
@@ -1,0 +1,142 @@
+// Fizzy: A fast WebAssembly interpreter
+// Copyright 2020 The Fizzy Authors.
+// SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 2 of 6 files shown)


lib/fizzy/span.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+public:
+    using value_type = std::remove_cv_t<T>;
+    using iterator = T*;
+    using reverse_iterator = std::reverse_iterator<iterator>;
+
+    constexpr span() noexcept = default;
+    constexpr span(const span&) = default;
+
+    constexpr span(const T* begin, std::size_t size) noexcept : m_begin{begin}, m_size{size} {}
```
gumb0, line 32: if `T` is already `const`, is `const` needed here or how is it allowed at all?
```suggestion
    constexpr span(T* begin, std::size_t size) noexcept : m_begin{begin}, m_size{size} {}
```

lib/fizzy/span.hpp
```diff
@@ -0,0 +1,55 @@
...
+/// The span describes an object that can refer to a contiguous sequence of objects with the first
+/// element of the sequence at position zero.
+///
+/// This is minimal implementation of C++20's std::span:
+/// https://en.cppreference.com/w/cpp/container/span
+/// Only `const T` is supported.
+template <typename T, typename = typename std::enable_if_t<std::is_const_v<T>>>
+class span
+{
+    T* const m_begin = nullptr;
```
gumb0, line 21: Did you remove first `const` here because `T` is already `const`?

lib/fizzy/span.hpp
```diff
@@ -0,0 +1,55 @@
...
+    const std::size_t m_size = 0;
+
+public:
+    using value_type = std::remove_cv_t<T>;
+    using iterator = T*;
+    using reverse_iterator = std::reverse_iterator<iterator>;
+
+    constexpr span() noexcept = default;
+    constexpr span(const span&) = default;
```
gumb0, line 30: `noexcept`?

chfast: I checked this and `noexcept` is redundant. The `= default` is `noexcept` if it can be.

lib/fizzy/span.hpp
chfast: Yes, but I'm trying to require `T` to be `const X` only by the `enable_if`. The C++20 can handle also mutable spans.

I will do one more round of constness correction. And some `constexpr` seem to be missing. But the goal is the have signatures of all methods be exactly as in `std::span`. E.g. `constexpr reverse_iterator rbegin() const noexcept;`.

lib/fizzy/span.hpp
gumb0: And for default constructor, too?

chfast: Ah, I see this is used inconsistently... I will remove the first one.

test/unittests/span_test.cpp
```diff
@@ -0,0 +1,102 @@
...
+    EXPECT_EQ(s[1], 2);
+    EXPECT_EQ(s[2], 3);
+
+    // For range loop also works.
+    uint64_t i = 0;
+    for (const auto& x : s)
+        EXPECT_EQ(x, ++i);
+}
+
+TEST(span, reverse_iterator)
```
axic, line 85: There doesn't seem to be a specific iterator test, just the range loop in the initialiser list.

chfast: True, but you cannot use for range loop with reverse iterator before C++20 where you can do `for (auto x : range::view::reverse(s)`.

axic: I mean would it be possible to have test cases for forward iterator like for the reverse iterator?

lib/fizzy/span.hpp
gumb0: Here I mean it looks at least inconsistent with the `m_begin` definition above, it should be either both `T*` or both `const T*`.
But not sure if it would somehow break anything?

lib/fizzy/span.hpp
```diff
@@ -0,0 +1,55 @@
...
+namespace fizzy
+{
+/// The span describes an object that can refer to a contiguous sequence of objects with the first
+/// element of the sequence at position zero.
+///
+/// This is minimal implementation of C++20's std::span:
+/// https://en.cppreference.com/w/cpp/container/span
+/// Only `const T` is supported.
+template <typename T, typename = typename std::enable_if_t<std::is_const_v<T>>>
+class span
```
axic, line 19: Is it possible to fall back to native const if compiling on C++20 or it is not worth the hassle (it is not any faster/better than this below)?

chfast: I'm not sure I understand the question.

This is compatible with C++20 and you can just start using `std::span<const uint64_t>`. The C++20 also supports the `std::span<uint64_t>`.

I disabled the usage of the later, because we don't need it. And also because of the implicit conversions involved. E.g. conversion from `const std::vector<uint64_t>&` to `span<uint64_t>` should be disallowed. I prefer not to handle this in the implementation as long as we don't need it.

If by native const you mean `const span<uint64_t>` it is not useful as an argument type (similar to `const int` argument type).
I think about span as a reference type similar to pointers:
- `span<int>` is as `int*`,
- `span<const int>` is as `const int*`,
- `const span<int>` is as `int* const`,
- `const span<const int>` is as `const int* const`.

axic: > This is compatible with C++20 and you can just start using std::span<const uint64_t>. The C++20 also supports the std::span<uint64_t>.

The question was can this header be done in a way such that any code using it will either use the native C++20 implementation or this fallback, depending on compiler settings?

lib/fizzy/span.hpp
chfast: Oh yes, fixed.

lib/fizzy/span.hpp
chfast: Yes, this can be done. I can do it here after #405.

lib/fizzy/span.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+    constexpr iterator begin() const noexcept { return m_begin; }
+    constexpr iterator end() const noexcept { return m_begin + m_size; }
+
+    constexpr reverse_iterator rbegin() const noexcept { return reverse_iterator{end()}; }
+    constexpr reverse_iterator rend() const noexcept { return reverse_iterator{begin()}; }
+};
+}  // namespace fizzy
+
+#endif
```
axic, line 69:
```suggestion
#endif /* __cplusplus > 201703L */
```

Helps for readability, but not important, given this file is small.

PR #407, opened 2020-07-06
Validate block result types

gumb0:
Depends on #403


gumb0:
There's an option to simplify `update_result_stack` reusing `drop_operand`, but the error messages won't be specific to blocks. Pushed it as a last commit, what do you think?

gumb0:
Rebased.

PR #376, opened 2020-06-08
Use only uint32_t to immediates items

chfast:
```
fizzy/execute/blake2b/512_bytes_rounds_1_mean                     -0.0186         -0.0185            73            71            73            71
fizzy/execute/blake2b/512_bytes_rounds_16_mean                    -0.0174         -0.0174          1082          1064          1082          1064
fizzy/execute/ecpairing/onepoint_mean                             +0.0151         +0.0151        394324        400277        394326        400280
fizzy/execute/keccak256/512_bytes_rounds_1_mean                   +0.1255         +0.1255            84            94            84            94
fizzy/execute/keccak256/512_bytes_rounds_16_mean                  +0.1382         +0.1382          1197          1363          1197          1363
fizzy/execute/memset/256_bytes_mean                               -0.0338         -0.0335             6             6             6             6
fizzy/execute/memset/60000_bytes_mean                             -0.0439         -0.0439          1304          1247          1304          1247
fizzy/execute/mul256_opt0/input0_mean                             -0.0412         -0.0410            25            24            25            24
fizzy/execute/mul256_opt0/input1_mean                             -0.0397         -0.0395            25            24            25            24
fizzy/execute/sha1/512_bytes_rounds_1_mean                        -0.0118         -0.0117            80            79            80            79
fizzy/execute/sha1/512_bytes_rounds_16_mean                       -0.0133         -0.0133          1100          1085          1100          1086
fizzy/execute/sha256/512_bytes_rounds_1_mean                      -0.0022         -0.0021            75            75            75            75
fizzy/execute/sha256/512_bytes_rounds_16_mean                     -0.0022         -0.0022          1022          1020          1022          1020
fizzy/execute/micro/eli_interpreter/halt_mean                     -0.0021         -0.0019             0             0             0             0
fizzy/execute/micro/eli_interpreter/exec105_mean                  -0.0118         -0.0120             4             4             4             4
fizzy/execute/micro/factorial/10_mean                             -0.0035         -0.0030             1             1             1             1
fizzy/execute/micro/factorial/20_mean                             -0.0069         -0.0069             2             2             2             2
fizzy/execute/micro/fibonacci/24_mean                             -0.0059         -0.0059          9108          9054          9108          9054
fizzy/execute/micro/host_adler32/1_mean                           -0.0027         -0.0023             1             1             1             1
fizzy/execute/micro/host_adler32/100_mean                         +0.0058         +0.0060             6             6             6             6
[...]

PR #408, opened 2020-07-06
Validate types at branch instructions

gumb0:


axic:
@gumb0 rebase please

lib/fizzy/parser_expr.cpp
```diff
@@ -155,34 +155,27 @@ inline uint8_t get_branch_arity(const ControlFrame& frame) noexcept
...
 
-
 void push_branch_immediates(const ControlFrame& branch_frame, int stack_height, bytes& immediates)
 {
-    const auto arity = get_branch_arity(branch_frame);
-
     // How many stack items to drop when taking the branch.
-    // This value can be negative for unreachable instructions.
-    const auto stack_drop = stack_height - branch_frame.parent_stack_height - arity;
+    const auto stack_drop = stack_height - branch_frame.parent_stack_height;
```
gumb0, line 171: It's not needed to subtract arity now, because now we pop the result from stack before calling this function (in `update_branch_stack`)

lib/fizzy/parser_expr.cpp
```diff
@@ -557,6 +550,14 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end, FuncIdx f
 
             if (instr == Instr::br)
                 mark_frame_unreachable(frame, operand_stack);
+            else
+            {
+                // For the case when branch is not taken for br_if,
```
gumb0, line 555: Validation Algorithm does something similar:

image

test/unittests/validation_test.cpp
```diff
@@ -455,7 +455,7 @@ TEST(validation, br_table_invalid_type)
     (func (param $x i32)
       (block $a (result i32)
         (block $b (result i64)
-          i32.const 1
+          i64.const 1
```
gumb0, line 458: This fix is needed because now it fist checks the return type on the stack against default label type before iterating over all labels and checking there consistency.
(Without the fix it's "type mismatch" error instead of "inconsistent labels")

gumb0:
Rebased.

lib/fizzy/parser_expr.cpp
```diff
@@ -157,34 +157,27 @@ inline uint8_t get_branch_arity(const ControlFrame& frame) noexcept
...
-    const auto stack_drop = stack_height - branch_frame.parent_stack_height - arity;
+    const auto stack_drop = stack_height - branch_frame.parent_stack_height;
 
     // Push frame start location as br immediates - these are final if frame is loop,
     // but for block/if/else these are just placeholders, to be filled at end instruction.
     push(immediates, static_cast<uint32_t>(branch_frame.code_offset));
     push(immediates, static_cast<uint32_t>(branch_frame.immediates_offset));
     push(immediates, static_cast<uint32_t>(stack_drop));
-    push(immediates, arity);
+    push(immediates, get_branch_arity(branch_frame));
```
axic, line 180: What is the return type of `get_branch_arity`?

gumb0: It's `uint8_t`

axic: I know the code prior was not clear too (with using auto), but can you leave a comment here stating it is `uint8_t`?

We have a separate PR which tries to change the types to unified `uint32_t`, but that showed slowdown: #376

gumb0: Added comment.

PR #410, opened 2020-07-07
Spectests code coverage

### LINKED ISSUE #406: CI: Get spectests coverage report

chfast:
This creates coverage report (same as for unittests) out of spectests execution. It is available as CircleCI artifact.

The codecov report can be split into two parts, but I'm leaving that for other time.

Closes #406

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -233,11 +233,10 @@ jobs:
       - build
       - test
 
-  coverage-tidy:
+  coverage:
     executor: linux-clang-latest
     environment:
       BUILD_TYPE: Coverage
-      CMAKE_OPTIONS: -DCMAKE_CXX_CLANG_TIDY=clang-tidy
     steps:
       - checkout
       - build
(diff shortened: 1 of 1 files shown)


chfast:
First spectests coverage report: https://21739-224892340-gh.circle-artifacts.com/0/coverage/index.html

circle.yml
```diff
@@ -86,6 +86,29 @@ commands:
         working_directory: ~/build
         command: ctest -E 'unittests|smoketests' -j4 --schedule-random --output-on-failure
 
+  collect_coverage_data:
```
gumb0, line 89: Other commands use hyphens for delimiters
```suggestion
    collect-coverage-data:
```

gumb0:
> The codecov report can be split into two parts, but I'm leaving that for other time.

Do you mean splitting into collecting coverage and uploading to codecov.io? If so, maybe it's better to keep `--required` till then, otherwise if it fails to upload, it will not be visible that previousle uploaded report was not updated?

chfast:
> Do you mean splitting into collecting coverage and uploading to codecov.io? If so, maybe it's better to keep `--required` till then, otherwise if it fails to upload, it will not be visible that previousle uploaded report was not updated?

I can keep it but in case of upload failure spectests will not run at all.

By codecov report I mean you can do two uploads - for unittests and spectests with different flags. Then you can filter by flags and have different stats. https://docs.codecov.io/docs/flags

circle.yml
chfast: I'm following CircleCI convention for command names, e.g. `store_artifacts`.

gumb0: ok, that were job names that use hyphens

gumb0:
> By codecov report I mean you can do two uploads - for unittests and spectests with different flags. Then you can filter by flags and have different stats. https://docs.codecov.io/docs/flags

This might be a fine plan, but maybe we don't need to collect/upload spectest coverage for each PR? Maybe it could be manually triggered like benchmarks.

chfast:
> This might be a fine plan, but maybe we don't need to collect/upload spectest coverage for each PR? Maybe it could be manually triggered like benchmarks.

It is better to keep it simple and collect every time.

PR #413, opened 2020-07-07
Validate types for select instruction

gumb0:
Depends on #408 

Some of related spec tests were already passing by accident, because the tests are not great:

https://github.com/WebAssembly/spec/blob/636b862b9c8a25ad65fb240fefd673e7f23bcdd0/test/core/select.wast#L296-L307
```wat
(assert_invalid
  (module (func $type-num-vs-num (select (i32.const 1) (i64.const 1) (i32.const 1))))
  "type mismatch"
)
(assert_invalid
  (module (func $type-num-vs-num (select (i32.const 1) (f32.const 1.0) (i32.const 1))))
  "type mismatch"
)
(assert_invalid
  (module (func $type-num-vs-num (select (i32.const 1) (f64.const 1.0) (i32.const 1))))
  "type mismatch"
)
```
^ These were invalid but for the wrong reason - functions are void but have result of `select` on stack in the end.

https://github.com/WebAssembly/spec/blob/636b862b9c8a25ad65fb240fefd673e7f23bcdd0/test/core/select.wast#L326-L333
```wat
(assert_invalid
  (module
    (func $type-3rd-operand-empty
      (i32.const 0) (i32.const 0) (select) (drop)
    )
  )
  "type mismatch"
)
```
^ This was also invalid for the wrong reason - `select` popped two values and pushed 0, so it was type mismatch for `drop` instruction.

(added to "To Upstream")

### DIFF
--- a/lib/fizzy/parser_expr.cpp
+++ b/lib/fizzy/parser_expr.cpp
@@ -227,6 +227,11 @@ inline void push_operand(Stack<OperandStackType>& operand_stack, ValType type)
     operand_stack.push(from_valtype(type));
 }
 
+inline void push_operand(Stack<OperandStackType>& operand_stack, OperandStackType type)
+{
+    operand_stack.push(type);
+}
+
 ValType find_local_type(
     const std::vector<ValType>& params, const std::vector<Locals>& locals, LocalIdx idx)
 {
@@ -376,10 +381,28 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end, FuncIdx f
             break;
 
         case Instr::drop:
-        case Instr::select:
-            // TODO: for select validate that two top operands are the same type
             drop_operand(frame, operand_stack, OperandStackType::Unknown);
             break;
+        case Instr::select:
+        {
+            const auto frame_stack_height = static_cast<int>(operand_stack.size());
+
+            // Two operands are expected, because the selector operand was already popped
+            // according to instruction type table
+            if (!frame.unreachable && frame_stack_height < frame.parent_stack_height + 2)
+                throw validation_error{"stack underflow"};
+
+
+            const auto operand_type = frame_stack_height > frame.parent_stack_height ?
+                                          operand_stack[0] :
+                                          OperandStackType::Unknown;
+
+            drop_operand(frame, operand_stack, operand_type);
+            drop_operand(frame, operand_stack, operand_type);
+            push_operand(operand_stack, operand_type);
+
+            break;
+        }
 
         case Instr::nop:
         case Instr::i32_eq:
--- a/lib/fizzy/parser_expr.cpp
+++ b/lib/fizzy/parser_expr.cpp
@@ -63,6 +63,36 @@ struct ControlFrame
     {}
 };
 
+enum class OperandStackType : uint8_t
+{
+    Unknown = 0,
+    i32 = static_cast<uint8_t>(ValType::i32),
+    i64 = static_cast<uint8_t>(ValType::i64),
+    f32 = static_cast<uint8_t>(ValType::f32),
+    f64 = static_cast<uint8_t>(ValType::f64),
+};
+
+inline OperandStackType from_valtype(ValType val_type) noexcept
+{
+    return static_cast<OperandStackType>(val_type);
+}
+
+inline bool type_matches(OperandStackType actual_type, ValType expected_type) noexcept
+{
+    if (actual_type == OperandStackType::Unknown)
+        return true;
+
+    return static_cast<ValType>(actual_type) == expected_type;
+}
+
+inline bool type_matches(OperandStackType actual_type, OperandStackType expected_type) noexcept
+{
+    if (expected_type == OperandStackType::Unknown || actual_type == OperandStackType::Unknown)
+        return true;
+
+    return expected_type == actual_type;
+}
+
 /// Parses blocktype.
 ///
 /// Spec: https://webassembly.github.io/spec/core/binary/types.html#binary-blocktype.
@@ -82,7 +112,7 @@ parser_result<std::optional<ValType>> parse_blocktype(const uint8_t* pos, const 
     return {validate_valtype(type), pos};
 }
 
-void update_operand_stack(const ControlFrame& frame, Stack<ValType>& operand_stack,
+void update_operand_stack(const ControlFrame& frame, Stack<OperandStackType>& operand_stack,
     span<const ValType> inputs, span<const ValType> outputs)
 {
     const auto frame_stack_height = static_cast<int>(operand_stack.size());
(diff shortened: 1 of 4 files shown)


lib/fizzy/parser_expr.cpp
```diff
@@ -63,6 +63,36 @@ struct ControlFrame
     {}
 };
 
+enum class OperandStackType : uint8_t
+{
+    Unknown,
```
chfast, line 68:
```suggestion
    Unknown = 0,
```

lib/fizzy/parser_expr.cpp
```diff
@@ -63,6 +63,36 @@ struct ControlFrame
...
+    f32 = static_cast<uint8_t>(ValType::f32),
+    f64 = static_cast<uint8_t>(ValType::f64),
+};
+
+inline OperandStackType from_valtype(ValType valType) noexcept
+{
+    return static_cast<OperandStackType>(valType);
+}
+
+inline bool type_matches(OperandStackType actual_type, ValType expected_type) noexcept
```
chfast, line 80:
```suggestion
inline bool operator==(OperandStackType actual_type, ValType expected_type) noexcept
```
?

lib/fizzy/parser_expr.cpp
```diff
@@ -63,6 +63,36 @@ struct ControlFrame
...
+{
+    if (actual_type == OperandStackType::Unknown)
+        return true;
+
+    return static_cast<ValType>(actual_type) == expected_type;
+}
+
+inline bool type_matches(OperandStackType actual_type, OperandStackType expected_type) noexcept
```
chfast, line 88:
```suggestion
inline bool operator==(OperandStackType actual_type, OperandStackType expected_type) noexcept
```
?

lib/fizzy/parser_expr.cpp
```diff
@@ -180,13 +215,19 @@ void push_branch_immediates(const ControlFrame& branch_frame, int stack_height,
...
-inline void mark_frame_unreachable(ControlFrame& frame, Stack<ValType>& operand_stack) noexcept
+inline void mark_frame_unreachable(
+    ControlFrame& frame, Stack<OperandStackType>& operand_stack) noexcept
 {
     frame.unreachable = true;
     operand_stack.shrink(static_cast<size_t>(frame.parent_stack_height));
 }
 
-inline void push_operand(Stack<ValType>& operand_stack, ValType type)
+inline void push_operand(Stack<OperandStackType>& operand_stack, ValType type)
```
chfast, line 225: Because of this overload, it may be better to just allow implicit conversion `ValType` -> `OperandStackType`.

lib/fizzy/parser_expr.cpp
gumb0: I didn't go with operators, because then we should probably also have `operator!=` for completeness, and for all pairs `OperansStackType`-`OperandStackType`, `OperandStackType`-`ValType`, `ValType`-`OperandStackType`, and unit tests for all of them...

Also it doesn't seem exacly like equivalence-kind of relation.

lib/fizzy/parser_expr.cpp
gumb0: How would you do that exactly? I can imagine only making `OperandStackType` a class, containing another enum type, and implicit conversion constructor.

lib/fizzy/parser_expr.cpp
chfast: I have similar doubts.

lib/fizzy/parser_expr.cpp
chfast: Yes, it looks like. Not sure this is worth it then.

gumb0: Yeah, it looks that it would be quite verbose then.

lib/fizzy/parser_expr.cpp
```diff
@@ -63,6 +63,36 @@ struct ControlFrame
     {}
 };
 
+enum class OperandStackType : uint8_t
+{
+    Unknown = 0,
+    i32 = static_cast<uint8_t>(ValType::i32),
```
axic, line 69: `enum class ValType : uint8_t` so why is there a need for typecasting here?

axic: Couldn't this just extend `ValType`?

lib/fizzy/parser_expr.cpp
```diff
@@ -103,16 +133,16 @@ void update_operand_stack(const ControlFrame& frame, Stack<ValType>& operand_sta
...
     // Push output values even if frame is unreachable.
     for (const auto output_type : outputs)
-        operand_stack.push(output_type);
+        operand_stack.push(from_valtype(output_type));
 }
 
-inline void drop_operand(
-    const ControlFrame& frame, Stack<ValType>& operand_stack, std::optional<ValType> expected_type)
+inline void drop_operand(const ControlFrame& frame, Stack<OperandStackType>& operand_stack,
+    OperandStackType expected_type)
```
axic, line 145: Wouldn't it be actually easier codewise to keep using `optional<ValType>`, not introducing `OperandStackType`, and comparing that way?

lib/fizzy/parser_expr.cpp
gumb0: Because `ValType` is still a different type from `uint8_t`, enum classes are strongly typed unlike plain enums.

lib/fizzy/parser_expr.cpp
gumb0: I did it, too, at some point. These two variants are quite close, but enum one is perhaps a bit more concise and closer to Validation Algorithm.

Also enum way guarantees stack operands to be 1 byte each, without "has_value" flag overhead in `optional`.

(Also "Unknown" maybe better expresses the notion of "value with unknown/opaque type" than empty `optional`, which is rather supposed to mean `no value/no type`)

lib/fizzy/parser_expr.cpp
gumb0: > Couldn't this just extend `ValType`?

No, not in C++

lib/fizzy/parser_expr.cpp
axic: We still use a mixture of `optional<ValType>` and `OperandStackType`. Perhaps we should aim to only use `OperandStackType` where possible?

gumb0: I don't think so, the remaining uses of `optional<ValType>` are all for block types, where empty one really means "no result value" (and so for example at jumps this means nothing is kept on stack; keeping `Unknown` would be incorrect)

lib/fizzy/parser_expr.cpp
```diff
@@ -340,10 +381,28 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end, FuncIdx f
...
+
+            // Two operands are expected, because the selector operand was already popped
+            // according to instruction type table
+            if (!frame.unreachable && frame_stack_height < frame.parent_stack_height + 2)
+                throw validation_error{"stack underflow"};
+
+
+            const auto operand_type = frame_stack_height > frame.parent_stack_height ?
+                                          operand_stack[0] :
+                                          OperandStackType::Unknown;
```
axic, line 398: So we get unknown in case of unreachable?

gumb0: Yes, if unreachable and nothing is pushed before, then `select` doesn't know types and will push `Unknown` in the end.

chfast: This comes from "polymorphic stack" concept in wasm. Even if unreachable, some information about instructions is validated.

PR #414, opened 2020-07-07
Remove instruction metrics (except max align)

gumb0:
Depends on #403


lib/fizzy/parser_expr.cpp
```diff
@@ -248,7 +248,6 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end, FuncIdx f
         std::tie(opcode, pos) = parse_byte(pos, end);
 
         auto& frame = control_stack.top();
-        const auto& metrics = metrics_table[opcode];
         const auto& type = type_table[opcode];
```
axic, line 251: I'd rather have `max_align` here.

gumb0: Why? It's used only in subset of instructions, which are handled all in one switch case, so why put it into wider scope?

axic: My reasoning is that none of the cases in the switch use the `opcode` variable and one must go up to the top to find it.

gumb0: Tried to benchmark, it doesn't make any difference.

axic: If there's no speed difference I'd prefer it at the top.

gumb0: Made `max_align` variable.

PR #415, opened 2020-07-08
Add constexpr_vector class

gumb0:
Required for #403

### DIFF
--- a/lib/fizzy/constexpr_vector.hpp
+++ b/lib/fizzy/constexpr_vector.hpp
@@ -1,0 +1,35 @@
+// Fizzy: A fast WebAssembly interpreter
+// Copyright 2019-2020 The Fizzy Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <utility>
+
+namespace fizzy
+{
+/// Constexpr vector is used to represent compile time constant value lists.
+/// It allows to have a collection (constexpr array) of such lists, where list lengths vary across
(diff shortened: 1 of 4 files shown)


lib/fizzy/constexpr_vector.hpp
```diff
@@ -0,0 +1,35 @@
...
+
+#include <utility>
+
+namespace fizzy
+{
+/// Constexpr vector is used to represent compile time constant value lists.
+/// It allows to have a collection (constexpr array) of such lists, where list lengths vary across
+/// the collection, but each length is still known at compile time (and doesn't exceed MaxSize
+/// limit).
+template <typename T, std::size_t MaxSize>
```
chfast, line 15:
```suggestion
template <typename T, std::size_t MaxCapacity>
```
Or just `Capacity`.

lib/fizzy/constexpr_vector.hpp
```diff
@@ -0,0 +1,35 @@
...
+class constexpr_vector
```
chfast, line 16: Can be also named `static_vector` as in https://github.com/gnzlbg/static_vector#3-existing-practice.
Also found a precedence of `constexpr_vector`: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0597r0.html

lib/fizzy/constexpr_vector.hpp
gumb0: Renamed to `Capacity`

PR #417, opened 2020-07-13
Validate missing else branch

gumb0:
Depends on #407

### DIFF
--- a/lib/fizzy/parser_expr.cpp
+++ b/lib/fizzy/parser_expr.cpp
@@ -41,8 +41,7 @@ struct ControlFrame
     const size_t code_offset{0};
 
     /// The immediates offset for block instructions.
-    /// Non-const because else block changes it (to resolve jumping over else block).
-    size_t immediates_offset{0};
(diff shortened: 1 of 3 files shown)


axic:
@gumb0 rebase plase

lib/fizzy/parser_expr.cpp
```diff
@@ -468,10 +467,15 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end, FuncIdx f
 
             update_result_stack(frame, operand_stack);  // else is the end of if.
 
-            // Reset frame after if.
-            frame.unreachable = false;
```
gumb0, line 472: I decided here at else instruction to pop the frame and push a new one instead of overwriting fields in the top one.

This way we can keep both `instruction` and `immediates_offset` fields `const` in `ControlFrame`, and it just seems more logical to push a new one, because it's different frame.

But this requires to transfer the contents of `br_immediate_offsets` from the old top frame to the new one.

PR #418, opened 2020-07-15
Suggestions made by CLion's analyzer

axic:


lib/fizzy/constexpr_vector.hpp
```diff
@@ -26,7 +26,7 @@ class constexpr_vector
     {}
 
     constexpr const T* data() const noexcept { return m_array; }
```
axic, line 28: Shouldn't we just mark all getters as nodiscard?

chfast: It is better to remove any current usage of `[[nodiscard]]`. This is just noise.

axic: Why? It supposed to warn if we don't use the result of those?

chfast: Yes. But this may be overpedantic. Even in C++20 they only marked `empty()` functions with it so far.

.clang-format
```diff
@@ -1,6 +1,6 @@
 ---
 Language:     Cpp
-Standard:     c++17
+Standard:     Cpp11
```
axic, line 3: Dammit, need to remove this from the commit.

PR #421, opened 2020-07-15
Rename instructions.{hpp,cpp} -> instruction_metrics.{hpp,cpp}

chfast:
This will make space for moving instructions implementations to
instructions.hpp file.


axic:
> This will make space for moving instructions implementations to
instructions.hpp file.

Why is that better than `execute.cpp`?

chfast:
> > This will make space for moving instructions implementations to
> > instructions.hpp file.
> 
> Why is that better than `execute.cpp`?

The `execute.cpp` is getting big.

PR #419, opened 2020-07-15
Introduce Value type for operands

chfast:
The `Value` is an union for now containing only `i64` value. Later to be extended with `f32`, `f64` and maybe `i32`.

I added implicit conversions from/to `uint64_t` to limit the number of required changes.

The conversion from `std::initializer_list<uint64_t>` not possible in a type safe manner. I guess we will need something like `template<typename T...> TypedArgs` to construct a container of `Value`s. This will also handle usage of different value types in execution invocation.

All will get more complicated if we want constructor from other types (e.g. `float`, `unsigned int`) as we will get ambiguities (e.g. `0` is of type `int` and will not know where to go).

PR #423, opened 2020-07-15
Value type improvements

chfast:
Depends on https://github.com/wasmx/fizzy/pull/419

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -1014,7 +1014,7 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value> 
         case Instr::i32_eqz:
         {
             const auto value = static_cast<uint32_t>(stack.pop());
-            stack.push(value == 0);
+            stack.push(uint32_t{value == 0});
             break;
         }
         case Instr::i32_eq:
@@ -1069,7 +1069,7 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value> 
         }
         case Instr::i64_eqz:
         {
-            stack.push(stack.pop() == 0);
+            stack.push(uint32_t{stack.pop() == 0});
             break;
         }
         case Instr::i64_eq:
@@ -1463,7 +1463,7 @@ end:
     if (trap)
         return Trap;
 
-    return stack.size() != 0 ? stack.pop() : Void;
+    return stack.size() != 0 ? ExecutionResult{stack.pop()} : Void;
 }
 
 std::vector<ExternalFunction> resolve_imported_functions(
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -359,17 +359,19 @@ inline bool store_into_memory(
 template <typename Op>
 inline void unary_op(OperandStack& stack, Op op) noexcept
 {
-    using T = decltype(op(stack.top()));
-    stack.top() = op(static_cast<T>(stack.top()));
+    using T = decltype(op({}));
+    const auto result = op(stack.top().as<T>());
+    stack.top() = Value{result};  // Convert to Value, also from signed integer types.
 }
 
 template <typename Op>
 inline void binary_op(OperandStack& stack, Op op) noexcept
 {
-    using T = decltype(op(stack.top(), stack.top()));
-    const auto val2 = static_cast<T>(stack.pop());
-    const auto val1 = static_cast<T>(stack.top());
-    stack.top() = static_cast<std::make_unsigned_t<T>>(op(val1, val2));
+    using T = decltype(op({}, {}));
+    const auto val2 = stack.pop().as<T>();
+    const auto val1 = stack.top().as<T>();
+    const auto result = op(val1, val2);
+    stack.top() = Value{result};  // Convert to Value, also from signed integer types.
 }
 
 template <typename T, template <typename> class Op>
--- a/lib/fizzy/value.hpp
+++ b/lib/fizzy/value.hpp
@@ -30,5 +30,34 @@ union Value
     constexpr Value(T) = delete;
 
     constexpr operator uint64_t() const noexcept { return i64; }
+
+    /// Get the value as the given type. Handy in templates.
+    /// Only required specializations are provided.
+    template <typename T>
+    constexpr T as() const noexcept;
 };
+
+template <>
+constexpr uint64_t Value::as<uint64_t>() const noexcept
+{
+    return i64;
+}
+
+template <>
+constexpr uint32_t Value::as<uint32_t>() const noexcept
+{
+    return static_cast<uint32_t>(i64);
+}
+
+template <>
+constexpr int64_t Value::as<int64_t>() const noexcept
+{
+    return static_cast<int64_t>(i64);
+}
+
+template <>
+constexpr int32_t Value::as<int32_t>() const noexcept
+{
+    return static_cast<int32_t>(i64);
+}
 }  // namespace fizzy
--- a/lib/fizzy/value.hpp
+++ b/lib/fizzy/value.hpp
@@ -13,7 +13,21 @@ union Value
     uint64_t i64;
 
     Value() = default;
-    constexpr Value(uint64_t v) noexcept : i64{v} {}
+
+    /// Converting constructors from integer types
+    ///
+    /// We need to support {signed,unsigned} x {32,64} integers. However, due to uint64_t being
+    /// defined differently in different implementations we need to avoid the alias and provide
+    /// constructors for unsigned long and unsigned long long independently.
+    constexpr Value(unsigned int v) noexcept : i64{v} {}
+    constexpr Value(unsigned long v) noexcept : i64{v} {}
+    constexpr Value(unsigned long long v) noexcept : i64{v} {}
+    constexpr Value(int64_t v) noexcept : i64{static_cast<uint64_t>(v)} {}
+    constexpr Value(int32_t v) noexcept : i64{static_cast<uint32_t>(v)} {}
+
+    /// Converting constructor from any other type (including smaller integer types) is deleted.
+    template <typename T>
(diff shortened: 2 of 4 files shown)


gumb0:
macOS builds failed

chfast:
> macOS builds failed

It looks on macOS the `uint64_t` is `unsigned long long` so we need more overloads.

lib/fizzy/execute.cpp
```diff
@@ -359,17 +359,17 @@ inline bool store_into_memory(
...
+    stack.top() = op(stack.top().as<T>());
 }
 
 template <typename Op>
 inline void binary_op(OperandStack& stack, Op op) noexcept
 {
-    using T = decltype(op(stack.top(), stack.top()));
-    const auto val2 = static_cast<T>(stack.pop());
-    const auto val1 = static_cast<T>(stack.top());
-    stack.top() = static_cast<std::make_unsigned_t<T>>(op(val1, val2));
```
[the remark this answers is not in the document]
gumb0: Ah I think now constructor `Value(int32_t v)` does this.

lib/fizzy/value.hpp
```diff
@@ -13,8 +13,19 @@ union Value
     uint64_t i64;
 
     Value() = default;
-    constexpr Value(uint64_t v) noexcept : i64{v} {}
+    constexpr Value(unsigned int v) noexcept : i64{v} {}
```
gumb0, line 16: Maybe a comment why we need these `unsigned` constuctors

chfast: Added.

chfast: Instead of having 5 constructor for ints + deleted constructors, I can do all of that with a template constructor with some enable_if and some other magic. But I'm not sure we want that now. Maybe it's better to wait for floats before more cleanups.

test/utils/fizzy_engine.cpp
```diff
@@ -135,8 +135,8 @@ WasmEngine::Result FizzyEngine::execute(
 {
     static_assert(sizeof(uint64_t) == sizeof(Value));
     const auto first_arg = reinterpret_cast<const Value*>(args.data());
-    const auto status =
-        fizzy::execute(*m_instance, static_cast<uint32_t>(func_ref), {first_arg, args.size()});
+    const auto status = fizzy::execute(
+        *m_instance, static_cast<uint32_t>(func_ref), span<const Value>(first_arg, args.size()));
```
gumb0, line 139: Is this change required?

chfast: Yes. Otherwise compiler thinks these are two values: `first_arg` -> `bool` -> `Value` and `args.size()` -> `size_t` -> `Value`.

gumb0: I see. Fo first_arg it's probably pointer to integer conversion, rather than `bool`, I think `bool` -> `Value` doesn't work anymore, you deleted the tests for it.

lib/fizzy/execute.cpp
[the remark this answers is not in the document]
axic: Is it better to have it in the constructor or make it more explicit here? From a reader's perspective explicitness here is nice.

Do we recall the test case while it had to be converted between signed/unsigned in this function?

lib/fizzy/execute.cpp
```diff
@@ -1014,7 +1014,7 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value>
         case Instr::i32_eqz:
         {
             const auto value = static_cast<uint32_t>(stack.pop());
-            stack.push(value == 0);
+            stack.push(uint32_t{value == 0});
```
axic, line 1017: Why do we need to make this clear? I know eqz returns 32-bit but we said we don't care because it is a union.

axic: Does this have any effect on speed?

chfast: The `Value(bool)` is deleted to avoid some surprises. So the bool expression must be converted to `uint32_t` (or `uint64_t`, but `uint32_t` seems nicer).

test/utils/fizzy_engine.cpp
chfast: Without the change it resolves overloading to `Value(bool)` what gives the compilation error (as this constructor is deleted).

lib/fizzy/execute.cpp
chfast: > Does this have any effect on speed?

Nope.

lib/fizzy/execute.cpp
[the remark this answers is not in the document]
chfast: All "signed integer" instructions relay on this.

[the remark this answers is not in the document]
gumb0: I think I like it better when `Value` itself encapsulates correct conversion `int32_t`->`Value`.

`make_unsigned_t` here looked to me more like a hack to handle an edge case...

chfast:
Cleaned up the git history.

chfast:
Benchmarks (no change):
```
fizzy/execute/blake2b/512_bytes_rounds_1_mean                     +0.0011         +0.0011            87            87            87            87
fizzy/execute/blake2b/512_bytes_rounds_16_mean                    +0.0006         +0.0006          1310          1311          1310          1311
fizzy/execute/ecpairing/onepoint_mean                             +0.0039         +0.0039        429238        430925        429239        430929
fizzy/execute/keccak256/512_bytes_rounds_1_mean                   -0.0056         -0.0056           105           104           105           104
fizzy/execute/keccak256/512_bytes_rounds_16_mean                  -0.0089         -0.0089          1530          1517          1530          1517
fizzy/execute/memset/256_bytes_mean                               -0.0033         -0.0033             7             7             7             7
fizzy/execute/memset/60000_bytes_mean                             +0.0004         +0.0004          1610          1610          1610          1610
fizzy/execute/mul256_opt0/input0_mean                             +0.0010         +0.0010            27            27            27            27
fizzy/execute/mul256_opt0/input1_mean                             +0.0015         +0.0015            27            27            27            27
fizzy/execute/sha1/512_bytes_rounds_1_mean                        -0.0011         -0.0011            95            95            95            95
fizzy/execute/sha1/512_bytes_rounds_16_mean                       +0.0004         +0.0004          1323          1324          1323          1324
fizzy/execute/sha256/512_bytes_rounds_1_mean                      +0.0064         +0.0064            99           100            99           100
fizzy/execute/sha256/512_bytes_rounds_16_mean                     +0.0025         +0.0025          1377          1380          1377          1380
fizzy/execute/micro/eli_interpreter/halt_mean                     +0.0018         +0.0018             0             0             0             0
fizzy/execute/micro/eli_interpreter/exec105_mean                  -0.0022         -0.0022             5             5             5             5
fizzy/execute/micro/factorial/10_mean                             +0.0020         +0.0020             0             0             0             0
fizzy/execute/micro/factorial/20_mean                             -0.0211         -0.0211             1             1             1             1
fizzy/execute/micro/fibonacci/24_mean                             -0.0103         -0.0103          7253          7178          7253          7178
fizzy/execute/micro/host_adler32/1_mean                           -0.0016         -0.0016             0             0             0             0
fizzy/execute/micro/host_adler32/100_mean                         +0.0013         +0.0013             3             3             3             3
fizzy/execute/micro/host_adler32/1000_mean                        +0.0086         +0.0086            30            30            30            30
fizzy/execute/micro/spinner/1_mean                                -0.0060         -0.0061             0             0             0             0
fizzy/execute/micro/spinner/1000_mean                             -0.0004         -0.0004            10            10            10            10
```

lib/fizzy/execute.cpp
```diff
@@ -359,17 +359,17 @@ inline bool store_into_memory(
...
 }
 
 template <typename Op>
 inline void binary_op(OperandStack& stack, Op op) noexcept
 {
-    using T = decltype(op(stack.top(), stack.top()));
-    const auto val2 = static_cast<T>(stack.pop());
-    const auto val1 = static_cast<T>(stack.top());
-    stack.top() = static_cast<std::make_unsigned_t<T>>(op(val1, val2));
+    using T = decltype(op({}, {}));
```
axic, line 369: Hm, what does this do with `{}, {}`?

chfast: The `{}`  in `f(T)` more/less mean "construct the required type T with default constructor".

E.g. The `void func(void*, Options, vector<int>, int)` can be invoked as `func(nullptr, {}, {}, 0)` or even `func({}, {}, {}, {})`.

axic: Yeah I get that, but in the case of `op`, which is `std::add`, `std::sub`, etc.

chfast: Well, `op` is an instance of e.g. `std::add<uint64_t>{}` which has `operator()` overloaded so `op(...)` expression is valid. The `decltype` gets the type of this expression.

axic: Okay so this line change is not related to the current PR. Could have been like this prior.

lib/fizzy/execute.cpp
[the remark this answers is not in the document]
axic: I'm fine keeping it this way but would prefer a simple comment that the result can be signed or unsigned and its conversion is handled in Value.

lib/fizzy/execute.cpp
chfast: Yes. The `{}` could have been use instead of `stack.top()` before. But now this simplifications is required as `stack.top()` has wrong type in general.

lib/fizzy/execute.cpp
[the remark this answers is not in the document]
chfast: Added comments and used `Value` explicitly (no implicit conversion although available).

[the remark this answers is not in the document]
axic: Hope it did not change runtimes lol.

PR #424, opened 2020-07-15
Implement floating point const and add

chfast:
A symbolic milestone starting floating point support implementation.

### DIFF
--- a/lib/fizzy/parser_expr.cpp
+++ b/lib/fizzy/parser_expr.cpp
@@ -27,6 +27,18 @@ inline void push(bytes& b, T value)
     b.append(storage, sizeof(storage));
 }
 
+template <typename T>
+inline parser_result<T> read_const(const uint8_t* pos, const uint8_t* end)
+{
+    constexpr auto size = sizeof(T);
+    if (pos + size > end)
+        throw parser_error{"unexpected EOF"};
+
+    T value;
+    __builtin_memcpy(&value, pos, size);
+    return {value, pos + size};
+}
+
 /// The control frame to keep information about labels and blocks as defined in
 /// Wasm Validation Algorithm https://webassembly.github.io/spec/core/appendix/algorithm.html.
 struct ControlFrame
@@ -305,12 +317,6 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end, FuncIdx f
             throw parser_error{"invalid instruction " + std::to_string(*(pos - 1))};
 
         // Floating point instructions are unsupported
-        case Instr::f32_const:
-            pos = skip(4, pos, end);
-            break;
-        case Instr::f64_const:
-            pos = skip(8, pos, end);
-            break;
         case Instr::f32_eq:
         case Instr::f32_ne:
         case Instr::f32_lt:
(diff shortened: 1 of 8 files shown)


lib/fizzy/value.hpp
```diff
@@ -0,0 +1,40 @@
...
+    constexpr Value(int64_t v) noexcept : i64{static_cast<uint64_t>(v)} {}
+    constexpr Value(int32_t v) noexcept : i64{static_cast<uint32_t>(v)} {}
+
+    constexpr Value(float v) noexcept : f32{v} {}
+
+    constexpr operator uint64_t() const noexcept { return i64; }
+
+    /// Get the value as the given type. Handy in templates.
+    template <typename T>
+    constexpr T as() const noexcept
```
gumb0, line 29: Can this be left without default implementation and then have specializations for `uint32_t` and `uint64_t` similar to `float`?

chfast:
Yes, makes sense.

lib/fizzy/value.hpp
chfast: Done in #423.

axic:
Doesnt this need spectest expectation update in CI?

chfast:
> Doesnt this need spectest expectation update in CI?

I was expecting that, but no. What is the conditions for skipping a spec test?

test/unittests/value_test.cpp
```diff
@@ -41,6 +41,12 @@ TEST(value, constructor_from_signed_ints)
     EXPECT_EQ(Value{int64_t{-3}}.i64, 0xfffffffffffffffd);
 }
 
+TEST(value, constructor_from_floating_points)
+{
+    EXPECT_EQ(Value{123.456f}.f32, 123.456f);
+    EXPECT_EQ(Value{123.456789001}.f64, 123.456789001);
```
axic, line 47: Would be nice to have a test for min/max.

chfast: You mean the total max/min, or the biggest mantissa only?

chfast: I used some special values in tests.

axic: Basically something which can not be represented in f32. `f64::max()` is fine, but not sure what that is actually storing, I believe a large exponent?

lib/fizzy/parser_expr.cpp
```diff
@@ -799,14 +805,27 @@ parser_result<Code> parse_expr(const uint8_t* pos, const uint8_t* end, FuncIdx f
...
         {
             int64_t value;
             std::tie(value, pos) = leb128s_decode<int64_t>(pos, end);
             push(code.immediates, static_cast<uint64_t>(value));
             break;
         }
+        case Instr::f32_const:
+        {
+            uint32_t value;
+            std::tie(value, pos) = read<decltype(value)>(pos, end);
```
gumb0, line 818: what is the point of using `decltype` here if `value` is declared on the previous line?

lib/fizzy/parser_expr.cpp
```diff
@@ -27,6 +27,18 @@ inline void push(bytes& b, T value)
     b.append(storage, sizeof(storage));
 }
 
+template <typename T>
+inline parser_result<T> read(const uint8_t* pos, const uint8_t* end)
```
gumb0, line 31: I would call it maybe `read_value` or `read_const`

test/unittests/value_test.cpp
```diff
@@ -50,6 +56,40 @@ TEST(value, as_integer)
...
+    f = std::numeric_limits<float>::min();
+    EXPECT_EQ(Value{f}.f32, f);
+    EXPECT_EQ(Value{f}.as<float>(), f);
+    f = std::numeric_limits<float>::max();
+    EXPECT_EQ(Value{f}.f32, f);
+    EXPECT_EQ(Value{f}.as<float>(), f);
+    f = std::numeric_limits<float>::denorm_min();
+    EXPECT_EQ(Value{f}.f32, f);
+    EXPECT_EQ(Value{f}.as<float>(), f);
+
```
gumb0, line 77: Could also add `numeric_limits<float>::lowest()`

lib/fizzy/parser_expr.cpp
chfast: Or `read_float`.

lib/fizzy/parser_expr.cpp
chfast: Easier to copy-paste to `Instr::f64_const`.

test/unittests/value_test.cpp
chfast: Done.

PR #428, opened 2020-07-16
Rename execution_result -> ExecutionResult

chfast:


axic:
Nice, I was about to ask why not rename it.

chfast:
Move it to separate header? I will use it also in `WasmEngine`.

axic:
> Move it to separate header? I will use it also in WasmEngine.

I think that is a bad idea -- we would be trying to apply Fizzy's API to other engines. Why can't we leave WasmEngine as is, already uses optional?

Probably I could / will change my mind if we would expose host functions in WasmEngine.

chfast:
> I think that is a bad idea -- we would be trying to apply Fizzy's API to other engines. Why can't we leave WasmEngine as is, already uses optional?

Sounds good.

PR #429, opened 2020-07-20
Remove redundant check from drop_operand

gumb0:


lib/fizzy/parser_expr.cpp
```diff
@@ -119,7 +119,7 @@ inline void drop_operand(
         static_cast<int>(operand_stack.size()) < frame.parent_stack_height + 1)
         throw validation_error{"stack underflow"};
 
-    if (frame.unreachable && static_cast<int>(operand_stack.size()) == frame.parent_stack_height)
```
gumb0, line 122: This check was redundant, becase the case of reachable frame + stack equal to parent stack is eliminated by the stack underflow condition above.

In other words, by this point not having a value on stack is possible only if it's unreachable.

axic: Why not leave an assertion here?

gumb0: I considered assertion, but it just seems too obvious, because the previous check implies it's true.

"(frame unreachable OR no stack underflow) AND stack underflow" implies "frame unreachable".

axic: Yeah if you read it carefully, but I think an assertion is much more explicit.

gumb0: Ok, will add an assertion.

PR #430, opened 2020-07-21
Validate types of instructions in constant expressions

gumb0:
Type mismatch when constant expression uses `global.get` looks to be not covered by spec tests (passed tests don't change here when I disable it).


axic:
I think it is a bit confusing that "constant constant expression" are validated in `parse_constant_expression` and global-based ones in `validate_constant_expression`. I think we should consider refactoring this, but for now it is fine.

PR #431, opened 2020-07-21
Add broken test cases for spectests

axic:


test/spectests/spectests.cpp
```diff
@@ -576,34 +576,25 @@ bool run_tests_from_dir(const fs::path& path, const test_settings& settings)
...
               << ", SKIPPED " << total.skipped << ".\n";
 
-    return (total.failed == 0 && !exception_thrown);
+    return total.failed == 0;
 }
 
 }  // namespace
 
-int main(int argc, char** argv)
+int main(int argc, char** argv) noexcept
```
axic, line 597: Probably this is a useless overkill.

test/spectests/spectests.cpp
```diff
@@ -576,34 +576,25 @@ bool run_tests_from_dir(const fs::path& path, const test_settings& settings)
...
-        try
-        {
-            const auto res = test_runner{settings}.run_from_file(f);
+        const auto res = test_runner{settings}.run_from_file(f);
 
-            total.passed += res.passed;
-            total.failed += res.failed;
-            total.skipped += res.skipped;
-        }
-        catch (const std::exception& ex)
```
axic, line 590: `main` has the very same catch all.

gumb0: Without it, it won't outuput `TOTAL` summary string in case of unexpected exception, but perhaps it's not important.

test/spectests/spectests.cpp
gumb0: Yeah, it looks it doesn't have any effect (uncaught exception terminates in both cases), and is not mentioned in the standard as a valid `main` signature, so I'd leave it without `noexcept`.

https://stackoverflow.com/a/47584093
https://en.cppreference.com/w/cpp/language/main_function

test/spectests/spectests.cpp
axic: I think if there's an exception we shouldn't try to recover, apart from displaying the exception.

test/spectests/spectests.cpp
axic: Removed.

PR #433, opened 2020-07-21
Validate functions' arity

gumb0:


test/unittests/validation_test.cpp
```diff
@@ -13,6 +13,15 @@
 using namespace fizzy;
 using namespace fizzy::test;
 
+TEST(validation, imported_function_type_invalid_arity)
+{
+    /* wat2wasm
```
gumb0, line 18: wabt doesn't throw an error unless function with this type is defined

axic: Can you add this to our list of "to reach out"? Or just ask the question on the wabt repo? It may be we are using an old wabt and it was fixed since, or perhaps they are more committed to the multi-return-value change (I believe that is implemented by wabt).

test/unittests/validation_test.cpp
```diff
@@ -13,6 +13,15 @@
 using namespace fizzy;
 using namespace fizzy::test;
 
+TEST(validation, imported_function_type_invalid_arity)
```
axic, line 16: This is not imported, just `function_type_invalid_arity`

test/unittests/validation_test.cpp
gumb0: I tried it with recent wabt 1.0.19, and it appears to be fixed: multi-value is enabled by default there, but with `--disable-multi-value` this correctly throws an error.

I think I'll add `--no-check` here

gumb0: 1.0.14 is the last one that has multi-value disabled by default, and it has this issue fixed also.

PR #434, opened 2020-07-22
ci: update wabt

gumb0:
Fixes https://github.com/wasmx/fizzy/pull/433#discussion_r458262639

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -184,7 +184,7 @@ jobs:
     - run:
         name: "Install wabt"
         working_directory: ~/bin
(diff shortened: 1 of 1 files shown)


gumb0:
This doesn't work: for some reason there are no linux binaries in 1.0.14 release.

gumb0:
> This doesn't work: for some reason there are no linux binaries in 1.0.14 release.

Fixed: 1.0.15 appears to be identical to 1.0.14 and has linux binaries.

PR #435, opened 2020-07-22
test: Add unit tests for ExecutionResult constructors

axic:
image

Either coverage is wrong, or the explicit constructor is not called. With these tests we make sure that values are always correct even if in the case `Trap`/`Void` don't get assigned via the explicit constructor, but by setting the `trapped` variable. We can figure out the case with coverage later (cc @chfast).

### DIFF
--- a/test/unittests/api_test.cpp
+++ b/test/unittests/api_test.cpp
@@ -25,6 +25,48 @@ ExecutionResult function_returning_void(Instance&, span<const uint64_t>, int) no
 }
 }  // namespace
 
+TEST(api, execution_result_trap)
+{
+    const auto result = Trap;
+    EXPECT_TRUE(result.trapped);
+    EXPECT_FALSE(result.has_value);
+    EXPECT_EQ(result.value, 0);
+}
+
+TEST(api, execution_result_void)
+{
+    const auto result = Void;
+    EXPECT_FALSE(result.trapped);
+    EXPECT_FALSE(result.has_value);
+    EXPECT_EQ(result.value, 0);
+}
+
+TEST(api, execution_result_value)
+{
+    const ExecutionResult result = 1234;
+    EXPECT_FALSE(result.trapped);
+    EXPECT_TRUE(result.has_value);
+    EXPECT_EQ(result.value, 1234);
+}
+
+TEST(api, execution_result_bool_constructor)
+{
+    bool success = false;
+    const ExecutionResult result{success};
+    EXPECT_TRUE(result.trapped);
+    EXPECT_FALSE(result.has_value);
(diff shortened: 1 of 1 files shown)


gumb0:
Do you want to try to cover the constexpr constructor line? I think this might do it
```
bool success = false; // non const variable
const ExecutionResult result{success};
EXPECT_TRUE(result.trapped);
```

test/unittests/api_test.cpp
```diff
@@ -25,6 +25,48 @@ ExecutionResult function_returning_void(Instance&, span<const uint64_t>, int) no
...
+{
+    const ExecutionResult result = 1234;
+    EXPECT_FALSE(result.trapped);
+    EXPECT_TRUE(result.has_value);
+    EXPECT_EQ(result.value, 1234);
+}
+
+TEST(api, execution_result_bool_constructor)
+{
+    bool success = false;
```
gumb0, line 54: Maybe it would execute the constructor with this being `const`, too, I'm not sure, just thought with non-const it's less probable to be optimized away.
Can be merged like this, anyway.

test/unittests/api_test.cpp
```diff
@@ -25,6 +25,48 @@ ExecutionResult function_returning_void(Instance&, span<const uint64_t>, int) no
...
+{
+    const auto result = Void;
+    EXPECT_FALSE(result.trapped);
+    EXPECT_FALSE(result.has_value);
+    EXPECT_EQ(result.value, 0);
+}
+
+TEST(api, execution_result_value)
+{
+    const ExecutionResult result = 1234;
```
axic, line 46:
```
const ExecutionResult result{1234};
```
results in
```
/Users/alex/Projects/fizzy/test/unittests/api_test.cpp:46:27: error: call to constructor of 'const fizzy::ExecutionResult' is ambiguous
    const ExecutionResult result{1234};
                          ^     ~~~~~~
/Users/alex/Projects/fizzy/lib/fizzy/execute.hpp:25:15: note: candidate constructor
    constexpr ExecutionResult(uint64_t _value) noexcept : has_value{true}, value{_value} {}
              ^
/Users/alex/Projects/fizzy/lib/fizzy/execute.hpp:29:24: note: candidate constructor
    constexpr explicit ExecutionResult(bool success) noexcept : trapped{!success} {}
                       ^
/Users/alex/Projects/fizzy/lib/fizzy/execute.hpp:18:8: note: candidate constructor (the implicit move constructor)
struct ExecutionResult
       ^
/Users/alex/Projects/fizzy/lib/fizzy/execute.hpp:18:8: note: candidate constructor (the implicit copy constructor)
```

PR #437, opened 2020-07-27
Validate alignment for floating point memory instructions

gumb0:


axic:
Looks okay to me but would wait for @chfast's review.

PR #445, opened 2020-07-31
Support floats in spectest runner

gumb0:
Depends on #424


test/smoketests/spectests/CMakeLists.txt
```diff
@@ -9,7 +9,7 @@ add_test(
 set_tests_properties(
     fizzy/smoketests/spectests/default
     PROPERTIES
-    PASS_REGULAR_EXPRESSION "PASSED 23, FAILED 0, SKIPPED 7"
+    PASS_REGULAR_EXPRESSION "PASSED 26, FAILED 1, SKIPPED 3"
```
gumb0, line 12: 1 failure is for float global, will be fixed in #446

test/spectests/spectests.cpp
```diff
@@ -51,11 +51,23 @@ const auto spectest_module = fizzy::parse(spectest_bin);
...
-uint64_t json_to_value(const json& v)
+fizzy::Value json_to_value(const json& v)
 {
     return static_cast<std::make_unsigned_t<T>>(std::stoull(v.get<std::string>()));
 }
 
+template <>
+fizzy::Value json_to_value<float>(const json& v)
+{
+    return std::stoull(v.get<std::string>());
```
chfast, line 62: Why using "to unsigned integer" for floating point values?

circle.yml
```diff
@@ -271,9 +271,9 @@ jobs:
       - benchmark:
           min_time: "0.01"
       - spectest:
-          expected_passed: 5468
-          expected_failed: 9
-          expected_skipped: 6336
+          expected_passed: 5577
+          expected_failed: 13
```
gumb0, line 275: 4 new failures are all in `globals.wast`, will be fixed in #446

test/spectests/spectests.cpp
gumb0: `wast2json` puts them as 64-bit integers to json

test/spectests/spectests.cpp
```diff
@@ -402,6 +396,20 @@ class test_runner
         return it_instance->second.get();
     }
 
+    std::optional<fizzy::Value> read_value(const json& v)
+    {
+        const auto arg_type = v.at("type").get<std::string>();
+        if (arg_type != "i32" && arg_type != "i64" && arg_type != "f32" && arg_type != "f64")
+        {
+            skip("Unsupported value type '" + arg_type + "'.");
```
axic, line 404: This is not covered by unit tests.

test/spectests/spectests.cpp
```diff
@@ -402,6 +396,20 @@ class test_runner
...
+            return std::nullopt;
+        }
+
+        // Values of all types are serialized to JSON as integers.
+        // Value type will handle correct conversions.
```
axic, line 409: How? Why not keep the old `json_to_value<int32_t>(arg.at("value"))` way?

gumb0: It's just simpler without additional `switch` and template.

Presumably `wast2json` does the opposite conversion - typed value into unsigned integer. We could reinterpret_cast it back to correct type, but `Value(correct_type_value)` constructor does the same as `Value(reinterpret_cast<unsigned long long>(correct_type_value))`, which is the idea of `Value` type...

test/spectests/spectests.cpp
gumb0: Added tests.

PR #446, opened 2020-07-31
Support floating point globals

gumb0:
Depends on #445

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -508,7 +508,9 @@ std::unique_ptr<Instance> instantiate(Module module,
     datasec_offsets.reserve(module.datasec.size());
     for (const auto& data : module.datasec)
     {
-        const uint64_t offset = eval_constant_expression(data.offset, imported_globals, globals);
+        // Offset is validated to be i32, but it's used in 64-bit calculation below.
+        const uint64_t offset =
+            eval_constant_expression(data.offset, imported_globals, globals).i64;
 
         if (offset + data.init.size() > memory->size())
             throw instantiate_error{"data segment is out of memory bounds"};
@@ -521,7 +523,9 @@ std::unique_ptr<Instance> instantiate(Module module,
     elementsec_offsets.reserve(module.elementsec.size());
     for (const auto& element : module.elementsec)
     {
-        const uint64_t offset = eval_constant_expression(element.offset, imported_globals, globals);
+        // Offset is validated to be i32, but it's used in 64-bit calculation below.
+        const uint64_t offset =
+            eval_constant_expression(element.offset, imported_globals, globals).i64;
 
         if (offset + element.init.size() > table->size())
             throw instantiate_error{"element segment is out of table bounds"};
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -214,8 +214,8 @@ std::tuple<bytes_ptr, Limits> allocate_memory(const std::vector<Memory>& module_
     }
 }
 
-uint64_t eval_constant_expression(ConstantExpression expr,
-    const std::vector<ExternalGlobal>& imported_globals, const std::vector<uint64_t>& globals)
+Value eval_constant_expression(ConstantExpression expr,
+    const std::vector<ExternalGlobal>& imported_globals, const std::vector<Value>& globals)
 {
     if (expr.kind == ConstantExpression::Kind::Constant)
         return expr.value.constant;
@@ -486,7 +486,7 @@ std::unique_ptr<Instance> instantiate(Module module,
     match_imported_globals(module.imported_global_types, imported_globals);
 
     // Init globals
-    std::vector<uint64_t> globals;
+    std::vector<Value> globals;
     globals.reserve(module.globalsec.size());
     for (auto const& global : module.globalsec)
     {
--- a/lib/fizzy/parser.cpp
+++ b/lib/fizzy/parser.cpp
@@ -174,17 +174,13 @@ inline parser_result<ConstantExpression> parse_constant_expression(
         break;
     }
     case Instr::f32_const:
-        // TODO: support this once floating points are implemented
         result.kind = ConstantExpression::Kind::Constant;
-        result.value.constant = 0;
-        pos = skip(4, pos, end);
+        std::tie(result.value.constant, pos) = parse_value<uint32_t>(pos, end);
         constant_actual_type = ValType::f32;
         break;
     case Instr::f64_const:
-        // TODO: support this once floating points are implemented
         result.kind = ConstantExpression::Kind::Constant;
-        result.value.constant = 0;
-        pos = skip(8, pos, end);
+        std::tie(result.value.constant, pos) = parse_value<uint64_t>(pos, end);
         constant_actual_type = ValType::f64;
         break;
     }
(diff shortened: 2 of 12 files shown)


lib/fizzy/parser.hpp
```diff
@@ -33,6 +33,18 @@ inline parser_result<uint8_t> parse_byte(const uint8_t* pos, const uint8_t* end)
 
     return {*pos, pos + 1};
 }
+template <typename T>
+
+inline parser_result<T> read_float_const(const uint8_t* pos, const uint8_t* end)
```
gumb0, line 38: Should this be better named `parse_float_const` ?

lib/fizzy/parser.cpp
```diff
@@ -126,6 +126,18 @@ inline parser_result<GlobalType> parse_global_type(const uint8_t* pos, const uin
     return {type, pos};
 }
 
+template <typename T>
+inline parser_result<T> read_const(const uint8_t* pos, const uint8_t* end)
```
axic: Is this `read_const` or just `read`/`read_value`/`read_type`?

lib/fizzy/parser.cpp
```diff
@@ -174,17 +186,14 @@ inline parser_result<ConstantExpression> parse_constant_expression(
...
-        pos = skip(4, pos, end);
+        std::tie(result.value.constant, pos) = read_const<uint32_t>(pos, end);
         constant_actual_type = ValType::f32;
         break;
     case Instr::f64_const:
-        // TODO: support this once floating points are implemented
         result.kind = ConstantExpression::Kind::Constant;
         result.value.constant = 0;
-        pos = skip(8, pos, end);
+        std::tie(result.value.constant, pos) = read_const<uint64_t>(pos, end);
```
axic: This first commit is titled "move read_const form header to parser.cpp", yet it also changes the implementation here.

lib/fizzy/parser.cpp
```diff
@@ -174,17 +186,14 @@ inline parser_result<ConstantExpression> parse_constant_expression(
...
         result.kind = ConstantExpression::Kind::Constant;
-        result.value.constant = 0;
-        pos = skip(4, pos, end);
+        std::tie(result.value.constant, pos) = read_const<uint32_t>(pos, end);
         constant_actual_type = ValType::f32;
         break;
     case Instr::f64_const:
-        // TODO: support this once floating points are implemented
         result.kind = ConstantExpression::Kind::Constant;
         result.value.constant = 0;
```
axic, line 183: If this is removed for f32, should also be removed here.

lib/fizzy/parser.cpp
gumb0: I'm considering `parse_float_value`/`parse_float_const` because it returns `parse_result` and is similar to other parse_* functions above and below.

gumb0:
Fixed the commits.

lib/fizzy/parser.cpp
axic: Reading `parse_float_const<uint32_t>(..)` is extremely confusing, so I would suggest to go with `parse_value<uint32_t>` or `parse_type<uint32_t>`, especially as this should do byteswapping should we support big endian targets. Right?

axic: If `parse_float_const<float>` and `parse_float_const<double>` works that is fine too.

gumb0: > If `parse_float_const<float>` and `parse_float_const<double>` works that is fine too.

Not without a cast on a calling side.

chfast: That would work (but I would need to check assembly of that to be sure). But might be confusing as you can/will implement `fN.const` as `iN.const`: https://github.com/wasmx/fizzy/blob/master/lib/fizzy/execute.cpp#L1004-L1005.

lib/fizzy/parser.hpp
chfast: `read_constant()`? `read_constant_bytes()`? But maybe keep the original name for now.

lib/fizzy/parser.cpp
```diff
@@ -174,17 +174,13 @@ inline parser_result<ConstantExpression> parse_constant_expression(
         break;
     }
     case Instr::f32_const:
-        // TODO: support this once floating points are implemented
         result.kind = ConstantExpression::Kind::Constant;
-        result.value.constant = 0;
-        pos = skip(4, pos, end);
```
chfast, line 180: The `skip()` function is probably unused now.

lib/fizzy/parser.cpp
gumb0: Renamed to `parse_value`

lib/fizzy/parser.cpp
gumb0: Removed

lib/fizzy/parser.cpp
gumb0: Removed.

lib/fizzy/execute.cpp
```diff
@@ -508,7 +508,8 @@ std::unique_ptr<Instance> instantiate(Module module,
     datasec_offsets.reserve(module.datasec.size());
     for (const auto& data : module.datasec)
     {
-        const uint64_t offset = eval_constant_expression(data.offset, imported_globals, globals);
+        const uint64_t offset =
+            eval_constant_expression(data.offset, imported_globals, globals).i64;
```
axic, line 512: Aren't these two validated to be `uin32_t`? And if so, the check below doesn't need to operate on 128-bit for overflow checks.

lib/fizzy/execute.cpp
```diff
@@ -521,7 +522,8 @@ std::unique_ptr<Instance> instantiate(Module module,
     elementsec_offsets.reserve(module.elementsec.size());
     for (const auto& element : module.elementsec)
     {
-        const uint64_t offset = eval_constant_expression(element.offset, imported_globals, globals);
+        const uint64_t offset =
```
axic, line 525: Same here.

lib/fizzy/execute.cpp
gumb0: It's validated to be 32 bit, and this, together with validation of `data.init` vector length to be 32 bit, guarantees that `offset + data.init.size()` won't overflow 64 bit.

If we do here `as<uint32_t>()`, the check below will implicitly cast it back to 64-bit anyway, so I'm not sure it's worth it.

lib/fizzy/parser.hpp
```diff
@@ -34,6 +34,18 @@ inline parser_result<uint8_t> parse_byte(const uint8_t* pos, const uint8_t* end)
     return {*pos, pos + 1};
 }
 
+template <typename T>
+inline parser_result<T> parse_value(const uint8_t* pos, const uint8_t* end)
```
axic, line 30: Since this needs to be rebased, you may want to update the commit message to read `parse_value` insead of `read_const`, but it is not important.

lib/fizzy/execute.cpp
axic: Perhaps adding a note that these are actually 32-bit could be useful?

lib/fizzy/parser.hpp
gumb0: It's "move and rename `read_const`" I would say

lib/fizzy/execute.cpp
gumb0: Added comment here and for elem section.

PR #447, opened 2020-07-31
Implement float->int trunc instructions

chfast:
The operator from the wasm spec is partial, i.e. it has some undefined results.
image
But, the execution spec requires to trap on operator's undefined result.
image

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -271,9 +271,9 @@ jobs:
       - benchmark:
           min_time: "0.01"
       - spectest:
-          expected_passed: 5581
+          expected_passed: 5589
           expected_failed: 9
-          expected_skipped: 6223
+          expected_skipped: 6215
 
   sanitizers-macos:
     executor: macos
@@ -289,9 +289,9 @@ jobs:
       - benchmark:
           min_time: "0.01"
       - spectest:
-          expected_passed: 5581
+          expected_passed: 5589
           expected_failed: 9
-          expected_skipped: 6223
+          expected_skipped: 6215
 
   benchmark:
     machine:
@@ -397,13 +397,13 @@ jobs:
           at: ~/build
       - spectest:
           skip_validation: true
-          expected_passed: 4639
+          expected_passed: 4647
           expected_failed: 9
-          expected_skipped: 7165
+          expected_skipped: 7157
       - spectest:
-          expected_passed: 5581
+          expected_passed: 5589
           expected_failed: 9
-          expected_skipped: 6223
+          expected_skipped: 6215
       - collect_coverage_data
 
 workflows:
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -1273,6 +1273,7 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value> 
             binary_op(stack, rotr<uint32_t>);
             break;
         }
+
         case Instr::i64_clz:
         {
             unary_op(stack, clz64);
@@ -1395,6 +1396,19 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value> 
             binary_op(stack, rotr<uint64_t>);
             break;
         }
+
+        case Instr::f32_add:
+        {
+            binary_op(stack, std::plus<float>{});
+            break;
+        }
+
+        case Instr::f64_add:
+        {
+            binary_op(stack, std::plus<double>{});
+            break;
+        }
+
         case Instr::i32_wrap_i64:
         {
             stack.push(stack.pop().as<uint32_t>());
@@ -1481,16 +1495,6 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value> 
                 trap = true;
                 goto end;
             }
-            break;
-        }
-        case Instr::f32_add:
-        {
-            binary_op(stack, std::plus<float>{});
-            break;
-        }
-        case Instr::f64_add:
-        {
-            binary_op(stack, std::plus<double>{});
             break;
         }
 
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -6,9 +6,11 @@
 #include "limits.hpp"
 #include "module.hpp"
 #include "stack.hpp"
+#include "trunc_boundaries.hpp"
 #include "types.hpp"
 #include <algorithm>
 #include <cassert>
+#include <cmath>
 #include <cstring>
 #include <stack>
 
@@ -322,6 +324,23 @@ inline DstT extend(SrcT in) noexcept
     }
     else
         return DstT{in};
+}
+
+template <typename SrcT, typename DstT>
+inline bool trunc(OperandStack& stack) noexcept
+{
+    using boundaries = trunc_boundaries<SrcT, DstT>;
+
+    const auto input = stack.top().as<SrcT>();
+    if (input > boundaries::lower && input < boundaries::upper)
+    {
+        assert(!std::isnan(input));
+        assert(input != std::numeric_limits<SrcT>::infinity());
+        assert(input != -std::numeric_limits<SrcT>::infinity());
+        stack.top() = static_cast<DstT>(input);
+        return true;
+    }
+    return false;
 }
 
 template <typename DstT, typename SrcT = DstT>
@@ -1381,6 +1400,42 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value> 
             stack.push(stack.pop().as<uint32_t>());
             break;
         }
+        case Instr::i32_trunc_f32_s:
+        {
+            if (!trunc<float, int32_t>(stack))
+            {
+                trap = true;
+                goto end;
+            }
+            break;
+        }
+        case Instr::i32_trunc_f32_u:
+        {
+            if (!trunc<float, uint32_t>(stack))
+            {
+                trap = true;
+                goto end;
+            }
+            break;
+        }
+        case Instr::i32_trunc_f64_s:
+        {
+            if (!trunc<double, int32_t>(stack))
+            {
+                trap = true;
+                goto end;
+            }
+            break;
+        }
+        case Instr::i32_trunc_f64_u:
+        {
+            if (!trunc<double, uint32_t>(stack))
+            {
+                trap = true;
+                goto end;
+            }
+            break;
+        }
         case Instr::i64_extend_i32_s:
         {
             const auto value = stack.pop().as<int32_t>();
@@ -1390,6 +1445,42 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value> 
         case Instr::i64_extend_i32_u:
         {
             // effectively no-op
+            break;
+        }
+        case Instr::i64_trunc_f32_s:
+        {
+            if (!trunc<float, int64_t>(stack))
+            {
+                trap = true;
+                goto end;
+            }
+            break;
+        }
+        case Instr::i64_trunc_f32_u:
+        {
+            if (!trunc<float, uint64_t>(stack))
+            {
+                trap = true;
+                goto end;
+            }
+            break;
+        }
+        case Instr::i64_trunc_f64_s:
+        {
+            if (!trunc<double, int64_t>(stack))
+            {
+                trap = true;
+                goto end;
+            }
+            break;
+        }
+        case Instr::i64_trunc_f64_u:
+        {
+            if (!trunc<double, uint64_t>(stack))
+            {
+                trap = true;
+                goto end;
+            }
             break;
         }
         case Instr::f32_add:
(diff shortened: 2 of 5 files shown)


gumb0:
Please rebase.

test/unittests/execute_floating_point_test.cpp
```diff
@@ -69,4 +70,174 @@ TEST(execute_floating_point, f64_add)
...
+        const auto arg =
+            std::nextafter(std::pow(src_type{2}, src_type{dst_limits::digits}), src_type{0});
+        const auto expected = static_cast<dst_type>(arg);
+        const auto result = execute(*instance, 0, {arg});
+        EXPECT_EQ(result.value.template as<dst_type>(), expected);
+    }
+
+    {  // Min defined: -2^N or 0.
+        const auto arg =
+            dst_limits::is_signed ? -std::pow(src_type{2}, src_type{dst_limits::digits}) : 0;
```
gumb0, line 206: Maybe would make sense to test 0 for signed types, too, and move `-2^N` under `if (dst_limits::is_signed)` at the bottom of the test.

axic:
How come this still does not affect spec tests expectations?

test/unittests/execute_floating_point_test.cpp

lib/fizzy/execute.cpp
```diff
@@ -1375,6 +1390,42 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value>
             stack.push(static_cast<uint32_t>(stack.pop()));
             break;
         }
+        case Instr::i32_trunc_f32_s:
+        {
+            if (!trunc<float, int32_t>(stack))
+            {
+                trap = true;
```
gumb0, line 1397: Do we trap instead of undefined result? Why? This seems against the spec.

lib/fizzy/execute.cpp
```diff
@@ -324,6 +326,19 @@ inline DstT extend(SrcT in) noexcept
         return DstT{in};
 }
 
+template <typename SrcT, typename DstT>
+inline bool trunc(OperandStack& stack) noexcept
+{
+    using boundaries = trunc_boundaries<SrcT, DstT>;
```
gumb0, line 332: Why would it not work with `static_cast<SrcT>(std::numeric_limits<Dst>::min())` / `static_cast<SrcT>(std::numeric_limits<Dst>::max())` ?

lib/fizzy/execute.cpp
chfast: I updated the description with the explanation. And spec tests confirm that.

lib/fizzy/execute.cpp
chfast: E.g.
- `-0.0 -> 0u`
- `-0.999 -> 0u`
- `-1.0 -> undefined`

test/unittests/execute_floating_point_test.cpp
```diff
@@ -66,4 +69,203 @@ TEST(execute_floating_point, f64_add)
...
+        // The lower boundary is:
+        // - for signed integers: -2^N - 1,
+        // - for unsigned integers: -1.
+        // However, the -2^N - 1 may be not representative in a float type so we compute it as
+        // floor(-2^N - epsilon).
+        const auto min_defined_int = IntLimits::min();
+        const auto lower_boundary =
+            std::floor(std::nextafter(FloatT{min_defined_int}, -FloatLimits::infinity()));
+        EXPECT_EQ(lower_boundary, expected_boundaries::lower);
+        EXPECT_THAT(execute(*instance, 0, {upper_boundary}), Traps());
```
gumb0, line 240:
```suggestion
        EXPECT_THAT(execute(*instance, 0, {lower_boundary}), Traps());
```

test/unittests/execute_floating_point_test.cpp
```diff
@@ -66,4 +69,203 @@ TEST(execute_floating_point, f64_add)
...
+    {
+        // BOUNDARIES OF DEFINITION
+        //
+        // Here we want to identify and test the boundary values of the defined behavior of the
+        // trunc instructions. For undefined results the execution must trap.
+        // Note that floating point type can represent any power of 2.
+
+        using expected_boundaries = trunc_boundaries<FloatT, IntT>;
+
+        // For iN with max value 2^N-1 the float(2^N) exists and trunc() from that value to iN
```
gumb0, line 218: To disambiguate from "max value" in the beginning of the sentence
```suggestion
        // For iN with max value 2^N-1 the float(2^N) exists and trunc(float(2^N)) to iN
```

test/unittests/execute_floating_point_test.cpp
```diff
@@ -66,4 +69,203 @@ TEST(execute_floating_point, f64_add)
...
+        // is undefined.
+        const auto upper_boundary = std::pow(FloatT{2}, FloatT{IntLimits::digits});
+        EXPECT_EQ(upper_boundary, expected_boundaries::upper);
+        EXPECT_THAT(execute(*instance, 0, {upper_boundary}), Traps());
+
+        // But the trunc() of the next float value smaller than 2^N is defined.
+        // Depending on the resolution of the floating point type, the result integer value may
+        // be other than 2^(N-1).
+        const auto max_defined = std::nextafter(upper_boundary, FloatT{0});
+        const auto max_defined_int = static_cast<IntT>(max_defined);
```
gumb0, line 228: This would be caught by UBSan in case the assumption that it's defined were wrong?

lib/fizzy/trunc_boundaries.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+namespace fizzy
+{
+template <typename SrcT, typename Dst>
+struct trunc_boundaries;
+
+template <>
+struct trunc_boundaries<float, int32_t>
+{
+    static constexpr float lower = -2147483904.0;
```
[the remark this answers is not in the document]
gumb0: I see explanation in the test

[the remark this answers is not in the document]
chfast: At this range `float` has big gaps in values. This is max value being smaller than `-2^31`.

test/unittests/execute_floating_point_test.cpp
chfast: Yes, UBSan catches the results that are undefined in wasm (because this is undefined behavior in C/C++).

lib/fizzy/execute.cpp
chfast: I think it is rather difficult to have generic formula for upper/lower boundaries. However, it is possible to compress these (e.g. for unsigned type these are `-1.0` and `2^N`. But I decided to leave these expanded at least until we have confirmation that all spectests are passed.

Having separate file for it is not ideal, but I wanted to check the values against math formulas used in unit tests.

lib/fizzy/trunc_boundaries.hpp
```diff
@@ -0,0 +1,69 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <cstdint>
+
+namespace fizzy
+{
+template <typename SrcT, typename Dst>
+struct trunc_boundaries;
```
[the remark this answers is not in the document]
chfast: Comment added.

test/unittests/execute_floating_point_test.cpp
```diff
@@ -66,4 +69,203 @@ TEST(execute_floating_point, f64_add)
...
+        EXPECT_THAT(execute(*instance, 0, {max_defined}), Result(max_defined_int));
+
+        // The lower boundary is:
+        // - for signed integers: -2^N - 1,
+        // - for unsigned integers: -1.
+        // However, the -2^N - 1 may be not representative in a float type so we compute it as
```
gumb0, line 234: I thnk this explanation how `lower_bound` is calculated should be in `trunc_boundaries.hpp`, too.

lib/fizzy/execute.cpp
gumb0: I think it looks fine now, I like that it's quite simple. The only thing is that  bounds look like magic values, they are not defined in spec. It needs a comment to explain where they come from.

test/unittests/execute_floating_point_test.cpp
```diff
@@ -66,4 +69,203 @@ TEST(execute_floating_point, f64_add)
...
+
+        const auto min_defined = std::nextafter(lower_boundary, FloatT{0});
+        EXPECT_THAT(execute(*instance, 0, {min_defined}), Result(min_defined_int));
+    }
+
+    {
+        // NaNs.
+        EXPECT_THAT(execute(*instance, 0, {FloatLimits::quiet_NaN()}), Traps());
+        EXPECT_THAT(execute(*instance, 0, {FloatLimits::signaling_NaN()}), Traps());
+        EXPECT_THAT(execute(*instance, 0, {nan<FloatT>()}), Traps());
```
gumb0, line 250: Can there be a NaN with sign bit set (`-nan<FloatT<()`)?

lib/fizzy/execute.cpp
chfast: Added comments.

test/unittests/execute_floating_point_test.cpp
chfast: Yes, it can be.

chfast: Added test cases for these.

circle.yml
```diff
@@ -271,9 +271,9 @@ jobs:
       - benchmark:
           min_time: "0.01"
       - spectest:
-          expected_passed: 5581
-          expected_failed: 9
-          expected_skipped: 6223
+          expected_passed: 5585
+          expected_failed: 13
```
axic, line 275: What are the new failed ones?

chfast: Should not be, I must update this.

test/unittests/execute_floating_point_test.cpp
chfast: Added documentation.

lib/fizzy/execute.cpp
```diff
@@ -324,6 +326,19 @@ inline DstT extend(SrcT in) noexcept
...
+
+    const auto in = stack.top().as<SrcT>();
```
axic, line 334:
```suggestion
    const auto input = stack.top().as<SrcT>();
```
?

test/unittests/execute_floating_point_test.cpp
```diff
@@ -66,4 +69,209 @@ TEST(execute_floating_point, f64_add)
...
+    (func (param f32) (result i32)
+      local.get 0
+      i32.trunc_f32_s
+    )
+    */
+    auto wasm = from_hex("0061736d0100000001060160017d017f030201000a070105002000a80b");
+
+    // Find and replace changeable values: types and the conversion instruction.
+    auto& param_type = wasm[wasm.find(static_cast<uint8_t>(ValType::f32))];
+    auto& result_type = wasm[wasm.find(static_cast<uint8_t>(ValType::i32))];
```
axic, line 187: Perhaps adding assertion that there are only a single occurrence of these valtypes?

lib/fizzy/trunc_boundaries.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+#pragma once
+
+#include <cstdint>
+
+namespace fizzy
+{
+/// Provides _exclusive_ lower and upper boundaries of the range
+/// for which SrcT -> DstT trunc operation has defined results.
+/// Theoretical values are (INTEGER_MIN - 1, INTEGER_MAX + 1) what is
```
gumb0, line 13:
```suggestion
/// Theoretical values are (INTEGER_MIN - 1, INTEGER_MAX + 1) which is
```

lib/fizzy/execute.cpp
```diff
@@ -324,6 +326,19 @@ inline DstT extend(SrcT in) noexcept
         return DstT{in};
 }
 
+template <typename SrcT, typename DstT>
+inline bool trunc(OperandStack& stack) noexcept
+{
+    using boundaries = trunc_boundaries<SrcT, DstT>;
+
+    const auto input = stack.top().as<SrcT>();
+    if (std::isnan(input) || input <= boundaries::lower || input >= boundaries::upper)
```
axic, line 335: Why doesn't this need to check for `inf`/`-inf`?

chfast: Because always: `-inf <= boundaries::lower` and `inf >= boundaries::upper`.

chfast: Oh, actually, this can be optimized :/

axic: Even if that is true (seemingly it is because the tests pass), we should have a static_assert here for that.

chfast: I'm not sure what assert you would like to have here.

axic:
```
-inf <= boundaries::lower
inf >= boundaries::lower
```

lib/fizzy/execute.cpp
```diff
@@ -324,6 +326,20 @@ inline DstT extend(SrcT in) noexcept
...
+{
+    using boundaries = trunc_boundaries<SrcT, DstT>;
+
+    const auto input = stack.top().as<SrcT>();
+    if (input > boundaries::lower && input < boundaries::upper)
+    {
+        stack.top() = static_cast<DstT>(input);
+        return true;
+    }
+    return false;
```
axic, line 340: After the optimisation I'd add `assert(std::isnan(input) || input == inf || input == -inf);` so that it matches the spec.

chfast: Ok, added correct variants of these asserts.

axic: Don't want to stretch this too much, but can we also have the static_assert for inf vs. boundaries? Obviously we cannot make the nan-check compile time :)

lib/fizzy/execute.cpp
chfast: No, this does not make sense to me. Every value is `>= -inf`.

chfast:
> Looks good to me otherwise.

So what now?

axic:
> So what now?

Dunno, merge?

PR #448, opened 2020-07-31
Replace static_casts with Value::as<>

axic:
This should have been done in #423.

Depends on #446.

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -691,7 +691,7 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value> 
             const auto br_table_size = read<uint32_t>(immediates);
             const auto arity = read<uint32_t>(immediates);
 
-            const auto br_table_idx = stack.pop();
+            const auto br_table_idx = static_cast<uint32_t>(stack.pop());
 
             const auto label_idx_offset = br_table_idx < br_table_size ?
                                               br_table_idx * BranchImmediateSize :
@@ -720,7 +720,7 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value> 
             const auto expected_type_idx = read<uint32_t>(immediates);
             assert(expected_type_idx < instance.module.typesec.size());
 
-            const auto elem_idx = stack.pop();
+            const auto elem_idx = static_cast<uint32_t>(stack.pop());
             if (elem_idx >= instance.table->size())
             {
                 trap = true;
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -328,7 +328,7 @@ template <typename DstT, typename SrcT = DstT>
 inline bool load_from_memory(
     bytes_view memory, OperandStack& stack, const uint8_t*& immediates) noexcept
 {
-    const auto address = static_cast<uint32_t>(stack.pop());
+    const auto address = stack.pop().as<uint32_t>();
     // NOTE: alignment is dropped by the parser
     const auto offset = read<uint32_t>(immediates);
     // Addressing is 32-bit, but we keep the value as 64-bit to detect overflows.
@@ -344,8 +344,10 @@ template <typename DstT>
 inline bool store_into_memory(
     bytes& memory, OperandStack& stack, const uint8_t*& immediates) noexcept
 {
-    const auto value = static_cast<DstT>(stack.pop());
-    const auto address = static_cast<uint32_t>(stack.pop());
+    // Could use `.as<DstT>()` would we have overloads for uint8_t/uint16_t,
+    // however it does not seem to be worth it for this single occasion.
+    const auto value = static_cast<DstT>(stack.pop().i64);
+    const auto address = stack.pop().as<uint32_t>();
     // NOTE: alignment is dropped by the parser
     const auto offset = read<uint32_t>(immediates);
     // Addressing is 32-bit, but we keep the value as 64-bit to detect overflows.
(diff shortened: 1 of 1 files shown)


lib/fizzy/execute.cpp
```diff
@@ -687,7 +688,8 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value>
             const auto br_table_size = read<uint32_t>(immediates);
             const auto arity = read<uint32_t>(immediates);
 
-            const auto br_table_idx = stack.pop();
+            // TODO: is uint64_t the correct stack type here?
+            const auto br_table_idx = stack.pop().as<uint64_t>();
```
axic, line 692: @gumb0 do you know this top of your head: what is the idx type for `br_table`?

lib/fizzy/execute.cpp
```diff
@@ -716,7 +718,8 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value>
             const auto expected_type_idx = read<uint32_t>(immediates);
             assert(expected_type_idx < instance.module.typesec.size());
 
-            const auto elem_idx = stack.pop();
+            // TODO: is this the correct stack type?
+            const auto elem_idx = stack.pop().as<uint64_t>();
```
axic, line 722: @gumb0 this here is elem_idx type for `call_indirect`

lib/fizzy/execute.cpp
```diff
@@ -716,7 +718,8 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value>
             const auto expected_type_idx = read<uint32_t>(immediates);
             assert(expected_type_idx < instance.module.typesec.size());
 
-            const auto elem_idx = stack.pop();
+            // TODO: is this the correct stack type?
```
axic, line 721: All these were implicitly `uint64_t` so far.

lib/fizzy/execute.cpp
```diff
@@ -344,8 +344,9 @@ template <typename DstT>
 inline bool store_into_memory(
     bytes& memory, OperandStack& stack, const uint8_t*& immediates) noexcept
 {
-    const auto value = static_cast<DstT>(stack.pop());
-    const auto address = static_cast<uint32_t>(stack.pop());
+    // TODO: consider adding uint8_t/uint16_t overloads to as()
```
chfast, line 347: Not good idea.

lib/fizzy/execute.cpp
```diff
@@ -344,8 +344,9 @@ template <typename DstT>
...
+    const auto value = static_cast<DstT>(stack.pop().as<uint64_t>());
```
chfast, line 348:
```suggestion
    const auto value = static_cast<DstT>(stack.pop().i64);
```

lib/fizzy/execute.cpp
chfast, line 692:
```suggestion
            const auto br_table_idx = stack.pop().i64;
```
Here and in other places.

gumb0: We have a nice table to look it up :)
https://github.com/wasmx/fizzy/blob/0728606b601960b51177f3fa7ed7cb2ef3e6110f/lib/fizzy/instructions.cpp#L29

gumb0: Here's the validation spec that defines input value as `i32`
https://webassembly.github.io/spec/core/valid/instructions.html#xref-syntax-instructions-syntax-instr-control-mathsf-br-table-l-ast-l-n

lib/fizzy/execute.cpp
gumb0: This one is also `i32` https://webassembly.github.io/spec/core/valid/instructions.html#xref-syntax-instructions-syntax-instr-control-mathsf-call-indirect-x

lib/fizzy/execute.cpp
```diff
@@ -806,14 +809,16 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value>
             if (idx < instance.imported_globals.size())
             {
                 assert(instance.imported_globals[idx].is_mutable);
-                *instance.imported_globals[idx].value = stack.pop();
+                // Using uint64_t as that equlas to the stack width.
+                *instance.imported_globals[idx].value = stack.pop().as<uint64_t>();
```
gumb0, line 813: After #446 left side will be `Value`

axic:
@chfast how about this now?

lib/fizzy/execute.cpp
```diff
@@ -1376,13 +1378,13 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value>
...
-            stack.push(static_cast<uint32_t>(stack.pop()));
+            stack.push(stack.pop().as<uint32_t>());
             break;
         }
         case Instr::i64_extend_i32_s:
         {
-            const auto value = static_cast<int32_t>(stack.pop());
-            stack.push(static_cast<uint64_t>(int64_t{value}));
+            const auto value = stack.pop().as<int32_t>();
+            stack.push(int64_t{value});
```
axic, line 1387: Do we actually need this `int64_t{}` here?

chfast: Yes, it is needed as this is the core of this instruction implementation (otherwise for `int32_t` type the value will be cap to 32 bits). But it may be better to move it to the live above: `int64_t{stack.pop().as<int32_t>()}`.

lib/fizzy/execute.cpp
```diff
@@ -691,7 +693,7 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value>
             const auto br_table_size = read<uint32_t>(immediates);
             const auto arity = read<uint32_t>(immediates);
 
-            const auto br_table_idx = stack.pop();
+            const auto br_table_idx = stack.pop().as<uint32_t>();
```
chfast, line 696: Make sure these functional changes are in a separate commit.

axic: Oh, the _idx stuff is a separate commit, also available as #461.

axic:
Squashed the commits. The first commit can be merged as part of this PR or separately under #461.

@chfast @gumb0 please give a final review

lib/fizzy/execute.cpp
axic: Could also squash it into a single one: `stack.push(int64_t{stack.pop().as<int32_t>()}`.

It all is a matter of preference, perhaps the current one is well balanced for readability.

chfast: Fine to leave it too.

PR #449, opened 2020-07-31
Implement f32/f64 comparison operators

axic:
Depends on #448.

Need to be cleaned up and unit tests.

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -1151,6 +1151,69 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value> 
             comparison_op(stack, std::greater_equal<uint64_t>());
             break;
         }
+
+        case Instr::f32_eq:
+        {
+            comparison_op(stack, std::equal_to<float>());
+            break;
+        }
+        case Instr::f32_ne:
+        {
+            comparison_op(stack, std::not_equal_to<float>());
+            break;
+        }
+        case Instr::f32_lt:
+        {
+            comparison_op(stack, std::less<float>());
+            break;
+        }
+        case Instr::f32_gt:
+        {
+            comparison_op<float>(stack, std::greater<float>());
+            break;
+        }
+        case Instr::f32_le:
+        {
+            comparison_op(stack, std::less_equal<float>());
+            break;
+        }
+        case Instr::f32_ge:
+        {
+            comparison_op(stack, std::greater_equal<float>());
+            break;
+        }
+
+        case Instr::f64_eq:
+        {
+            comparison_op(stack, std::equal_to<double>());
+            break;
+        }
+        case Instr::f64_ne:
+        {
+            comparison_op(stack, std::not_equal_to<double>());
+            break;
+        }
+        case Instr::f64_lt:
+        {
+            comparison_op(stack, std::less<double>());
+            break;
+        }
+        case Instr::f64_gt:
+        {
+            comparison_op<double>(stack, std::greater<double>());
+            break;
+        }
+        case Instr::f64_le:
+        {
+            comparison_op(stack, std::less_equal<double>());
+            break;
+        }
+        case Instr::f64_ge:
+        {
+            comparison_op(stack, std::greater_equal<double>());
+            break;
+        }
+
         case Instr::i32_clz:
         {
             unary_op(stack, clz32);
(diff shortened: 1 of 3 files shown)


test/unittests/execute_floating_point_test.cpp
```diff
@@ -66,4 +66,129 @@ TEST(execute_floating_point, f64_add)
...
+        std::nextafter(-Limits::max(), TypeParam{0}),
+        std::nextafter(-TypeParam{1.0}, -Limits::infinity()),
+        -TypeParam{1.0},
+        std::nextafter(-TypeParam{1.0}, TypeParam{0}),
+        std::nextafter(-Limits::min(), -Limits::infinity()),
+        -Limits::min(),
+        std::nextafter(-Limits::min(), TypeParam{0}),
+        std::nextafter(-Limits::denorm_min(), -Limits::infinity()),
+        -Limits::denorm_min(),
+        TypeParam{0},  // FIXME.
```
axic, line 140: What is the fixme here?

test/unittests/execute_floating_point_test.cpp
```diff
@@ -66,4 +66,129 @@ TEST(execute_floating_point, f64_add)
...
+                EXPECT_THAT(execute(*inst, ne, {a, b}), Result(uint32_t{i != j})) << a << "!=" << b;
+                EXPECT_THAT(execute(*inst, lt, {a, b}), Result(uint32_t{i < j})) << a << "<" << b;
+                EXPECT_THAT(execute(*inst, gt, {a, b}), Result(uint32_t{i > j})) << a << ">" << b;
+                EXPECT_THAT(execute(*inst, le, {a, b}), Result(uint32_t{i <= j})) << a << "<=" << b;
+                EXPECT_THAT(execute(*inst, ge, {a, b}), Result(uint32_t{i >= j})) << a << ">=" << b;
+            }
+        }
+    }
+
+    // Negative zero.
```
axic, line 187: Why are these needed if we have `-0` and `0` in `ordered_values`?

test/unittests/execute_floating_point_test.cpp
chfast: It was to add "negative zero" comparison.

test/unittests/execute_floating_point_test.cpp
chfast: We don't.

axic: Ah, its -1 and 0. Can we just have -0 there too?

lib/fizzy/execute.cpp
```diff
@@ -1392,33 +1392,81 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value>
             // effectively no-op
             break;
         }
-        case Instr::f32_add:
+        case Instr::f32_eq:
```
chfast, line 1395: Should be placed after `i64_ge_u`.

test/unittests/execute_floating_point_test.cpp
chfast: No.

lib/fizzy/execute.cpp

test/unittests/execute_floating_point_test.cpp
axic: Why?

gumb0: Why?

chfast: The values in the array must be strictly ordered, otherwise we don't know what result to expect. But `-0 == 0`.

gumb0: I don't see why they need to be strictly ordered, to get the expected result you always calculate it as `i op j` anyway. Also as at some iterations `i < j`, at others `j < i` (and at some `i == j`)

chfast: Please commit a change doing that because I don't see how this suppose to work.

gumb0: Oh I see, I misunderstood how it works first. But I pushed the way I thought it works first, it doesn't rely on strict ordering, what do you think?

test/unittests/execute_floating_point_test.cpp
```diff
@@ -185,12 +186,12 @@ TYPED_TEST(execute_floating_point_types, compare)
             }
             else
             {
-                EXPECT_THAT(execute(*inst, eq, {a, b}), Result(uint32_t{i == j})) << a << "==" << b;
-                EXPECT_THAT(execute(*inst, ne, {a, b}), Result(uint32_t{i != j})) << a << "!=" << b;
-                EXPECT_THAT(execute(*inst, lt, {a, b}), Result(uint32_t{i < j})) << a << "<" << b;
-                EXPECT_THAT(execute(*inst, gt, {a, b}), Result(uint32_t{i > j})) << a << ">" << b;
-                EXPECT_THAT(execute(*inst, le, {a, b}), Result(uint32_t{i <= j})) << a << "<=" << b;
-                EXPECT_THAT(execute(*inst, ge, {a, b}), Result(uint32_t{i >= j})) << a << ">=" << b;
+                EXPECT_THAT(execute(*inst, eq, {a, b}), Result(uint32_t{a == b})) << a << "==" << b;
```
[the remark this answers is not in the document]
axic: That is a good point, I'd rather document these requirements/decision as comments in the test.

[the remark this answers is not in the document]
gumb0: Ok, feel free to delete my commit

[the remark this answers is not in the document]
chfast: Added comments and dropped the commit.

PR #451, opened 2020-08-03
test: Support floating point results in Result matcher

gumb0:


test/utils/asserts.hpp
```diff
@@ -20,7 +20,15 @@ MATCHER(Result, "empty result")
 
 MATCHER_P(Result, value, "")  // NOLINT(readability-redundant-string-init)
 {
-    return !arg.trapped && arg.has_value && arg.value == uint64_t(value);
+    if (arg.trapped || !arg.has_value)
+        return false;
+
+    if constexpr (std::is_same_v<value_type, float>)
```
chfast, line 26: FYI, this can be also `if constexpr (std::is_floating_point_v<value_type>) { return arg.value.as<value_type>() == value; }`. But I think the current one is fine and probably easier to debug.

PR #452, opened 2020-08-04
test: Fix Result() for int32_t type

chfast:


gumb0:
`execute_numeric_test.cpp` has many cases like `Result(uint32_t(-42)))`, can we simplify it now?

chfast:
> `execute_numeric_test.cpp` has many cases like `Result(uint32_t(-42)))`, can we simplify it now?

Yes, done.

PR #454, opened 2020-08-04
test: Add floating-point utils

chfast:
These are helpers mostly to inspect and build NaN "values". Wasm has concepts of canonical and arithmetic NaNs, which are not present in C/C++.

Some future usage examples:
- https://github.com/wasmx/fizzy/pull/457/files#diff-5c88f1e384667cb0ce959addfcea39e3R101-R127
- https://github.com/wasmx/fizzy/pull/458/files#diff-5c88f1e384667cb0ce959addfcea39e3R166-R183

### DIFF
--- a/test/unittests/floating_point_utils_test.cpp
+++ b/test/unittests/floating_point_utils_test.cpp
@@ -1,0 +1,192 @@
+// Fizzy: A fast WebAssembly interpreter
+// Copyright 2020 The Fizzy Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+#include <gtest/gtest.h>
+#include <test/utils/floating_point_utils.hpp>
+
+using namespace fizzy::test;
+
+TEST(floating_point_utils, double_as_uint)
+{
+    EXPECT_EQ(FP(0.0).as_uint(), 0x0000000000000000);
+    EXPECT_EQ(FP(-0.0).as_uint(), 0x8000000000000000);
+    EXPECT_EQ(FP(FP64::Limits::infinity()).as_uint(), 0x7FF'0000000000000);
+    EXPECT_EQ(FP(-FP64::Limits::infinity()).as_uint(), 0xFFF'0000000000000);
+    EXPECT_EQ(FP(FP64::Limits::max()).as_uint(), 0x7FE'FFFFFFFFFFFFF);
+    EXPECT_EQ(FP(-FP64::Limits::max()).as_uint(), 0xFFE'FFFFFFFFFFFFF);
+    EXPECT_EQ(FP(FP64::Limits::min()).as_uint(), 0x001'0000000000000);
+    EXPECT_EQ(FP(-FP64::Limits::min()).as_uint(), 0x801'0000000000000);
+    EXPECT_EQ(FP(FP64::Limits::denorm_min()).as_uint(), 0x000'0000000000001);
+    EXPECT_EQ(FP(-FP64::Limits::denorm_min()).as_uint(), 0x800'0000000000001);
+    EXPECT_EQ(FP(1.0).as_uint(), 0x3FF'0000000000000);
+    EXPECT_EQ(FP(-1.0).as_uint(), 0xBFF'0000000000000);
+    EXPECT_EQ(FP(std::nextafter(1.0, 0.0)).as_uint(), 0x3FE'FFFFFFFFFFFFF);
+    EXPECT_EQ(FP(std::nextafter(-1.0, 0.0)).as_uint(), 0xBFE'FFFFFFFFFFFFF);
+    EXPECT_EQ(FP(FP64::nan(FP64::canon)).as_uint(), 0x7FF'8000000000000);
+    EXPECT_EQ(FP(-FP64::nan(FP64::canon)).as_uint(), 0xFFF'8000000000000);
+}
+
+TEST(floating_point_utils, binary_representation_implementation_defined)
+{
+    EXPECT_EQ(FP(FP64::Limits::quiet_NaN()).as_uint(), 0x7FF'8000000000000);
+    EXPECT_EQ(FP(FP64::Limits::quiet_NaN()).nan_payload(), 0x8000000000000);
+    EXPECT_EQ(FP(FP64::Limits::signaling_NaN()).nan_payload(), 0x4000000000000);
+
+    EXPECT_EQ(FP(FP32::Limits::quiet_NaN()).as_uint(), 0x7FC00000);
+    EXPECT_EQ(FP(FP32::Limits::quiet_NaN()).nan_payload(), 0x400000);
+    EXPECT_EQ(FP(FP32::Limits::signaling_NaN()).nan_payload(), 0x200000);
+}
+
+TEST(floating_point_utils, float_as_uint)
+{
+    EXPECT_EQ(FP(0.0f).as_uint(), 0x00000000);
+    EXPECT_EQ(FP(-0.0f).as_uint(), 0x80000000);
+    EXPECT_EQ(FP(FP32::Limits::infinity()).as_uint(), 0x7F800000);
+    EXPECT_EQ(FP(-FP32::Limits::infinity()).as_uint(), 0xFF800000);
+    EXPECT_EQ(FP(FP32::Limits::max()).as_uint(), 0x7F7FFFFF);
+    EXPECT_EQ(FP(-FP32::Limits::max()).as_uint(), 0xFF7FFFFF);
+    EXPECT_EQ(FP(FP32::Limits::min()).as_uint(), 0x00800000);
+    EXPECT_EQ(FP(-FP32::Limits::min()).as_uint(), 0x80800000);
+    EXPECT_EQ(FP(FP32::Limits::denorm_min()).as_uint(), 0x00000001);
+    EXPECT_EQ(FP(-FP32::Limits::denorm_min()).as_uint(), 0x80000001);
+    EXPECT_EQ(FP(1.0f).as_uint(), 0x3F800000);
+    EXPECT_EQ(FP(-1.0f).as_uint(), 0xBF800000);
+    EXPECT_EQ(FP(std::nextafter(1.0f, 0.0f)).as_uint(), 0x3F7FFFFF);
+    EXPECT_EQ(FP(std::nextafter(-1.0f, 0.0f)).as_uint(), 0xBF7FFFFF);
+    EXPECT_EQ(FP(FP32::nan(FP32::canon)).as_uint(), 0x7FC00000);
+    EXPECT_EQ(FP(-FP32::nan(FP32::canon)).as_uint(), 0xFFC00000);
+}
+
+TEST(floating_point_utils, double_from_uint)
+{
+    EXPECT_EQ(FP(uint64_t{0x0000000000000000}).value, 0.0);
+    EXPECT_EQ(FP(uint64_t{0x8000000000000000}).value, -0.0);
+    EXPECT_EQ(FP(uint64_t{0x3FF'000000000DEAD}).value, 0x1.000000000DEADp0);
+    EXPECT_EQ(FP(uint64_t{0xBFF'000000000DEAD}).value, -0x1.000000000DEADp0);
+    EXPECT_EQ(FP(uint64_t{0x7FF'0000000000000}).value, FP64::Limits::infinity());
+    EXPECT_EQ(FP(uint64_t{0xFFF'0000000000000}).value, -FP64::Limits::infinity());
+}
+
+TEST(floating_point_utils, float_from_uint)
+{
+    EXPECT_EQ(FP(uint32_t{0x00000000}).value, 0.0f);
+    EXPECT_EQ(FP(uint32_t{0x80000000}).value, -0.0f);
+    EXPECT_EQ(FP(uint32_t{0x3FEF5680}).value, 0x1.DEADp0f);
+    EXPECT_EQ(FP(uint32_t{0xBFEF5680}).value, -0x1.DEADp0f);
+    EXPECT_EQ(FP(uint32_t{0x7F800000}).value, FP32::Limits::infinity());
+    EXPECT_EQ(FP(uint32_t{0xFF800000}).value, -FP32::Limits::infinity());
+}
+
+TEST(floating_point_utils, double_nan_payload)
+{
+    constexpr auto inf = FP64::Limits::infinity();
+    const auto qnan = FP64::nan(FP64::canon);
+
+    EXPECT_EQ(FP(0.0).nan_payload(), 0);
+    EXPECT_EQ(FP(FP64::nan(1)).nan_payload(), 1);
+    EXPECT_EQ(FP(FP64::nan(FP64::canon + 1)).nan_payload(), FP64::canon + 1);
+    EXPECT_EQ(FP(qnan).nan_payload(), FP64::canon);
+    EXPECT_EQ(FP(qnan + 1.0).nan_payload(), FP64::canon);
+    EXPECT_EQ(FP(inf - inf).nan_payload(), FP64::canon);
+    EXPECT_EQ(FP(inf * 0.0).nan_payload(), FP64::canon);
+}
+
+TEST(floating_point_utils, float_nan_payload)
+{
+    constexpr auto inf = FP32::Limits::infinity();
+    const auto qnan = FP32::nan(FP32::canon);
+
+    EXPECT_EQ(FP(0.0f).nan_payload(), 0);
+    EXPECT_EQ(FP(FP32::nan(1)).nan_payload(), 1);
+    EXPECT_EQ(FP(FP32::nan(FP32::canon + 1)).nan_payload(), FP32::canon + 1);
+    EXPECT_EQ(FP(qnan).nan_payload(), FP32::canon);
+    EXPECT_EQ(FP(qnan + 1.0f).nan_payload(), FP32::canon);
+    EXPECT_EQ(FP(inf - inf).nan_payload(), FP32::canon);
+    EXPECT_EQ(FP(inf * 0.0f).nan_payload(), FP32::canon);
+}
+
+TEST(floating_point_utils, double_nan)
+{
+    EXPECT_TRUE(std::isnan(FP64::nan(FP64::canon)));
+    EXPECT_TRUE(std::isnan(FP64::nan(1)));
+    EXPECT_TRUE(std::isnan(FP64::nan(0xDEADBEEF)));
+    EXPECT_TRUE(std::isnan(FP64::nan(0xDEADBEEFBEEEF)));
+    EXPECT_FALSE(std::isnan(FP64::nan(0)));
+
+    EXPECT_EQ(FP{FP64::nan(FP64::canon)}.nan_payload(), FP64::canon);
+
+    EXPECT_EQ(FP{FP64::nan(FP64::canon)}.as_uint(), 0x7FF'8000000000000);
+    EXPECT_EQ(FP{FP64::nan(0xDEADBEEF)}.as_uint(), 0x7FF'00000DEADBEEF);
+    EXPECT_EQ(FP{FP64::nan(0xDEADBEEFBEEEF)}.as_uint(), 0x7FF'DEADBEEFBEEEF);
+}
+
+TEST(floating_point_utils, float_nan)
+{
+    EXPECT_TRUE(std::isnan(FP32::nan(FP32::canon)));
+    EXPECT_TRUE(std::isnan(FP32::nan(1)));
+    EXPECT_TRUE(std::isnan(FP32::nan(0x7fffff)));
+    EXPECT_TRUE(std::isnan(FP32::nan(0x400001)));
+    EXPECT_FALSE(std::isnan(FP32::nan(0)));
+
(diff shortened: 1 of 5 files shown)


test/utils/floating_point_utils.hpp
```diff
@@ -0,0 +1,69 @@
...
+{
+template <typename T, typename = std::enable_if<std::is_floating_point_v<T>>>
+struct FPUtils
+{
+    using Limits = std::numeric_limits<T>;
+
+    static constexpr auto NumMantissaBits = Limits::digits - 1;
+    static constexpr auto MantissaMask = (uint64_t{1} << NumMantissaBits) - 1;
+
+    static constexpr auto CanonicalNaNPayload = uint64_t{1} << (NumMantissaBits - 1);
```
gumb0, line 22: Do you have a good link about where it's defined as canonical?

test/utils/floating_point_utils.hpp
```diff
@@ -0,0 +1,69 @@
...
+uint64_t get_nan_payload(T value) noexcept
+{
+    if (!std::isnan(value))
+        return 0;  // NaN payload is never 0.
+
+    return as_uint64(value) & FPUtils<T>::MantissaMask;
+}
+
+template <typename T, typename = std::enable_if<std::is_floating_point_v<T>>>
+T nan(uint64_t payload = FPUtils<T>::CanonicalNaNPayload) noexcept
```
gumb0, line 62: Couldn't we use `std::nan`/`std::nanf` for this?

test/utils/floating_point_utils.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+    static constexpr auto CanonicalNaNPayload = uint64_t{1} << (NumMantissaBits - 1);
+};
+
+template <typename T, typename = std::enable_if<std::is_floating_point_v<T>>>
+uint64_t as_uint64(T value) noexcept
+{
+    constexpr auto num_bytes = sizeof(T);
+    uint8_t bytes[num_bytes];
+    __builtin_memcpy(bytes, &value, num_bytes);
```
gumb0, line 30: Do we need to copy, cant' we just `reinterpret_cast<uint8_t*>(&value)`?

test/utils/floating_point_utils.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+    uint64_t u = 0;
+    for (size_t i = 0; i < num_bytes; ++i)
+        u |= uint64_t{bytes[i]} << (i * 8);  // little-endian
```
gumb0, line 34: Could do `i += 8` each iteration instead of `i * 8`, but maybe not very important

test/utils/floating_point_utils.hpp
chfast: True, but they do it in an implementation-defined manner related to printing NaN "values". Maybe there is a way to find a combination of chars that gives wasm canonical and arithmetic NaNs, but I decided to go with my own NaN inspection helpers. Also to learn the internal structure of floats.

test/utils/floating_point_utils.hpp
chfast: That's actually undefined behavior, but I don't think compiler mind it. After revisiting, there should be simpler implementations ignoring endianness.

test/utils/floating_point_utils.hpp
chfast: Nah.

test/utils/floating_point_utils.hpp
```diff
@@ -0,0 +1,81 @@
...
+}
+
+
+template <typename T>
+struct FP
+{
+    static_assert(std::is_floating_point_v<T>);
+
+    /// Shortcut to numeric_limits.
+    using limits = std::numeric_limits<T>;
```
gumb0, line 31: Why not `Limits`, `UintType`? I think it would be easier to see, that these are types.

test/utils/floating_point_utils.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+    /// The unsigned integer type matching the size of this floating-point type.
+    using uint_type = std::conditional_t<std::is_same_v<T, float>, uint32_t, uint64_t>;
+    static_assert(sizeof(T) == sizeof(uint_type));
```
gumb0, line 35: As there's `long double`, maybe change the `static_assert(std::is_floating_point_v<T>);` above to allow only `float` and `double`?

Also maybe `static_assert(std::numeric_limits<T>::is_iec559)`?

test/utils/floating_point_utils.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+    /// The number of mantissa bits in the binary representation.
+    static constexpr auto num_mantissa_bits = limits::digits - 1;
+
+    /// The binary mask of the mantissa part of the binary representation.
+    static constexpr auto mantissa_mask = (uint_type{1} << num_mantissa_bits) - 1;
+
+    /// The payload of the canonical NaN (only the top bit set).
+    /// See: https://webassembly.github.io/spec/core/syntax/values.html#canonical-nan.
+    static constexpr auto canon = uint_type{1} << (num_mantissa_bits - 1);
```
gumb0, line 45: Why not `nan_canonical_payload`?

test/utils/floating_point_utils.hpp
```diff
@@ -0,0 +1,81 @@
...
+    uint_type nan_payload() const noexcept
+    {
+        return std::isnan(value) ? (as_uint() & mantissa_mask) : 0;
+    }
+
+    /// Build the NaN value with the given payload.
+    static T nan(uint_type payload = canon) noexcept
+    {
+        const auto nan_payload = payload & mantissa_mask;
+        const auto nan_exponent = (uint_type{limits::max_exponent} << 1) - 1;
```
gumb0, line 66: This line is a bit confusing to me...
Why `<< 1`?

test/utils/floating_point_utils.hpp
chfast: Can be. It just looks little weird when mixing with stuff from `std::numeric_limits`. Can constants stay lowercase?

test/utils/floating_point_utils.hpp
chfast: The size comparison should eliminated `long double`, but I will add this check also. Checking `is_iec559` may be also good for main implementation.

test/utils/floating_point_utils.hpp
chfast: Can be, but is rather annoying in places where it is used. And wasm spec names this constants as `canon`: https://webassembly.github.io/spec/core/syntax/values.html#aux-canon

test/utils/floating_point_utils.hpp
chfast: The exponent is kinda signed value, e.g. for `float32` exponent has 8 bits with values range `[-127, 128]`. To get exponent with all bits set you need something like `2 * max_exponent - 1`. Can change to multiplication if that's better.

test/utils/floating_point_utils.hpp
gumb0: Yes, my preference would be CamelCase for types, lowercase for constants is fine

test/utils/floating_point_utils.hpp
gumb0: Ok, fine to keep `canon`

test/utils/floating_point_utils.hpp
chfast: Added some unit tests comparing with `std::nan()`. It produces only quiet (aka arithmetic) nans. I prefer testing others (i.e. signaling) nans too.

test/utils/floating_point_utils.hpp
chfast: Link added.

test/utils/floating_point_utils.hpp
chfast: Done.

test/utils/floating_point_utils.hpp
chfast: Added `static_assert`.

test/utils/floating_point_utils.hpp
```diff
@@ -0,0 +1,82 @@
...
+    UintType nan_payload() const noexcept
+    {
+        return std::isnan(value) ? (as_uint() & mantissa_mask) : 0;
+    }
+
+    /// Build the NaN value with the given payload.
+    static T nan(UintType payload = canon) noexcept
+    {
+        const auto nan_payload = payload & mantissa_mask;
+        const auto nan_exponent = (UintType{Limits::max_exponent} << 1) - 1;
```
gumb0, line 67: `constexpr`?
Also maybe put it outside the function, close to `mantissa_mask`

chfast: OK

chfast: Done.

gumb0:
Tests for `FP(UintType u)` are missing

test/utils/floating_point_utils.hpp
```diff
@@ -0,0 +1,89 @@
...
+    static_assert(sizeof(T) == sizeof(UintType));
+
+    /// The number of mantissa bits in the binary representation.
+    static constexpr auto num_mantissa_bits = Limits::digits - 1;
+
+    /// The binary mask of the mantissa part of the binary representation.
+    static constexpr auto mantissa_mask = (UintType{1} << num_mantissa_bits) - 1;
+
+    /// The number of exponent bits (E) in the binary representation.
+    /// Produced out of Limits::max_exponent which is 2^
```
gumb0, line 45: which is `E / 2` ? maybe it doesn't explain much...

test/utils/floating_point_utils.hpp
```diff
@@ -0,0 +1,89 @@
...
+    static constexpr auto num_exponent_bits = 2 * Limits::max_exponent;
```
gumb0, line 46: Could be also `sizeof(T) * 8 - num_mantissa_bits -1`

test/utils/floating_point_utils.hpp
```diff
@@ -0,0 +1,89 @@
...
+
+    /// The binary mask of the exponent part of the binary representation, i.e. 2^E - 1.
+    /// This is also exponent value (all exponent bits set) for NaNs.
+    /// Produced from Limits::max_exponent which is 2^(E-1).
+    static constexpr auto exponent_mask = 2 * UintType{Limits::max_exponent} - 1;
```
gumb0, line 51: did you want to use `num_exponent_bits` here?

test/utils/floating_point_utils.hpp
chfast: I included something what was suppose to be deleted. Anyway, I used your suggestion to compute `num_exponent_bits`.

test/utils/floating_point_utils.hpp
chfast: No, it was suppose to be the mask, but in the end I changed that to `nan_exponent` as the mask would be at the wrong position.

chfast:
> Tests for `FP(UintType u)` are missing

Added some tests for that. They are also used in `nan()`.

test/utils/asserts.hpp
```diff
@@ -23,10 +24,8 @@ MATCHER_P(Result, value, "")  // NOLINT(readability-redundant-string-init)
     if (arg.trapped || !arg.has_value)
         return false;
 
-    if constexpr (std::is_same_v<value_type, float>)
-        return arg.value.f32 == value;
-    else if constexpr (std::is_same_v<value_type, double>)
-        return arg.value.f64 == value;
+    if constexpr (std::is_floating_point_v<value_type>)
+        return arg.value.template as<value_type>() == fizzy::test::FP{value};
```
axic, line 28: This C++ syntax of `template` just to be able to use `as<value_type>` is terribly, but it is what it is.

axic: Why is this beneficial over the previous one, btw?

test/utils/floating_point_utils.hpp
```diff
@@ -0,0 +1,89 @@
...
+    /// Return unsigned integer with the binary representation of the value.
+    UintType as_uint() const noexcept { return bit_cast<UintType>(value); }
+
+    /// Returns NaN payload if the value is a NaN, otherwise 0 (NaN payload is never 0).
+    UintType nan_payload() const noexcept
+    {
+        return std::isnan(value) ? (as_uint() & mantissa_mask) : 0;
+    }
+
+    /// Build the NaN value with the given payload.
```
axic, line 69: Is this the canonical one?

axic:
Out of curiousity: could we add tests to the main cmake (next to the bigendian tests) which ensures the platform has the proper IEEE-754 / NaN support?

test/utils/floating_point_utils.hpp
chfast: I think I can make the argument explicit.

test/utils/asserts.hpp
chfast: The `FP` is comparing binary representation.

The `template` is a mandatory hint to the compiler that the `<` in `as<value_type>` is a template bracket. Fortunately, GCC is giving nice hints when you forget it.

chfast:
> Out of curiousity: could we add tests to the main cmake (next to the bigendian tests) which ensures the platform has the proper IEEE-754 / NaN support?

See https://en.cppreference.com/w/cpp/language/class_template_argument_deduction.

test/unittests/floating_point_utils_test.cpp
```diff
@@ -0,0 +1,162 @@
...
+    EXPECT_EQ(FP(-FP64::Limits::infinity()).as_uint(), 0xFFF'0000000000000);
+    EXPECT_EQ(FP(FP64::Limits::max()).as_uint(), 0x7FE'FFFFFFFFFFFFF);
+    EXPECT_EQ(FP(-FP64::Limits::max()).as_uint(), 0xFFE'FFFFFFFFFFFFF);
+    EXPECT_EQ(FP(FP64::Limits::min()).as_uint(), 0x001'0000000000000);
+    EXPECT_EQ(FP(-FP64::Limits::min()).as_uint(), 0x801'0000000000000);
+    EXPECT_EQ(FP(FP64::Limits::denorm_min()).as_uint(), 0x000'0000000000001);
+    EXPECT_EQ(FP(-FP64::Limits::denorm_min()).as_uint(), 0x800'0000000000001);
+    EXPECT_EQ(FP(1.0).as_uint(), 0x3FF'0000000000000);
+    EXPECT_EQ(FP(-1.0).as_uint(), 0xBFF'0000000000000);
+    EXPECT_EQ(FP(std::nextafter(1.0, 0.0)).as_uint(), 0x3FE'FFFFFFFFFFFFF);
```
axic, line 24: Damn these helpers, so this is the `1.0 - <smallest representable value>`

chfast: Not exactly. This is the next representable value after `1.0` towards `0.0`. But should be something like `0.9999...`.

test/utils/floating_point_utils.hpp
```diff
@@ -0,0 +1,95 @@
...
+#include <limits>
+#include <type_traits>
+
+namespace fizzy::test
+{
+/// Simple implementation of C++20's std::bit_cast.
+template <typename DstT, typename SrcT>
+DstT bit_cast(SrcT x) noexcept
+{
+    DstT z;
```
axic, line 18: x, z are an interesting choice of combination :)

test/unittests/floating_point_utils_test.cpp
```diff
@@ -0,0 +1,162 @@
...
+    EXPECT_NE(FP{one}, inf);
+    EXPECT_NE(one, FP{inf});
+
+    EXPECT_NE(FP{one}, FP{nan});
+    EXPECT_NE(FP{one}, nan);
+    EXPECT_NE(one, FP{nan});
+
+    EXPECT_NE(FP{inf}, FP{nan});
+    EXPECT_NE(FP{inf}, nan);
+    EXPECT_NE(inf, FP{nan});
```
gumb0, line 161: Maybe also compare NaNs with different payload

test/unittests/floating_point_utils_test.cpp
```diff
@@ -0,0 +1,162 @@
...
+    EXPECT_EQ(FP(qnan + 1.0f).nan_payload(), FP32::canon);
+    EXPECT_EQ(FP(inf - inf).nan_payload(), FP32::canon);
+    EXPECT_EQ(FP(inf * 0.0f).nan_payload(), FP32::canon);
+}
+
+TEST(floating_point_utils, double_nan)
+{
+    EXPECT_TRUE(std::isnan(FP64::nan(FP64::canon)));
+    EXPECT_TRUE(std::isnan(FP64::nan(1)));
+    EXPECT_TRUE(std::isnan(FP64::nan(0xdeadbeef)));
```
gumb0, line 100: Nitpick: most numbers in this file use uppercase hex digits, but some are lowercase

test/utils/floating_point_utils.hpp
chfast: Took it from Go's BigInt, where `z` is always the return value. Want me to change it?

test/unittests/floating_point_utils_test.cpp
chfast: Fixed.

test/unittests/floating_point_utils_test.cpp
chfast: Added.

test/utils/floating_point_utils.hpp
axic: It is fine, does not matter too much. Would have used something like from/to, a/b, x/y, in/out -- but many of these could be subject to some clashes with certain compilers, so not worth the hassle.

test/unittests/floating_point_utils_test.cpp
```diff
@@ -0,0 +1,179 @@
...
+    EXPECT_EQ(FP(std::nan("1")).nan_payload(), FP64::canon + 1);
+    EXPECT_EQ(FP(std::nan("0xDEAD")).nan_payload(), FP64::canon + 0xDEAD);
+}
+
+TEST(floating_point_utils, compare_double)
+{
+    const auto one = 1.0;
+    const auto inf = FP64::Limits::infinity();
+    const auto cnan = FP64::nan(FP64::canon);
+    const auto snan = FP64::nan(1);
```
axic, line 138: What is snan?

chfast: signaling nan

test/utils/floating_point_utils.hpp
```diff
@@ -0,0 +1,95 @@
...
+    static_assert(sizeof(x) == sizeof(z));
+    __builtin_memcpy(&z, &x, sizeof(x));
+    return z;
+}
+
+/// A wrapper for floating-point types with inspection/construction/comparison utilities.
+template <typename T>
+struct FP
+{
+    static_assert(std::is_same_v<T, float> || std::is_same_v<T, double>);
```
axic, line 28: Hm, do you also want to add a `static_assert` for float/double being iee754 or fine as we have it in value.h and/or the limits check below?

chfast: The `static_assert(Limits::is_iec559);` does exactly that, except only when `FP` is instantiated.

test/unittests/floating_point_utils_test.cpp
```diff
@@ -0,0 +1,179 @@
...
+    EXPECT_EQ(FP(inf * 0.0).nan_payload(), FP64::canon);
+}
+
+TEST(floating_point_utils, float_nan_payload)
+{
+    constexpr auto inf = FP32::Limits::infinity();
+    constexpr auto qnan = FP32::Limits::quiet_NaN();
+
+    EXPECT_EQ(FP(0.0f).nan_payload(), 0);
+    EXPECT_EQ(FP(FP32::Limits::signaling_NaN()).nan_payload(), 0x200000);
```
gumb0, line 89: The exact payloads of `Limits::quiet_NaN()` and `Limits::signaling_NaN()` are implementation defined, aren't they? So maybe not worth it to put it into tests? Or maybe add a comment about this.

test/utils/floating_point_utils.hpp
```diff
@@ -0,0 +1,95 @@
...
+    friend bool operator==(FP a, FP b) noexcept { return a.as_uint() == b.as_uint(); }
+    friend bool operator==(FP a, T b) noexcept { return a == FP{b}; }
+    friend bool operator==(T a, FP b) noexcept { return FP{a} == b; }
+
+    friend bool operator!=(FP a, FP b) noexcept { return !(a == b); }
+    friend bool operator!=(FP a, T b) noexcept { return a != FP{b}; }
+    friend bool operator!=(T a, FP b) noexcept { return FP{a} != b; }
+};
+
+FP(uint64_t)->FP<double>;
```
gumb0, line 90: Nitpick: these two could be ordered float, then double, like `using` below

test/unittests/floating_point_utils_test.cpp
chfast: Moved to separate tests.

test/utils/floating_point_utils.hpp
chfast: Reordered.

PR #455, opened 2020-08-04
Implement int->float convert instructions

chfast:
Here situation is simple - we convert an integer to the closer floating point value. No special cases as range of `f32` is greater than `i64`.

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -328,9 +328,12 @@ inline DstT extend(SrcT in) noexcept
         return DstT{in};
 }
 
+/// Converts the top stack item by truncating a float value to an integer value.
 template <typename SrcT, typename DstT>
 inline bool trunc(OperandStack& stack) noexcept
 {
+    static_assert(std::is_floating_point_v<SrcT>);
+    static_assert(std::is_integral_v<DstT>);
     using boundaries = trunc_boundaries<SrcT, DstT>;
 
     const auto input = stack.top().as<SrcT>();
@@ -343,6 +346,15 @@ inline bool trunc(OperandStack& stack) noexcept
         return true;
     }
     return false;
+}
+
+/// Converts the top stack item from an integer value to a float value.
+template <typename SrcT, typename DstT>
+inline void convert(OperandStack& stack) noexcept
+{
+    static_assert(std::is_integral_v<SrcT>);
+    static_assert(std::is_floating_point_v<DstT>);
+    stack.top() = static_cast<DstT>(stack.top().as<SrcT>());
 }
 
 template <typename DstT, typename SrcT = DstT>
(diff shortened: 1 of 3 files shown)


lib/fizzy/execute.cpp
```diff
@@ -1473,6 +1479,46 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value>
...
+        {
+            convert<int64_t, double>(stack);
+            break;
+        }
+        case Instr::f64_convert_i64_u:
+        {
+            convert<uint64_t, double>(stack);
+            break;
+        }
         case Instr::f32_add:
```
axic, line 1522: Hm, these seem to be out of order. We try to maintain the opcode layout in the switch statement.

test/unittests/execute_floating_point_test.cpp
```diff
@@ -66,4 +69,326 @@ TEST(execute_floating_point, f64_add)
...
+    opcode = static_cast<uint8_t>(TypeParam::opcode);
+
+    auto instance = instantiate(parse(wasm));
+
+    EXPECT_THAT(execute(*instance, 0, {IntT{0}}), Result(FloatT{0}));
+    EXPECT_THAT(execute(*instance, 0, {IntT{1}}), Result(FloatT{1}));
+
+    const auto max = IntLimits::max();
+    const auto exact = IntLimits::digits < FloatLimits::digits;
+    const auto max_expected = std::pow(FloatT{2}, FloatT{IntLimits::digits}) - FloatT{exact};
```
gumb0, line 383: Can you explain this trick?

test/unittests/execute_floating_point_test.cpp
```diff
@@ -66,4 +69,326 @@ TEST(execute_floating_point, f64_add)
...
+    EXPECT_THAT(execute(*instance, 0, {max}), Result(max_expected));
+
+    if constexpr (IntLimits::is_signed)
+    {
+        static_assert(std::is_same_v<decltype(-max), IntT>);
+        EXPECT_THAT(execute(*instance, 0, {-max}), Result(-max_expected));
+
+        const auto min_expected = -std::pow(FloatT{2}, FloatT{IntLimits::digits});
```
gumb0, line 391: I wonder how `std::pow` is implemented, is it guaranteed to always have the exact result for power of 2?

chfast: I don't think `pow()` matters here, but floating point types can represent all powers of 2 around integer ranges due to having the exponent part which produces `2^e` values.

test/unittests/execute_floating_point_test.cpp
```diff
@@ -66,4 +69,326 @@ TEST(execute_floating_point, f64_add)
...
+    result_type = static_cast<uint8_t>(TypeParam::dst_valtype);
+    opcode = static_cast<uint8_t>(TypeParam::opcode);
+
+    auto instance = instantiate(parse(wasm));
+
+    EXPECT_THAT(execute(*instance, 0, {IntT{0}}), Result(FloatT{0}));
+    EXPECT_THAT(execute(*instance, 0, {IntT{1}}), Result(FloatT{1}));
+
+    const auto max = IntLimits::max();
+    const auto exact = IntLimits::digits < FloatLimits::digits;
```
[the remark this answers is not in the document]
chfast: Changed.

test/unittests/execute_floating_point_test.cpp
chfast: Explained.

test/unittests/execute_floating_point_test.cpp
```diff
@@ -279,3 +343,61 @@ TYPED_TEST(execute_floating_point_trunc, trunc)
...
+      local.get 0
+      f32.convert_i32_s
+    )
+    */
+    auto wasm = from_hex("0061736d0100000001060160017f017d030201000a070105002000b20b");
+
+    // Find and replace changeable values: types and the conversion instruction.
+    auto& param_type = wasm[wasm.find(static_cast<uint8_t>(ValType::i32))];
+    auto& result_type = wasm[wasm.find(static_cast<uint8_t>(ValType::f32))];
+    auto& opcode = wasm[wasm.find(static_cast<uint8_t>(Instr::f32_convert_i32_s))];
```
axic, line 377: Can you add the asserts for count like in the other PR?

lib/fizzy/execute.cpp

test/unittests/execute_floating_point_test.cpp

lib/fizzy/execute.cpp
```diff
@@ -343,6 +343,12 @@ inline bool trunc(OperandStack& stack) noexcept
     return false;
 }
 
+template <typename SrcT, typename DstT>
+inline void convert(OperandStack& stack) noexcept
+{
+    stack.top() = static_cast<DstT>(stack.top().as<SrcT>());
```
gumb0, line 349: It's getting difficult to see which of these small helpers are suitable for which types. Maybe we should add `static_assert`  here that `SrcT` is integer and `DstT` is floating point?

chfast: That's why I used `FloatT` and `IntT` in tests. I will try doing something about it here.

axic: These static asserts for types are nice, we should have them in each helper where it makes sense.

PR #456, opened 2020-08-04
Implement floating point memory instructions

gumb0:


lib/fizzy/execute.cpp
```diff
@@ -340,12 +345,21 @@ inline bool load_from_memory(
...
+        return static_cast<T>(stack.pop().i64);
+}
+
 template <typename DstT>
 inline bool store_into_memory(
     bytes& memory, OperandStack& stack, const uint8_t*& immediates) noexcept
 {
-    const auto value = static_cast<DstT>(stack.pop());
-    const auto address = static_cast<uint32_t>(stack.pop());
+    const auto value = pop_store_memory_arg<DstT>(stack);
```
chfast, line 361: Why `stack.pop().template as<DstT>()` does not work here?

lib/fizzy/execute.cpp
```diff
@@ -315,7 +315,12 @@ inline T load(bytes_view input, size_t offset) noexcept
 template <typename DstT, typename SrcT>
 inline DstT extend(SrcT in) noexcept
 {
-    if constexpr (std::is_signed<SrcT>::value)
+    if constexpr (std::is_floating_point_v<SrcT>)
```
chfast, line 318: I don't think I like additional levels of generic helpers. But in this case I propose simpler implementation.
```
if constexpr (std::is_same_v<SrcT, DstT>)
    return in;

// Handle else...
```

lib/fizzy/execute.cpp
gumb0: There's no `as` for `uint8_t`, `uint16_t`

lib/fizzy/execute.cpp
```diff
@@ -340,12 +342,21 @@ inline bool load_from_memory(
     return true;
 }
 
+template <typename T>
+inline T pop_store_memory_arg(OperandStack& stack) noexcept
```
chfast, line 346: I don't see reason to pass `OperandStack` in. Better to take `Value` as argument and make it similar to `extend`. Maybe call it `shrink` or `trunc` then?

test/spectests/spectests.cpp
```diff
@@ -58,6 +58,22 @@ fizzy::bytes load_wasm_file(const fs::path& json_file_path, std::string_view fil
         std::istreambuf_iterator<char>{wasm_file}, std::istreambuf_iterator<char>{});
 }
 
+bool float_values_equal(fizzy::Value value1, fizzy::Value value2) noexcept
+{
+    if (std::isnan(value1.as<float>()))
+        return value1.as<uint32_t>() == value2.as<uint32_t>();  // compare binary representations
```
axic, line 64: Why not always compare the binary representation?

gumb0: This was changed to using `FP` utility, please review https://github.com/wasmx/fizzy/pull/460 first

axic:
Please rebase

test/unittests/execute_floating_point_test.cpp
```diff
@@ -417,3 +417,229 @@ TYPED_TEST(execute_floating_point_trunc, trunc)
...
+{
+    /* wat2wasm
+    (memory 1 1)
+    (data (i32.const 0) "\00\00\00\00") ;; 0
+    (data (i32.const 4) "\b6\f3\9d\3f") ;; 1.234
+    (data (i32.const 8) "\b6\f3\9d\bf") ;; -1.234
+    (data (i32.const 12) "\00\00\c0\7f") ;; canonical NaN
+    (data (i32.const 16) "\01\00\c0\7f") ;; arithmetic NaN
+    (data (i32.const 20) "\01\00\80\7f") ;; signaling NaN
+    (func (param i32) (result f32)
```
axic, line 431: Can you add all the representations from `floating_point_utils`, i.e. include infinity?

test/unittests/execute_floating_point_test.cpp
```diff
@@ -417,3 +417,229 @@ TYPED_TEST(execute_floating_point_trunc, trunc)
...
+{
+    /* wat2wasm
+    (memory 1 1)
+    (data (i32.const 0)  "\00\00\00\00\00\00\00\00") ;; 0
+    (data (i32.const 8)  "\58\39\b4\c8\76\be\f3\3f") ;; 1.234
+    (data (i32.const 16) "\58\39\b4\c8\76\be\f3\bf") ;; -1.234
+    (data (i32.const 24) "\00\00\00\00\00\00\f8\7f") ;; canonical NaN
+    (data (i32.const 32) "\01\00\00\00\00\00\f8\7f") ;; arithmetic NaN
+    (data (i32.const 40) "\01\00\00\00\00\00\f0\7f") ;; signaling NaN
+    (func (param i32) (result f64)
```
axic, line 484: Same here.

test/unittests/execute_floating_point_test.cpp
gumb0: Added -0 and infinities, also to store tests

axic: Don't want to stretch it too much, but would it make sense having every case form` floating_point_utils:double_as_uint`, etc?

gumb0: More cases would ask for a general table approach... I'm inclined to say it's not worth it

chfast: > Don't want to stretch it too much, but would it make sense having every case form` floating_point_utils:double_as_uint`, etc?

These values are not very specific. As it is is fine to me.

axic: I think they are nice examples though, and shows if any part (beginning or end) is cut from loading/storing.

lib/fizzy/execute.cpp
```diff
@@ -359,13 +361,24 @@ inline bool load_from_memory(
...
 }
 
+template <typename DstT>
+inline DstT shrink(Value value) noexcept
+{
+    if constexpr (std::is_floating_point_v<DstT>)
+        return value.as<DstT>();
+    else
+    {
+        // Could use `.as<DstT>()` would we have overloads for uint8_t/uint16_t,
```
axic, line 371: Perhaps add a static assert for is_integral? Not sure it is important.

PR #458, opened 2020-08-04
Implement f32.demote_f64 instruction

chfast:
image

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -541,6 +541,15 @@ inline constexpr T fmax(T a, T b) noexcept
     return a < b ? b : a;
 }
 
+__attribute__((no_sanitize("float-cast-overflow"))) inline constexpr float demote(
+    double value) noexcept
+{
+    // The float-cast-overflow UBSan check disabled for this conversion. In older clang versions
+    // (up to 8.0) it reports a failure when non-infinity f64 value is converted to f32 infinity.
+    // Such behavior is expected.
+    return static_cast<float>(value);
+}
+
 std::optional<uint32_t> find_export(const Module& module, ExternalKind kind, std::string_view name)
 {
     const auto it = std::find_if(module.exportsec.begin(), module.exportsec.end(),
@@ -1759,6 +1768,11 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value> 
             convert<uint64_t, float>(stack);
             break;
         }
+        case Instr::f32_demote_f64:
+        {
+            stack.top() = demote(stack.top().f64);
+            break;
+        }
         case Instr::f64_convert_i32_s:
         {
             convert<int32_t, double>(stack);
@@ -1795,7 +1809,6 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value> 
         case Instr::f64_trunc:
         case Instr::f64_nearest:
         case Instr::f64_sub:
-        case Instr::f32_demote_f64:
         case Instr::i32_reinterpret_f32:
         case Instr::i64_reinterpret_f64:
         case Instr::f32_reinterpret_i32:
--- a/test/unittests/execute_floating_point_test.cpp
+++ b/test/unittests/execute_floating_point_test.cpp
@@ -92,7 +92,7 @@ struct WasmTypeName
 template <typename T>
 class execute_floating_point_types : public testing::Test
 {
-protected:
+public:
     using L = typename FP<T>::Limits;
 
     // The list of positive floating-point values without zeros, infinities and NaNs.
@@ -848,6 +848,13 @@ TEST(execute_floating_point, f64_promote_f32)
     ASSERT_TRUE(!res4.trapped && res4.has_value);
     EXPECT_EQ(std::signbit(res4.value.f64), 1);
     EXPECT_GT(FP{res4.value.f64}.nan_payload(), FP64::canon);
+
+    // Any input NaN other than canonical must result in an arithmetic NaN.
+    for (const auto nan : execute_floating_point_types<float>::positive_noncanonical_nans)
+    {
+        EXPECT_THAT(execute(*instance, 0, {nan}), ArithmeticNaN(double{}));
+        EXPECT_THAT(execute(*instance, 0, {-nan}), ArithmeticNaN(double{}));
+    }
 }
 
 TEST(execute_floating_point, f32_demote_f64)
@@ -944,10 +951,11 @@ TEST(execute_floating_point, f32_demote_f64)
     }
 
     // Any input NaN other than canonical must result in an arithmetic NaN.
-    EXPECT_THAT(execute(*instance, 0, {FP64::nan(FP64::canon + 1)}), ArithmeticNaN(float{}));
-    EXPECT_THAT(execute(*instance, 0, {-FP64::nan(FP64::canon + 1)}), ArithmeticNaN(float{}));
-    EXPECT_THAT(execute(*instance, 0, {FP64::nan(1)}), ArithmeticNaN(float{}));
-    EXPECT_THAT(execute(*instance, 0, {-FP64::nan(1)}), ArithmeticNaN(float{}));
+    for (const auto nan : execute_floating_point_types<double>::positive_noncanonical_nans)
+    {
+        EXPECT_THAT(execute(*instance, 0, {nan}), ArithmeticNaN(float{}));
+        EXPECT_THAT(execute(*instance, 0, {-nan}), ArithmeticNaN(float{}));
+    }
 }
 
 
--- a/test/unittests/execute_floating_point_test.cpp
+++ b/test/unittests/execute_floating_point_test.cpp
@@ -850,6 +850,106 @@ TEST(execute_floating_point, f64_promote_f32)
     EXPECT_GT(FP{res4.value.f64}.nan_payload(), FP64::canon);
 }
 
+TEST(execute_floating_point, f32_demote_f64)
+{
+    /* wat2wasm
+    (func (param f64) (result f32)
+      local.get 0
+      f32.demote_f64
+    )
+    */
+    const auto wasm = from_hex("0061736d0100000001060160017c017d030201000a070105002000b60b");
+    auto instance = instantiate(parse(wasm));
+
+    constexpr double f32_max = FP32::Limits::max();
+    ASSERT_EQ(f32_max, 0x1.fffffep127);
+
+    // The "artificial" f32 range limit: the next f32 number that could be represented
+    // if the exponent had a larger range.
+    // Wasm spec Rounding section denotes this as the limit_N in the float_N function (for N=32).
+    // https://webassembly.github.io/spec/core/exec/numerics.html#rounding
+    constexpr double f32_limit = 0x1p128;  // 2**128.
+
+    // The lower boundary input value that results in the infinity. The number is midway between
+    // f32_max and f32_limit. For this value rounding prefers infinity, because f32_limit is even.
+    constexpr double lowest_to_inf = (f32_max + f32_limit) / 2;
+    ASSERT_EQ(lowest_to_inf, 0x1.ffffffp127);
+
+    const std::pair<double, float> test_cases[] = {
+        // demote(+-0) = +-0
+        {0.0, 0.0f},
+        {-0.0, -0.0f},
+
+        {1.0, 1.0f},
+        {-1.0, -1.0f},
+        {double{FP32::Limits::lowest()}, FP32::Limits::lowest()},
+        {double{FP32::Limits::max()}, FP32::Limits::max()},
+        {double{FP32::Limits::min()}, FP32::Limits::min()},
+        {double{-FP32::Limits::min()}, -FP32::Limits::min()},
+        {double{FP32::Limits::denorm_min()}, FP32::Limits::denorm_min()},
+        {double{-FP32::Limits::denorm_min()}, -FP32::Limits::denorm_min()},
+
+        // Some special f64 values.
+        {FP64::Limits::lowest(), -FP32::Limits::infinity()},
+        {FP64::Limits::max(), FP32::Limits::infinity()},
+        {FP64::Limits::min(), 0.0f},
+        {-FP64::Limits::min(), -0.0f},
+        {FP64::Limits::denorm_min(), 0.0f},
+        {-FP64::Limits::denorm_min(), -0.0f},
+
+        // Out of range values rounded to max/lowest.
+        {std::nextafter(f32_max, FP64::Limits::infinity()), FP32::Limits::max()},
+        {std::nextafter(double{FP32::Limits::lowest()}, -FP64::Limits::infinity()),
+            FP32::Limits::lowest()},
+
+        {std::nextafter(lowest_to_inf, 0.0), FP32::Limits::max()},
+        {std::nextafter(-lowest_to_inf, 0.0), FP32::Limits::lowest()},
+
+        // The smallest of range values rounded to infinity.
+        {lowest_to_inf, FP32::Limits::infinity()},
+        {-lowest_to_inf, -FP32::Limits::infinity()},
+
+        {std::nextafter(lowest_to_inf, FP64::Limits::infinity()), FP32::Limits::infinity()},
+        {std::nextafter(-lowest_to_inf, -FP64::Limits::infinity()), -FP32::Limits::infinity()},
+
+        // float_32(r) = +inf  (if r >= +limit_32)
+        {f32_limit, FP32::Limits::infinity()},
+
+        // float_32(r) = -inf  (if r <= -limit_32)
+        {-f32_limit, -FP32::Limits::infinity()},
+
+        // demote(+-inf) = +-inf
+        {FP64::Limits::infinity(), FP32::Limits::infinity()},
+        {-FP64::Limits::infinity(), -FP32::Limits::infinity()},
+
+        // Rounding.
+        {0x1.fffffefffffffp0, 0x1.fffffep0f},  // round down
+        {0x1.fffffe0000000p0, 0x1.fffffep0f},  // exact (odd)
+        {0x1.fffffd0000001p0, 0x1.fffffep0f},  // round up
+
+        {0x1.fffff8p0, 0x1.fffff8p0f},                       // exact (even)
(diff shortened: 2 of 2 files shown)


lib/fizzy/execute.cpp
```diff
@@ -510,6 +510,12 @@ inline uint64_t popcnt64(uint64_t value) noexcept
     return static_cast<uint64_t>(__builtin_popcountll(value));
 }
 
+__attribute__((no_sanitize("float-cast-overflow"))) inline constexpr float demote(
```
gumb0, line 513: Some comment explaining why it's needed?

chfast: Yes, added.

test/unittests/execute_floating_point_test.cpp
```diff
@@ -465,6 +465,62 @@ TEST(execute_floating_point, f64_promote_f32)
...
+        {double{FP32::Limits::lowest()} * 0x1.0000008000008p0, -FP32::Limits::infinity()},
+
+        // demote(+-inf) = +-inf
+        {FP64::Limits::infinity(), FP32::Limits::infinity()},
+        {-FP64::Limits::infinity(), -FP32::Limits::infinity()},
+
+        // The canonical NaN must result in canonical NaN (only the top bit set).
+        {FP32::nan(FP32::canon), FP64::nan(FP64::canon)},
+        {-FP32::nan(FP32::canon), -FP64::nan(FP64::canon)},
+    };
```
axic, line 510: Can you add the usual manual test case here too?

test/unittests/execute_floating_point_test.cpp
```diff
@@ -465,6 +465,62 @@ TEST(execute_floating_point, f64_promote_f32)
...
+        {std::nextafter(double{FP32::Limits::max()}, FP64::Limits::infinity()),
+            FP32::Limits::max()},
+        {std::nextafter(double{FP32::Limits::lowest()}, -FP64::Limits::infinity()),
+            FP32::Limits::lowest()},
+        {double{FP32::Limits::max()} * 0x1.0000008000007p0, FP32::Limits::max()},
+        {double{FP32::Limits::lowest()} * 0x1.0000008000007p0, FP32::Limits::lowest()},
+
+        // The smallest of range values rounded to infinity.
+        {double{FP32::Limits::max()} * 0x1.0000008000008p0, FP32::Limits::infinity()},
+        {double{FP32::Limits::lowest()} * 0x1.0000008000008p0, -FP32::Limits::infinity()},
```
axic, line 501: What are these values?

test/unittests/execute_floating_point_test.cpp
```diff
@@ -465,6 +465,62 @@ TEST(execute_floating_point, f64_promote_f32)
...
+        {double{FP32::Limits::min()}, FP32::Limits::min()},
+        {double{FP32::Limits::denorm_min()}, FP32::Limits::denorm_min()},
+
+        // Out of range values rounded to max/lowest.
+        {std::nextafter(double{FP32::Limits::max()}, FP64::Limits::infinity()),
+            FP32::Limits::max()},
+        {std::nextafter(double{FP32::Limits::lowest()}, -FP64::Limits::infinity()),
+            FP32::Limits::lowest()},
+        {double{FP32::Limits::max()} * 0x1.0000008000007p0, FP32::Limits::max()},
+        {double{FP32::Limits::lowest()} * 0x1.0000008000007p0, FP32::Limits::lowest()},
```
axic, line 497: What is this value?

test/unittests/execute_floating_point_test.cpp
chfast: Boundaries values that convert to max or infinity. I will rework it to be more readable.

test/unittests/execute_floating_point_test.cpp
```diff
@@ -551,6 +551,69 @@ TEST(execute_floating_point, f64_promote_f32)
...
+    (func (param f64) (result f32)
+      local.get 0
+      f32.demote_f64
+    )
+    */
+    const auto wasm = from_hex("0061736d0100000001060160017c017d030201000a070105002000b60b");
+    auto instance = instantiate(parse(wasm));
+
+    // The boundary input value that results in the infinity.
+    constexpr double lowest_to_inf = 0x1.ffffffp127;
```
gumb0, line 566: I'm a bit confused and might be wrong, but cases `p > lowest_to_inf` and `p < -lowest_to_inf` are missing, and I think they are undefined behavior in C++, but defined to result in infinities in wasm.

chfast: I think it is defined if IEEE 754 is supported (as for some other cases) because then infinities also counts.

> The minimum range of representable values for a floating type is the most negative finite floating-point number representable in that type through the most positive finite floating-point number representable in that type. In addition, if negative infinity is representable in a type, the range of that type is extended to all negative real numbers; likewise, if positive infinity is representable in a type, the range of that type is extended to all positive real numbers.

https://stackoverflow.com/a/25831946/725174

Although, I'm confused how to chose nearest value between `max` and `infinity`...

The cases you mentions are not present because `lowest_to_inf` already yields infinity, so for any `p > lowest_to_inf` we get infinity too. But I can add test for `nextafter(lowest_to_inf)`.

chfast: This answers my question.

> In the default IEEE 754 round-to-nearest mode, a few double values above the maximum finite float (that is, FLT_MAX) convert to FLT_MAX. The exact limit is the number midway between FLT_MAX (0x1.fffffep127 in C99 hexadecimal representation) and the next float number that could be represented if the exponent in the single-precision format had a larger range, 0x2.0p127. The limit is thus 0x1.ffffffp127 or approximately 3.4028235677973366e+38 in decimal.

https://stackoverflow.com/a/17751145/725174

Although I wander where wasm spec handles this, if anywhere.

chfast: I finally have explanation for the `lowest_to_inf` value (see comments).
I also added two cases for `p > lowest_to_inf`.

test/unittests/execute_floating_point_test.cpp
```diff
@@ -700,6 +700,89 @@ TEST(execute_floating_point, f64_promote_f32)
...
+    ASSERT_EQ(f32_max, 0x1.fffffep127);
+
+    // The "artificial" f32 range limit: the next f32 number that could be represented
+    // if the exponent had a larger range.
+    // Wasm spec Rounding section denotes this as the limit_N in the float_N function (for N=32).
+    // https://webassembly.github.io/spec/core/exec/numerics.html#rounding
+    constexpr double f32_limit = 0x1p128;  // 2**128.
+
+    // The lower boundary input value that results in the infinity. The number is midway between
+    // f32_max and f32_limit. For this value rounding prefers infinity, because f32_limit is even.
```
gumb0, line 724: I don't quite understand the last sentence. `f32_limit` would indeed be even if casted to int, but for floats there's no even/odd ?

test/unittests/execute_floating_point_test.cpp
```diff
@@ -700,6 +700,89 @@ TEST(execute_floating_point, f64_promote_f32)
...
+
+        {std::nextafter(lowest_to_inf, 0.0), FP32::Limits::max()},
+        {-std::nextafter(lowest_to_inf, 0.0), FP32::Limits::lowest()},
+
+        // The smallest of range values rounded to infinity.
+        {lowest_to_inf, FP32::Limits::infinity()},
+        {-lowest_to_inf, -FP32::Limits::infinity()},
+
+        {std::nextafter(lowest_to_inf, FP64::Limits::infinity()), FP32::Limits::infinity()},
+        {-std::nextafter(lowest_to_inf, FP64::Limits::infinity()), -FP32::Limits::infinity()},
```
gumb0, line 753: shouldn't this be
```suggestion
        {std::nextafter(-lowest_to_inf, -FP64::Limits::infinity()), -FP32::Limits::infinity()},
```

test/unittests/execute_floating_point_test.cpp
gumb0: Ah I guess it's about evenness of the mantissa part

image

test/unittests/execute_floating_point_test.cpp
chfast: It is the same. floats has symmetrical positive and negative ranges.

test/unittests/execute_floating_point_test.cpp
chfast: Of course there is (at least in Wasm spec).
image
In this case `f32_limit` is even by definition.

gumb0: But still confused, mantissa of `f32_max` is even, manitssa of `f32_limit` is odd

test/unittests/execute_floating_point_test.cpp
```diff
@@ -700,6 +700,89 @@ TEST(execute_floating_point, f64_promote_f32)
...
+        EXPECT_THAT(execute(*instance, 0, {arg}), Result(expected)) << arg << " -> " << expected;
+    }
+
+    // Any input NaN other than canonical must result in an arithmetic NaN.
+    EXPECT_THAT(execute(*instance, 0, {FP64::nan(FP64::canon + 1)}), ArithmeticNaN(float{}));
+    EXPECT_THAT(execute(*instance, 0, {-FP64::nan(FP64::canon + 1)}), ArithmeticNaN(float{}));
+    EXPECT_THAT(execute(*instance, 0, {FP64::nan(1)}), ArithmeticNaN(float{}));
+    EXPECT_THAT(execute(*instance, 0, {-FP64::nan(1)}), ArithmeticNaN(float{}));
+
+    EXPECT_THAT(execute(*instance, 0, {0x1.fffffefffffffp0}), Result(0x1.fffffep0f));  // round down
```
gumb0, line 781: Also confused, why is this one rounded down, it's closer to `0x1.ffffffp0f`

test/unittests/execute_floating_point_test.cpp
chfast: I think in both cases you are confused by the size of the f32 mantisa which is 23 bits. When written as 6 hex digits the last bit should always be 0. The mantissa full of ones is `1.fffffe` (leading 1 is implicit). Then `m = 0x7fffff` and is odd.

The `f32_limit` is even by definition: `even_N(+-limit_N) <=> true`. But also mantissa of `1p128` is all zero therefore even.

test/unittests/execute_floating_point_test.cpp
chfast: See comment above.

test/unittests/execute_floating_point_test.cpp
chfast: Changed to use only single variant consistently.

test/unittests/execute_floating_point_test.cpp
```diff
@@ -685,6 +685,89 @@ TEST(execute_floating_point, f64_promote_f32)
...
+        // demote(+-0) = +-0
+        {0.0, 0.0f},
+        {-0.0, -0.0f},
+
+        {1.0, 1.0f},
+        {-1.0, -1.0f},
+        {double{FP32::Limits::lowest()}, FP32::Limits::lowest()},
+        {double{FP32::Limits::max()}, FP32::Limits::max()},
+        {double{FP32::Limits::min()}, FP32::Limits::min()},
+        {double{FP32::Limits::denorm_min()}, FP32::Limits::denorm_min()},
```
axic, line 723: Can we also have test cases for `FP64::Limits` so we know how they are truncated?

test/unittests/execute_floating_point_test.cpp
gumb0: I see, yes, I was confused by hex notation.

test/unittests/execute_floating_point_test.cpp
```diff
@@ -779,6 +779,97 @@ TEST(execute_floating_point, f64_promote_f32)
...
+    // Any input NaN other than canonical must result in an arithmetic NaN.
+    for (const auto nan : execute_floating_point_types<double>::positive_noncanonical_nans)
+    {
+        EXPECT_THAT(execute(*instance, 0, {nan}), ArithmeticNaN(float{}));
+        EXPECT_THAT(execute(*instance, 0, {-nan}), ArithmeticNaN(float{}));
+    }
+
+    EXPECT_THAT(execute(*instance, 0, {0x1.fffffefffffffp0}), Result(0x1.fffffep0f));  // round down
+    EXPECT_THAT(execute(*instance, 0, {0x1.fffffe0000000p0}), Result(0x1.fffffep0f));  // exact
+    EXPECT_THAT(execute(*instance, 0, {0x1.fffffd0000001p0}), Result(0x1.fffffep0f));  // round up
```
gumb0, line 872: Maye also add the case with double value in the middle between two floats, to show tie-to-even

test/unittests/execute_floating_point_test.cpp
```diff
@@ -92,7 +92,7 @@ struct WasmTypeName
 template <typename T>
 class execute_floating_point_types : public testing::Test
 {
-protected:
+public:
```
gumb0, line 95: why?

test/unittests/execute_floating_point_test.cpp
chfast: Added. But they are boring.

test/unittests/execute_floating_point_test.cpp
chfast: I want to access the list of NaNs from other test suite.

test/unittests/execute_floating_point_test.cpp
chfast: Added.

test/unittests/execute_floating_point_test.cpp
```diff
@@ -781,6 +781,106 @@ TEST(execute_floating_point, f64_promote_f32)
...
+    (func (param f64) (result f32)
+      local.get 0
+      f32.demote_f64
+    )
+    */
+    const auto wasm = from_hex("0061736d0100000001060160017c017d030201000a070105002000b60b");
+    auto instance = instantiate(parse(wasm));
+
+    constexpr double f32_max = FP32::Limits::max();
+    ASSERT_EQ(f32_max, 0x1.fffffep127);
```
axic, line 803: Why is this assert here and not in limits? If here, can't this be made into a static_assert?

chfast: Can be. I just want to see what the value is.

test/unittests/execute_floating_point_test.cpp
```diff
@@ -781,6 +781,106 @@ TEST(execute_floating_point, f64_promote_f32)
...
+    // The "artificial" f32 range limit: the next f32 number that could be represented
+    // if the exponent had a larger range.
+    // Wasm spec Rounding section denotes this as the limit_N in the float_N function (for N=32).
+    // https://webassembly.github.io/spec/core/exec/numerics.html#rounding
+    constexpr double f32_limit = 0x1p128;  // 2**128.
+
+    // The lower boundary input value that results in the infinity. The number is midway between
+    // f32_max and f32_limit. For this value rounding prefers infinity, because f32_limit is even.
+    constexpr double lowest_to_inf = (f32_max + f32_limit) / 2;
+    ASSERT_EQ(lowest_to_inf, 0x1.ffffffp127);
```
axic, line 814: It doesn't really matter, but is there a reason this should assert_eq and not static_assert?

chfast: I started with `static_assert`, then decided that maybe it should build anyway even if not true. I guess static_assert may be less confusing for readers.

test/unittests/execute_floating_point_test.cpp
```diff
@@ -779,6 +779,114 @@ TEST(execute_floating_point, f64_promote_f32)
...
+        {0x1.fffffe0000000p0, 0x1.fffffep0f},  // exact (odd)
+        {0x1.fffffd0000001p0, 0x1.fffffep0f},  // round up
+
+        {0x1.fffff8p0, 0x1.fffff8p0f},                      // exact (even)
+        {(0x1.fffff8p0 + 0x1.fffffap0) / 2, 0x1.fffff8p0},  // tie-to-even down
+        {0x1.fffffap0, 0x1.fffffap0f},                      // exact (odd)
+        {(0x1.fffffap0 + 0x1.fffffcp0) / 2, 0x1.fffffcp0},  // tie-to-even up
+        {0x1.fffffcp0, 0x1.fffffcp0f},                      // exact (even)
+
+        // The canonical NaN must result in canonical NaN (only the top bit set).
```
gumb0, line 874:
```suggestion
        // The canonical NaN must result in canonical NaN (only the top bit of payload set).
```

test/unittests/execute_floating_point_test.cpp
```diff
@@ -779,6 +779,114 @@ TEST(execute_floating_point, f64_promote_f32)
...
+        {FP64::Limits::infinity(), FP32::Limits::infinity()},
+        {-FP64::Limits::infinity(), -FP32::Limits::infinity()},
+
+        // Rounding.
+        {0x1.fffffefffffffp0, 0x1.fffffep0f},  // round down
+        {0x1.fffffe0000000p0, 0x1.fffffep0f},  // exact (odd)
+        {0x1.fffffd0000001p0, 0x1.fffffep0f},  // round up
+
+        {0x1.fffff8p0, 0x1.fffff8p0f},                      // exact (even)
+        {(0x1.fffff8p0 + 0x1.fffffap0) / 2, 0x1.fffff8p0},  // tie-to-even down
```
axic, line 869:
```suggestion
        {(0x1.fffff8p0 + 0x1.fffffap0) / 2, 0x1.fffff8p0f},  // tie-to-even down
```

test/unittests/execute_floating_point_test.cpp
```diff
@@ -779,6 +779,114 @@ TEST(execute_floating_point, f64_promote_f32)
...
+        {0x1.fffffap0, 0x1.fffffap0f},                      // exact (odd)
+        {(0x1.fffffap0 + 0x1.fffffcp0) / 2, 0x1.fffffcp0},  // tie-to-even up
```
axic, line 871:
```suggestion
        {(0x1.fffffap0 + 0x1.fffffcp0) / 2, 0x1.fffffcp0f},  // tie-to-even up
```

PR #459, opened 2020-08-04
Implement reinterpret instructions

chfast:
image

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -357,6 +357,20 @@ inline void convert(OperandStack& stack) noexcept
     stack.top() = static_cast<DstT>(stack.top().as<SrcT>());
 }
 
+/// Performs a bit_cast from SrcT type to DstT type.
+///
+/// This should be optimized to empty function in assembly. Except for f32 -> i32 where pushing
+/// the result i32 value to the stack requires zero-extension to 64-bit.
+template <typename SrcT, typename DstT>
+inline void reinterpret(OperandStack& stack) noexcept
+{
+    static_assert(sizeof(SrcT) == sizeof(DstT));
+    const auto src = stack.top().as<SrcT>();
+    DstT dst;
+    __builtin_memcpy(&dst, &src, sizeof(dst));
+    stack.top() = dst;
+}
+
 template <typename DstT, typename SrcT = DstT>
 inline bool load_from_memory(
     bytes_view memory, OperandStack& stack, const uint8_t*& immediates) noexcept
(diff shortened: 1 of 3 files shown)


lib/fizzy/execute.cpp
```diff
@@ -1680,6 +1708,26 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value>
...
+            break;
+        }
+        case Instr::f32_reinterpret_i32:
+        {
+            reinterpret<uint32_t, float>(stack);
+            break;
+        }
+        case Instr::f64_reinterpret_i64:
+        {
+            reinterpret<uint64_t, uint64_t>(stack);
```
gumb0, line 1728:
```suggestion
            reinterpret<uint64_t, double>(stack);
```

lib/fizzy/execute.cpp
```diff
@@ -357,6 +357,20 @@ inline void convert(OperandStack& stack) noexcept
     stack.top() = static_cast<DstT>(stack.top().as<SrcT>());
 }
 
+/// Performs a bitcast from SrcT type to DstT type.
+///
+/// This is a formality (except for f32 -> i32 where zero-extension is needed)
```
gumb0, line 362: What zero-extension?

chfast: When we push a `i32` value to the stack we actually zero-extend it to `i64` in `Value`.

lib/fizzy/execute.cpp

lib/fizzy/execute.cpp
chfast: I improved the explanation.

lib/fizzy/execute.cpp
```diff
@@ -357,6 +357,20 @@ inline void convert(OperandStack& stack) noexcept
     stack.top() = static_cast<DstT>(stack.top().as<SrcT>());
 }
 
+/// Performs a bit_cast from SrcT type to DstT type.
+///
+/// This is a formality and should be optimized to empty function in assembly.
+/// Except for f32 -> i32 where pushing the result i32 value to the stack requires zero-extension.
```
gumb0, line 363:
```suggestion
/// Except for f32 -> i32 where pushing the result i32 value to the stack requires zero-extension to 64-bit.
```

lib/fizzy/execute.cpp
```diff
@@ -357,6 +357,20 @@ inline void convert(OperandStack& stack) noexcept
     stack.top() = static_cast<DstT>(stack.top().as<SrcT>());
 }
 
+/// Performs a bit_cast from SrcT type to DstT type.
+///
+/// This is a formality and should be optimized to empty function in assembly.
```
gumb0, line 362: Nit: I wouldn't call it a "formality". Compiler can optimize it away, but from C++ point of view it's required to avoid UB, right?

chfast: That's what I call formality.

lib/fizzy/execute.cpp
```diff
@@ -357,6 +357,20 @@ inline void convert(OperandStack& stack) noexcept
...
+/// Except for f32 -> i32 where pushing the result i32 value to the stack requires zero-extension.
+template <typename SrcT, typename DstT>
+inline void reinterpret(OperandStack& stack) noexcept
+{
+    static_assert(sizeof(SrcT) == sizeof(DstT));
+    const auto src = stack.top().as<SrcT>();
+    DstT dst;
+    __builtin_memcpy(&dst, &src, sizeof(dst));
```
axic, line 370: Why couldn't `reinterpret_cast` work here instead?

chfast: 1. You cannot do `reinterpret_cast<uint32_t>(f32)`.
2. Using pointers would work: `*reinterpreter_cast<uint32_t*>(&f32)`, but in theory that is not valid use of `reinterpret_cast` and undefined behavior. Using `memcpy` is the safest way of doing type pruning. Although looks ugly.
3. The C++20 there will be `std::bit_cast` - exactly what we need here.

PR #460, opened 2020-08-05
test: Fix comparing of floating point results in spectest runner

gumb0:


test/spectests/spectests.cpp
```diff
@@ -58,6 +58,22 @@ fizzy::bytes load_wasm_file(const fs::path& json_file_path, std::string_view fil
         std::istreambuf_iterator<char>{wasm_file}, std::istreambuf_iterator<char>{});
 }
 
+bool float_values_equal(fizzy::Value value1, fizzy::Value value2) noexcept
+{
+    if (std::isnan(value1.as<float>()))
+        return value1.as<uint32_t>() == value2.as<uint32_t>();  // compare binary representations
```
chfast, line 64: This is quite close to undefined behavior: only single union member may be active at any point in time and you are allowed to only access the active one. Here you are accessing two members is a row - one of the accesses must be wrong.

But also to simplify: you can always compare binary representation of floats. I would just leave `return value1.as<uint32_t>() == value2.as<uint32_t>();`.

Also: add the following check in `value.hpp`: `static_assert(offsetof(Value, f32) == offsetof(Value, i32));`.

gumb0: (now I see why you needed `to_uint64`/`from_uint64`)

Well ok, then the current way here it will be UB only for NaN case, but if I do always `as<uint32_t>()` then always...

chfast: That's actually not true, because you always set `i64` to active in `read_value()`. In theory, you are never allowed to access `f32` nor `f64`. But I doubt this will cause issues in practice. However, I had issues with performing type punning using unions in GCC before (when the misuse is obvious, in the same function).

But my motivation to always use "binary" representation for comparison is to make sure this is a reliable method and there are no regressions in running spectests.

chfast:
No spectests results change is expected here?

gumb0:
> No spectests results change is expected here?

No, I guess because all NaN tests are skipped on master so far.

test/spectests/spectests.cpp
gumb0: This was replaced to comparison using `FP` utility from #454

test/spectests/spectests.cpp
```diff
@@ -404,16 +422,22 @@ class test_runner
...
     {
-        const auto arg_type = v.at("type").get<std::string>();
-        if (arg_type != "i32" && arg_type != "i64" && arg_type != "f32" && arg_type != "f64")
+        const auto type = v.at("type").get<std::string>();
+
+        // JSON tests have all values including floats serialized as 64-bit unsigned integers
+        const uint64_t uint_value = std::stoull(v.at("value").get<std::string>());
+
+        if (type == "i32" || type == "i64")
+            return uint_value;
```
gumb0, line 431: I decided for now not to complicate it with checking if the value declared as 32-bit is really not wider. It won't result in UB if it happens.

test/smoketests/spectests/failures/failures.wast
```diff
@@ -40,9 +40,15 @@
 ;; invalid result
 (module
     (func (export "foo.i32") (result i32) (i32.const 1))
+    (func (export "foo.i64") (result i64) (i64.const 1))
+    (func (export "foo.f32") (result f32) (f32.const 1))
+    (func (export "foo.f64") (result f64) (f64.const 1))
```
axic, line 45: I think it would be better to use actual floats for these, to be sure. Such as 1.234.

test/spectests/spectests.cpp
```diff
@@ -404,16 +422,22 @@ class test_runner
 
     std::optional<fizzy::Value> read_value(const json& v)
     {
-        const auto arg_type = v.at("type").get<std::string>();
-        if (arg_type != "i32" && arg_type != "i64" && arg_type != "f32" && arg_type != "f64")
+        const auto type = v.at("type").get<std::string>();
+
+        // JSON tests have all values including floats serialized as 64-bit unsigned integers
```
axic, line 427: Not that JSON supports the whole 64-bit 😉

test/smoketests/spectests/failures/failures.wast
gumb0: Changed

test/spectests/spectests.cpp
chfast: The format supports arbitrary long literals, in theory.

PR #461, opened 2020-08-06
Properly use i32 for br_table_idx and elem_idx in execute

axic:
Fix based on https://github.com/wasmx/fizzy/pull/448#discussion_r463763921.

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -687,7 +687,7 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value> 
             const auto br_table_size = read<uint32_t>(immediates);
             const auto arity = read<uint32_t>(immediates);
 
-            const auto br_table_idx = stack.pop();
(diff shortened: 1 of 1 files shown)


lib/fizzy/execute.cpp
```diff
@@ -687,7 +687,7 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value>
             const auto br_table_size = read<uint32_t>(immediates);
             const auto arity = read<uint32_t>(immediates);
 
-            const auto br_table_idx = stack.pop();
+            const auto br_table_idx = static_cast<uint32_t>(stack.pop());
```
axic, line 690: This is more pedantic, than anything, because after validation, even if we do not clean up the stack, the only way to end up here is either using 32-bit ops or after `i32_wrap_i64`, which does the cleanup. I wonder if we miss something.

Could also fold this into being the first commit in #448.

axic:
Replaced by #448.

PR #462, opened 2020-08-07
Check floats compliance with IEEE 754

chfast:
This checks if the floating point types used to implement `f32` and `f64` Wasm types are compliant with IEEE 754.

The CMake check is rather difficult to do. We would need to check if a test program compiles.

### DIFF
--- a/lib/fizzy/value.hpp
+++ b/lib/fizzy/value.hpp
@@ -5,6 +5,7 @@
 #pragma once
 
 #include <cstdint>
+#include <limits>
 
 namespace fizzy
 {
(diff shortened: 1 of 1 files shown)


axic:
Well at least with this they'll get a compile time error.

chfast:
> Well at least with this they'll get a compile time error.

How is this different from CMake error?

axic:
> > Well at least with this they'll get a compile time error.
>
> How is this different from CMake error?

By "cmake error" I meant which presents itself during configuration, e.g. `cmake ..`, while this will present itself during build `cmake --build`. I'm fine with either if the configuration part is too complex.

PR #464, opened 2020-08-07
tests: Remove explicit uint64_t cast from Value

axic:


test/unittests/parser_test.cpp
```diff
@@ -661,15 +661,15 @@ TEST(parser, global_float)
...
     EXPECT_TRUE(module.globalsec[1].type.is_mutable);
     EXPECT_EQ(module.globalsec[1].type.value_type, ValType::f64);
     EXPECT_EQ(module.globalsec[1].expression.kind, ConstantExpression::Kind::Constant);
-    EXPECT_EQ(Value(module.globalsec[1].expression.value.constant).f64, 3.4);
+    EXPECT_EQ(module.globalsec[1].expression.value.constant.f64, 3.4);
     EXPECT_FALSE(module.globalsec[2].type.is_mutable);
     EXPECT_EQ(module.globalsec[2].type.value_type, ValType::f64);
     EXPECT_EQ(module.globalsec[2].expression.kind, ConstantExpression::Kind::GlobalGet);
-    EXPECT_EQ(module.globalsec[2].expression.value.constant, 1);
+    EXPECT_EQ(module.globalsec[2].expression.value.constant.f64, 1);
```
axic, line 672: This fails tests.

test/unittests/stack_test.cpp
```diff
@@ -281,7 +281,10 @@ TEST(operand_stack, to_vector)
     stack.push(2);
     stack.push(3);
 
-    EXPECT_THAT(std::vector(stack.rbegin(), stack.rend()), ElementsAre(1, 2, 3));
```
axic, line 284: Tried `ElementsAre(fizzy::Value{1}, ...)` but that complained there's no `operator=(const Value, const Value)` and it had to be fully qualified with `fizzy::` because otherwise conflicted with `Value` in gmock/gtest -- all in all, this is simpler.

test/unittests/parser_test.cpp
gumb0: This is an index of `GlobalGet`, it should be `.i64`

test/unittests/api_test.cpp
```diff
@@ -315,22 +315,22 @@ TEST(api, find_exported_global)
 
     auto opt_global = find_exported_global(*instance, "g1");
     ASSERT_TRUE(opt_global);
-    EXPECT_EQ(*opt_global->value, 0);
+    EXPECT_EQ(opt_global->value->as<uint32_t>(), 0);
```
gumb0, line 318: You change the test here to check only 32-bit, I thinks it's better to check `.i64`

axic: Changed it to i32 because that's what the sources use above.

test/unittests/parser_test.cpp
axic: The other cases above use `f32` and `f64`.

axic: Ah never mind, didn't see the `GlobalGet` vs `Constant`, will fix.

axic: Btw, the test below uses `value.global_index` and not `value.constant` for that case, so I think this test here is a bit wrong.

gumb0: Oh right, should be `value.global_index`

axic: Fixed.

axic:
Should squash the separate commits of `test` before merging.

test/unittests/execute_call_test.cpp
```diff
@@ -286,10 +288,10 @@ TEST(execute_call, imported_functions_call_indirect)
     ASSERT_EQ(module.codesec.size(), 2);
 
     constexpr auto sqr = [](Instance&, span<const Value> args, int) {
-        return Value{args[0] * args[0]};
+        return Value{uint64_t{args[0].as<uint32_t>()} * args[0].as<uint32_t>()};
```
gumb0, line 291: A bit hard to read
```suggestion
        const auto x = args[0].as<uint32_t>();
        return Value{uint64_t{x} * x};
```

test/unittests/api_test.cpp
gumb0: In case `find_exported_global` would return some incorrect `i64` global with the equal lower word, we would miss it.

test/unittests/instantiate_test.cpp
```diff
@@ -14,12 +14,12 @@ using namespace fizzy::test;
...
 namespace
 {
-uint64_t call_table_func(Instance& instance, size_t idx)
+uint32_t call_table_func(Instance& instance, size_t idx)
 {
     const auto& elem = (*instance.table)[idx];
     const auto res = elem->function(instance, {}, 0);
     EXPECT_TRUE(res.has_value);
-    return res.value;
+    return res.value.as<uint32_t>();
```
gumb0, line 22: Here, too: if table function would return incorrect `i64` value, we could miss it, if we check only lower 32 bit.

axic:
@gumb0 this is still waiting for a thorough review from you, I'd like to merge this.

test/unittests/api_test.cpp
chfast: I'm for staying with `i64` just because it tests more bits and goes along with the current implementation.

axic: With this reasoning, are we also using `.i64` for inputs even if the function type is defined as i32?

chfast: > With this reasoning, are we also using `.i64` for inputs even if the function type is defined as i32?

Not for computation, but you can add asserts there.

axic: That is a tad inconsistent. Perhaps the best really is a helper which asserts there are no dirty higher bits.

chfast: Testing more than is defined by wasm type system is fine to me.

test/utils/asserts.hpp
```diff
@@ -54,3 +55,18 @@ namespace fizzy
...
+{
+inline uint32_t as_uint32(fizzy::Value value)
+{
+    EXPECT_EQ(value.i64 & 0xffffffff00000000ULL, 0);
+    return static_cast<uint32_t>(value.i64);
+}
+
+inline int32_t as_int32(fizzy::Value value)
+{
+    EXPECT_EQ(value.i64 & 0xffffffff00000000ULL, 0);
```
chfast, line 69:
```suggestion
    EXPECT_EQ(value.i64 & 0xffffffff00000000, 0);
```
Yes, this looks good now. The suggestion should work too I think.

Alternatively, we may consider to add the assert to `Value::as<uint32_t>()`.

axic: I had it without the postfix, but wanted to be sure. Perhaps not needed in C++.

axic: > Alternatively, we may consider to add the assert to Value::as<uint32_t>().

Maybe we should do that :wink:

axic: Added it, but we already have a trap with the unit tests:
```
[ RUN      ] end_to_end.milestone2
Assertion failed: ((i64 & 0xffffffff00000000) == 0), function as, file /Users/alex/Projects/fizzy/lib/fizzy/value.hpp, line 59.
Abort trap: 6
```

test/unittests/execute_call_test.cpp
```diff
@@ -286,10 +288,12 @@ TEST(execute_call, imported_functions_call_indirect)
...
 
     constexpr auto sqr = [](Instance&, span<const Value> args, int) {
-        return Value{args[0] * args[0]};
+        const auto x = as_uint32(args[0]);
+        return Value{uint64_t{x} * x};
     };
     constexpr auto isqrt = [](Instance&, span<const Value> args, int) {
-        return Value{(11 + args[0] / 11) / 2};
+        const auto x = as_uint32(args[0]);
+        return Value{(11 + uint64_t{x} / 11) / 2};
```
axic, line 296: These two (sqr, isqrt) expect i64 as output, should we explicitly cast them?

chfast: No. This is fine.

test/unittests/parser_test.cpp
```diff
@@ -661,15 +661,15 @@ TEST(parser, global_float)
...
     EXPECT_TRUE(module.globalsec[1].type.is_mutable);
     EXPECT_EQ(module.globalsec[1].type.value_type, ValType::f64);
     EXPECT_EQ(module.globalsec[1].expression.kind, ConstantExpression::Kind::Constant);
-    EXPECT_EQ(Value(module.globalsec[1].expression.value.constant).f64, 3.4);
+    EXPECT_EQ(module.globalsec[1].expression.value.constant.f64, 3.4);
     EXPECT_FALSE(module.globalsec[2].type.is_mutable);
     EXPECT_EQ(module.globalsec[2].type.value_type, ValType::f64);
     EXPECT_EQ(module.globalsec[2].expression.kind, ConstantExpression::Kind::GlobalGet);
-    EXPECT_EQ(module.globalsec[2].expression.value.constant, 1);
+    EXPECT_EQ(module.globalsec[2].expression.value.global_index, 1);
```
axic, line 672: Hm, this was a typo iirc, we did not fix it in another PR? Will make sure this has its own commit.

test/unittests/stack_test.cpp
axic: Could also add check for size of `result` to equal 3 here.

test/utils/asserts.hpp
```diff
@@ -54,3 +55,18 @@ namespace fizzy
...
+
+namespace fizzy::test
+{
+inline uint32_t as_uint32(fizzy::Value value)
+{
+    EXPECT_EQ(value.i64 & 0xffffffff00000000ULL, 0);
+    return static_cast<uint32_t>(value.i64);
+}
+
+inline int32_t as_int32(fizzy::Value value)
```
chfast, line 67: This is not used anywhere.

test/unittests/parser_test.cpp
```diff
@@ -624,11 +624,11 @@ TEST(parser, global_multi_const_inited)
     EXPECT_FALSE(module.globalsec[0].type.is_mutable);
     EXPECT_EQ(module.globalsec[0].type.value_type, ValType::i32);
     EXPECT_EQ(module.globalsec[0].expression.kind, ConstantExpression::Kind::Constant);
-    EXPECT_EQ(module.globalsec[0].expression.value.constant, 0x01);
+    EXPECT_EQ(as_uint32(module.globalsec[0].expression.value.constant), 1);
     EXPECT_TRUE(module.globalsec[1].type.is_mutable);
     EXPECT_EQ(module.globalsec[1].type.value_type, ValType::i32);
     EXPECT_EQ(module.globalsec[1].expression.kind, ConstantExpression::Kind::Constant);
-    EXPECT_EQ(module.globalsec[1].expression.value.constant, uint32_t(-1));
+    EXPECT_EQ(as_uint32(module.globalsec[1].expression.value.constant), -1);
```
chfast, line 631:
```suggestion
    EXPECT_EQ(as_uint32(module.globalsec[1].expression.value.constant), uint32_t(-1));
```
This should stay as otherwise implicit conversion is needed.

test/unittests/parser_test.cpp
axic: We actually have its own commit already.

test/unittests/parser_test.cpp
axic: Couldn't we just use `as_int32(..), -1` ?

chfast: In theory you can, but I don't see strong point for it. I prefer to refer to `i32` as `uint32_t` unless some heavy signed arithmetic is involved. Here you can just replace it with `0xffffffff`.

axic: Alright, removed it.

test/utils/asserts.hpp
axic: Created an issue to track this.

test/unittests/api_test.cpp
axic: Solved this by adding the helper.

test/unittests/stack_test.cpp
axic: Done.

chfast (approved):

PR #467, opened 2020-08-10
test: Add proper f{32,64}_add tests

chfast:
It adds/fixes some testing helpers for floating point instructions.

Then it adds wasm spec-driver tests for `fadd`, i.e. it adds checks for these points from the definition:
image

### DIFF
--- a/test/unittests/execute_floating_point_test.cpp
+++ b/test/unittests/execute_floating_point_test.cpp
@@ -9,6 +9,7 @@
 #include <test/utils/asserts.hpp>
 #include <test/utils/floating_point_utils.hpp>
 #include <test/utils/hex.hpp>
+#include <array>
 #include <cmath>
 
 using namespace fizzy;
@@ -61,36 +62,6 @@ TEST(execute_floating_point, f64_const)
     EXPECT_THAT(execute(*instance, 0, {}), Result(8589934592.1));
 }
 
-TEST(execute_floating_point, f32_add)
-{
-    /* wat2wasm
-    (func (param f32 f32) (result f32)
-      local.get 0
-      local.get 1
-      f32.add
-    )
-    */
-    const auto wasm = from_hex("0061736d0100000001070160027d7d017d030201000a0901070020002001920b");
-
-    auto instance = instantiate(parse(wasm));
-    EXPECT_THAT(execute(*instance, 0, {1.001f, 6.006f}), Result(7.007f));
-}
-
-TEST(execute_floating_point, f64_add)
-{
-    /* wat2wasm
-    (func (param f64 f64) (result f64)
-      local.get 0
-      local.get 1
-      f64.add
-    )
-    */
-    const auto wasm = from_hex("0061736d0100000001070160027c7c017c030201000a0901070020002001a00b");
-
-    auto instance = instantiate(parse(wasm));
-    EXPECT_THAT(execute(*instance, 0, {1.0011, 6.0066}), Result(7.0077));
-}
-
 /// Compile-time information about a Wasm type.
 template <typename T>
 struct WasmTypeTraits;
@@ -121,6 +92,18 @@ template <typename T>
 class execute_floating_point_types : public testing::Test
 {
 protected:
+    using L = typename FP<T>::Limits;
+
+    // The list of positive floating-point values without zeros, infinities and NaNs.
+    inline static const std::array positive_special_values{
+        L::denorm_min(),
+        L::min(),
+        std::nextafter(T{1.0}, T{0.0}),
+        T{1.0},
+        std::nextafter(T{1.0}, L::infinity()),
+        L::max(),
+    };
+
     /// Creates a wasm module with a single function for the given instructions opcode.
     /// The opcode is converted to match the type, e.g. f32_add -> f64_add.
     static bytes get_binop_code(Instr opcode)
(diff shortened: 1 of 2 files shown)


test/unittests/execute_floating_point_test.cpp
```diff
@@ -100,9 +117,69 @@ struct WasmTypeName
...
 
+template <typename T>
+class execute_floating_point_types : public testing::Test
+{
+};
+
+template <typename T>
+bytes get_binop_code(Instr opcode)
+{
+    constexpr auto opcode_offset =
```
gumb0: Minor suggestion: I would just calculate result opcode here, perhaps it would be easier to understand, without guessing what this offset is.

test/unittests/execute_floating_point_test.cpp
```diff
@@ -100,9 +117,69 @@ struct WasmTypeName
...
 TYPED_TEST_SUITE(execute_floating_point_types, FloatingPointTypes, WasmTypeName);
 
+TYPED_TEST(execute_floating_point_types, binop_nan_propagation)
+{
+    // Tests NaN propagation in binary instructions (binop).
+    // If all NaN inputs are canonical the result must be the canonical NaN.
+    // Otherwise, the result must be an arithmetic NaN.
+
+    constexpr Instr opcodes[] = {
+        Instr::f32_add,
```
gumb0, line 177: Perhaps add a comment, that these will be replaced with 64-bit versions when `TypeParam` is `double`

chfast: Done.

test/unittests/execute_floating_point_test.cpp

test/unittests/execute_floating_point_test.cpp
```diff
@@ -14,6 +14,26 @@
 using namespace fizzy;
 using namespace fizzy::test;
 
+MATCHER_P(CanonicalNaN, value, "result with an arithmetic NaN")
```
axic, line 18:
```suggestion
MATCHER_P(CanonicalNaN, value, "result with a canonical NaN")
```

chfast: Ooops, that has been fixed somewhere...

test/unittests/execute_floating_point_test.cpp
```diff
@@ -100,13 +88,120 @@ struct WasmTypeName
...
+    // If all NaN inputs are canonical the result must be the canonical NaN.
+    // Otherwise, the result must be an arithmetic NaN.
+
+    constexpr Instr opcodes[] = {
+        Instr::f32_add,
+    };
 
+    for (const auto op : opcodes)
+    {
+        auto instance = instantiate(parse(this->get_binop_code(op)));
```
gumb0, line 185: It's static
```suggestion
        auto instance = instantiate(parse(get_binop_code(op)));
```

test/unittests/execute_floating_point_test.cpp
```diff
@@ -100,9 +117,71 @@ struct WasmTypeName
     }
 };
 
+template <typename T>
+class execute_floating_point_types : public testing::Test
+{
+protected:
+    /// Creates a wasm module with a single function for the given instructions opcode.
+    /// The opcode is converted to match the type, e.g. f32_add -> f64_add.
+    bytes get_binop_code(Instr opcode)
```
axic: Nit: why not mark this static in this commit?

test/unittests/execute_floating_point_test.cpp
```diff
@@ -121,9 +92,21 @@ template <typename T>
 class execute_floating_point_types : public testing::Test
 {
 protected:
+    using L = typename FP<T>::Limits;
+
+    // The q is any positive floating-point value without zeros, infinities and NaNs.
```
gumb0, line 97: Why "q"?

test/unittests/execute_floating_point_test.cpp
```diff
@@ -288,6 +271,76 @@ TYPED_TEST(execute_floating_point_types, compare)
...
+    EXPECT_THAT(exec(-Limits::infinity(), TypeParam{0.0}), Result(-Limits::infinity()));
+    EXPECT_THAT(exec(-Limits::infinity(), -TypeParam{0.0}), Result(-Limits::infinity()));
+
+    // fadd(+-0, z2) = z2  (for z2 = +-inf)
+    EXPECT_THAT(exec(TypeParam{0.0}, Limits::infinity()), Result(Limits::infinity()));
+    EXPECT_THAT(exec(TypeParam{0.0}, -Limits::infinity()), Result(-Limits::infinity()));
+    EXPECT_THAT(exec(-TypeParam{0.0}, Limits::infinity()), Result(Limits::infinity()));
+    EXPECT_THAT(exec(-TypeParam{0.0}, -Limits::infinity()), Result(-Limits::infinity()));
+
+    for (const auto q : this->q_values)
```
gumb0, line 308:
```suggestion
    for (const auto q : q_values)
```

chfast: Does not work because of whatever is under GTest macros. You need `this->` or `execute_floating_point_types::`.

test/unittests/execute_floating_point_test.cpp
chfast: > The meta variable q is used to range over (signless) rational magnitudes, excluding nan or ∞.
https://webassembly.github.io/spec/core/exec/numerics.html#numerics

test/unittests/execute_floating_point_test.cpp
chfast: Fixed.

test/unittests/execute_floating_point_test.cpp
gumb0: That includes zeroes, but it's not so important

test/unittests/execute_floating_point_test.cpp
```diff
@@ -100,13 +88,140 @@ struct WasmTypeName
...
 };
 
+template <typename T>
+class execute_floating_point_types : public testing::Test
+{
+protected:
+    using L = typename FP<T>::Limits;
+
+    // The q is any positive floating-point value without zeros, infinities and NaNs.
+    inline static const std::array q_values{
```
gumb0, line 98: What's the point of making it inline?

chfast: I can define it here instead of doing so outside of the class.

test/unittests/execute_floating_point_test.cpp
```diff
@@ -100,13 +88,140 @@ struct WasmTypeName
     }
 };
 
+template <typename T>
+class execute_floating_point_types : public testing::Test
+{
+protected:
+    using L = typename FP<T>::Limits;
```
gumb0, line 95: I would prefer this to be `Limits`

test/unittests/execute_floating_point_test.cpp
gumb0: (I meant closer to this array, because now it looks like the tests checks only `f32_add`)

test/unittests/execute_floating_point_test.cpp
gumb0: I would call these two maybe `positive_special_values` and `ordered_special_values`

test/unittests/execute_floating_point_test.cpp
chfast: It cannot be because it will conflict of other `using Limits` in the tests.

gumb0: Why can't it use this one in the tests?

test/unittests/execute_floating_point_test.cpp
chfast: Done.

test/unittests/execute_floating_point_test.cpp

test/unittests/execute_floating_point_test.cpp
chfast: Because each `TYPED_TEST` inherits a template `execute_floating_point_types<TypeParam>` you have to access it as `typename execute_floating_point_types<TypeParam>::Limits`. Not very short.

PR #468, opened 2020-08-10
Implement f{32,64}.div instructions

chfast:
image

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -510,6 +510,13 @@ inline uint64_t popcnt64(uint64_t value) noexcept
     return static_cast<uint64_t>(__builtin_popcountll(value));
 }
 
+template <typename T>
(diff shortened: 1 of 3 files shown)


gumb0:
Some UBSan errors on macOS

lib/fizzy/execute.cpp
```diff
@@ -510,6 +510,13 @@ inline uint64_t popcnt64(uint64_t value) noexcept
     return static_cast<uint64_t>(__builtin_popcountll(value));
 }
 
+template <typename T>
+__attribute__((no_sanitize("float-divide-by-zero"))) inline constexpr T fdiv(T a, T b) noexcept
```
chfast, line 514: The attribute disables UBSan check in both GCC and Clang. Using C++ attributes (`[[gnu::no_sanitize("")]]`) is also possible but requires some changes to compiler's warning options. So I think it is fine to leave it as is for now.

chfast:
@axic, I added spec annotations in tests.

PR #469, opened 2020-08-10
ci: New sanitizers config

chfast:

### DIFF
--- a/test/unittests/execute_floating_point_test.cpp
+++ b/test/unittests/execute_floating_point_test.cpp
@@ -102,7 +102,15 @@ public:
 
         constexpr Iterator begin() const { return m_begin; }
         constexpr Iterator end() const { return m_end; }
-        constexpr size_t size() const { return static_cast<size_t>(m_end - m_begin); }
+
+        [[gnu::no_sanitize("pointer-subtract")]] constexpr size_t size() const
+        {
+            // The "pointer-subtract" sanitizer is disabled because GCC fails to compile
+            // constexpr function with pointer subtraction.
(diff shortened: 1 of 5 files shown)


gumb0 (approved):

test/unittests/execute_floating_point_test.cpp
```diff
@@ -102,7 +102,15 @@ class TestValues
 
         constexpr Iterator begin() const { return m_begin; }
         constexpr Iterator end() const { return m_end; }
-        constexpr size_t size() const { return static_cast<size_t>(m_end - m_begin); }
+
+        [[gnu::no_sanitize("pointer-subtract")]] constexpr size_t size() const
+        {
+            // The "pointer-subtract" sanitizer is disabled because GCC fails to compiler
```
axic, line 108:
```suggestion
            // The "pointer-subtract" sanitizer is disabled because GCC fails to compile
```

chfast:
This requires to lower the call depth limit to `512`.

axic:
What potential issues would gcc find over clang?

Can we do this limit conditionally when building for sanitizer-gcc target?

chfast:
> What potential issues would gcc find over clang?

I don't know really. They seem to have similar features, but have different bugs and versions of asan. E.g. I can enable `pointer-compare` check in GCC while it is broken in Clang.

> Can we do this limit conditionally when building for sanitizer-gcc target?

Please, no. I don't want to build and test different versions of Fizzy depending on debug options, sanitizer configuration, etc.

axic:
512 as a call depth seems a bit on the slim side though.

PR #470, opened 2020-08-10
Some changes recommended by CLion's analyzer

axic:


lib/fizzy/types.hpp
```diff
@@ -294,7 +294,7 @@ struct ConstantExpression
     Kind kind = Kind::Constant;
     union
     {
-        Value constant = 0;
+        Value constant{0};
```
chfast, line 297:
```suggestion
        Value constant{};
```

lib/fizzy/instructions.cpp
```diff
@@ -448,12 +448,12 @@ constexpr uint8_t instruction_max_align_table[256] = {
 
 const InstructionType* get_instruction_type_table() noexcept
 {
-    return instruction_type_table;
+    return static_cast<const InstructionType*>(instruction_type_table);
```
chfast, line 451: This does not make much sense.

lib/fizzy/trunc_boundaries.hpp
```diff
@@ -51,24 +50,22 @@ template <>
...
+    static constexpr float lower = -1.0f;
+    static constexpr float upper = 18446744073709551616.0f;
 };
 
 template <>
 struct trunc_boundaries<double, int64_t>
 {
     /// The first representable value lower than theoretical -9223372036854775809.
     static constexpr double lower = -9223372036854777856.0;
-
```
chfast, line 71: What is the problem here?

axic: Here, nothing, just some cases had an empty line other don't, made that consistent.

chfast: To separate elements with comments.

axic: What do you mean? Doxygen?

lib/fizzy/value.hpp
```diff
@@ -71,13 +71,13 @@ constexpr int32_t Value::as<int32_t>() const noexcept
 }
 
 template <>
-constexpr float Value::as() const noexcept
+constexpr float Value::as<float>() const noexcept
```
chfast, line 74: I'm surprised it worked before.

lib/fizzy/execute.hpp
```diff
@@ -20,7 +20,7 @@ struct ExecutionResult
 {
     const bool trapped = false;
     const bool has_value = false;
-    const Value value = 0;
+    const Value value{};
```
gumb0, line 23: @chfast After this default-initialization which member of `Value` is it allowed to access without UB?

lib/fizzy/trunc_boundaries.hpp
gumb0: I understood that it's to show that the comment refers to the next line only, and doesn't to the line after.

lib/fizzy/execute.hpp
chfast: > If T is a union type, the first non-static named data member is zero-initialized and all padding is initialized to zero bits. 

Maybe that is not the best idea in the end.

lib/fizzy/value.hpp
axic: There's a lot of surprising things

lib/fizzy/execute.hpp
axic: So what should be done here?

chfast: I did some experiments and this is fine.

lib/fizzy/trunc_boundaries.hpp
chfast: Yes, but the new way is fine too.

lib/fizzy/execute.hpp
gumb0: At some point documentation for `ExecutionResult` shoud be added, explaining how it is allowed to access `value` (only when  `!trapped & has_value` and only as the correct type)

PR #471, opened 2020-08-11
Implement f{32,64}.copysign instructions

chfast:
image

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -1554,6 +1554,16 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value> 
             binary_op(stack, fdiv<float>);
             break;
         }
+        case Instr::f32_copysign:
+        {
+            // TODO: This is not optimal implementation. The std::copysign() is inlined, but
+            //       it affects the compiler to still use SSE vectors (probably due to C++ ABI)
+            //       while this can be implemented with just generic registers and integer
+            //       instructions: (a & ABS_MASK) | (b & SIGN_MASK).
+            //       https://godbolt.org/z/aPqvfo
+            binary_op(stack, static_cast<float (*)(float, float)>(std::copysign));
+            break;
+        }
 
         case Instr::f64_abs:
         {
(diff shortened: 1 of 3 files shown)


lib/fizzy/execute.cpp
```diff
@@ -1470,6 +1470,16 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value>
             binary_op(stack, std::divides<float>{});
             break;
         }
+        case Instr::f32_copysign:
+        {
+            // TODO: This is not optimal implementation. The std::copysign() is inlined, but
```
axic, line 1475: Why not implement it then? Is that planned after merging this? Did you only see this for `copysign`?

chfast: Only for `copysign` and `abs`. Also theoretical UB will be involved, so I prefer to have all spec tests passed before doing that.

test/unittests/execute_floating_point_test.cpp
```diff
@@ -408,6 +408,35 @@ TYPED_TEST(execute_floating_point_types, div)
...
+    using FP = FP<TypeParam>;
+    using Limits = typename FP::Limits;
+
+    auto instance = instantiate(parse(this->get_binop_code(Instr::f32_copysign)));
+    const auto exec = [&](auto arg1, auto arg2) { return execute(*instance, 0, {arg1, arg2}); };
+
+    std::vector p_values(
+        std::begin(this->positive_special_values), std::end(this->positive_special_values));
+    for (const auto x :
+        {TypeParam{0}, Limits::infinity(), FP::nan(FP::canon), FP::nan(FP::canon + 1), FP::nan(1)})
```
gumb0, line 422:
```suggestion
        {TypeParam{0}, Limits::infinity(), FP::nan(FP::canon), -FP::nan(FP::canon), FP::nan(FP::canon + 1), -FP::nan(FP::canon + 1), FP::nan(1), -FP::nan(1)})
```

lib/fizzy/execute.cpp
```diff
@@ -1538,6 +1538,16 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value>
             binary_op(stack, fdiv<float>);
             break;
         }
+        case Instr::f32_copysign:
+        {
+            // TODO: This is not optimal implementation. The std::copysign() is inlined, but
+            //       it afects the compiler to still use SSE vectors (probably due to C++ABI)
```
axic, line 1544: Nit: affects

chfast: Where is spellcheck???

axic: I'm here :wave:

PR #472, opened 2020-08-11
Implement f{32,64}.{min,max} instructions

chfast:
image

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -517,6 +517,18 @@ __attribute__((no_sanitize("float-divide-by-zero"))) inline constexpr T fdiv(T a
     return a / b;  // For IEC 559 (IEEE 754) floating-point types division by 0 is defined.
 }
 
+template <typename T>
+inline constexpr T fmin(T a, T b) noexcept
+{
+    if (std::isnan(a) || std::isnan(b))
+        return std::numeric_limits<T>::quiet_NaN();
+
+    if (a == 0 && b == 0 && (std::signbit(a) == 1 || std::signbit(b) == 1))
+        return -T{0};
+
+    return b < a ? b : a;
+}
+
 std::optional<uint32_t> find_export(const Module& module, ExternalKind kind, std::string_view name)
 {
     const auto it = std::find_if(module.exportsec.begin(), module.exportsec.end(),
@@ -1554,6 +1566,11 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value> 
             binary_op(stack, fdiv<float>);
             break;
         }
+        case Instr::f32_min:
+        {
+            binary_op(stack, fmin<float>);
+            break;
+        }
         case Instr::f32_copysign:
         {
             // TODO: This is not optimal implementation. The std::copysign() is inlined, but
@@ -1589,6 +1606,11 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value> 
         case Instr::f64_div:
         {
             binary_op(stack, fdiv<double>);
+            break;
+        }
+        case Instr::f64_min:
+        {
+            binary_op(stack, fmin<double>);
             break;
         }
         case Instr::f64_copysign:
@@ -1737,7 +1759,6 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value> 
         case Instr::f32_nearest:
         case Instr::f32_sub:
         case Instr::f32_mul:
-        case Instr::f32_min:
         case Instr::f32_max:
         case Instr::f64_ceil:
         case Instr::f64_floor:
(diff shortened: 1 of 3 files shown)


lib/fizzy/execute.cpp
```diff
@@ -485,6 +485,26 @@ inline uint64_t popcnt64(uint64_t value) noexcept
     return static_cast<uint64_t>(__builtin_popcountll(value));
 }
 
+template <typename T>
+inline constexpr T fmin(T a, T b) noexcept
+{
+    if (std::isnan(a))
+        return a;
+    if (std::isnan(b))
```
axic, line 493: Are there any tests deciding this order? e.g. what happens if both are nan?

chfast:
Has bugs.

lib/fizzy/execute.cpp
chfast: Yes. The standard NaN propagation rules apply.

test/unittests/execute_floating_point_test.cpp
```diff
@@ -264,6 +264,8 @@ TYPED_TEST(execute_floating_point_types, binop_nan_propagation)
     constexpr Instr opcodes[] = {
         Instr::f32_add,
         Instr::f32_div,
+        Instr::f32_min,
+        Instr::f32_max,
```
axic, line 268: The spec says nothing about canonical nans, yet all these tests below pass?

test/unittests/execute_floating_point_test.cpp
```diff
@@ -581,6 +583,54 @@ TYPED_TEST(execute_floating_point_types, div)
...
+            const auto a = this->ordered_special_values[i];
+            const auto b = this->ordered_special_values[j];
+            if (!std::isnan(a) && !std::isnan(b))
+            {
+                EXPECT_THAT(execute(*instance, 0, {a, b}), Result(i < j ? a : b)) << a << ", " << b;
+            }
+        }
+    }
+
+    // fmin(+-0, -+0) = -O
```
axic, line 604:
```suggestion
    // fmin(+-0, -+0) = -0
```

test/unittests/execute_floating_point_test.cpp
```diff
@@ -581,6 +583,54 @@ TYPED_TEST(execute_floating_point_types, div)
     EXPECT_THAT(exec(TypeParam{0xABCD.01p7}, TypeParam{4}), Result(TypeParam{0x1.579A02p20}));
 }
 
+TYPED_TEST(execute_floating_point_types, min)
+{
+    auto instance = instantiate(parse(this->get_binop_code(Instr::f32_min)));
+
+    // Check every pair from cartesian product of the list of values.
+    for (size_t i = 0; i < std::size(this->ordered_special_values); ++i)
```
axic, line 591: While using this list is nice, I think having explicit test cases based on the spec (in the opening commit) would be useful.

test/unittests/execute_floating_point_test.cpp
chfast: It does. Implicitly, by using `nans_N{}` pseudo-function. I.e. if all NaN inputs are canonical NaNs the output must be a canonical NaN as well. Otherwise, an arithmetic NaN is fine.

test/unittests/execute_floating_point_test.cpp
chfast: Duplicating the same checks is less confusing?

axic: I'm not sold on this list, it is a pain looking it up every single time.

chfast: Done.

lib/fizzy/execute.cpp
```diff
@@ -534,10 +532,8 @@ inline constexpr T fmin(T a, T b) noexcept
 template <typename T>
 inline constexpr T fmax(T a, T b) noexcept
 {
-    if (std::isnan(a))
-        return a;
-    if (std::isnan(b))
-        return b;
+    if (std::isnan(a) || std::isnan(b))
+        return std::numeric_limits<T>::quiet_NaN();
```
axic, line 536: Where did this manifest itself as an error?

chfast: In the NaN propagation test - for signaling NaN inputs, we must return an arithmetic NaN.

lib/fizzy/execute.cpp
```diff
@@ -517,6 +517,30 @@ __attribute__((no_sanitize("float-divide-by-zero"))) inline constexpr T fdiv(T a
     return a / b;  // For IEC 559 (IEEE 754) floating-point types division by 0 is defined.
 }
 
+template <typename T>
+inline constexpr T fmin(T a, T b) noexcept
+{
+    if (std::isnan(a) || std::isnan(b))
+        return std::numeric_limits<T>::quiet_NaN();
```
gumb0, line 524: Is it guaranteed that it's a canonical NaN? I know we have a test for that, but maybe it's worth a `static_assert`?

chfast: No. But I don't have a way to check it here.

PR #475, opened 2020-08-12
test: Add negative NaN test cases

gumb0:


test/unittests/execute_floating_point_test.cpp
```diff
@@ -213,14 +222,33 @@ TYPED_TEST(execute_floating_point_types, binop_nan_propagation)
...
+
         EXPECT_THAT(execute(*instance, 0, {anan, cnan}), ArithmeticNaN(TypeParam{}));
+        EXPECT_THAT(execute(*instance, 0, {anan, -cnan}), ArithmeticNaN(TypeParam{}));
+        EXPECT_THAT(execute(*instance, 0, {-anan, cnan}), ArithmeticNaN(TypeParam{}));
+        EXPECT_THAT(execute(*instance, 0, {-anan, -cnan}), ArithmeticNaN(TypeParam{}));
+
         EXPECT_THAT(execute(*instance, 0, {cnan, anan}), ArithmeticNaN(TypeParam{}));
+        EXPECT_THAT(execute(*instance, 0, {cnan, -anan}), ArithmeticNaN(TypeParam{}));
+        EXPECT_THAT(execute(*instance, 0, {-cnan, anan}), ArithmeticNaN(TypeParam{}));
+        EXPECT_THAT(execute(*instance, 0, {-cnan, -anan}), ArithmeticNaN(TypeParam{}));
```
axic, line 251: Shouldn't we check the sign of the nan on the result side?

chfast: That is not requires as any sign is fine.

gumb0: The spec restricts only the payload of the propagated NaN.

axic: > The spec restricts only the payload of the propagated NaN.

That means it could be any sign?

gumb0: Yes.

axic: I mean in the above example, it is CPU specific whether `-cnan, -anan` will result in `anan` or `-anan`?

chfast: Both `cnan` and `-cnan` are wasm canonical NaNs. Similarly for arithmetic NaNs. The sign is ignored.

gumb0: I don't know how CPUs propagate the sign bit, it probably depends on the opcode, too.

axic: I'd add a todo here to investigate this further.

chfast: I added a TODO comment.

PR #476, opened 2020-08-12
Implement f{32,64}.abs instructions

gumb0:
image

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -1521,9 +1521,23 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value> 
             break;
         }
 
+        case Instr::f32_abs:
+        {
+            // TODO: This can be optimized https://godbolt.org/z/aPqvfo
+            unary_op(stack, static_cast<float (*)(float)>(std::fabs));
+            break;
+        }
+
         case Instr::f32_add:
         {
             binary_op(stack, std::plus<float>{});
+            break;
+        }
+
+        case Instr::f64_abs:
+        {
+            // TODO: This can be optimized https://godbolt.org/z/aPqvfo
+            unary_op(stack, static_cast<double (*)(double)>(std::fabs));
             break;
         }
 
@@ -1667,7 +1681,6 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value> 
             break;
         }
 
-        case Instr::f32_abs:
         case Instr::f32_neg:
         case Instr::f32_ceil:
         case Instr::f32_floor:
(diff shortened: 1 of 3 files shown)


lib/fizzy/execute.cpp
```diff
@@ -1521,12 +1521,26 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value>
             break;
         }
 
+        case Instr::f32_abs:
+        {
+            // TODO: This can be optimized https://godbolt.org/z/aPqvfo
```
gumb0, line 1526: Included link to the same goldbolt as in #471

lib/fizzy/execute.cpp
```diff
@@ -1521,12 +1521,26 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value>
...
+            unary_op(stack, static_cast<float (*)(float)>(std::fabs));
```
gumb0, line 1527: `std::fabs` is unfortunately not a template, but several overloaded variants.

gumb0: Also would work without casts: `fabsf` and `fabs` C functions.

chfast: > (Also would work without casts: `fabsf` and `fabs` C functions.)

That would only work for `fabsf`...

test/unittests/execute_floating_point_test.cpp
```diff
@@ -363,6 +385,27 @@ TYPED_TEST(execute_floating_point_types, add)
     EXPECT_THAT(exec(TypeParam{0x0.287p2}, TypeParam{0x1.FFp4}), Result(TypeParam{0x1.048Ep5}));
 }
 
+TYPED_TEST(execute_floating_point_types, abs)
```
chfast, line 388: Put the test by the opcode value, i.e. before `add`.

gumb0: Moved.

test/unittests/execute_floating_point_test.cpp
```diff
@@ -293,6 +315,28 @@ TYPED_TEST(execute_floating_point_types, compare)
...
+    using FP = FP<TypeParam>;
+    using Limits = typename FP::Limits;
+
+    auto instance = instantiate(parse(this->get_unop_code(Instr::f32_abs)));
+    const auto exec = [&](auto arg) { return execute(*instance, 0, {arg}); };
+
+    std::vector p_values(
+        std::begin(this->positive_special_values), std::end(this->positive_special_values));
+    for (const auto x :
+        {TypeParam{0}, Limits::infinity(), FP::nan(FP::canon), FP::nan(FP::canon + 1), FP::nan(1)})
```
axic, line 329: Based on the spec, can you add `-0`?

PR #477, opened 2020-08-12
Implement f{32,64}.neg instructions

gumb0:
image

### DIFF
--- a/test/unittests/execute_floating_point_test.cpp
+++ b/test/unittests/execute_floating_point_test.cpp
@@ -338,6 +338,28 @@ TYPED_TEST(execute_floating_point_types, abs)
     }
 }
 
+TYPED_TEST(execute_floating_point_types, neg)
+{
+    using FP = FP<TypeParam>;
+    using Limits = typename FP::Limits;
+
+    auto instance = instantiate(parse(this->get_unop_code(Instr::f32_neg)));
+    const auto exec = [&](auto arg) { return execute(*instance, 0, {arg}); };
+
+    std::vector p_values(
+        std::begin(this->positive_special_values), std::end(this->positive_special_values));
+    for (const auto x :
+        {TypeParam{0}, Limits::infinity(), FP::nan(FP::canon), FP::nan(FP::canon + 1), FP::nan(1)})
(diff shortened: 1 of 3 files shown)


test/unittests/execute_floating_point_test.cpp
```diff
@@ -407,6 +407,28 @@ TYPED_TEST(execute_floating_point_types, abs)
     }
 }
 
+TYPED_TEST(execute_floating_point_types, neg)
```
chfast, line 410: Order the tests by opcode, if in the same fixture.

gumb0: Moved.

test/unittests/execute_floating_point_test.cpp
```diff
@@ -337,6 +337,28 @@ TYPED_TEST(execute_floating_point_types, abs)
...
+        std::begin(this->positive_special_values), std::end(this->positive_special_values));
+    for (const auto x :
+        {TypeParam{0}, Limits::infinity(), FP::nan(FP::canon), FP::nan(FP::canon + 1), FP::nan(1)})
+        p_values.push_back(x);
+
+    for (const auto p : p_values)
+    {
+        // fneg(+-p) = -+p
+        EXPECT_THAT(exec(p), Result(-p));
+        EXPECT_THAT(exec(-p), Result(p));
```
axic, line 358: Aren't in these tests practically trusting the compiler compared to the others where we have explicit constants?

chfast: We trust that negation work everywhere.

axic: If we wanted to be pedantic we'd always test against the encoded representation.

chfast: Possible in the future... i.e. we would need to extend `FP` to do some software floating-point operations.

axic: Something like `wast2wasm4cpp` script could work, which evaluates the C++ code and puts in the hard constants. We could manually verify that when committed and at every change.

PR #481, opened 2020-08-13
Implement f{32,64}.{ceil,floor,trunc} instructions

chfast:
image

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -543,7 +543,15 @@ inline T ffloor(T value) noexcept
         return std::numeric_limits<T>::quiet_NaN();  // Positive canonical NaN.
 
     // The FE_INEXACT error is ignored (whenever the implementation reports it at all).
-    return std::floor(value);
+    const auto result = std::floor(value);
+
+    // TODO: GCC BUG WORKAROUND:
+    // GCC implements std::floor() with  __builtin_floor().
+    // When rounding direction is set to FE_DOWNWARD
+    // the __builtin_floor() outputs -0 where it should +0.
+    // The following workarounds the issue by using the fact that the sign of
+    // the output must always match the sign of the input value.
+    return std::copysign(result, value);
 }
 
 template <typename T>
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -525,6 +525,39 @@ inline uint64_t popcnt64(uint64_t value) noexcept
 }
 
 template <typename T>
+inline T fceil(T value) noexcept
+{
+    static_assert(std::is_floating_point_v<T>);
+    if (std::isnan(value))
+        return std::numeric_limits<T>::quiet_NaN();  // Positive canonical NaN.
+
+    // The FE_INEXACT error is ignored (whenever the implementation reports it at all).
+    return std::ceil(value);
+}
+
+template <typename T>
+inline T ffloor(T value) noexcept
+{
+    static_assert(std::is_floating_point_v<T>);
+    if (std::isnan(value))
+        return std::numeric_limits<T>::quiet_NaN();  // Positive canonical NaN.
+
+    // The FE_INEXACT error is ignored (whenever the implementation reports it at all).
+    return std::floor(value);
+}
+
+template <typename T>
+inline T ftrunc(T value) noexcept
+{
+    static_assert(std::is_floating_point_v<T>);
+    if (std::isnan(value))
+        return std::numeric_limits<T>::quiet_NaN();  // Positive canonical NaN.
+
+    // The FE_INEXACT error is ignored (whenever the implementation reports it at all).
+    return std::trunc(value);
+}
+
+template <typename T>
 __attribute__((no_sanitize("float-divide-by-zero"))) inline constexpr T fdiv(T a, T b) noexcept
 {
     static_assert(std::numeric_limits<T>::is_iec559);
@@ -535,7 +568,7 @@ template <typename T>
 inline constexpr T fmin(T a, T b) noexcept
 {
     if (std::isnan(a) || std::isnan(b))
-        return std::numeric_limits<T>::quiet_NaN();
+        return std::numeric_limits<T>::quiet_NaN();  // Positive canonical NaN.
 
     if (a == 0 && b == 0 && (std::signbit(a) == 1 || std::signbit(b) == 1))
         return -T{0};
@@ -547,7 +580,7 @@ template <typename T>
 inline constexpr T fmax(T a, T b) noexcept
 {
     if (std::isnan(a) || std::isnan(b))
-        return std::numeric_limits<T>::quiet_NaN();
+        return std::numeric_limits<T>::quiet_NaN();  // Positive canonical NaN.
 
     if (a == 0 && b == 0 && (std::signbit(a) == 0 || std::signbit(b) == 0))
         return T{0};
@@ -1586,11 +1619,27 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value> 
             unary_op(stack, std::negate<float>{});
             break;
         }
+        case Instr::f32_ceil:
+        {
+            unary_op(stack, fceil<float>);
+            break;
+        }
+        case Instr::f32_floor:
+        {
+            unary_op(stack, ffloor<float>);
+            break;
+        }
+        case Instr::f32_trunc:
+        {
+            unary_op(stack, ftrunc<float>);
+            break;
+        }
         case Instr::f32_sqrt:
         {
             unary_op(stack, static_cast<float (*)(float)>(std::sqrt));
             break;
         }
+
         case Instr::f32_add:
         {
             binary_op(stack, std::plus<float>{});
@@ -1643,11 +1692,27 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value> 
             unary_op(stack, std::negate<double>{});
             break;
         }
+        case Instr::f64_ceil:
+        {
+            unary_op(stack, fceil<double>);
+            break;
+        }
+        case Instr::f64_floor:
+        {
+            unary_op(stack, ffloor<double>);
+            break;
+        }
+        case Instr::f64_trunc:
+        {
+            unary_op(stack, ftrunc<double>);
+            break;
+        }
         case Instr::f64_sqrt:
         {
             unary_op(stack, static_cast<double (*)(double)>(std::sqrt));
             break;
         }
+
         case Instr::f64_add:
         {
             binary_op(stack, std::plus<double>{});
@@ -1843,13 +1908,7 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, span<const Value> 
             break;
         }
 
-        case Instr::f32_ceil:
-        case Instr::f32_floor:
-        case Instr::f32_trunc:
         case Instr::f32_nearest:
-        case Instr::f64_ceil:
-        case Instr::f64_floor:
-        case Instr::f64_trunc:
         case Instr::f64_nearest:
             throw unsupported_feature("Floating point instruction.");
         default:
--- a/test/unittests/execute_floating_point_test.cpp
+++ b/test/unittests/execute_floating_point_test.cpp
@@ -11,7 +11,13 @@
 #include <test/utils/floating_point_utils.hpp>
 #include <test/utils/hex.hpp>
 #include <array>
+#include <cfenv>
 #include <cmath>
+
+// Enables access and modification of the floating-point environment.
+// Here we use it to change rounding direction in tests.
+// Although required by the C standard, neither GCC nor Clang supports it.
+#pragma STDC FENV_ACCESS ON
 
 using namespace fizzy;
 using namespace fizzy::test;
@@ -95,6 +101,9 @@ class execute_floating_point_types : public testing::Test
 public:
     using L = typename FP<T>::Limits;
 
+    static constexpr auto all_rounding_directions = {
+        FE_TONEAREST, FE_DOWNWARD, FE_UPWARD, FE_TOWARDZERO};
+
     // The list of positive floating-point values without zeros, infinities and NaNs.
     inline static const std::array positive_special_values{
         L::denorm_min(),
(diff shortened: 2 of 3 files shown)


test/unittests/execute_floating_point_test.cpp
```diff
@@ -438,6 +441,86 @@ TYPED_TEST(execute_floating_point_types, neg)
...
+{
+    using Limits = typename FP<T>::Limits;
+
+    // The "int only" is the range of the floating-point type of only consecutive integer values.
+    // (Integer shift is used instead of std::pow() because of the clang compiler bug).
+    static constexpr auto int_only_begin = T{uint64_t{1} << (Limits::digits - 1)};
+    static constexpr auto int_only_end = T{uint64_t{1} << Limits::digits};
+
+    // The list of rounding test cases as pairs (input, expected_trunc) with only positive inputs.
+    inline static const std::pair<T, T> tests[]{
```
gumb0, line 456:
```suggestion
    inline static const std::pair<T, T> positive_trunc_tests[]{
```

test/unittests/execute_floating_point_test.cpp
```diff
@@ -438,6 +441,86 @@ TYPED_TEST(execute_floating_point_types, neg)
...
+        {Limits::infinity(), Limits::infinity()},
+    };
+};
+
+TYPED_TEST(execute_floating_point_types, ceil)
+{
+    auto instance = instantiate(parse(this->get_unop_code(Instr::f32_ceil)));
+    const auto exec = [&](auto arg) { return execute(*instance, 0, {arg}); };
+
+    for (const auto& [arg, expected_trunc] : rounding_test_cases<TypeParam>::tests)
```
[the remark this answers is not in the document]
chfast: True. I don't need specialization any more.

test/unittests/execute_floating_point_test.cpp
```diff
@@ -155,6 +155,48 @@ class execute_floating_point_types : public testing::Test
         FP<T>::nan(1),
     };
 
+    // The "int only" is the range of the floating-point type of only consecutive integer values.
```
gumb0, line 158: I'm not sure I understand this, is it the range where precision is not enough to represent non-integers, but still is enough to represent every consecutive integer in the range?

gumb0: Why is it special for these opcodes?

chfast: The `int_only_end - 1` is the JavaScript's "safe max integer". Outside this range not every integer is representable. Just that.

Maybe also that it is easy to make sure the input literals are actually representable.

chfast: I think I expected a bit more out of these when I first started computing the input values.

gumb0: I think I would be fine with just 0, 1, 2 cases, these are just confusing...

axic: > The int_only_end - 1 is the JavaScript's "safe max integer". Outside this range not every integer is representable. Just that.

JavaScript safe max integer is 53-bits, this seems to be not the case here?

test/unittests/execute_floating_point_test.cpp
```diff
@@ -439,6 +484,56 @@ TYPED_TEST(execute_floating_point_types, neg)
     }
 }
 
+TYPED_TEST(execute_floating_point_types, ceil)
+{
+    auto instance = instantiate(parse(this->get_unop_code(Instr::f32_ceil)));
+    const auto exec = [&](auto arg) { return execute(*instance, 0, {arg}); };
+
```
axic, line 491: I think there should be a specific case for `-q = -0 (if -1 < -q < 0)` here.

The rules about infinity and zero matches that of C++ too (https://en.cppreference.com/w/cpp/numeric/math/ceil), but stil having tests here would be needed.

Not sure the rule requires quiet_nan, but needs a test too, because C++ just passes through any nan.

lib/fizzy/execute.cpp
```diff
@@ -524,6 +524,33 @@ inline uint64_t popcnt64(uint64_t value) noexcept
     return static_cast<uint64_t>(__builtin_popcountll(value));
 }
 
+template <typename T>
+inline T fceil(T value) noexcept
+{
+    static_assert(std::is_floating_point_v<T>);
+    if (std::isnan(value))
+        return std::numeric_limits<T>::quiet_NaN();
+    return std::ceil(value);
```
axic, line 533: From https://en.cppreference.com/w/cpp/numeric/math/ceil:

Screenshot 2020-08-14 at 19 31 28

What about this? The same applies for `std::floor`, and `std::trunc`.

test/unittests/execute_floating_point_test.cpp
chfast: 1. I can the special case.
2. Zeros and infinities are included in the `positive_trunc_tests`. Rules are simple and the same for all 4 rounding instructions, so I don't see point of having them duplicated everywhere.
3. NaNs are checked in `unop_nan_propagation` as this is common rules for all instructions (except some which ignore NaNs).

lib/fizzy/execute.cpp
axic: Okay, the documentation just kept calling it "floating point exception" and I thought it is an actual exception, not just a flag.

Can you leave perhaps a comment above ceil/floor that `FE_INEXACT` has no bearing on the correctness of this?

test/unittests/execute_floating_point_test.cpp
axic: Those special cases also exists for floor/trunc (actually 2 for trunc).

Will check `positive_trunc_tests`.

chfast: I added this comment in the `positive_trunc_tests`.

test/unittests/execute_floating_point_test.cpp
```diff
@@ -440,6 +492,72 @@ TYPED_TEST(execute_floating_point_types, neg)
...
+    EXPECT_THAT(exec(std::nextafter(TypeParam{1}, TypeParam{0})), Result(TypeParam{0}));
+
+    for (const auto& [arg, expected_trunc] :
+        execute_floating_point_types<TypeParam>::positive_trunc_tests)
+    {
+        // For positive values, the floor() is trunc().
+        EXPECT_THAT(exec(arg), Result(expected_trunc)) << arg << ": " << expected_trunc;
+
+        // For negative values, the floor() is trunc() - 1, unless the input is already an integer.
+        const auto expected_neg =
```
axic, line 533: First I thought `expected_neg` is negation and `expected_pos` is some kind of position, maybe calling them `expected_negative` and `expected_positive` is better.

test/unittests/execute_floating_point_test.cpp
```diff
@@ -440,6 +492,72 @@ TYPED_TEST(execute_floating_point_types, neg)
...
+}
+
+TYPED_TEST(execute_floating_point_types, trunc)
+{
+    auto instance = instantiate(parse(this->get_unop_code(Instr::f32_trunc)));
+    const auto exec = [&](auto arg) { return execute(*instance, 0, {arg}); };
+
+    // ftrunc(+q) = +0  (if 0 < +q < 1)
+    EXPECT_THAT(exec(FP<TypeParam>::Limits::denorm_min()), Result(TypeParam{0}));
+    EXPECT_THAT(exec(std::nextafter(TypeParam{1}, TypeParam{0})), Result(TypeParam{0}));
```
axic, line 546: Should these have swapped operand: `nextafter(0, 1)` e.g. the smallest number larger than 0 ?

chfast: The `denorm_min` is the smallest > 0.
The `nextafter(0, 1)` is the largest < 1.

test/unittests/execute_floating_point_test.cpp
chfast: Will do.

test/unittests/execute_floating_point_test.cpp
axic: If `nextafter(0, 1)` is the largest < 1, then why do you have `nextafter(0, 1)` here?

lib/fizzy/execute.cpp
chfast: This is GCC documentation about the floating point exceptions: https://www.gnu.org/software/libc/manual/html_node/FP-Exceptions.html
They can trigger SIGFPE signal, but this is not default behavior.

test/unittests/execute_floating_point_test.cpp
chfast: Wrong comment.
`nextafter(0,1)` is `denorm_min` - i.e. it is the next after 0 towards 1.
`nextafter(1,0` is the largest < 1.

test/unittests/execute_floating_point_test.cpp
gumb0: > The `int_only_end - 1` is the JavaScript's "safe max integer". Outside this range not every integer is representable. Just that.

Ok, and what is `int_only_begin` ?

lib/fizzy/execute.cpp
axic: Apparently can be forcefully disabled with `feenableexcept(0)`, see https://en.cppreference.com/w/cpp/numeric/fenv:

Screenshot 2020-08-17 at 10 52 10

axic: And there's the `non-call-exceptions` but that apparently also catches SIGSEGV.

axic: I think eventually we could have an enter/leave function for setting up FP properties. "enter" in the beginning of `execute` and then "enter"/"leave" wrapping on host functions.

Depends on which is more likely: heavy use of ceil/floor/trunc/round or not, because these are concerned by the rounding settings and the exceptions.

axic: Alternatively using the builtins wouldn't relieve us from this stdlib madness?

test/unittests/execute_floating_point_test.cpp
chfast: The beginning of this section. This in this section the fp value increases by `1.0`, before by `0.5`. I prefer testing around such "special" values instead of random values.

chfast: The main point of using `int_only_begin` is to tests `nearest()` as for `int_only_begin - 0.5` and `int_only_begin -1.5` we have equal distance to nearby integer values.

gumb0: I would reword/explain with something like this
```suggestion
    // The [int_only_begin; int_only_end) is the range of floating-point numbers, where each representable number is an integer and there are no fractional numbers between them. These numbers are represented as mantissa [0x1.00..0; 0x1.ff..f] and exponent 2**(mantissa digits without implicit leading 1).
```

test/unittests/execute_floating_point_test.cpp
```diff
@@ -440,6 +492,72 @@ TYPED_TEST(execute_floating_point_types, neg)
     }
 }
 
+TYPED_TEST(execute_floating_point_types, ceil)
+{
+    auto instance = instantiate(parse(this->get_unop_code(Instr::f32_ceil)));
+    const auto exec = [&](auto arg) { return execute(*instance, 0, {arg}); };
+
+    // fceil(-q) = -0  (if -1 < -q < 0)
```
gumb0, line 500: I'm not a fan of duplicating these cases, it might be confusing why it's tested two times.

Perhaps a comment, that it's tested in the loop over `positive_trunc_tests`, but added here for explicitness, would help.

test/unittests/execute_floating_point_test.cpp
gumb0: > The main point of using `int_only_begin` is to tests `nearest()` as for `int_only_begin - 0.5` and `int_only_begin -1.5` we have equal distance to nearby integer values.

This can also be added to a comment.

lib/fizzy/execute.cpp
chfast: > Alternatively using the builtins wouldn't relieve us from this stdlib madness?

This is not portable and we are using them currently anyway. This is the implementation of `std::ceil()` in libstdc++:
```cpp
  inline _GLIBCXX_CONSTEXPR float
  ceil(float __x)
  { return __builtin_ceilf(__x); }
```

I think sticking to the C stdlib is fine. If someone wants to enable GNU's extension to trap on FP exceptions be my guest.

test/unittests/execute_floating_point_test.cpp
axic: These are not tested two times, these are some edge cases.

gumb0: These are the same cases as `positive_trunc_tests` items 
`{std::nextafter(T{1}, T{0}), T{0}}` 
`{L::denorm_min(), 0}`  tested in the loop as
```
EXPECT_THAT(exec(-arg), Result(-expected_trunc)) << -arg << ": " << -expected_trunc;
```

similar in tests for other instructions

chfast: Are we fine here?

lib/fizzy/execute.cpp
axic: > If someone wants to enable GNU's extension to trap on FP exceptions be my guest.

Behaviour should be documented, and I'm leaning towards we should mention these in the README (since we don't have a complete API doc yet).

test/unittests/execute_floating_point_test.cpp
```diff
@@ -440,6 +497,76 @@ TYPED_TEST(execute_floating_point_types, neg)
...
+    // (also included in positive_trunc_tests, here for explicitness).
+    EXPECT_THAT(exec(std::nextafter(-TypeParam{1}, TypeParam{0})), Result(-TypeParam{0}));
+    EXPECT_THAT(exec(-FP<TypeParam>::Limits::denorm_min()), Result(-TypeParam{0}));
```
axic, line 508: So this is `-nextafter(0, 1)` ?

test/unittests/execute_floating_point_test.cpp
axic: @gumb0 you are correct. I still prefer if we have clear single test cases for the edge cases listed in the specific for the given instruction. It is much easier to review that every case is tested.

Stuff hidden in those loops is extremely hard to review, in my opinion.

gumb0: I'm fine with including it, already approved.

test/unittests/execute_floating_point_test.cpp
chfast: Yes.

chfast:
There is an issue with `std::floor` implementation.
When rounding direction is set to `FE_DOWNWARD`, `std::floor()` returns `-0` instead of `0` for the following configurations:
- GCC, Release, {float, double},
- GCC, Debug, float,
but GCC, Release, native build is fine.

lib/fizzy/execute.cpp
```diff
@@ -524,6 +524,45 @@ inline uint64_t popcnt64(uint64_t value) noexcept
...
+        return std::numeric_limits<T>::quiet_NaN();  // Positive canonical NaN.
+
+    // The FE_INEXACT error is ignored (whenever the implementation reports it at all).
+    const auto result = std::floor(value);
+
+    // TODO: GCC / glibc BUG WORKAROUND:
+    // When rounding direction is set to FE_DOWNWARD the std::floor outputs -0 where it should +0.
+    // The following workarounds the issue by using the fact that the sign of the output must always
+    // match the sign of the input value.
+    return std::copysign(result, value);
```
axic, line 552: Perhaps also leave the gcc bug link here (I know it was reported for ppc, but this seems similar).

test/unittests/execute_floating_point_test.cpp
```diff
@@ -11,8 +11,11 @@
 #include <test/utils/floating_point_utils.hpp>
 #include <test/utils/hex.hpp>
 #include <array>
+#include <cfenv>
 #include <cmath>
 
+#pragma STDC FENV_ACCESS ON
```
axic, line 17: I know it is possible to look what this means -- I just did that --, but perhaps nicer to leave a small warning here so that it is not removed in the future:
> Note: this is important because our tests need complete control over floating point settings.

test/unittests/execute_floating_point_test.cpp
```diff
@@ -155,6 +161,62 @@ class execute_floating_point_types : public testing::Test
         FP<T>::nan(1),
     };
 
+    // The [int_only_begin; int_only_end) is the range of floating-point numbers, where each
+    // representable number is an integer and there are no fractional numbers between them.
+    // These numbers are represented as mantissa [0x1.00..0; 0x1.ff..f]
+    // and exponent 2**(mantissa digits without implicit leading 1).
+    // The main point of using int_only_begin is to tests nearest() as for int_only_begin - 0.5 and
+    // int_only_begin - 1.5 we have equal distance to nearby integer values.
+    // (Integer shift is used instead of std::pow() because of the clang compiler bug).
```
axic, line 170: What clang compiler bug is this? I guess we have it linked in the hackmd, do we want to link to it here too?

test/unittests/execute_floating_point_test.cpp
```diff
@@ -440,6 +505,100 @@ TYPED_TEST(execute_floating_point_types, neg)
...
+
+        // ftrunc(+q) = +0  (if 0 < +q < 1)
+        // (also included in positive_trunc_tests, here for explicitness).
+        EXPECT_THAT(exec(FP<TypeParam>::Limits::denorm_min()), Result(TypeParam{0}));
+        EXPECT_THAT(exec(std::nextafter(TypeParam{1}, TypeParam{0})), Result(TypeParam{0}));
+
+        // ftrunc(-q) = -0  (if -1 < -q < 0)
+        // (also included in positive_trunc_tests, here for explicitness).
+        EXPECT_THAT(exec(std::nextafter(-TypeParam{1}, TypeParam{0})), Result(-TypeParam{0}));
+        EXPECT_THAT(exec(-FP<TypeParam>::Limits::denorm_min()), Result(-TypeParam{0}));
```
axic, line 589: Potentially all these would be easier to follow by using `nextafter(0, -1)` ?

lib/fizzy/execute.cpp
chfast: That one is fixed. We need to report new one, but we need a reproducer first.

test/unittests/execute_floating_point_test.cpp
chfast: I only have the `clang_bug` branch for now.

test/unittests/execute_floating_point_test.cpp
chfast: I don't think so. I prefer using named values. And `nextafter(0, -1)` is probably `-0`.

lib/fizzy/execute.cpp
gumb0: `ceil` that results in `-0` doesn't have this problem?

test/unittests/execute_floating_point_test.cpp
gumb0: I would comment what pragma does (enables access and modification of the floating-point environment (rounding directions in particular) ).

chfast: Yes, I will add it.

chfast: Added.

lib/fizzy/execute.cpp
chfast: `ceil` is tested the same way, and no issues.

Do you mean `ceil()` returning +0 where it should -0? The test cases of values from `[-1,-0]` are included.

axic (approved):
Looks good to me, but please squash the appropriate commits.

chfast:
> Looks good to me, but please squash the appropriate commits.

Done. I left the "GCC bug workaround" commit separated.

gumb0 (approved):

PR #483, opened 2020-08-13
benchmarks: add Ramunajan's formula for calculating pi

axic:


test/benchmarks/ramunajan_pi.inputs
```diff
@@ -0,0 +1,14 @@
+pi_10_runs
+ramunajan
+i:I
+10
+
+31415926535897932
+
+pi_30_runs
```
chfast, line 8: I don't think adding more than one execution benchmark adds any additional information. It rather just make running benchmarks take longer.

axic:
The wat can be found here: https://gist.github.com/axic/1b330d744148e2aa46085f0f69521b40

test/benchmarks/ramanujan_pi.c
```diff
@@ -0,0 +1,34 @@
...
+    for (unsigned i = n; i > 1; --i)
+        ret *= i;
+    return ret;
+}
+
+static double _ramanujan_pi(unsigned n)
+{
+    // This dumb implementation of Ramanujan series calculates 1/pi
+    double ret = 0.0;
+    for (unsigned k = 0; k < n; k++) {
```
chfast, line 17: For "more floating point" and fun, you can make it `double k`.

test/benchmarks/ramanujan_pi.c
```diff
@@ -0,0 +1,34 @@
+#include <math.h>
+
+#define WASM_EXPORT __attribute__((visibility("default")))
+
+static unsigned factorial(unsigned n)
```
chfast, line 5: Can be also implemented with `double n`.

axic: Yeah thought about those, but kind of resorted that perhaps for forcefully having "more floating point" perhaps a hand crafted synthetic benchmark using all ops is better?

chfast: So what's the point of this benchmark if only 10-20% of instructions is floating-point?

axic: Absolute number of instructions or instructions weighted by the number of times they are executed?

If you profile it it should be clear the heavy part is the floating point calculation (pow, etc.). The factorial part only works with small number (<30).

chfast: Ok then.

PR #484, opened 2020-08-13
test: Fix running spec tests that expect NaN result

gumb0:


test/spectests/spectests.cpp
```diff
@@ -420,26 +403,57 @@ class test_runner
...
+    static fizzy::Value read_floating_point_value(const json& v)
+    {
+        assert(
+            v.at("type").get<std::string>() == "f32" || v.at("type").get<std::string>() == "f64");
+        assert(!is_canonical_nan(v) && !is_arithmetic_nan(v));
+
+        // JSON tests have all values including floats serialized as 64-bit unsigned integers.
         const uint64_t uint_value = std::stoull(v.at("value").get<std::string>());
+        /// The unsigned integer type matching the size of this floating-point type.
+        using UintType = std::conditional_t<std::is_same_v<T, float>, uint32_t, uint64_t>;
```
chfast, line 426:
```suggestion
        using UintType = FP<T>::UintType;
```

test/utils/floating_point_utils.hpp
```diff
@@ -66,6 +66,10 @@ struct FP
         return std::isnan(value) ? (as_uint() & mantissa_mask) : 0;
     }
 
+    bool is_canonical_nan() const noexcept { return nan_payload() == canon; }
```
chfast, line 69: Please also update `CanonicalNaN` and `ArithemeticNaN` matcher.

gumb0: I did already.

test/spectests/spectests.cpp
gumb0: Fixed.

test/spectests/spectests.cpp
```diff
@@ -420,26 +403,57 @@ class test_runner
...
+
+    template <typename T>
+    static fizzy::Value read_floating_point_value(const json& v)
+    {
+        assert(
+            v.at("type").get<std::string>() == "f32" || v.at("type").get<std::string>() == "f64");
+        assert(!is_canonical_nan(v) && !is_arithmetic_nan(v));
+
+        // JSON tests have all values including floats serialized as 64-bit unsigned integers.
         const uint64_t uint_value = std::stoull(v.at("value").get<std::string>());
```
axic, line 424: @gumb0 Did not mention before merging, because this question affected `master` too (e.g. not introduced via this PR): do we want to wrap any of these in proper exception handling for invalid inputs (see https://en.cppreference.com/w/cpp/string/basic_string/stoul for `invalid_argument` or `out_of_range`) or it is fine leaving it to the general exception catching we have?

gumb0: Actually it would be nice to output a line number in the case of exception, now you see only that exception happened somewhere in the file.

chfast: Wait until it happens again.

PR #485, opened 2020-08-13
ci: run full spec test suite

gumb0:
Depends on #484

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -140,13 +140,13 @@ commands:
         default: false
       expected_passed:
         type: integer
-        default: 11409
+        default: 18695
       expected_failed:
         type: integer
         default: 3
       expected_skipped:
         type: integer
-        default: 401
+        default: 681
 
     steps:
       - run:


circle.yml
```diff
@@ -140,21 +140,21 @@ commands:
...
+        default: 681
 
     steps:
       - run:
           name: "Download spectest files"
           working_directory: ~/build
           command: |
             if [ ! -d wasm-spec ]; then
-              git clone https://github.com/wasmx/wasm-spec --branch vanilla-json --depth 1
+              git clone https://github.com/wasmx/wasm-spec --branch vanilla-json-full --depth 1
```
chfast, line 157: Why I'm not seeing this branch in wasmx/wasm-spec?

Can't we just use the same branch?

gumb0: It's there https://github.com/wasmx/wasm-spec/tree/vanilla-json-full

Maybe we can use the same one, if we merge test expectations update immediately after I push to it.

chfast: Ok, please do it now.

gumb0: Pushed to `vanilla-json`

PR #486, opened 2020-08-13
Implement f{32,64}.nearest instructions

chfast:


lib/fizzy/execute.cpp
```diff
@@ -551,6 +552,18 @@ inline T ftrunc(T value) noexcept
     return std::trunc(value);
 }
 
+template <typename T>
+T fnearest(T a) noexcept
```
axic, line 556: This doesn't seem to enforce the same NaN rules as #481, yet the spec seems to be the same regarding it?

test/unittests/execute_test.cpp
```diff
@@ -829,223 +829,6 @@ TEST(execute, memory_copy_32bytes)
     EXPECT_EQ(output, input);
 }
 
-TEST(execute, fp_instructions)
```
axic, line 832: This is a landmark event!

test/unittests/execute_floating_point_test.cpp
```diff
@@ -534,6 +535,43 @@ TYPED_TEST(execute_floating_point_types, trunc)
     }
 }
 
+TYPED_TEST(execute_floating_point_types, nearest)
+{
+    auto instance = instantiate(parse(this->get_unop_code(Instr::f32_nearest)));
+    const auto exec = [&](auto arg) { return execute(*instance, 0, {arg}); };
+
+    for (const auto& [arg, expected_trunc] : this->positive_trunc_tests)
```
axic, line 543: Note to self:
```
nearest(+-inf) = +-inf
nearest(+-0) = +-0
```
are included in `positive_trunc_tests`.

test/unittests/execute_floating_point_test.cpp
```diff
@@ -534,6 +535,43 @@ TYPED_TEST(execute_floating_point_types, trunc)
     }
 }
 
+TYPED_TEST(execute_floating_point_types, nearest)
+{
+    auto instance = instantiate(parse(this->get_unop_code(Instr::f32_nearest)));
+    const auto exec = [&](auto arg) { return execute(*instance, 0, {arg}); };
+
```
axic, line 542: Can you add specific cases for
```
fnearest(+q) = +0 (if 0 < +q <= 0.5)
fnearest(-q) = -0 (if -0.5 <= -q < 0)
```

lib/fizzy/execute.cpp
chfast: We have them tested. I was expecting this will fail and I will need `if (isnan())` case, but seems not required. Will check against the C spec though.

test/unittests/execute_floating_point_test.cpp
axic: Did you add these yet?

lib/fizzy/execute.cpp
axic: Did you check?

test/unittests/execute_floating_point_test.cpp
axic: Can you leave this here as a comment?

lib/fizzy/execute.cpp
```diff
@@ -565,6 +566,37 @@ inline T ftrunc(T value) noexcept
...
+{
+    // This implementation uses std::nearbyint() which rounds to integer using current rounding
```
axic, line 572: Add static assert that T is floating point?

chfast: Added.

test/unittests/execute_floating_point_test.cpp
chfast: Added.

lib/fizzy/execute.cpp
chfast: This is the same as in other related functions: for NaN input it returns unspecified NaN.

chfast:
This is a working version with `volatile` use "synchronize" calls. This is not great, but works at least for the tested compiler configurations.

We may spend another week to:
- try to synchronize with atomics,
- report compiler bugs.
But I think we should go with this version for now.

PR #491, opened 2020-08-17
ci: don't run fizzy-spectests --skip-validation

gumb0:
Not needed anymore.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -140,12 +140,13 @@ commands:
         default: false
       expected_passed:
         type: integer
+        default: 11409
       expected_failed:
         type: integer
-        default: 0
+        default: 3
       expected_skipped:
         type: integer
-        default: 0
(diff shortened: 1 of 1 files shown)


axic:
Why only remove on ci?

gumb0:
> Why only remove on ci?

It might make sense to keep it in the tool, to easily see how many validation tests are being run.

circle.yml
```diff
@@ -395,11 +395,6 @@ jobs:
       - checkout
       - attach_workspace:
           at: ~/build
-      - spectest:
-          skip_validation: true
-          expected_passed: 10467
-          expected_failed: 3
-          expected_skipped: 1343
       - spectest:
           expected_passed: 11409
```
chfast, line 399: At least set default values for these so we don't need to update them in 3 places.

PR #492, opened 2020-08-17
FP rounding mode

chfast:


lib/fizzy/execute.cpp
```diff
@@ -555,13 +556,33 @@ inline constexpr T fmax(T a, T b) noexcept
...
-    double value) noexcept
+__attribute__((no_sanitize("float-cast-overflow"))) inline float demote(double value) noexcept
 {
+    // Save current rounding mode.
+    // Negative value denote error. On architectures without rounding mode support glibc always
+    // returns FE_TONEAREST, but there are C standard library implementations that return
+    // error code, e.g. -ENOTSUP.
+    const auto current_rounding_mode = std::fegetround();
+
+    // Fast path if the rounding more is the desired or could not be determined.
```
gumb0, line 571:
```suggestion
    // Fast path if the rounding mode is the desired or could not be determined.
```

lib/fizzy/execute.cpp
```diff
@@ -555,13 +556,33 @@ inline constexpr T fmax(T a, T b) noexcept
...
+    // The returned status code is ignored because it means the input value is invalid (unlikely)
+    // or the target architecture does not support changing rounding mode.
+    std::fesetround(FE_TONEAREST);
+
     // The float-cast-overflow UBSan check disabled for this conversion. In older clang versions
     // (up to 8.0) it reports a failure when non-infinity f64 value is converted to f32 infinity.
     // Such behavior is expected.
-    return static_cast<float>(value);
+    // The result is volatile to enforce it being computed before the ending std::fesetround().
+    const volatile auto result = static_cast<float>(value);
```
gumb0, line 588: Does this `volatile` trick always guarantee the right order?

test/unittests/execute_floating_point_test.cpp
```diff
@@ -1072,7 +1073,16 @@ TEST(execute_floating_point, f32_demote_f64)
 
     for (const auto& [arg, expected] : test_cases)
     {
-        EXPECT_THAT(execute(*instance, 0, {arg}), Result(expected)) << arg << " -> " << expected;
+#pragma STDC FENV_ACCESS ON
```
gumb0, line 1076: Why don't we have this in the implementation?

lib/fizzy/execute.cpp
chfast: Hard to tell.

test/unittests/execute_floating_point_test.cpp
chfast: Because it seems it changes absolutely nothing.

lib/fizzy/execute.cpp
axic: Couldn't you instead add a branch to trigger different layout?

```
if (current_rounding_mode  != status_code || status_code == -1) std::fesetround(current_rounding_mode);
```

axic: Getting to a point inline assembly would be easier/nicer here for the entire function lol.

chfast: > Couldn't you instead add a branch to trigger different layout?
> 
> ```
> if (current_rounding_mode  != status_code || status_code == -1) std::fesetround(current_rounding_mode);
> ```

This is done some comments earlier.

test/unittests/execute_floating_point_test.cpp
chfast: Neither GCC nor Clang support this pragma.

PR #493, opened 2020-08-17
Add value type to ExternalGlobal and check it at instantiate

gumb0:
This fixes the disabled instantiation test.

Not covered by spec tests.

### DIFF
--- a/test/unittests/instantiate_test.cpp
+++ b/test/unittests/instantiate_test.cpp
@@ -378,7 +378,7 @@ TEST(instantiate, imported_globals_mismatched_mutability)
         "global 0 mutability doesn't match module's global mutability");
 }
 
-TEST(instantiate, DISABLED_imported_globals_mismatched_type)
+TEST(instantiate, imported_globals_mismatched_type)
 {
     /* wat2wasm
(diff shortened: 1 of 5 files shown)


test/unittests/instantiate_test.cpp
```diff
@@ -392,8 +410,8 @@ TEST(instantiate, DISABLED_imported_globals_mismatched_type)
     const auto bin2 = from_hex("0061736d01000000020b01036d6f64026731037f00");
     const auto module2 = parse(bin2);
 
-    EXPECT_THROW_MESSAGE(
-        instantiate(module2, {}, {}, {}, {*g}), instantiate_error, "type mismatch");
+    EXPECT_THROW_MESSAGE(instantiate(module2, {}, {}, {}, {*g}), instantiate_error,
+        "global 0 value type doesn't match module's global type");
```
chfast, line 414: Modify or and a test where non-0 global is wrong.

gumb0: Added a case to `imported_globals_mismatched_type` test

chfast (approved):

axic (approved):

PR #494, opened 2020-08-18
ci: Add native builds

chfast:
On architectures with SSE4 and AVX, the floating point operations use different set of CPU instructions. Enabling native builds should tests these.

### DIFF
--- a/test/unittests/validation_test.cpp
+++ b/test/unittests/validation_test.cpp
@@ -406,7 +406,7 @@ TEST(validation, load_alignment)
         {Instr::f64_load, 3},
     };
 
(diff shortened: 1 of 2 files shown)


test/unittests/validation_test.cpp
```diff
@@ -406,7 +406,7 @@ TEST(validation, load_alignment)
         {Instr::f64_load, 3},
     };
 
-    for (const auto test_case : test_cases)
+    for (const auto& test_case : test_cases)
```
chfast, line 409: Found by XCode 12, but I removed usage of this compiler because of issues with container overflow detection.

axic (approved):

PR #495, opened 2020-08-18
Refactor Stack utility class

chfast:
I started this to investigate "containter-overflow" issues reported by ASan in XCode 12.0 beta. This does not fix the issue, but at least I minimized the Stack interface and we don't use inheritance any more.

### DIFF
--- a/lib/fizzy/stack.hpp
+++ b/lib/fizzy/stack.hpp
@@ -13,47 +13,39 @@
 namespace fizzy
 {
 template <typename T>
-class Stack : public std::vector<T>
+class Stack
 {
+    std::vector<T> m_container;
+
 public:
-    using difference_type = typename std::vector<T>::difference_type;
-
-    using std::vector<T>::vector;
-
-    using std::vector<T>::back;
-    using std::vector<T>::emplace_back;
-    using std::vector<T>::pop_back;
-    using std::vector<T>::resize;
-    using std::vector<T>::size;
-
-    // Also used: size(), resize(), clear(), empty(), end()
-
-    void push(T val) { emplace_back(val); }
+    void push(T val) { m_container.emplace_back(val); }
 
     template <typename... Args>
     void emplace(Args&&... args)
     {
-        std::vector<T>::emplace_back(std::forward<Args>(args)...);
+        m_container.emplace_back(std::forward<Args>(args)...);
     }
 
-    T pop()
+    T pop() noexcept
     {
-        const auto res = back();
-        pop_back();
+        assert(!m_container.empty());
+        const auto res = m_container.back();
+        m_container.pop_back();
         return res;
     }
 
-    T& operator[](size_t index) noexcept { return std::vector<T>::operator[](size() - index - 1); }
+    bool empty() const noexcept { return m_container.empty(); }
 
-    T& top() noexcept { return (*this)[0]; }
+    size_t size() const noexcept { return m_container.size(); }
 
-    /// Drops @a num_elements elements from the top of the stack.
-    void drop(size_t num_elements = 1) noexcept { resize(size() - num_elements); }
+    T& operator[](size_t index) noexcept { return m_container[size() - index - 1]; }
+
+    T& top() noexcept { return m_container.back(); }
 
     void shrink(size_t new_size) noexcept
     {
         assert(new_size <= size());
-        resize(new_size);
+        m_container.resize(new_size);
     }
 };
 
--- a/test/unittests/stack_test.cpp
+++ b/test/unittests/stack_test.cpp
@@ -19,6 +19,7 @@ TEST(stack, push_and_pop)
     stack.push('b');
     stack.push('c');
 
+    EXPECT_FALSE(stack.empty());
     EXPECT_EQ(stack.size(), 3);
 
     EXPECT_EQ(stack.pop(), 'c');
@@ -40,6 +41,7 @@ TEST(stack, emplace)
     stack.emplace('b');
     stack.emplace('c');
 
+    EXPECT_FALSE(stack.empty());
     EXPECT_EQ(stack.size(), 3);
 
     EXPECT_EQ(stack.pop(), 'c');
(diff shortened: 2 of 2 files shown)


axic:
I'd prefer this PR, unless there is some significant parser slowdown.

test/unittests/stack_test.cpp
```diff
@@ -80,94 +80,6 @@ TEST(stack, drop_and_peek)
...
-    stack.drop(0);
-    EXPECT_FALSE(stack.empty());
-    EXPECT_EQ(stack.size(), 2);
-
-    stack.clear();
-    EXPECT_TRUE(stack.empty());
-    EXPECT_EQ(stack.size(), 0);
-}
-
-TEST(stack, resize)
```
axic, line 101: There's no test for `shrink` besides this.

axic:
Could rename the `drop_and_peek` test to `drop`. And maybe group the `struct_item` test with the others.

test/unittests/stack_test.cpp
chfast: https://github.com/wasmx/fizzy/pull/495/files#diff-931a4a112d70874574895c172f9bfeefL245

lib/fizzy/stack.hpp
```diff
@@ -13,48 +13,42 @@
...
     {
-        const auto res = back();
-        pop_back();
+        const auto res = m_container.back();
+        m_container.pop_back();
         return res;
     }
 
-    T& operator[](size_t index) noexcept { return std::vector<T>::operator[](size() - index - 1); }
+    bool empty() const { return m_container.empty(); }
```
axic, line 36: `noexcept`.

lib/fizzy/stack.hpp
```diff
@@ -13,48 +13,42 @@
...
 
-    T& top() noexcept { return (*this)[0]; }
+    size_t size() const { return m_container.size(); }
```
axic, line 38: `noexcept`.

lib/fizzy/stack.hpp
```diff
@@ -13,48 +13,42 @@
...
 
     void shrink(size_t new_size) noexcept
     {
         assert(new_size <= size());
-        resize(new_size);
+        m_container.resize(new_size);
     }
+
+    /// Drops @a num_elements elements from the top of the stack.
+    void drop(size_t num_elements = 1) noexcept { shrink(size() - num_elements); }
```
axic, line 51: Hmm, shouldn't this need `assert(size() >= num_elements)`, since `size_t` is unsigned, the value to `shrink` could wrap around?

axic: `OperandStack::drop` has `assert(num_elements <= size())` so maybe follow that style.

axic:
The rest looks good to me.

lib/fizzy/stack.hpp
chfast: I removed drop().

gumb0 (approved):

chfast:
Not really a change performance-wise
```
fizzy/parse/blake2b_mean                                -0.0114         -0.0114            23            23            23            23
fizzy/parse/ecpairing_mean                              -0.0003         -0.0003          1338          1337          1338          1337
fizzy/parse/keccak256_mean                              -0.0095         -0.0094            42            42            42            42
fizzy/parse/memset_mean                                 -0.0034         -0.0034             6             6             6             6
fizzy/parse/mul256_opt0_mean                            -0.0269         -0.0269             8             8             8             8
fizzy/parse/ramanujan_pi_mean                           +0.0160         +0.0160            24            24            24            24
fizzy/parse/sha1_mean                                   -0.0008         -0.0008            38            38            38            38
fizzy/parse/sha256_mean                                 +0.0031         +0.0031            63            63            63            63
fizzy/parse/taylor_pi_mean                              -0.0262         -0.0262             3             3             3             3
fizzy/parse/micro/eli_interpreter_mean                  -0.0167         -0.0167             4             4             4             4
fizzy/parse/micro/factorial_mean                        -0.0392         -0.0392             1             1             1             1
fizzy/parse/micro/fibonacci_mean                        -0.0473         -0.0473             1             1             1             1
fizzy/parse/micro/host_adler32_mean                     -0.0250         -0.0250             2             2             2             2
fizzy/parse/micro/spinner_mean                          -0.0160         -0.0160             1             1             1             1
```

lib/fizzy/stack.hpp
```diff
@@ -13,47 +13,39 @@
...
 
     template <typename... Args>
     void emplace(Args&&... args)
     {
-        std::vector<T>::emplace_back(std::forward<Args>(args)...);
+        m_container.emplace_back(std::forward<Args>(args)...);
     }
 
-    T pop()
+    T pop() noexcept
```
axic, line 29: Is this marked noexcept because we manually ensure the caller will call it in a bad state?

chfast: Assert is not relevant. This is `noexcept` because `.back()` and `.pop_back()` are `noexcept` too.

axic: I checked it when I commented on empty/size, but to me back/pop_back did not seem to be noexcept:
- http://cplusplus.com/reference/vector/vector/back/
- http://cplusplus.com/reference/vector/vector/pop_back/

axic: Ah okay noexcept since C++20:
- https://en.cppreference.com/w/cpp/container/vector/back
- https://en.cppreference.com/w/cpp/container/vector/pop_back

axic: Were these behaving as noexcept pre-C++20 just not marked so?

chfast: Officially they are not `noexcept` (even in C++20), but in practice they are.

axic (approved):
Looks okay to me, but would prefer a second look from @gumb0 on the noexcept question.

lib/fizzy/stack.hpp
gumb0: This looks good to me.

PR #497, opened 2020-08-19
Update the readme with floating point

axic:


README.md
```diff
@@ -66,6 +68,34 @@ Follow [this guide](./test/spectests/README.md) for using the tool.
...
+In glibc this can be controlled via the [`feenableexcept` and `fedisableexcept` functions](https://www.gnu.org/savannah-checkouts/gnu/libc/manual/html_node/Control-Functions.html).
+
+The caller has to ensure this is not set to *trap* mode. We strongly suggest not to enable it and catch the signal,
+because in that case correct behaviour can not be guaranteed.
+
+### Rounding
+
+The IEEE 754 standard also defines four different rounding directions (or modes):
+- to nearest, tie-to-even,
+- downward (towards -inf),
```
gumb0, line 90: maybe better
```suggestion
- downward (towards -infinity),
```

README.md
```diff
@@ -66,6 +68,34 @@ Follow [this guide](./test/spectests/README.md) for using the tool.
...
+
+Certain floating point operations can emit *exceptions* as defined by the IEEE 754 standard.
+(Here is a [summary from glibc](https://www.gnu.org/software/libc/manual/html_node/FP-Exceptions.html)).
+It is however up to the language how these manifest, and in C/C++ depending on the settings they can result in
+a) setting of a flag; b) in *traps*, such as the `SIGFPE` signal.
+
+Fizzy does not manipulate this setting, but expects it to be left at option a) of above, which is the default.
+In glibc this can be controlled via the [`feenableexcept` and `fedisableexcept` functions](https://www.gnu.org/savannah-checkouts/gnu/libc/manual/html_node/Control-Functions.html).
+
+The caller has to ensure this is not set to *trap* mode. We strongly suggest not to enable it and catch the signal,
```
gumb0, line 83: Nitpick: maybe it's only me, but maybe it could be unclear whether we suggest to catch the signal or not to catch
```suggestion
The caller has to ensure this is not set to *trap* mode. We strongly suggest to avoid enabling it and catching the signal,
```

gumb0 (approved):

README.md
```diff
@@ -66,6 +68,34 @@ Follow [this guide](./test/spectests/README.md) for using the tool.
 
 This is the unit tests suite of Fizzy.
 
+## Special notes about floating point
+
+### Exceptions
+
+Certain floating point operations can emit *exceptions* as defined by the IEEE 754 standard.
+(Here is a [summary from glibc](https://www.gnu.org/software/libc/manual/html_node/FP-Exceptions.html)).
```
chfast, line 76:
```suggestion
(Here is a [summary from GNU C Library](https://www.gnu.org/software/libc/manual/html_node/FP-Exceptions.html)).
```

README.md
```diff
@@ -66,6 +68,34 @@ Follow [this guide](./test/spectests/README.md) for using the tool.
...
+It is however up to the language how these manifest, and in C/C++ depending on the settings they can result in
+a) setting of a flag; b) in *traps*, such as the `SIGFPE` signal.
+
+Fizzy does not manipulate this setting, but expects it to be left at option a) of above, which is the default.
+In glibc this can be controlled via the [`feenableexcept` and `fedisableexcept` functions](https://www.gnu.org/savannah-checkouts/gnu/libc/manual/html_node/Control-Functions.html).
```
chfast, line 81:
```suggestion
In GNU C Library this can be controlled via the [`feenableexcept` and `fedisableexcept` functions](https://www.gnu.org/savannah-checkouts/gnu/libc/manual/html_node/Control-Functions.html).
```

README.md
chfast, line 83:
```suggestion
The user of the Fizzy library has to ensure this is not set to *trap* mode. We strongly suggest not to enable it and catch the signal,
```

chfast: For me the second sentence is also confusing. Maybe just "The behavior with traps enabled is unpredictable/not guaranteed/...".

README.md
```diff
@@ -66,6 +68,34 @@ Follow [this guide](./test/spectests/README.md) for using the tool.
...
+
+The caller has to ensure this is not set to *trap* mode. We strongly suggest not to enable it and catch the signal,
+because in that case correct behaviour can not be guaranteed.
+
+### Rounding
+
+The IEEE 754 standard also defines four different rounding directions (or modes):
```
chfast, line 88:
```suggestion
The IEEE 754 standard defines four rounding directions (or modes):
```

README.md
```diff
@@ -66,6 +68,34 @@ Follow [this guide](./test/spectests/README.md) for using the tool.
...
+- to nearest, tie-to-even,
```
chfast, line 89:
```suggestion
- to nearest, tie to even (default)
```

README.md
chfast, line 90:
```suggestion
- toward -∞ (rounding down, floor)
```

README.md
```diff
@@ -66,6 +68,34 @@ Follow [this guide](./test/spectests/README.md) for using the tool.
...
+- downward (towards -inf),
+- upward (towards inf),
```
chfast, line 91:
```suggestion
- toward +∞ (rounding up, ceiling)
```

README.md
```diff
@@ -66,6 +68,34 @@ Follow [this guide](./test/spectests/README.md) for using the tool.
...
+- towards zero (truncate).
```
chfast, line 92:
```suggestion
- toward  0 (truncation)
```

axic:
@axic @gumb0 please also review the first commit which changes the goals section.

README.md
```diff
@@ -66,6 +68,34 @@ Follow [this guide](./test/spectests/README.md) for using the tool.
...
+In the GNU C Library this can be controlled via the [`feenableexcept` and `fedisableexcept` functions](https://www.gnu.org/savannah-checkouts/gnu/libc/manual/html_node/Control-Functions.html).
+
+The user of the Fizzy library has to ensure this is not set to *trap* mode.
+The behavior with traps enabled is unpredictable and correct behaviour is not guaranteed, thus we strongly advise against using them.
+
+### Rounding
+
+The IEEE 754 standard defines four rounding directions (or modes):
+- to nearest, tie to even (default),
+- toward -∞ (rounding down, floor),
```
axic, line 90:
```suggestion
- toward -∞ (rounding down aka floor),
- toward +∞ (rounding up aka ceiling),
```

@chfast how about this?

chfast: Both "rounding down" and "floor" are aka. Just drop "floor" and "ceiling".

README.md
```diff
@@ -66,6 +68,34 @@ Follow [this guide](./test/spectests/README.md) for using the tool.
...
+- toward +∞ (rounding up, ceiling),
+- toward  0 (truncation).
+
+WebAssembly requires "to nearest" rounding, which is the first option, and is the default.
```
chfast, line 94:
```suggestion
The semantic of the WebAssembly floating-point instructions is based on the default "no nearest" rounding direction. Therefore, Fizzy expects this rounding direction to be selected. Otherwise, the result of some instructions may be different than what the WebAssembly specification predicts, i.e. the behavior will follow the selected rounding direction.
```

The list of these instructions will be provided later when we have tests in place.

The `floor` instruction should be not on the list, that's why we are having the workaround for the GCC bug already in place.

README.md
```diff
@@ -66,6 +68,34 @@ Follow [this guide](./test/spectests/README.md) for using the tool.
...
+In the GNU C Library the rounding mode can be controlled via the [`fesetround` and `fegetround` functions](https://www.gnu.org/software/libc/manual/html_node/Rounding.html).
+
+Fizzy does not manipulate this setting, but expects it to be left at the default, otherwise correct behaviour can not be guaranteed.
```
chfast, line 97:
```suggestion

```
Should be drop if the above change is accepted.

README.md
gumb0: To me this `floor` hack is an inconsistency and I think it should be mentioned, if we want to keep it.

gumb0:
> @axic @gumb0 please also review the first commit which changes the goals section.

that looks good to me

README.md
axic: Well the current one was your suggestion: https://github.com/wasmx/fizzy/pull/497#discussion_r473072120 :wink:

README.md
gumb0: Now I see how the hack is consistent with keeping the instructions affected by rounding mode limited to a logical set. Ok to add a list of affected instructions later.

README.md
chfast: I took it from wikipedia, but it is more readable without the last alias.

README.md
```diff
@@ -17,12 +17,14 @@ I) Code quality
...
 
 II) Simplicity
+- [x] Only implement WebAssembly 1.0 (and none of the proposals)
 - [x] Interpreter only
-- [x] Provide no support for floating point operations (initially)
 - [x] Support only WebAssembly binary encoding as an input (no support for the text format (`.wat`/`.wast`))
 
 III) Conformance
-- [ ] Should pass the official WebAssembly test suite
+- [x] Should have 99+% unit test coverage
```
chfast, line 25: I'd say we aim for 100% although it may be difficult to get precisely that because of the toolchain bugs and limitations.

axic: That's why I put 99%, but I'm fine either way.

gumb0: "99+%" looks good to me

PR #499, opened 2020-08-20
Add static_asserts for template types where possible

axic:


lib/fizzy/execute.cpp
```diff
@@ -367,6 +367,8 @@ inline void convert(OperandStack& stack) noexcept
 template <typename SrcT, typename DstT>
 inline void reinterpret(OperandStack& stack) noexcept
 {
+    static_assert(std::is_integral_v<SrcT> == std::is_floating_point_v<DstT> ||
```
axic, line 370: Not sure this one is useful, but it is not apparent based on comments that this is only used to convert between floating point <-> integral.

chfast (approved):
I can live with these.

lib/fizzy/execute.cpp
axic, line 370:
```suggestion
    static_assert((std::is_integral_v<SrcT> && std::is_floating_point_v<DstT>) ||
                      (std::is_floating_point_v<SrcT> && std::is_integral_v<DstT>));
```

PR #507, opened 2020-08-24
Support fp in wasm_engine/fizzy_engine

axic:


axic:
I see little importance in implementing this, given it is only used by benchmarks, where we worked around the floating point in/out question.

PR #508, opened 2020-08-24
cmake: enable more strict compiler warnings

axic:


chfast:
Add `-Wmissing-declarations` (may be GCC only)

lib/fizzy/execute.cpp
```diff
@@ -1963,7 +1963,7 @@ ExecutionResult execute(
         }
         case Instr::f64_promote_f32:
         {
-            stack.top() = double{stack.top().f32};
+            stack.top() = static_cast<double>(stack.top().f32);
```
axic, line 1966: @chfast do you want this to go into master?

cmake/CableCompilerSettings.cmake
```diff
@@ -87,6 +87,12 @@ macro(cable_configure_compiler)
             # Enable basing warnings set and treat them as errors.
             add_compile_options(-Werror -Wall -Wextra -Wshadow)
 
+            add_compile_options(-Wdouble-promotion -Wcast-qual -Wcast-align -Wdangling-else -Wlogical-not-parentheses)
```
axic, line 90:
```suggestion
            add_compile_options(-Wdouble-promotion -Wcast-qual -Wcast-align -Wdangling-else -Wlogical-not-parentheses -Wmissing-declarations)
```

lib/fizzy/execute.cpp
chfast: No.

cmake/CableCompilerSettings.cmake
```diff
@@ -87,6 +87,12 @@ macro(cable_configure_compiler)
             # Enable basing warnings set and treat them as errors.
             add_compile_options(-Werror -Wall -Wextra -Wshadow)
 
+            add_compile_options(-Wdouble-promotion -Wcast-qual -Wcast-align -Wdangling-else -Wlogical-not-parentheses -Wmissing-declarations)
```
chfast, line 90: I think this must be checked against documentation.
E.g. `-Wlogical-not-parentheses`
- is enabled in Clang by default: https://clang.llvm.org/docs/DiagnosticsReference.html#wlogical-not-parentheses
- is enabled in GCC with `-Wall`: https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wall

chfast (changes requested):

cmake/CableCompilerSettings.cmake
axic: I added these based on documentation: first added a lot of the things I liked, then removed those which are included in `-Wall` (and other settings we have). Seems like this one slipped through the cracks.

chfast: I picked this one randomly.

The `-Wdouble-promotion` should be remove as I don't want the change https://github.com/wasmx/fizzy/pull/508#discussion_r478300508.

I will check the rest.

chfast: - It seems `-Wcast-qual` is only for C-style casts of pointers. Unlikely we will see such thing in our code, but the warning option can stay.

axic: > The -Wdouble-promotion should be remove as I don't want the change #508 (comment).

Do you have a reasoning? I'd be also interested to hear @gumb0's view.

axic: @chfast @gumb0 any other warnings we should/could enable?

chfast: > > The -Wdouble-promotion should be remove as I don't want the change #508 (comment).
> 
> Do you have a reasoning? I'd be also interested to hear @gumb0's view.

The `double{x}` is superior to `static_cast<double>(x)` and the `-Wdouble-promotion` forbids using the first syntax. It is rather a compiler bug due to fact that `-Wdouble-promotion` predates the `double{x}` syntax.

There is even Clang bug for this: https://bugs.llvm.org/show_bug.cgi?id=34062.

chfast: > @chfast @gumb0 any other warnings we should/could enable?

Yes, try
```
-Weverything -Wno-c++98-compat -Wno-c++98-compat-pedantic -Wno-padded -Wno-switch-enum -Wno-float-equal -Wno-double-promotion
```
I found many easy fixes with these already.

gumb0: `-Wlogical-not-parentheses` looks to be included in `-Wall`?

test/utils/wasm_engine.cpp
```diff
@@ -2,6 +2,8 @@
 // Copyright 2020 The Fizzy Authors.
 // SPDX-License-Identifier: Apache-2.0
 
+#include <test/utils/wasm_engine.hpp>
```
gumb0, line 5:
```suggestion
#include "wasm_engine.hpp"
```

test/utils/wasm3_engine.cpp
```diff
@@ -41,8 +41,8 @@ namespace
 const void* env_adler32(IM3Runtime /*runtime*/, uint64_t* stack, void* mem)
 {
     uint64_t* ret = stack;
-    const uint32_t offset = *(uint32_t*)(stack++);
-    const uint32_t length = *(uint32_t*)(stack++);
+    const uint32_t offset = static_cast<uint32_t>(*stack++);
+    const uint32_t length = static_cast<uint32_t>(*stack);
```
axic, line 45: I used `stack[0]` and `stack[1]` for readability here.

axic: And to that effect we could use `stack[0] = fizzy::tests:adler32`, couldn't we?

test/utils/wasm_engine.cpp
axic: Used this style because it is used in other files in this directory.

test/utils/wasm3_engine.cpp
chfast: Done, thanks.

cmake/CableCompilerSettings.cmake
chfast: Move the changes to the main CMakeLists.txt, do not modify the Cable file.

test/utils/wasm_engine.cpp
chfast: It should be always "" if including the header matching the .cpp file (clang-format will place it as the first one). Unless the file is not in the same dir.

cmake/CableCompilerSettings.cmake
chfast: This is cleaned up now.

lib/fizzy/execute.cpp
chfast: Do we have consensus here?

chfast (approved):

CMakeLists.txt
```diff
@@ -35,6 +35,34 @@ if(is_big_endian)
...
+cable_add_cxx_compiler_flag_if_supported(-Wnewline-eof)
+cable_add_cxx_compiler_flag_if_supported(-Wsuggest-destructor-override)
+cable_add_cxx_compiler_flag_if_supported(-Wunreachable-code-break)
+cable_add_cxx_compiler_flag_if_supported(-Wduplicated-cond)
+cable_add_cxx_compiler_flag_if_supported(-Wduplicate-enum)
+cable_add_cxx_compiler_flag_if_supported(-Wlogical-op)
+
+option(WEVERYTHING "Enable almost all compiler warnings" OFF)
+if(WEVERYTHING)
+    add_compile_options(-Weverything)
```
gumb0, line 55: Why do I not see these on CI?
`Build Tidy (Debug)` step outputs
```
----
Generator:  Ninja
Compiler:   Clang
Build Type: Debug
LTO:        OFF
Flags:          -Dd_m3HasTracer -Wall -Wextra -Wparentheses -Wundef -Wpointer-arith -Wstrict-aliasing=2 -Werror=implicit-function-declaration -Wno-unused-function -Wno-unused-variable -Wno-unused-parameter -Wno-missing-field-initializers -Werror=shadow
Debug flags:   -g -DDEBUG=1 -ggdb -O0
Release flags: -O3 -Wfatal-errors -fomit-frame-pointer -fno-stack-check -fno-stack-protector
----
```

are these not all of the flags?

chfast: These are wasm3 flags. We don't add ours to theirs.

test/bench_internal/parser_benchmarks.hpp
```diff
@@ -0,0 +1,15 @@
+// Fizzy: A fast WebAssembly interpreter
```
gumb0, line 1: Maybe this file should be `experimental.hpp`, because two of its three functions are defined in `experimental.cpp`?

Also not sure if the third one - `leb128u_decode_u64_noinline` - needs to be in its own cpp.

test/unittests/execute_floating_point_test.cpp
```diff
@@ -137,8 +137,8 @@ class TestValues
             std::array<T, ps.size() * 2 + 1> a;
 
             auto it = std::begin(a);
-            it = std::transform(std::reverse_iterator{std::end(ps)},
-                std::reverse_iterator{std::begin(ps)}, it, std::negate<T>{});
+            it = std::transform(std::make_reverse_iterator(std::end(ps)),
```
gumb0, line 140: Does this resolve some warning?

CMakeLists.txt
gumb0: I don't see any warning emitted in `Build Tidy`, do we not generate any currently? If we did, would CI fail?

test/unittests/execute_floating_point_test.cpp
chfast: There is warning that `reverse_iterator` is may not to designed for CTAD (as it does not have any explicit CTAD hints). I think it is ok to change it here as this is the only case for this warning. And I'm also considering rewriting these test arrays to constexpr version where STL algorithms are going to be less useful.

test/bench_internal/parser_benchmarks.hpp
chfast: I would go with something less specific than `parser_benchmarks.cpp`: `experimental.cpp` is fine.

This is not explicitly needed now as I discovered `[[noinline]]` attribute...

CMakeLists.txt
chfast: All warnings are fixed.
But here is older build with the warning for https://github.com/wasmx/fizzy/pull/508#discussion_r484497148: https://app.circleci.com/pipelines/github/wasmx/fizzy/4478/workflows/b12515da-c249-4903-aa9f-8d3d1d5e7cbb/jobs/31809

gumb0: I mean, if it fails CI, what's the point of making them optional?

chfast: People can still build it locally without a struggle.

gumb0: Do you mean they can build it locally without -Weverything for compilers that we don't test?

chfast: Yes. Clang discourages usage of `-Weverything` regularly for projects so I want to follow this rule for now. While Fizzy is actively maintained it is probably not big deal, but future versions of Clang may enable something we don't want.

gumb0 (approved):

test/bench_internal/parser_benchmarks.hpp
chfast: This is resolved now to use `experimental.hpp` for alternative LEB128 implementation. The other functions were moved to `parser_benchmarks.cpp` with `[[gnu::noinline]]` attribute.

PR #511, opened 2020-08-25
test: Add fizzy-testfloat tool

chfast:
This expects input generated by the `testfloat_gen` tool. See section 5 in http://www.jhauser.us/arithmetic/TestFloat-3/doc/TestFloat-general.html.

The `testfloat_gen` usage:
```
testfloat_gen [<option>...] <type>|<function>
  <option>:  (* is default)
    -help            --Write this message and exit.
    -prefix <text>   --Write <text> as a line of output before any test cases.
    -seed <num>      --Set pseudo-random number generator seed to <num>.
 *  -seed 1
    -level <num>     --Testing level <num> (1 or 2).
 *  -level 1
    -n <num>         --Generate <num> test cases.
    -forever         --Generate test cases indefinitely (implies '-level 2').
 *  -rnear_even      --Round to nearest/even.
    -rminMag         --Round to minimum magnitude (toward zero).
    -rmin            --Round to minimum (down).
    -rmax            --Round to maximum (up).
    -rnear_maxMag    --Round to nearest/maximum magnitude (nearest/away).
    -tininessbefore  --Detect underflow tininess before rounding.
 *  -tininessafter   --Detect underflow tininess after rounding.
 *  -notexact        --Rounding to integer is not exact (no inexact
                         exceptions).
    -exact           --Rounding to integer is exact (raising inexact
                         exceptions).
  <type>:
    <int>            --Generate test cases with one integer operand.
    <float>          --Generate test cases with one floating-point operand.
    <float> <num>    --Generate test cases with <num> (1, 2, or 3)
                         floating-point operands.
  <function>:
    <int>_to_<float>     <float>_add      <float>_eq
    <float>_to_<int>     <float>_sub      <float>_le
    <float>_to_<float>   <float>_mul      <float>_lt
    <float>_roundToInt   <float>_mulAdd   <float>_eq_signaling
                         <float>_div      <float>_le_quiet
                         <float>_rem      <float>_lt_quiet
                         <float>_sqrt
  <int>:
    ui32             --Unsigned 32-bit integer.
    ui64             --Unsigned 64-bit integer.
    i32              --Signed 32-bit integer.
    i64              --Signed 64-bit integer.
  <float>:
    f32              --Binary 32-bit floating-point (single-precision).
    f64              --Binary 64-bit floating-point (double-precision).
```


circle.yml
```diff
@@ -23,13 +23,48 @@ executors:
       xcode: 11.6.0
 
 commands:
-  description: "Install macOS system dependencies"
   install_macos_deps:
+    description: "Install macOS system dependencies"
```
axic, line 27: What is the difference?

circle.yml
```diff
@@ -278,7 +319,14 @@ jobs:
           command: |
             # Convert to relative paths
             sed -i 's|$(pwd)/||' ~/build/fizzy.lcov
-            codecov --required --file ~/build/fizzy.lcov -X gcov
+
+            counter=1
+            until codecov --required --file ~/build/fizzy.lcov -X gcov || [ $counter = 5 ]; do
```
axic, line 324: Seems to be an unrelated fix.

circle.yml
```diff
@@ -420,19 +469,19 @@ workflows:
...
-      - release-native-macos
-      - coverage
-      - cxx20
-      - sanitizers
-      - sanitizers-macos
-      - fuzzing
-      - spectest:
-          requires:
-            - coverage
+        - lint
```
axic, line 472: Assuming the whitespace change will be removed prior to merge?

test/testfloat/testfloat.cpp
```diff
@@ -0,0 +1,395 @@
...
+
+Tool for testing IEEE 754 floating-point compliance.
+Expects inputs from TestFloat's testfloat_gen.
+See http://www.jhauser.us/arithmetic/TestFloat-3/doc/TestFloat-general.html.
+
+Usage:
+  fizzy-testfloat [options] <function>
+
+  -rnear_even     Round to nearest/even [default].
+  -rminMag        Round to minimum magnitude (toward zero).
```
axic, line 30: I think these options are inconsistent, underscore vs camelcase.

I'd suggest: `-rnearest` and `-rmin-mag`.

test/testfloat/testfloat.cpp
```diff
@@ -0,0 +1,395 @@
...
+        fizzy::FuncIdx func_idx, Type result, Type param, Options opts) noexcept
+      : idx{func_idx},
+        result_type{result},
+        param_types{param, Type::f32},
+        num_arguments{1},
+        options{opts}
+    {}
+};
+
+auto get_wasm_code()
```
axic, line 93: Oh, almost like JS at this point.

test/testfloat/testfloat.cpp
```diff
@@ -0,0 +1,395 @@
...
+    (func $f64_mul (param f64 f64) (result f64) (f64.mul (local.get 0) (local.get 1)))
+    (func $f32_div (param f32 f32) (result f32) (f32.div (local.get 0) (local.get 1)))
+    (func $f64_div (param f64 f64) (result f64) (f64.div (local.get 0) (local.get 1)))
+    (func $f32_sqrt (param f32) (result f32) (f32.sqrt (local.get 0)))
+    (func $f64_sqrt (param f64) (result f64) (f64.sqrt (local.get 0)))
+
+    (func $i32_to_f32 (param i32) (result f32) (f32.convert_i32_s (local.get 0)))
+    (func $ui32_to_f32 (param i32) (result f32) (f32.convert_i32_u (local.get 0)))
+    (func $i64_to_f32 (param i64) (result f32) (f32.convert_i64_s (local.get 0)))
+    (func $ui64_to_f32 (param i64) (result f32) (f32.convert_i64_u (local.get 0)))
```
axic, line 110: Hm, why not `u64_` or use `i64_s_to` ?

test/testfloat/testfloat.cpp
```diff
@@ -0,0 +1,395 @@
...
+};
+
+bool check(const FunctionDescription& func, fizzy::Instance& instance, const uint64_t inputs[],
+    bool ignore_nans)
+{
+    const auto report_failure = [&](auto result, auto expected) {
+        std::cerr << "FAILURE: " << result << " <-";
+        for (size_t i = 0; i < func.num_arguments; ++i)
+            std::cerr << ' ' << TypedValue{func.param_types[i], inputs[i]};
+        std::cerr << "\n   exp.: " << expected << "\n";
```
axic, line 288: I'd use `expected:` here.

circle.yml
chfast: It was in wrong place. This is `description` of `install_macos_deps` command.

test/testfloat/testfloat.cpp
chfast: These follow options from `testfloat_gen` so you don't need to remember two names doing the same thing. This usually goes like this
```
testfloat_gen -rminMag f32_add | fizzy-testfloat -rminMag f32_add
```

See https://github.com/wasmx/fizzy/pull/511#issue-473460885.

test/testfloat/testfloat.cpp
chfast: Similarly, the function names follow the function names from `testfloat_gen`, if possible. We could just use Wasm instruction names (each function is one instruction), just you would need to know which testfloat function name maps to which Wasm instruction.

test/testfloat/testfloat.cpp
chfast: The only point of this function is to have wast code just before the `FunctionDescription` map.

test/testfloat/testfloat.cpp
chfast: This looks like that:
```
FAILURE: 00000000 <- 00000000 00000000
   exp.: 00000001
```
Not the best in the world, but does its job without need of computing offsets.

testfloat.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env sh
```
chfast, line 1: This file should be deleted.

test/testfloat/testfloat.cpp
axic: Can you leave a note in this source file?

test/testfloat/testfloat.cpp
axic: Likewise, can you leave a comment explaining this?

test/testfloat/testfloat.cpp
axic: Not that much important, but `exp.` to me looks like exponent, though I have been working too much EVM. How about something like:
```
FAILURE:   0000
 expected: 0000
```

circle.yml
chfast: Moved to #516.

test/testfloat/testfloat.cpp
chfast:
```
FAILURE: 00000000 <- 00000000 00000000
         00000001 (expected)
```
This better?

axic: Looks good.

test/testfloat/testfloat.cpp
```diff
@@ -0,0 +1,423 @@
...
+        __builtin_unreachable();
+    }
+
+    friend std::ostream& operator<<(std::ostream& os, const TypedValue& x)
+    {
+        os << std::hex << std::uppercase << std::setfill('0');
+        switch (x.type_)
+        {
+        case Type::i32:
+            return os << std::setw(8) << x.value_.as<uint32_t>();
```
axic, line 305: I'm not a fan on `std::hex` and `std::setw`, there's always some mess with it in Solidity at least. Could we use a new version of the `hex` helper we have in `test/utils/hex`?

test/testfloat/testfloat.cpp
chfast: Added comment.

test/testfloat/testfloat.cpp
chfast: Done.

test/testfloat/testfloat.cpp
chfast: The `setw` is fine as it only applies to the next number.
The `hex` is also fine because we display all in hex anyway.

axic: I know it is "fine" but somehow there are differences with these on different architectures. Solidity has weird issues with these time to time, between the different platforms.

test/testfloat/testfloat.cpp
```diff
@@ -0,0 +1,423 @@
...
+struct FunctionDescription
+{
+    const fizzy::FuncIdx idx;
+    const Type result_type;
+    const Type param_types[2];
+    const size_t num_arguments;
+    const Options options = Options::None;
+
+    constexpr FunctionDescription(fizzy::FuncIdx func_idx, Type result, Type param) noexcept
+      : idx{func_idx}, result_type{result}, param_types{param, Type::f32}, num_arguments{1}
```
gumb0, line 86: Is it required to init both elements always?
```suggestion
      : idx{func_idx}, result_type{result}, param_types{param}, num_arguments{1}
```

test/testfloat/testfloat.cpp
```diff
@@ -0,0 +1,423 @@
...
+
+        std::fesetround(rounding_direction);
+
+        int num_failures = 0;
+        std::cin >> std::hex;
+        while (num_failures < max_failures)
+        {
+            uint64_t inputs[4]{};
+            for (size_t i = 0; i < func.num_arguments + 2; ++i)
+                std::cin >> inputs[i];
```
gumb0, line 405: Maybe add a comment what's input format?

test/testfloat/testfloat.cpp
```diff
@@ -0,0 +1,423 @@
...
+        {
+            const FP value{value_.f32};
+            const FP expected{static_cast<uint32_t>(expected_bits)};
+            if (ignore_nans && std::isnan(expected.value))
+            {
+                if (expected.is_canonical_nan())
+                    return value.is_canonical_nan();
+                if (expected.is_arithmetic_nan())
+                    return value.is_arithmetic_nan();
+                return false;  // Signaling NaN should never happen.
```
gumb0, line 276: Maybe this should throw "invalid input" rather than test failure

test/testfloat/testfloat.cpp
```diff
@@ -0,0 +1,423 @@
...
+        const bool invalid_operation = (expected_exceptions & InvalidOperation) != 0;
+
+        if (r.trapped == invalid_operation)
+            return true;
+
+        report_failure(r.trapped, invalid_operation);
+        return false;
+    }
+
+    const auto result = TypedValue{func.result_type, r.value};
```
gumb0, line 346: You don't check `r.trapped` in this case.
(also worth to check `r.has_value)`

test/testfloat/testfloat.cpp
```diff
@@ -0,0 +1,423 @@
...
+See http://www.jhauser.us/arithmetic/TestFloat-3/doc/TestFloat-general.html.
+
+Usage:
+  fizzy-testfloat [options] <function>
+
+  -rnear_even     Round to nearest/even [default].
+  -rminMag        Round to minimum magnitude (toward zero).
+  -rmin           Round to minimum (down).
+  -rmax           Round to maximum (up).
+  -ignoreNaNs     Do not check for specific NaN results ("is NaN" is still checked).
```
gumb0, line 41: Maybe it could be `-ignoreNaNPayloads`

chfast: Makes more sense, but `testfloat` has `-checkNaNs`...

test/testfloat/testfloat.cpp
chfast: Because `constexpr` requires all being initialized... I will recheck if you suggestion inits with zero.

test/testfloat/testfloat.cpp
chfast: Good point.

test/testfloat/testfloat.cpp
```diff
@@ -0,0 +1,423 @@
...
+        case Type::i32:
+            return os << std::setw(8) << x.value_.as<uint32_t>();
+        case Type::i64:
+            return os << std::setw(16) << x.value_.i64;
+        case Type::f32:
+            return os << std::setw(8) << FP{x.value_.f32}.as_uint();
+        case Type::f64:
+            return os << std::setw(16) << FP{x.value_.f64}.as_uint();
+        }
+        __builtin_unreachable();
```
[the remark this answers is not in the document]
chfast: Ok, fine.

test/testfloat/testfloat.cpp
chfast: Good point.

test/testfloat/testfloat.cpp
```diff
@@ -0,0 +1,433 @@
...
+    for (size_t i = 0; i < func.num_arguments; ++i)
+        args[i] = make_value(func.param_types[i], inputs[i]);
+
+    const auto r = fizzy::execute(
+        instance, func.idx, fizzy::span<const fizzy::Value>(args, func.num_arguments));
+
+    if (func.options == Options::TrapIsInvalidOperation)
+    {
+        const auto expected_exceptions = inputs[func.num_arguments + 1];
+        const bool invalid_operation = (expected_exceptions & InvalidOperation) != 0;
```
gumb0, line 337: Should we check that no other exceptions are expected? (if they are, `throw std::invalid_argument`)

test/testfloat/testfloat.cpp
chfast: I added an assert because this is not testable.

test/testfloat/testfloat.cpp
chfast: Done.

test/testfloat/testfloat.cpp
chfast: Done.

test/testfloat/testfloat.cpp
chfast: Changed.

test/testfloat/testfloat.cpp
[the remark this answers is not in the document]
chfast: Done.

test/testfloat/testfloat.cpp
chfast: I'm restoring the ostream state at exit, but still `std::hex` and others are used (no arguments to assume they don't work).

test/testfloat/testfloat.cpp
```diff
@@ -0,0 +1,441 @@
...
+            }
+            return value == expected;
+        }
+        }
+        __builtin_unreachable();
+    }
+
+    friend std::ostream& operator<<(std::ostream& os, const TypedValue& x)
+    {
+        std::ios os_state{nullptr};
```
gumb0, line 301: Here we you did that using `flags()`, I think it would be enough here, too
https://github.com/wasmx/fizzy/blob/5e6b32c30e8a754ad1479feb66043b2daa99f1fd/test/utils/asserts.cpp#L15-L20

test/testfloat/testfloat.cpp
```diff
@@ -0,0 +1,441 @@
...
+    for (size_t i = 0; i < func.num_arguments; ++i)
+        args[i] = make_value(func.param_types[i], inputs[i]);
+
+    const auto r = fizzy::execute(
+        instance, func.idx, fizzy::span<const fizzy::Value>(args, func.num_arguments));
+
+    if (func.options == Options::TrapIsInvalidOperation)
+    {
+        const auto expected_exceptions = inputs[func.num_arguments + 1];
+        const bool invalid_operation = expected_exceptions == InvalidOperation;
```
gumb0, line 344: still could add `assert(expected_exceptions == InvalidOperation || expected_exceptions == 0)`

test/testfloat/testfloat.cpp
chfast: The "flags" will not restore the `setfill`...

gumb0 (approved):

axic:
Should this add a `testfloat/README.md` and link to it from the main readme just like spectests/bench is doing?

chfast:
> Should this add a `testfloat/README.md` and link to it from the main readme just like spectests/bench is doing?

Added README.

chfast:
Please check these remaining TODOs: https://github.com/wasmx/fizzy/pull/511/files#diff-5f073e51994ab67385e8c72e34141c60R24-R27

test/testfloat/README.md
```diff
@@ -0,0 +1,28 @@
+# Fizzy TestFloat
+
+The `fizzy-testfloat` is a CLI tool that cooperates with the [Berkeley TestFloat] floating-point
+implementation test suite and tools. 
+
+The TestFloat's `testfloat_gen` which generates operand values and expected result for selected
```
gumb0, line 6:
```suggestion
The TestFloat's `testfloat_gen` generates operand values and expected result for selected
```

test/testfloat/README.md
```diff
@@ -0,0 +1,28 @@
+# Fizzy TestFloat
+
+The `fizzy-testfloat` is a CLI tool that cooperates with the [Berkeley TestFloat] floating-point
+implementation test suite and tools. 
+
+The TestFloat's `testfloat_gen` generates operand values and expected result for selected
+floating-point instruction. Then this generated input is processed by `fizzy-testfloat` to check if
+the Fizzy's implementation of the matching WebAssembly instruction produces the same result.
```
axic, line 8:
```suggestion
Fizzy's implementation of the matching WebAssembly instruction produces the same result.
```

circle.yml
```diff
@@ -41,6 +41,41 @@ commands:
...
+      - restore_cache:
+          name: "Get TestFloat from cache"
+          key: &testfloat-cache-key TestFloat-3e-{{arch}}-v2
+      - run:
+          name: "Install TestFloat"
+          working_directory: /tmp
+          command: |
+            if [ ! -f testfloat_gen ]; then
+              sudo apt-get -q update
+              sudo apt-get -qy install --no-install-recommends unzip
```
axic, line 56: Could have added unzip to cpp-build-env 😝

test/testfloat/CMakeLists.txt
```diff
@@ -0,0 +1,157 @@
...
+endif()
+
+add_executable(fizzy-testfloat testfloat.cpp)
+target_link_libraries(fizzy-testfloat PRIVATE fizzy::fizzy fizzy::test-utils)
+target_include_directories(fizzy-testfloat PRIVATE ${PROJECT_SOURCE_DIR}/lib)
+
+find_program(TESTFLOAT_GEN testfloat_gen)
+if(TESTFLOAT_GEN)
+    # TODO: Use random seed
+    # TODO: Make level explicit, add option to test level 2.
```
axic, line 18: What is level?

test/testfloat/CMakeLists.txt
```diff
@@ -0,0 +1,157 @@
...
+    set(TESTFLOAT_REQUIRED TRUE)
+endif()
+
+add_executable(fizzy-testfloat testfloat.cpp)
+target_link_libraries(fizzy-testfloat PRIVATE fizzy::fizzy fizzy::test-utils)
+target_include_directories(fizzy-testfloat PRIVATE ${PROJECT_SOURCE_DIR}/lib)
+
+find_program(TESTFLOAT_GEN testfloat_gen)
+if(TESTFLOAT_GEN)
+    # TODO: Use random seed
```
axic, line 17: Where to use random seed?

test/testfloat/testfloat.cpp
```diff
@@ -0,0 +1,442 @@
...
+
+namespace
+{
+/// The "usage" help message.
+///
+/// The option names match the ones from the testfloat_gen / testfloat tools.
+///
+/// TODO: Rename <float>_{ceil,floor,trunc,nearest} to <float>_roundToInt_r_<round>
+///       as in the testfloat tool?
+/// TODO: Rename <float>_trunc_to_<int> to <float>_to_<int>_r_<round>
```
axic, line 26: I'd say we merge as is and can review/reconsider these later.

test/testfloat/testfloat.cpp
```diff
@@ -0,0 +1,442 @@
...
+        switch (m_type)
+        {
+        case Type::i32:
+        case Type::i64:
+            return m_value.i64 == expected_bits;
+        case Type::f32:
+        {
+            const FP value{m_value.f32};
+            const FP expected{static_cast<uint32_t>(expected_bits)};
+            if (ignore_nans && std::isnan(expected.value))
```
axic, line 270: So what does `ignore_nans` means? To ignore payloads?

test/testfloat/CMakeLists.txt
chfast: `testfloat_gen -seed 1`

test/testfloat/testfloat.cpp
```diff
@@ -0,0 +1,442 @@
...
+                    return value.is_arithmetic_nan();
+                throw std::invalid_argument{"invalid input: unexpected signaling NaN"};
+            }
+            return value == expected;
+        }
+        }
+        __builtin_unreachable();
+    }
+
+    friend std::ostream& operator<<(std::ostream& os, const TypedValue& x)
```
axic, line 299: I would name `x` as `value`.

test/testfloat/CMakeLists.txt
chfast: `-level 1` - less tests; `-level 2` - more tests.

circle.yml
chfast: Fixed with github.

test/testfloat/testfloat.cpp
```diff
@@ -0,0 +1,442 @@
...
+{
+    const auto report_failure = [&](auto result, auto expected) {
+        std::cerr << "FAILURE: " << result << " <-";
+        for (size_t i = 0; i < func.num_arguments; ++i)
+            std::cerr << ' ' << TypedValue{func.param_types[i], inputs[i]};
+        std::cerr << "\n         " << expected << " (expected)\n";
+    };
+
+    fizzy::Value args[2]{};
+    for (size_t i = 0; i < func.num_arguments; ++i)
```
axic, line 335: Nit: could add an assert for `num_arguments <= 2`, but I know it is ensured by the declaration in `FunctionDescription`.

test/testfloat/testfloat.cpp
```diff
@@ -0,0 +1,442 @@
...
+  The function names match the ones from the testfloat_gen tool, with the following exceptions:
+  - <float>_to_<int> is <float>_trunc_to_<int> as the result is always rounded with -rminMag,
+  - <float>_roundToInt is <float>_{ceil,floor,trunc,nearest} as they have the rounding direction fixed.
+
+Example:
+  testfloat_gen f32_add | fizzy-testfloat f32_add
+)";
+
+// The maximum number of failures reported.
+constexpr auto max_failures = 5;
```
axic, line 53: Why not move this into `main` and perhaps make it configurable in the future? I'm not sure where it makes better sense, but was easier for reading `main` if it were there.

test/testfloat/testfloat.cpp
```diff
@@ -0,0 +1,442 @@
...
+        //   015E834A C700FFBF C700FFBF 01
+        // f64_to_f32:
+        //   B68FFFF8000000FF 80000000 03
+
+        int num_failures = 0;
+        std::cin >> std::hex;
+        while (num_failures < max_failures)
+        {
+            uint64_t inputs[4]{};
+            for (size_t i = 0; i < func.num_arguments + 2; ++i)
```
axic, line 423: Ah ok, this reads *up to* 4 inputs.

test/testfloat/testfloat.cpp
chfast: Yes.

test/testfloat/testfloat.cpp
chfast: Because it is not an option, it is easier to find it and change when at the top of the file.

test/testfloat/CMakeLists.txt
axic: So perhaps move this comment above the `set(TESTFLOAT_LEVEL 1` line.

axic:
Can you also squash some commits (like the two ci commits, and the review ones)?

test/testfloat/CMakeLists.txt
chfast: Ah, ok. This is already solved.

axic (approved):

PR #515, opened 2020-09-02
Make hard memory limit configurable

### LINKED ISSUE #509: Extend instantiate() with a memory limit
This would replace the hard `MemoryPagesLimit` in `limits.h`. If no limit is set by the user, it defaults to `MemoryPagesLimit` (256Mb). If it is set higher than 2^32-1 return an error (or just make the limit uint32).

Also set this limit to 4Gb in spectests in order to pass them.

gumb0:
Closes #509

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -3,7 +3,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include "execute.hpp"
-#include "limits.hpp"
 #include "module.hpp"
 #include "stack.hpp"
 #include "trunc_boundaries.hpp"
@@ -172,7 +171,7 @@ std::tuple<table_ptr, Limits> allocate_table(
 }
 
 std::tuple<bytes_ptr, Limits> allocate_memory(const std::vector<Memory>& module_memories,
-    const std::vector<ExternalMemory>& imported_memories)
+    const std::vector<ExternalMemory>& imported_memories, uint32_t memory_pages_limit)
 {
     static const auto bytes_delete = [](bytes* b) noexcept { delete b; };
     static const auto null_delete = [](bytes*) noexcept {};
@@ -185,11 +184,11 @@ std::tuple<bytes_ptr, Limits> allocate_memory(const std::vector<Memory>& module_
         const auto memory_max = module_memories[0].limits.max;
 
         // TODO: better error handling
-        if ((memory_min > MemoryPagesLimit) ||
-            (memory_max.has_value() && *memory_max > MemoryPagesLimit))
+        if ((memory_min > memory_pages_limit) ||
+            (memory_max.has_value() && *memory_max > memory_pages_limit))
         {
             throw instantiate_error{"cannot exceed hard memory limit of " +
-                                    std::to_string(MemoryPagesLimit * PageSize) + " bytes"};
+                                    std::to_string(memory_pages_limit * PageSize) + " bytes"};
         }
 
         // NOTE: fill it with zeroes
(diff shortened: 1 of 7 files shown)


lib/fizzy/execute.cpp
```diff
@@ -1229,9 +1230,10 @@ ExecutionResult execute(
             uint32_t ret = static_cast<uint32_t>(cur_pages);
             try
             {
+                // TODO use memory_pages_limit passed to instantiate
                 const size_t memory_max_pages =
```
axic, line 1234: Couldn't we update `instance.memory_pages_limit` with the imported max?

test/unittests/instantiate_test.cpp
```diff
@@ -297,7 +297,7 @@ TEST(instantiate, imported_memory_invalid)
     const auto bin_without_max = from_hex("0061736d01000000020a01036d6f64016d020001");
     const auto module_without_max = parse(bin_without_max);
     EXPECT_THROW_MESSAGE(
-        instantiate(module_without_max, {}, {}, {{&memory, {1, MemoryPagesLimit + 1}}}),
+        instantiate(module_without_max, {}, {}, {{&memory, {1, DefaultMemoryPagesLimit + 1}}}),
         instantiate_error,
         "imported memory limits cannot exceed hard memory limit of 268435456 bytes");
```
axic, line 302: Text update?

axic:
Looks good in general, this is what I meant.

lib/fizzy/execute.cpp
gumb0: I'll try it

lib/fizzy/execute.cpp
```diff
@@ -701,7 +701,13 @@ std::unique_ptr<Instance> instantiate(Module module,
 
     auto [table, table_limits] = allocate_table(module.tablesec, imported_tables);
 
-    auto [memory, memory_limits] = allocate_memory(module.memorysec, imported_memories);
+    auto [memory, memory_limits] =
+        allocate_memory(module.memorysec, imported_memories, memory_pages_limit);
+    if (memory_limits.max.has_value())
```
axic, line 706: Would add something like:
```suggestion
    // If imported memory is used, we need to adjust the memory limit.
    // Note: allocate_memory ensures this limit is always below memory_pages_limit.
    if (memory_limits.max.has_value())
```

test/spectests/spectests.cpp
```diff
@@ -126,9 +127,10 @@ class test_runner
                         continue;
                     }
 
-                    m_instances[name] = fizzy::instantiate(std::move(module),
-                        std::move(imports.functions), std::move(imports.tables),
-                        std::move(imports.memories), std::move(imports.globals));
+                    m_instances[name] =
+                        fizzy::instantiate(std::move(module), std::move(imports.functions),
+                            std::move(imports.tables), std::move(imports.memories),
+                            std::move(imports.globals), TestMemoryPagesLimit);
```
axic, line 133: I'll actually override the default in #329 too.

lib/fizzy/execute.cpp
gumb0: I added the comment, but it applies to non-imported memory, too.

axic: Ah right, it is the memory section setting.

lib/fizzy/execute.cpp
```diff
@@ -173,7 +172,7 @@ std::tuple<table_ptr, Limits> allocate_table(
 }
 
 std::tuple<bytes_ptr, Limits> allocate_memory(const std::vector<Memory>& module_memories,
-    const std::vector<ExternalMemory>& imported_memories)
+    const std::vector<ExternalMemory>& imported_memories, unsigned memory_pages_limit)
```
chfast, line 175:
```suggestion
    const std::vector<ExternalMemory>& imported_memories, uint32_t memory_pages_limit)
```
Always use uint32_t for memory indexing and sizes?

lib/fizzy/execute.cpp
```diff
@@ -677,7 +676,8 @@ std::optional<uint32_t> find_export(const Module& module, ExternalKind kind, std
 
 std::unique_ptr<Instance> instantiate(Module module,
     std::vector<ExternalFunction> imported_functions, std::vector<ExternalTable> imported_tables,
-    std::vector<ExternalMemory> imported_memories, std::vector<ExternalGlobal> imported_globals)
+    std::vector<ExternalMemory> imported_memories, std::vector<ExternalGlobal> imported_globals,
+    unsigned memory_pages_limit /* = DefaultMemoryPagesLimit */)
```
chfast, line 680:
```suggestion
    unsigned memory_pages_limit /*= DefaultMemoryPagesLimit*/)
```
To be consistent with `/*unused*/`?

lib/fizzy/execute.cpp
axic: This is number of pages, the limits sections uses `uint32_t`, and we actually use `size_t` in multiple places regarding this.

chfast (approved):

axic (approved):

PR #516, opened 2020-09-02
CI improvements

chfast:
Ordered some jobs like this:
image

The idea is to run more important jobs first, and in case of failures not to run others.

Besides, some other small improvements are included.


axic (approved):
Looks okay to me. We could bikeshed about what order is the best, to me this looks better than what we have today.

gumb0 (approved):

PR #517, opened 2020-09-02
Add uint32_t i32 field to Value union

chfast:
Problems this introduces
- `stack.top() = 0` only zeros the `.i32` now.
- `FizzyValue{1}` only inits the `.i32` part. This is excessively used in C API tests. And cannot be disabled. In C++20 you can do `FizzyValue{.i64 = 1}`.

### DIFF
--- a/include/fizzy/fizzy.h
+++ b/include/fizzy/fizzy.h
@@ -18,11 +18,11 @@ typedef struct FizzyModule FizzyModule;
 typedef struct FizzyInstance FizzyInstance;
 
 /// The data type representing numeric values.
-///
-/// The #i64 member is used to represent values of both i32 and i64 type.
 typedef union FizzyValue
 {
-    /// Integer value.
+    /// 32-bit integer value.
+    uint32_t i32;
+    /// 64-bit integer value.
     uint64_t i64;
     /// 32-bit floating-point value.
     float f32;
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -1089,7 +1089,7 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, const Value* args,
                 goto trap;
             const auto lhs = stack.top().as<int32_t>();
             if (lhs == std::numeric_limits<int32_t>::min() && rhs == -1)
-                stack.top() = 0;
+                stack.top() = int32_t{0};
             else
                 stack.top() = rem(lhs, rhs);
             break;
@@ -1199,7 +1199,7 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, const Value* args,
                 goto trap;
             const auto lhs = stack.top().as<int64_t>();
             if (lhs == std::numeric_limits<int64_t>::min() && rhs == -1)
-                stack.top() = 0;
+                stack.top() = int64_t{0};
             else
                 stack.top() = rem(lhs, rhs);
             break;
@@ -1434,7 +1434,7 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, const Value* args,
         }
         case Instr::i64_extend_i32_u:
         {
-            // effectively no-op
+            stack.top() = uint64_t{stack.top().i32};
             break;
         }
         case Instr::i64_trunc_f32_s:
--- a/lib/fizzy/stack.hpp
+++ b/lib/fizzy/stack.hpp
@@ -116,7 +116,7 @@ public:
         m_top = m_bottom - 1;
 
         const auto local_variables = std::copy_n(args, num_args, m_locals);
-        std::fill_n(local_variables, num_local_variables, 0);
+        std::fill_n(local_variables, num_local_variables, Value{});
     }
 
     OperandStack(const OperandStack&) = delete;
--- a/lib/fizzy/value.hpp
+++ b/lib/fizzy/value.hpp
@@ -4,7 +4,6 @@
 
 #pragma once
 
-#include <cassert>
 #include <cstdint>
 #include <limits>
 
@@ -12,6 +11,7 @@ namespace fizzy
 {
 union Value
 {
+    uint32_t i32;
     uint64_t i64;
     float f32;
     double f64;
@@ -26,11 +26,11 @@ union Value
     /// We need to support {signed,unsigned} x {32,64} integers. However, due to uint64_t being
     /// defined differently in different implementations we need to avoid the alias and provide
     /// constructors for unsigned long and unsigned long long independently.
-    constexpr Value(unsigned int v) noexcept : i64{v} {}
+    constexpr Value(unsigned int v) noexcept : i32{v} {}
     constexpr Value(unsigned long v) noexcept : i64{v} {}
     constexpr Value(unsigned long long v) noexcept : i64{v} {}
     constexpr Value(int64_t v) noexcept : i64{static_cast<uint64_t>(v)} {}
-    constexpr Value(int32_t v) noexcept : i64{static_cast<uint32_t>(v)} {}
+    constexpr Value(int32_t v) noexcept : i32{static_cast<uint32_t>(v)} {}
 
     constexpr Value(float v) noexcept : f32{v} {}
     constexpr Value(double v) noexcept : f64{v} {}
@@ -54,8 +54,7 @@ constexpr uint64_t Value::as<uint64_t>() const noexcept
 template <>
 constexpr uint32_t Value::as<uint32_t>() const noexcept
 {
-    assert((i64 & 0xffffffff00000000) == 0);
-    return static_cast<uint32_t>(i64);
+    return i32;
 }
 
 template <>
@@ -67,8 +66,7 @@ constexpr int64_t Value::as<int64_t>() const noexcept
 template <>
 constexpr int32_t Value::as<int32_t>() const noexcept
 {
-    assert((i64 & 0xffffffff00000000) == 0);
-    return static_cast<int32_t>(i64);
+    return static_cast<int32_t>(i32);
 }
 
 template <>
--- a/test/unittests/api_test.cpp
+++ b/test/unittests/api_test.cpp
@@ -47,7 +47,7 @@ TEST(api, execution_result_value)
     const ExecutionResult result = Value{1234_u32};
     EXPECT_FALSE(result.trapped);
     EXPECT_TRUE(result.has_value);
-    EXPECT_EQ(result.value.i64, 1234_u32);
+    EXPECT_EQ(result.value.i32, 1234_u32);
 }
 
 TEST(api, execution_result_bool_constructor)
@@ -65,7 +65,7 @@ TEST(api, execution_result_value_constructor)
     const ExecutionResult result{value};
     EXPECT_FALSE(result.trapped);
     EXPECT_TRUE(result.has_value);
-    EXPECT_EQ(result.value.i64, 1234_u32);
+    EXPECT_EQ(result.value.i32, 1234_u32);
 }
 
 TEST(api, resolve_imported_functions)
--- a/test/unittests/capi_test.cpp
+++ b/test/unittests/capi_test.cpp
@@ -378,7 +378,7 @@ TEST(capi, find_exported_global)
     ASSERT_TRUE(fizzy_find_exported_global(instance, "g1", &global));
     EXPECT_EQ(global.type.value_type, FizzyValueTypeI32);
     EXPECT_FALSE(global.type.is_mutable);
-    EXPECT_EQ(global.value->i64, 42);
+    EXPECT_EQ(global.value->i32, 42);
 
     EXPECT_FALSE(fizzy_find_exported_global(instance, "g2", &global));
     EXPECT_FALSE(fizzy_find_exported_global(instance, "foo", &global));
@@ -729,7 +729,7 @@ TEST(capi, memory_access)
     memory[0] = 0xaa;
     memory[1] = 0xbb;
 
-    EXPECT_EQ(fizzy_execute(instance, 0, nullptr, 0).value.i64, 0x22bbaa);
+    EXPECT_THAT(fizzy_execute(instance, 0, nullptr, 0), CResult(0x22bbaa_u32));
 
     fizzy_free_instance(instance);
 }
@@ -771,7 +771,7 @@ TEST(capi, imported_memory_access)
     auto* instance = fizzy_instantiate(module, nullptr, 0, nullptr, &memory, nullptr, 0);
     ASSERT_NE(instance, nullptr);
 
-    EXPECT_EQ(fizzy_execute(instance, 0, nullptr, 0).value.i64, 0x221100);
+    EXPECT_EQ(fizzy_execute(instance, 0, nullptr, 0).value.i32, 0x221100);
 
     EXPECT_EQ(fizzy_get_instance_memory_size(instance), 65536);
 
@@ -781,8 +781,8 @@ TEST(capi, imported_memory_access)
     memory_data[0] = 0xaa;
     memory_data[1] = 0xbb;
 
-    EXPECT_EQ(fizzy_execute(instance_memory, 0, nullptr, 0).value.i64, 0x22bbaa);
-    EXPECT_EQ(fizzy_execute(instance, 0, nullptr, 0).value.i64, 0x22bbaa);
+    EXPECT_EQ(fizzy_execute(instance_memory, 0, nullptr, 0).value.i32, 0x22bbaa);
+    EXPECT_EQ(fizzy_execute(instance, 0, nullptr, 0).value.i32, 0x22bbaa);
 
     fizzy_free_instance(instance);
     fizzy_free_instance(instance_memory);
@@ -838,7 +838,9 @@ TEST(capi, execute_with_host_function)
                                               nullptr},
         {{FizzyValueTypeI32, &inputs[0], 2},
             [](void*, FizzyInstance*, const FizzyValue* args, int) {
-                return FizzyExecutionResult{false, true, {args[0].i64 / args[1].i64}};
+                FizzyValue v;
+                v.i32 = args[0].i32 / args[1].i32;
+                return FizzyExecutionResult{false, true, {v}};
             },
             nullptr}};
 
--- a/test/unittests/execute_numeric_test.cpp
+++ b/test/unittests/execute_numeric_test.cpp
@@ -483,6 +483,16 @@ TEST(execute_numeric, i64_extend_i32_s)
     EXPECT_THAT(i64_extend_i32_s(0x80000001_u32), Result(0xffffffff80000001_u64));
     EXPECT_THAT(i64_extend_i32_s(0xfffffffe_u32), Result(0xfffffffffffffffe_u64));
     EXPECT_THAT(i64_extend_i32_s(0xffffffff_u32), Result(0xffffffffffffffff_u64));
+
+    // Put some garbage in the Value's high bits.
+    TypedValue v{ValType::i32, {}};
+    v.value.i64 = 0xdeaddeaddeaddead;
+    v.value.i32 = 0x80000000;
+    EXPECT_THAT(i64_extend_i32_s(v), Result(0xffffffff80000000_u64));
+
+    v.value.i64 = 0xdeaddeaddeaddead;
+    v.value.i32 = 0x40000000;
+    EXPECT_THAT(i64_extend_i32_s(v), Result(0x0000000040000000_u64));
 }
 
 TEST(execute_numeric, i64_extend_i32_u)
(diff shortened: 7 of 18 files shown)


axic:
What was the blocker on this? It would make the Rust API a tiny bit nicer too.

chfast:
> What was the blocker on this? It would make the Rust API a tiny bit nicer too.

Unit tests expectations update.

lib/fizzy/value.hpp
```diff
@@ -54,8 +55,7 @@ constexpr uint64_t Value::as<uint64_t>() const noexcept
 template <>
 constexpr uint32_t Value::as<uint32_t>() const noexcept
 {
-    assert((i64 & 0xffffffff00000000) == 0);
```
axic, line 57: Do we want to keep this?

chfast: No, as this is invalid now.

axic:
This is now blocked on explicitly typing number literals in the tests.

lib/fizzy/value.hpp
```diff
@@ -57,18 +57,6 @@ constexpr uint32_t Value::as<uint32_t>() const noexcept
     return i32;
 }
 
-template <>
-constexpr unsigned long Value::as<unsigned long>() const noexcept
```
axic: Added these as "quick fix" to see which ones the compiler couldn't correctly match :)

CMakeLists.txt
```diff
@@ -43,6 +43,7 @@ add_compile_options(
     -Wmissing-declarations
     $<$<COMPILE_LANGUAGE:CXX>:-Wextra-semi>
     $<$<COMPILE_LANGUAGE:CXX>:-Wold-style-cast>
+    -Wno-invalid-constexpr
```
axic, line 46: This is just stray from my local needs. (Need to update xcode, the current one still doesn't work without this.).

test/utils/asserts.hpp
```diff
@@ -96,7 +97,6 @@ namespace fizzy::test
 {
 inline uint32_t as_uint32(fizzy::Value value)
```
axic, line 98: Eventually we should also remove this, if we merge this PR.

lib/fizzy/stack.hpp
```diff
@@ -115,7 +115,7 @@ class OperandStack
         m_top = m_bottom - 1;
 
         const auto local_variables = std::copy_n(args, num_args, m_locals);
-        std::fill_n(local_variables, num_local_variables, 0);
+        std::fill_n(local_variables, num_local_variables, Value{});
```
gumb0, line 118: Hm, I'm confused by union initialization again. Does default initialization like this set the active member of the union?

chfast: I'm confused too. This was only found by GCC coverage build.

The `0` is definitely wrong because it calls the special constructor and inits `.i32` only.

But the default constructor should be wrong too because it should only init the first field and `.i32` is the first one. This works however for current compilers. I wander if MSan can help here. Would be good if you can check the spec yourself.

One solution to solve this is to provide own implementation of the default constructor which inits `.i64` (or define `i64 = 0;`). But with that the `Value` looses some properties like "is trivially constructable".

Second solution is to move `.i64` to the first position. This guarantees the `Value{}` to be fully zero-initialized. Same for `FizzyValue` in C API.

But I don't see any sanity solution to init locals with zero without knowing the type.

chfast: I don't believe I'm doing it again to myself, but:
1. The `Value{}` is value-initialization.
2. if T is a class type with a default constructor that is neither user-provided nor deleted (that is, it may be a class with an implicitly-defined or defaulted default constructor), the object is zero-initialized.
3. Zero-initialization: If T is a union type, the first non-static named data member is zero-initialized and all padding is initialized to zero bits.

So I think the `Value{}` is correct here as it zeros all object's bytes.

lib/fizzy/execute.cpp
```diff
@@ -1199,7 +1199,7 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, const Value* args,
                 goto trap;
             const auto lhs = stack.top().as<int64_t>();
             if (lhs == std::numeric_limits<int64_t>::min() && rhs == -1)
-                stack.top() = 0;
+                stack.top().i64 = 0;  // FIXME: Review all cases like this.
```
gumb0, line 1202: I don't insist it's much better, but this way it would look more like you're just returning 64-bit value, without assuming internal representation of stack item
```suggestion
                stack.top() = uint64_t{0};  // FIXME: Review all cases like this.
```

test/spectests/spectests.cpp
```diff
@@ -485,7 +485,24 @@ class test_runner
         return fizzy::execute(*instance, *func_idx, args.data());
     }
 
-    bool check_integer_result(fizzy::Value actual_value, const json& expected)
+    bool check_i32_result(fizzy::Value actual_value, const json& expected)
```
gumb0, line 488: I think `read_integer_value` function has to be modified, too, because now it produces always 64-bit `Value`

test/utils/asserts.hpp
```diff
@@ -100,6 +100,8 @@ MATCHER_P(CResult, value, "")  // NOLINT(readability-redundant-string-init)
             return arg.value.f64 == fizzy::test::FP{value};
         }
     }
+    else if constexpr (std::is_same_v<value_type, uint32_t>)
```
gumb0, line 103: I think this should also include `int32_t`?

test/utils/asserts.hpp
```diff
@@ -100,6 +100,8 @@ MATCHER_P(CResult, value, "")  // NOLINT(readability-redundant-string-init)
...
+        return arg.value.i32 == value;
     else  // always check 64 bit of result for all integers, including 32-bit results
```
gumb0, line 105: comment is outdated

test/utils/fizzy_c_engine.cpp
```diff
@@ -108,13 +108,15 @@ WasmEngine::Result FizzyCEngine::execute(
     WasmEngine::FuncRef func_ref, const std::vector<uint64_t>& args)
 {
     static_assert(sizeof(uint64_t) == sizeof(FizzyValue));
+    const auto func_idx = static_cast<uint32_t>(func_ref);
+    const auto* module = fizzy_get_instance_module(m_instance.get());
+    const auto func_type = fizzy_get_function_type(module, func_idx);
```
gumb0, line 113: You can add now
```
    assert(args.size() == func_type.inputs_size);
```

test/unittests/value_test.cpp
```diff
@@ -15,9 +15,9 @@ TEST(value, value_initialization)
 
 TEST(value, constructor_from_i64)
```
gumb0, line 16: Test for constructor from i32 is now needed.

test/unittests/capi_test.cpp
```diff
@@ -593,7 +593,7 @@ TEST(capi, memory_access)
     memory[0] = 0xaa;
     memory[1] = 0xbb;
 
-    EXPECT_EQ(fizzy_execute(instance, 0, nullptr, 0).value.i64, 0x22bbaa);
+    EXPECT_EQ(fizzy_execute(instance, 0, nullptr, 0).value.i32, 0x22bbaa);
```
gumb0, line 596:
```suggestion
    EXPECT_EQ(fizzy_execute(instance, 0, nullptr, 0), CResult(0x22bbaa_u32));
```

lib/fizzy/stack.hpp
gumb0: It seems to depend on the interpretation of what is "padding" for a union - whether unused bits of unactive members are unions. I see some people having doubts about this here: https://stackoverflow.com/questions/11812555/how-to-zero-initialize-an-union#comment15704643_11812606 I have not found a definite answer to this yet.

However, I had a concern about something else (perhaps not directly related to this PR): we zero-initialize `Value`s of locals here, which might or might not choose the active member of the union (I have not found the answer to this either). Then function may access them right away with `local.get`, then some other operation may access some member of it - can be any type member. Then this may or may not be UB.

We can add a test for this, but as I remember sanitizers didn't catch union type punning.

And if we're required to set the right type, then there's no way around needing the types of locals when initializing.

gumb0: > We can add a test for this, but as I remember sanitizers didn't catch union type punning.

Well, the tests you added here (`execute.local_init`) already check this actually. Because checking the result of a function reads some specific member of `Value` union.

chfast: I did some experiments and it looks like GCC and MSVC inits the whole object while Clang only the first element: https://godbolt.org/z/cWscY8.

gumb0: This may be a good reason to make `i64` the first member. Are there any advantages for it to be the second other than aesthetical?

axic:
Needs rebase after similar PRs (#655, #659) are merged.

lib/fizzy/execute.cpp
```diff
@@ -1429,12 +1429,13 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, const Value* args,
         }
         case Instr::i64_extend_i32_s:
         {
+            // FIXME: Add more tests and make sure this is correct.
```
axic, line 1432: Do you want to add tests first before this PR?

axic: Done in #680 ?

chfast: Yes, although this PR will extend the tests anyway by polluting the Value with some garbage (not possible before).

test/unittests/api_test.cpp
```diff
@@ -47,7 +47,7 @@ TEST(api, execution_result_value)
     const ExecutionResult result = Value{1234};
```
gumb0, line 47:
```suggestion
    const ExecutionResult result = Value{1234_u32};
```

lib/fizzy/execute.cpp
axic: I'd agree this is cleaner, since the `rem()` case below is defined as `T rem(T a, T b)`, so both would be assigning a `int64_t` to `top()`.

axic: In that sense also I do not get the FIXME, having `top() = uint64_t{0}` seems to be the correct approach.

lib/fizzy/execute.cpp
```diff
@@ -1429,12 +1429,13 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, const Value* args,
...
             stack.top() = int64_t{stack.top().as<int32_t>()};
             break;
         }
         case Instr::i64_extend_i32_u:
         {
-            // effectively no-op
+            stack.top().i64 = uint64_t{stack.top().i32};
```
axic, line 1438: Why not?

```suggestion
            stack.top() = uint64_t{stack.top().i32};
```

test/unittests/api_test.cpp
```diff
@@ -65,7 +65,7 @@ TEST(api, execution_result_value_constructor)
     const ExecutionResult result{value};
```
axic, line 65: @gumb0 How about the `value{1234}` above?

test/unittests/capi_test.cpp
```diff
@@ -380,7 +380,7 @@ TEST(capi, instantiate_imported_globals)
     auto instance = fizzy_instantiate(module, nullptr, 0, nullptr, nullptr, globals, 4);
     EXPECT_NE(instance, nullptr);
 
-    EXPECT_THAT(fizzy_execute(instance, 0, nullptr, 0), CResult(42));
+    EXPECT_THAT(fizzy_execute(instance, 0, nullptr, 0), CResult(42_u32));
     EXPECT_THAT(fizzy_execute(instance, 1, nullptr, 0), CResult(uint64_t{43}));
```
axic, line 384:
```suggestion
    EXPECT_THAT(fizzy_execute(instance, 1, nullptr, 0), CResult(43_u64));
```

test/unittests/execute_numeric_test.cpp
```diff
@@ -481,6 +481,12 @@ TEST(execute_numeric, i64_extend_i32_s)
     EXPECT_THAT(i64_extend_i32_s(0x80000001_u32), Result(0xffffffff80000001_u64));
     EXPECT_THAT(i64_extend_i32_s(0xfffffffe_u32), Result(0xfffffffffffffffe_u64));
     EXPECT_THAT(i64_extend_i32_s(0xffffffff_u32), Result(0xffffffffffffffff_u64));
+
+    // Put some garbage in the Value's high bits.
```
axic, line 485: Hmm, what about a case where `i64` has top bit set, but `i32` does not have top bit set. That should clear the top bits.

test/utils/asserts.hpp
axic: @chfast do you want to remove it in a separate commit, or a separate PR? I can do the removal too.

axic: Well, actually shouldn't all the test code be using `as<uint32_t>()`? We can apply that parallel to this PR.

test/unittests/capi_test.cpp
chfast: Well the `CResult()` and `Result()` are pretty useless now as they try to compare all 64 bits without knowing the type.

lib/fizzy/execute.cpp
chfast: Done.

test/unittests/execute_numeric_test.cpp
chfast: Done.

test/unittests/value_test.cpp
chfast: Done.

test/utils/asserts.hpp
chfast: As changed in #687, only `_u32` and `_u64` are accepted for `CResult()`.

test/utils/fizzy_c_engine.cpp
chfast: Done.

test/unittests/api_test.cpp
gumb0: The way `Value` constuctor behaves with `int` should be checked by `Value` unit tests.

chfast:
I don't see any significant speed difference except for noise. The GCC LTO build is much noisier.
```
fizzy/execute/blake2b/512_bytes_rounds_1_mean                     +0.0095         +0.0095            80            81            80            81
fizzy/execute/blake2b/512_bytes_rounds_16_mean                    +0.0085         +0.0085          1207          1217          1207          1217
fizzy/execute/ecpairing/onepoint_mean                             +0.0072         +0.0072        404734        407646        404736        407648
fizzy/execute/keccak256/512_bytes_rounds_1_mean                   +0.0201         +0.0201           101           103           101           103
fizzy/execute/keccak256/512_bytes_rounds_16_mean                  +0.0131         +0.0131          1489          1508          1489          1508
fizzy/execute/memset/256_bytes_mean                               -0.0167         -0.0167             7             7             7             7
fizzy/execute/memset/60000_bytes_mean                             -0.0209         -0.0209          1539          1507          1539          1507
fizzy/execute/mul256_opt0/input1_mean                             +0.0347         +0.0347            29            30            29            30
fizzy/execute/ramanujan_pi/33_runs_mean                           -0.0274         -0.0274           114           111           114           111
fizzy/execute/sha1/512_bytes_rounds_1_mean                        +0.0303         +0.0303            88            91            88            91
fizzy/execute/sha1/512_bytes_rounds_16_mean                       +0.0334         +0.0334          1225          1266          1225          1266
fizzy/execute/sha256/512_bytes_rounds_1_mean                      +0.0075         +0.0075            80            81            80            81
fizzy/execute/sha256/512_bytes_rounds_16_mean                     +0.0087         +0.0087          1101          1111          1101          1111
fizzy/execute/taylor_pi/pi_1000000_runs_mean                      -0.0011         -0.0011         40619         40573         40620         40573
fizzy/execute/micro/eli_interpreter/exec105_mean                  -0.0150         -0.0150             5             4             5             4
fizzy/execute/micro/factorial/20_mean                             +0.0207         +0.0206             1             1             1             1
fizzy/execute/micro/fibonacci/24_mean                             +0.0175         +0.0175          5048          5137          5048          5137
fizzy/execute/micro/host_adler32/1_mean                           +0.0211         +0.0211             0             0             0             0
fizzy/execute/micro/host_adler32/1000_mean                        -0.0042         -0.0042            31            31            31            31
fizzy/execute/micro/icall_hash/1000_steps_mean                    +0.0033         +0.0033            69            69            69            69
fizzy/execute/micro/spinner/1_mean                                -0.0603         -0.0603             0             0             0             0
fizzy/execute/micro/spinner/1000_mean                             -0.0134         -0.0134             9             9             9             9
```

test/utils/asserts.hpp
gumb0: Hm, writing a test expecting a negative value will be a bit awkward then, istn't it?

chfast: `-1_u32`.

test/unittests/value_test.cpp
```diff
@@ -13,11 +13,38 @@ TEST(value, value_initialization)
...
+
+TEST(value, constructor_from_int)
+{
+    // The int type is likely the int32_t.
+    Value v = 0;
+    EXPECT_EQ(v.i32, 0);
+    v = -3;
+    EXPECT_EQ(v.i32, 0xfffffffd);
+
+    constexpr auto max_int = std::numeric_limits<uint32_t>::max();
```
gumb0, line 38: This is the same as in the test above, did you mean:
```suggestion
    constexpr auto max_int = std::numeric_limits<int>::max();
```

test/unittests/value_test.cpp
```diff
@@ -13,11 +13,38 @@ TEST(value, value_initialization)
...
+    EXPECT_EQ(v.i32, max_int);
+}
+
 TEST(value, constructor_from_i64)
 {
-    Value v = 1;
+    Value v = int64_t{1};
     EXPECT_EQ(v.i64, 1);
-    v = 2;
+    v = int64_t{2};
```
gumb0, line 47: You could add here a negative value, as in new tests above

test/utils/asserts.hpp
gumb0: But isn't it equivalent to `-uint32_t{1}`, so signed value?

gumb0: I'm confused about why `Result` and `CResult` are different in this regard.
Is `Result` allowing signed values because otherwise there're many tests to change?

chfast: > But isn't it equivalent to `-uint32_t{1}`, so signed value?

It is, but it stays `uint32_t` type.



> I'm confused about why `Result` and `CResult` are different in this regard.
> Is `Result` allowing signed values because otherwise there're many tests to change?

The `Result()` requires type information. I restricted values to `CResult()` more because the `EXPECT_THAT(Value{0x100000000_u64}, CResult(0_u32))` is a match.

gumb0: > `EXPECT_THAT(Value{0x100000000_u64}, CResult(0_u32))` is a match.

How does restricting to only unsigned values help this?

chfast: You can't put `CResult(0)` so you need to thing if it should be `_u32` or `_u64`. This also affects existing tests.

gumb0: I see. And what is the disadvantage of making `Result()` accept only unsigned expected types?

chfast: It may be good idea, but this requires 176 additional test changes. So I went with adding type checking instead. Can be done later if desired.

We can also require type in `CResult` or even merge these two. But I believe there is additional research needed here. E.g. we should try combining matchers: `Result: Not(Traps) && HasResult && ResultValueIs(x)`.

test/utils/asserts.hpp
chfast: > Well, actually shouldn't all the test code be using `as<uint32_t>()`? We can apply that parallel to this PR.

The `Result()` was checking this to some distinct. But I went with additional type checking to solved it for good.

test/utils/asserts.hpp
gumb0: I see, that's what I guessed in the beginning, too many tests to change to make `Result` accept only unsigned,

gumb0 (approved):
Looks fine to me :shipit:

test/utils/asserts.hpp
chfast: > @chfast do you want to remove it in a separate commit, or a separate PR? I can do the removal too.

In https://github.com/wasmx/fizzy/pull/691.

test/unittests/api_test.cpp
```diff
@@ -47,7 +47,7 @@ TEST(api, execution_result_value)
     const ExecutionResult result = Value{1234_u32};
     EXPECT_FALSE(result.trapped);
     EXPECT_TRUE(result.has_value);
-    EXPECT_EQ(result.value.i64, 1234_u32);
+    EXPECT_EQ(result.value.i32, 1234_u32);
```
axic, line 50: Notpick, we do not actually need `_u32` here, right?

test/utils/fizzy_c_engine.cpp
```diff
@@ -108,13 +108,16 @@ WasmEngine::Result FizzyCEngine::execute(
...
+    assert(args.size() == func_type.inputs_size);
     const auto first_arg = reinterpret_cast<const FizzyValue*>(args.data());
-    const auto status =
-        fizzy_execute(m_instance.get(), static_cast<uint32_t>(func_ref), first_arg, 0);
+    const auto status = fizzy_execute(m_instance.get(), func_idx, first_arg, 0);
     if (status.trapped)
         return {true, std::nullopt};
     else if (status.has_value)
-        return {false, status.value.i64};
+        return {false, func_type.output == FizzyValueTypeI32 ? status.value.i32 : status.value.i64};
```
axic, line 120: Hmm both of these (fizzy_c_engine and fizzy_engine) don't validate that f32/f64 is, but `bench` ensures that. Anyway perhaps we should have an assert somewhere, can be done in a new PR.

axic (approved):

test/unittests/api_test.cpp
chfast: No. But using `_u32` gives higher confidence (i.e. someone has thought what the result type is expected). Still gray area though.

test/utils/fizzy_c_engine.cpp
chfast: Added asserts.

axic (approved):

include/fizzy/fizzy.h
```diff
@@ -18,11 +18,12 @@ typedef struct FizzyModule FizzyModule;
...
 
 /// The data type representing numeric values.
-///
-/// The #i64 member is used to represent values of both i32 and i64 type.
 typedef union FizzyValue
 {
-    /// Integer value.
+    /// 32-bit integer value.
+    uint32_t i32;
+
```
axic, line 25: Nit: extra new line between these two, but not the others :)

gumb0 (approved):

PR #519, opened 2020-09-04
ci: Coverage with GCC

chfast:
I'd like to switch to using GCC for line-based coverage. The Clang source-based coverage will stay but will not be used on codecov.

There are two issues with Clang: it has some bugs (e.g. https://codecov.io/gh/wasmx/fizzy/src/3b19fb019b2d204ebe4cb7256d0cfb5db318b78e/lib/fizzy/parser.cpp#L155) and in the source-based coverage to line-based coverage conversion the comments are also marked as executed. The later screws the line count stats.

The plan is to prepare LCOV summary of the coverage on CI and only upload this single file (similarly to what we currently do with Clang).


chfast:
I'm getting nice results locally, but something does not work on CI (maybe because of different build dir location).

axic:
I like the idea, but the codecov link still displays weird results, the files are missing most of the code lines and have empty lines instead (for example `execute.cpp`).

axic (approved):

PR #520, opened 2020-09-04
ci: Combine linter/style/tidy/portability checks

chfast:
This moves clang-tidy checking and C++20 build to "lint" job.

The `-Weverything` build will be added to "lint" too in #508.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -6,9 +6,6 @@ parameters:
     default: false
 
 executors:
-  lint:
-    docker:
-      - image: ethereum/cpp-build-env:14-lint
   linux-gcc-9:
     docker:
       - image: ethereum/cpp-build-env:14-gcc-9
@@ -17,7 +14,7 @@ executors:
       - image: ethereum/cpp-build-env:14-gcc-10
   linux-clang-latest:
     docker:
-      - image: ethereum/cpp-build-env:14-clang-10
+      - image: ethereum/cpp-build-env:15-clang-10
   macos:
     macos:
       xcode: 11.6.0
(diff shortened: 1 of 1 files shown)


circle.yml
```diff
@@ -17,7 +14,7 @@ executors:
       - image: ethereum/cpp-build-env:14-gcc-10
   linux-clang-latest:
     docker:
-      - image: ethereum/cpp-build-env:14-clang-10
+      - image: ethereum/cpp-build-env:15-clang-10
```
axic, line 17: What is the update in 15? Why not update the others for consistency too?

circle.yml
```diff
@@ -304,13 +304,13 @@ jobs:
   sanitizers-macos:
     executor: macos
     environment:
-      BUILD_TYPE: RelWithDebInfo
-      CMAKE_OPTIONS: -DENABLE_ASSERTIONS=ON -DSANITIZE=address,undefined,nullability,implicit-unsigned-integer-truncation,implicit-signed-integer-truncation
       UBSAN_OPTIONS: halt_on_error=1
```
axic, line 308: This is not a build, but runtime parameter, hence not move under `build` settings?

circle.yml
```diff
@@ -440,7 +441,6 @@ workflows:
           requires:
             - sanitizers-macos
       - coverage
-      - cxx20
```
axic, line 442: You did not remove the `cxx20` task above. Should it be kept?

circle.yml
chfast: I only updated this one to include `clang-format`.

circle.yml
chfast: It must be the environment variable anyway (libubsan checks it), so this way of setting it is fine.

axic (approved):
Looks good to me, just remove the stray cxx20 task.

PR #528, opened 2020-09-10
Improve documentation of internal C++ API

axic:


lib/fizzy/execute.hpp
```diff
@@ -33,12 +36,30 @@ struct ExecutionResult
     constexpr explicit ExecutionResult(bool success) noexcept : trapped{!success} {}
 };
 
+/// Shortcut for execution that resulted in successful execution, but without a result.
 constexpr ExecutionResult Void{true};
+/// Shortcut for execution that resulted in a trap.
 constexpr ExecutionResult Trap{false};
 
-// Execute a function on an instance.
+/// Execute a function on an instance.
```
chfast, line 44:
```suggestion
/// Execute a function from an instance.
```

lib/fizzy/execute.hpp
```diff
@@ -33,12 +36,30 @@ struct ExecutionResult
...
+///
+/// @param instance    The instance, see `instance.hpp`.
```
chfast, line 46:
```suggestion
/// @param instance    The instance.
```

lib/fizzy/execute.hpp
```diff
@@ -18,11 +18,14 @@
 
 namespace fizzy
 {
-// The result of an execution.
+/// The result of an execution.
 struct ExecutionResult
 {
+    /// This is true if the execution has trapped. No other fields are valid.
     const bool trapped = false;
+    /// This is true if the `value()` returns valid data.
```
gumb0, line 26: (there's no `value()`)
```suggestion
    /// This is true if value contains valid data.
```

lib/fizzy/execute.hpp
```diff
@@ -18,11 +18,14 @@
 
 namespace fizzy
 {
-// The result of an execution.
+/// The result of an execution.
 struct ExecutionResult
 {
+    /// This is true if the execution has trapped. No other fields are valid.
```
gumb0, line 24: I suggest to remove this, as Rust part relies on `has_value` being valid in this case now.
```suggestion
    /// This is true if the execution has trapped.
```

lib/fizzy/execute.hpp
```diff
@@ -33,12 +36,30 @@ struct ExecutionResult
...
 constexpr ExecutionResult Trap{false};
 
-// Execute a function on an instance.
+/// Execute a function on an instance.
+///
+/// @param instance    The instance, see `instance.hpp`.
+/// @param func_idx    The function index. MUST be a valid index, otherwise undefined behaviour
+///                    (including crash) happens.
+/// @param args        The pointer to the arguments. The number of items must match the expected
+///                    number of input parameters of the function, otherwide underfiend behaviour
```
gumb0, line 50:
```suggestion
///                    number of input parameters of the function, otherwise underfined behaviour
```
and the same below

lib/fizzy/execute.hpp
```diff
@@ -33,12 +36,30 @@ struct ExecutionResult
...
+/// Shortcut for execution that resulted in a trap.
 constexpr ExecutionResult Trap{false};
 
-// Execute a function on an instance.
+/// Execute a function on an instance.
+///
+/// @param instance    The instance, see `instance.hpp`.
+/// @param func_idx    The function index. MUST be a valid index, otherwise undefined behaviour
+///                    (including crash) happens.
+/// @param args        The pointer to the arguments. The number of items must match the expected
```
gumb0, line 49:
```suggestion
/// @param args        The pointer to the arguments. The number of items and their types must match the expected
```

lib/fizzy/parser.hpp
```diff
@@ -17,6 +17,11 @@ constexpr bytes_view wasm_prefix{wasm_prefix_data, sizeof(wasm_prefix_data)};
 template <typename T>
 using parser_result = std::pair<T, const uint8_t*>;
 
+/// Parses `input` into a Module.
```
gumb0, line 20: Do backticks work in doxygen?

lib/fizzy/parser.hpp
```diff
@@ -17,6 +17,11 @@ constexpr bytes_view wasm_prefix{wasm_prefix_data, sizeof(wasm_prefix_data)};
...
+///
+/// @param input The WebAssembly binary. No need to persist by the caller, since all relevant parts
```
gumb0, line 22: Did you want 4 spaces between argument name and comment?

lib/fizzy/execute.hpp
axic: Why?

lib/fizzy/execute.hpp
axic: I think there may have been some rebases earlier.

lib/fizzy/execute.hpp
axic, line 50:
```suggestion
///                    number of input parameters of the function, otherwide undefined behaviour
```

None of you caught this 😅

lib/fizzy/parser.hpp
axic: Apparently, I think so.

lib/fizzy/parser.hpp
axic: I'll copy your style from #664.

lib/fizzy/execute.hpp
gumb0: Also codespell thinks `otherwide` and `underfiend` are ok, but catches `undefiend`

lib/fizzy/execute.hpp
chfast: You can follow the type to go there.

lib/fizzy/execute.hpp
axic: Yeah.. fixed all.

gumb0 (approved):

lib/fizzy/parser_expr.cpp
```diff
@@ -95,7 +95,7 @@ inline bool type_matches(OperandStackType actual_type, OperandStackType expected
 /// @return optional type of the block result.
 parser_result<std::optional<ValType>> parse_blocktype(const uint8_t* pos, const uint8_t* end)
 {
-    // The byte meaning an empty wasm result type.
+    /// The byte meaning an empty wasm result type.
```
gumb0, line 98: Not sure if it makes sense to use doxygen comments inside functions? (it's an implementation detail after all)

But if it does, then next line should be doxygen, too.

axic: Yes, it shouldn't be in a function, for a second thought it was a struct. Will remove.

lib/fizzy/stack.hpp
```diff
@@ -89,13 +89,14 @@ class OperandStack
...
-    /// @param max_stack_height     The maximum operand stack height in the function. This excludes
-    ///                             args and num_local_variables.
+    ///
+    /// @param  args                   Function arguments. Values are copied at the beginning of the
+    ///                                storage space.
+    /// @param  num_local_variables    The number of the function local variables (excluding
+    ///                                arguments). This number of values is zeroed in the storage
+    ///                                space after the arguments.
+    /// @param  max_stack_height       The maximum operand stack height in the function. This
+    ///                                excludes args and num_local_variables.
```
gumb0, line 99:
```suggestion
    ///                                excludes @a args and @a num_local_variables.
```

lib/fizzy/parser_expr.cpp
axic: Fixed.

lib/fizzy/stack.hpp
axic: Fixed.

gumb0 (approved):

PR #530, opened 2020-09-10
Public C API (validation only)

chfast:


test/compilation/compilation_test.c
```diff
@@ -0,0 +1,10 @@
+// Fizzy: A fast WebAssembly interpreter
+// Copyright 2020 The Fizzy Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+#include <fizzy/fizzy.h>
+
+bool dummy()
```
gumb0, line 7: `-Wmissing-declarations` complains about this

gumb0 (approved):

axic:
Should add an install target to this, or in a new PR to be merged before.

include/fizzy/fizzy.h
```diff
@@ -0,0 +1,18 @@
...
+
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+bool fizzy_validate(const uint8_t* wasm_binary, size_t wasm_binary_size);
```
axic, line 14: What C std are we targeting? C99? EVMC supports C99+C11. We should probably also have a ci run for that.

test/compilation/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+# EVMC: Ethereum Client-VM Connector API.
+# Copyright 2018-2020 The EVMC Authors.
+# Licensed under the Apache License, Version 2.0.
+
+# This CMake script creates multiple additional targets to test the compilation of public headers
+# with different C standards.
+
+set(STANDARDS c_std_99 c_std_11)
```
axic, line 8: Never mind, this does it, nice!

chfast:
> Should add an install target to this, or in a new PR to be merged before.

Added here.

axic:
Do we want to add a `FIZZY_VERSION` macro in this PR?

chfast:
> Do we want to add a `FIZZY_VERSION` macro in this PR?

Add it where?

axic:
`fizzy.h`, the public API. Or no need yet and assume the integrator knows what version they are using? It may be premature anyway.

chfast:
So far I was successfully avoiding version macros.

test/unittests/capi_test.cpp
```diff
@@ -0,0 +1,14 @@
+// Fizzy: A fast WebAssembly interpreter
+// Copyright 2020 The Fizzy Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+#include <fizzy/fizzy.h>
+#include <gtest/gtest.h>
+
+TEST(capi, fizzy_validate)
```
axic, line 8: #533 does not use the `fizzy_` prefix here. Do we want to follow?

chfast: I'm forward-following now.

gumb0 (approved):

axic (approved):

PR #532, opened 2020-09-15
build: Pass sanitizer flags to WABT

chfast:


axic (approved):
Ok this way too, but most of these commits we use the `cmake:` prefix.

PR #533, opened 2020-09-17
C API execute

gumb0:
Possible options to deal with module-instance relationship.

PR currently uses 2.b approach.

1. Copy module during instantiate
    ```
    FizzyInstance* instance = fizzy_instantiate(module, NULL, 0);
    // can use module, instance has it's own copy
    ...
    fizzy_free_module(module);
    fizzy_free_instance(instance);
    ```
2. Invalidate module in instantiate
    2.a. `fizzy_free_module` required after invalidation, so that any module allocated by `fizzy_parse` must be freed in the end by the user.
    ```
    FizzyInstance* instance = fizzy_instantiate(module, NULL, 0);
    // can't use module
    ...
    fizzy_free_module(module);
    fizzy_free_instance(instance);
    ``` 
    2.b. instatiate calls `fizzy_free_module`, so it must not be called by user afterwards.
    
    ```
    FizzyInstance* instance = fizzy_instantiate(module, NULL, 0);
    // can't use module
    ...
    fizzy_free_instance(instance);
    ``` 

3. Reference counting
    ```
    FizzyInstance* instance = fizzy_instantiate(module, NULL, 0);
    // can use module, instance has a reference to the same module
    ...
    fizzy_free_module(module);
    fizzy_free_instance(instance);
    ```
    Different ways to implement this on the C++ side:
    3.a. double indirection
    ```
    struct FizzyModule
    {
        shared_ptr<fizzy::Module> module;
    };
    
    FizzyInstance* fizzy_instantiate(FizzyModule* cmod, ...)
    {
      return fizzy::instantiate(cmod->module, ...)
    }
    ```
    3.b. `Module` with enabled `shared_from_this`:
    ```
    class Module : public std::enable_shared_from_this<Module> { ... };
    
    fizzy::Module* unwrap(FizzyModule* cmod) { return reinterpret_cast<fizzy::Module*>(cmod); }
    
    FizzyInstance* fizzy_instantiate(FizzyModule* cmod, ...)
    {
      return fizzy::instantiate(unwrap(cmod)->shared_from_this(), ...)
    }
    ```

### DIFF
--- a/include/fizzy/fizzy.h
+++ b/include/fizzy/fizzy.h
@@ -11,7 +11,85 @@
 extern "C" {
 #endif
 
+/// The opaque data type representing a module.
+struct FizzyModule;
+
+/// The opaque data type representing an instance (instantiated module).
+struct FizzyInstance;
+
+/// The data type representing numeric values.
+///
+/// i64 member is used to represent values of both i32 and i64 type.
+union FizzyValue
+{
+    uint64_t i64;
+    float f32;
+    double f64;
+};
+
+/// Result of execution of a function.
+typedef struct FizzyExecutionResult
+{
+    /// Whether execution ended with a trap.
+    bool trapped;
+    /// Whether function returned a value. Valid only if trapped == false
+    bool has_value;
+    /// Value returned from a function. Valid only if has_value == true
+    union FizzyValue value;
+} FizzyExecutionResult;
+
+
+/// Pointer to external function.
+///
+/// @param context      Opaque pointer to execution context.
+/// @param instance     Pointer to module instance.
+/// @param args         Pointer to the argument array.
+/// @param args_size   Size of the  argument array.
+/// @param depth        Call stack depth.
+typedef struct FizzyExecutionResult (*FizzyExternalFn)(void* context,
+    struct FizzyInstance* instance, const union FizzyValue* args, size_t args_size, int depth);
+
+/// External function.
+typedef struct FizzyExternalFunction
+{
+    // TODO function type
+
+    /// Pointer to function.
+    FizzyExternalFn function;
+    /// Opaque pointer to execution context, that will be passed to function.
+    void* context;
+} FizzyExternalFunction;
+
+/// Validate binary module.
 bool fizzy_validate(const uint8_t* wasm_binary, size_t wasm_binary_size);
+
+/// Parse binary module.
+struct FizzyModule* fizzy_parse(const uint8_t* wasm_binary, size_t wasm_binary_size);
+
+/// Free resources associated with the module.
+///
+/// Should be called unless @p module was passed to fizzy_instantiate.
+void fizzy_free_module(struct FizzyModule* module);
+
+/// Instantiate a module.
+/// Takes ownership of module, i.e. @p module is invalidated after this call.
+///
+/// @param module                   Pointer to module.
+/// @param imported_functions       Pointer to the imported function array.
+/// @param imported_functions_size  Size of the imported function array.
+struct FizzyInstance* fizzy_instantiate(struct FizzyModule* module,
+    const struct FizzyExternalFunction* imported_functions, size_t imported_functions_size);
+
+/// Free resources associated with the instance.
+void fizzy_free_instance(struct FizzyInstance* instance);
+
+/// Execute module function.
+///
+/// @param instance     Pointer to module instance.
+/// @param args         Pointer to the argument array.
+/// @param depth        Call stack depth.
+FizzyExecutionResult fizzy_execute(
+    struct FizzyInstance* instance, uint32_t func_idx, const union FizzyValue* args, int depth);
 
 #ifdef __cplusplus
 }
--- a/lib/fizzy/capi.cpp
+++ b/lib/fizzy/capi.cpp
@@ -2,10 +2,76 @@
 // Copyright 2020 The Fizzy Authors.
 // SPDX-License-Identifier: Apache-2.0
 
+#include "cxx20/bit.hpp"
+#include "execute.hpp"
+#include "instantiate.hpp"
 #include "parser.hpp"
 #include <fizzy/fizzy.h>
+#include <memory>
+
+namespace
+{
+inline FizzyValue wrap(fizzy::Value value) noexcept
+{
+    return fizzy::bit_cast<FizzyValue>(value);
+}
+
+inline fizzy::Value unwrap(FizzyValue value) noexcept
+{
+    return fizzy::bit_cast<fizzy::Value>(value);
+}
+
+inline const FizzyValue* wrap(const fizzy::Value* values) noexcept
+{
+    return reinterpret_cast<const FizzyValue*>(values);
+}
+
+inline const fizzy::Value* unwrap(const FizzyValue* values) noexcept
+{
+    return reinterpret_cast<const fizzy::Value*>(values);
+}
+
+inline FizzyInstance* wrap(fizzy::Instance* instance) noexcept
+{
+    return reinterpret_cast<FizzyInstance*>(instance);
+}
+
+inline fizzy::Instance* unwrap(FizzyInstance* instance) noexcept
+{
+    return reinterpret_cast<fizzy::Instance*>(instance);
+}
+
+inline FizzyExecutionResult wrap(const fizzy::ExecutionResult& result) noexcept
+{
+    return {result.trapped, result.has_value, wrap(result.value)};
+}
+
+inline fizzy::ExecutionResult unwrap(FizzyExecutionResult result) noexcept
+{
+    if (result.trapped)
+        return fizzy::Trap;
+    else if (!result.has_value)
+        return fizzy::Void;
+    else
+        return unwrap(result.value);
+}
+
+inline auto unwrap(FizzyExternalFn func, void* context) noexcept
+{
+    return [func, context](fizzy::Instance& instance, fizzy::span<const fizzy::Value> args,
+               int depth) noexcept -> fizzy::ExecutionResult {
+        const auto result = func(context, wrap(&instance), wrap(args.data()), args.size(), depth);
+        return unwrap(result);
+    };
+}
+}  // namespace
 
 extern "C" {
+struct FizzyModule
+{
+    fizzy::Module module;
+};
+
 bool fizzy_validate(const uint8_t* wasm_binary, size_t wasm_binary_size)
 {
     try
@@ -17,5 +83,68 @@ bool fizzy_validate(const uint8_t* wasm_binary, size_t wasm_binary_size)
     {
         return false;
     }
+}
+
+FizzyModule* fizzy_parse(const uint8_t* wasm_binary, size_t wasm_binary_size)
+{
+    try
+    {
+        auto cmodule = std::make_unique<FizzyModule>();
+        cmodule->module = fizzy::parse({wasm_binary, wasm_binary_size});
+        return cmodule.release();
+    }
+    catch (...)
+    {
+        return nullptr;
+    }
+}
+
+void fizzy_free_module(FizzyModule* module)
+{
+    delete module;
+}
+
+FizzyInstance* fizzy_instantiate(FizzyModule* module,
+    const FizzyExternalFunction* imported_functions, size_t imported_functions_size)
+{
+    try
+    {
+        // Convert fizzy_external_function to fizzy::ExternalFunction
+        std::vector<fizzy::ExternalFunction> functions(imported_functions_size);
+        for (size_t imported_func_idx = 0; imported_func_idx < imported_functions_size;
+             ++imported_func_idx)
+        {
+            const auto& cfunc = imported_functions[imported_func_idx];
+
+            auto func = unwrap(cfunc.function, cfunc.context);
+            // TODO get type from input array
+            auto func_type = module->module.imported_function_types[imported_func_idx];
+
+            functions[imported_func_idx] =
+                fizzy::ExternalFunction{std::move(func), std::move(func_type)};
+        }
+
+        auto instance = fizzy::instantiate(std::move(module->module), std::move(functions));
+
+        fizzy_free_module(module);
+        return wrap(instance.release());
+    }
+    catch (...)
+    {
+        fizzy_free_module(module);
+        return nullptr;
+    }
+}
+
+void fizzy_free_instance(FizzyInstance* instance)
+{
+    delete unwrap(instance);
+}
+
+FizzyExecutionResult fizzy_execute(
+    FizzyInstance* instance, uint32_t func_idx, const FizzyValue* args, int depth)
+{
+    const auto result = fizzy::execute(*unwrap(instance), func_idx, unwrap(args), depth);
+    return wrap(result);
 }
 }
--- a/test/compilation/compilation_test.c
+++ b/test/compilation/compilation_test.c
@@ -4,9 +4,86 @@
 
 #include <fizzy/fizzy.h>
 
-bool dummy(void);
+bool validate(const uint8_t* binary, size_t size);
+bool parse(const uint8_t* binary, size_t size);
+bool instantiate(const uint8_t* binary, size_t size);
+struct FizzyExecutionResult dummy_host_func(void* context, struct FizzyInstance* instance,
+    const union FizzyValue* args, size_t args_size, int depth);
+bool instantiate_with_host_func(const uint8_t* binary, size_t size);
+bool execute(const uint8_t* binary, size_t size);
 
-bool dummy()
+bool validate(const uint8_t* binary, size_t size)
 {
-    return fizzy_validate(NULL, 0);
+    return fizzy_validate(binary, size);
+}
+
+bool parse(const uint8_t* binary, size_t size)
+{
+    struct FizzyModule* module = fizzy_parse(binary, size);
+    if (!module)
+        return false;
+
+    fizzy_free_module(module);
+    return true;
+}
+
+bool instantiate(const uint8_t* binary, size_t size)
+{
+    struct FizzyModule* module = fizzy_parse(binary, size);
+    if (!module)
+        return false;
+
+    struct FizzyInstance* instance = fizzy_instantiate(module, NULL, 0);
+    if (!instance)
+        return false;
+
+    fizzy_free_instance(instance);
+    return true;
+}
+
+struct FizzyExecutionResult dummy_host_func(void* context, struct FizzyInstance* instance,
+    const union FizzyValue* args, size_t args_size, int depth)
+{
+    (void)context;
+    (void)instance;
+    (void)args;
+    (void)args_size;
+    (void)depth;
+    struct FizzyExecutionResult res = {true, false, {0}};
+    return res;
+}
+
+bool instantiate_with_host_func(const uint8_t* binary, size_t size)
+{
+    struct FizzyModule* module = fizzy_parse(binary, size);
+    if (!module)
+        return false;
+
+    FizzyExternalFunction host_funcs[] = {{dummy_host_func, NULL}};
+
+    struct FizzyInstance* instance = fizzy_instantiate(module, host_funcs, 1);
+    if (!instance)
+        return false;
+
+    fizzy_free_instance(instance);
+    return true;
+}
+
+bool execute(const uint8_t* binary, size_t size)
+{
+    struct FizzyModule* module = fizzy_parse(binary, size);
+    if (!module)
+        return false;
+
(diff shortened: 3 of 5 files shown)


lib/fizzy/capi.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+        auto cinstance = std::make_unique<fizzy_instance>();
+        cinstance->instance = instance.release();
+
+        fizzy_free_module(module);
+        return cinstance.release();
+    }
+    catch (...)
+    {
+        return nullptr;
```
axic, line 109: This is the annoying part, if this function consumes `module`, I think it should consume it in every case. This means `free_module` should be called here.

test/unittests/capi_test.cpp
```diff
@@ -0,0 +1,157 @@
...
+    auto module = fizzy_parse(wasm_prefix, sizeof(wasm_prefix));
+    EXPECT_TRUE(module);
+    fizzy_free_module(module);
+}
+
+TEST(capi, instantiate)
+{
+    uint8_t wasm_prefix[]{0x00, 0x61, 0x73, 0x6d, 0x01, 0x00, 0x00, 0x00};
+    auto module = fizzy_parse(wasm_prefix, sizeof(wasm_prefix));
+
```
axic, line 31: Should add expect for module.

test/unittests/capi_test.cpp
```diff
@@ -0,0 +1,157 @@
...
+    EXPECT_TRUE(fizzy_validate(wasm_prefix, sizeof(wasm_prefix)));
+    wasm_prefix[7] = 1;
+    EXPECT_FALSE(fizzy_validate(wasm_prefix, sizeof(wasm_prefix)));
+}
+
+TEST(capi, parse)
+{
+    uint8_t wasm_prefix[]{0x00, 0x61, 0x73, 0x6d, 0x01, 0x00, 0x00, 0x00};
+    auto module = fizzy_parse(wasm_prefix, sizeof(wasm_prefix));
+    EXPECT_TRUE(module);
```
axic, line 23: I'd prefer `EQ` or `NE` with `nullptr` for pointers for clarity.

lib/fizzy/capi.cpp
```diff
@@ -0,0 +1,129 @@
...
+    }
+}
+
+void fizzy_free_instance(fizzy_instance* instance)
+{
+    delete instance->instance;
+    delete instance;
+}
+
+fizzy_execution_result fizzy_execute(fizzy_instance* instance, uint32_t func_idx,
```
axic, line 119: I don't think depth should be part of the public API, @chfast ?

chfast: I think it must be. My plan is to convert it to a more generic opaque "ThreadExecutionContext". This is a sneak peak: https://github.com/wasmx/fizzy/pull/529/commits/5f08ee277a7959252b9dfd5763ce66dfc6b89b5d.

Of course this is up to discussion.

lib/fizzy/capi.cpp
gumb0: Yeah, agree. It's a bit weird that function modifies it's argument in case of failure, but it reflects the way it is currently in C++ API: `fizzy::instantiate(std::move(module))` would leave `module` empty even if it fails.

On the other hand, in C++ the caller has the freedom to copy the module before instantiate if they want to. Maybe we should add `fizzy_copy_module` to make this an option in C API, too.

lib/fizzy/capi.cpp
gumb0: Yeah, we discussed it a bit when introducing `depth` in C++ API. My argument for it is that without it host functions don't have an option to respect the call depth limit (calls to imported functions would reset depth to 0). For example, this test would fail (go into infinite recursion): https://github.com/wasmx/fizzy/blob/84abe560eaf30ccb1a1164e3290d8effdba8ad24/test/unittests/execute_call_test.cpp#L506

lib/fizzy/capi.cpp
axic: Yeah I thought about duplication too.

axic:
@gumb0 can you rebase this on #530 ?

lib/fizzy/capi.cpp
```diff
@@ -2,8 +2,25 @@
...
 #include "parser.hpp"
 #include <fizzy/fizzy.h>
+#include <memory>
+
+namespace
+{
+// TODO replace with the version from cxx/bit.h
+/// Simple implementation of C++20's std::bit_cast.
+template <typename DstT, typename SrcT>
+DstT bit_cast(SrcT x) noexcept
```
axic, line 16: This is not needed anymore here.

axic:
We need to think about a similar feature as "resolve imports" in C++. wabt/wasm3 have something called"add host function".

include/fizzy/fizzy.h
```diff
@@ -11,8 +11,87 @@
 extern "C" {
 #endif
 
+/// The opaque data type representing a module.
+struct fizzy_module;
```
chfast, line 15: I'd prefer to name types as `FizzyModule`. In any case, it is good to make the decision now.

lib/fizzy/capi.cpp
```diff
@@ -18,4 +32,88 @@ bool fizzy_validate(const uint8_t* wasm_binary, size_t wasm_binary_size)
...
+            // TODO get type from input array
+            auto func_type = module->module.imported_function_types[imported_func_idx];
+
+            functions[imported_func_idx] =
+                fizzy::ExternalFunction{std::move(func), std::move(func_type)};
+        }
+
+        auto instance = fizzy::instantiate(std::move(module->module), std::move(functions));
+
+        auto cinstance = std::make_unique<fizzy_instance>();
```
chfast, line 90: In case C++ function returns a pointer, it will be better to just convert it to opaque type pointer.
`reinterpret_cast<fizzy_instance*>(instance.release())`. This is correct, although looks ugly. You can additionally add `wrap`/`unwrap` functions to have ugliness under control.

```
fizzy_instance* wrap(Instance* x) { return reinterpret_cast<fizzy_instance*>(x); }
Instance* unwrap(fizzy_instance* x) { return reinterpret_cast<instance*>(x); }
```
Or alternatively,

```
fizzy_instance* wrap(std::unique_ptr<Instance> x) { return reinterpret_cast<fizzy_instance*>(x.release()); }
std::unique_ptr<Instance> unwrap(fizzy_instance* x) { return {reinterpret_cast<instance*>(x)}; }
```

The second one looks better to me as it is clear that after `unwrap()` you own the Instance.

Obviously, the `Module` is an exception here, so I'm ok with inspecting it some other day.

lib/fizzy/capi.cpp
```diff
@@ -18,4 +32,88 @@ bool fizzy_validate(const uint8_t* wasm_binary, size_t wasm_binary_size)
...
+    catch (...)
+    {
+        fizzy_free_module(module);
+        return nullptr;
+    }
+}
+
+void fizzy_free_instance(fizzy_instance* instance)
+{
+    delete instance->instance;
```
chfast, line 105: Then the whole function becomes.

```suggestion
    unwrap(instance);
```

test/compilation/compilation_test.c
```diff
@@ -4,9 +4,86 @@
 
 #include <fizzy/fizzy.h>
 
-bool dummy(void);
+bool validate(const uint8_t* binary, size_t size);
```
chfast, line 7: For the future, the "compilation test" is to check if the code in the header is valid `C99`, `C11`. The is no need to add code to it.

test/unittests/capi_test.cpp
```diff
@@ -12,3 +16,224 @@ TEST(capi, validate)
     wasm_prefix[7] = 1;
     EXPECT_FALSE(fizzy_validate(wasm_prefix, sizeof(wasm_prefix)));
 }
+
+TEST(capi, parse)
+{
+    uint8_t wasm_prefix[]{0x00, 0x61, 0x73, 0x6d, 0x01, 0x00, 0x00, 0x00};
+    auto module = fizzy_parse(wasm_prefix, sizeof(wasm_prefix));
```
chfast, line 23:
```suggestion
    auto module = fizzy_parse(wasm_prefix, std::size(wasm_prefix));
```

test/unittests/capi_test.cpp
```diff
@@ -12,3 +16,224 @@ TEST(capi, validate)
...
+    const auto wasm = from_hex("0061736d010000000105016000017f020d01046d6f643104666f6f310000");
+    auto module = fizzy_parse(wasm.data(), wasm.size());
+    EXPECT_NE(module, nullptr);
+
+    EXPECT_EQ(fizzy_instantiate(module, nullptr, 0), nullptr);
+
+    module = fizzy_parse(wasm.data(), wasm.size());
+    EXPECT_NE(module, nullptr);
+
+    fizzy_external_function host_funcs[] = {
```
chfast, line 56: I think this would be more readable:
```suggestion
    FizzyExternalFunction host_funcs[] = {
```

include/fizzy/fizzy.h
```diff
@@ -11,8 +11,87 @@
...
+
+/// Pointer to external function.
+///
+/// @param context      Opaque pointer to execution context.
+/// @param instance     Pointer to module instance.
+/// @param args         Pointer to the argument array.
+/// @param args_size   Size of the  argument array.
+/// @param depth        Call stack depth.
+typedef struct fizzy_execution_result (*fizzy_external_fn)(void* context,
+    struct fizzy_instance* instance, const union fizzy_value* args, uint32_t args_size, int depth);
```
chfast, line 50: The `uint32_t` will be annoying to use, let's use `size_t` for array sizes for now (as in `fizzy_validate`, `fizzy_parse`, etc).

```suggestion
    struct fizzy_instance* instance, const union fizzy_value* args, size_t args_size, int depth);
```

test/compilation/compilation_test.c
gumb0: I know, but I can imagine the case when the header compiles, but then calling the function doesn't compile in C like you imagine it would.

test/unittests/capi_test.cpp
```diff
@@ -12,3 +16,224 @@ TEST(capi, validate)
...
+      )
+    */
+    const auto wasm =
+        from_hex("0061736d01000000010401600000020901016d03666f6f0000030201000a0601040010000b");
+    auto module = fizzy_parse(wasm.data(), wasm.size());
+    EXPECT_NE(module, nullptr);
+
+    bool called = false;
+    fizzy_external_function host_funcs[] = {
+        {[](void* context, fizzy_instance*, const fizzy_value*, uint32_t, int) {
```
chfast, line 172: What is this `context` doing?

test/compilation/compilation_test.c
chfast: There are unittests for that.

lib/fizzy/capi.cpp
gumb0: Yeah this is better. For some reason I thought having `Instance&` in other parts of C++ API would make this not possible, but now I think it should work.

test/compilation/compilation_test.c
gumb0: The ones compiled in `C99` and `C11`?

lib/fizzy/capi.cpp
```diff
@@ -9,6 +9,20 @@
 #include <fizzy/fizzy.h>
 #include <memory>
 
+namespace
+{
+fizzy_instance* wrap(std::unique_ptr<fizzy::Instance> x)
```
chfast, line 14: I think it is better not to hake `unique_ptr` because the `unwrap` cannot return one.

test/unittests/capi_test.cpp
gumb0: This is some opaque context from the host side that host can set during instantiate.

Here's an example where it's needed - calling a function from another instance:
https://github.com/wasmx/fizzy/blob/4cce6ee6a7a66b29cbc107808851454f30981c1e/test/unittests/capi_test.cpp#L226

(However, in this example called function's index is hardcoded, I expect it's going to be inconvenient in case you would need to pass both instance and function index via one `context` pointer - say if you want a single general host function dispatching the calls to another instance)

include/fizzy/fizzy.h
gumb0: Yeah, I'm thinking now, there's not much sense in having very different naming in C++ and C.

test/unittests/capi_test.cpp
gumb0: Why is this better?

include/fizzy/fizzy.h
```diff
@@ -11,7 +11,86 @@
...
+    double f64;
+};
+
+/// Result of execution of a function.
+typedef struct FizzyExecutionResult
+{
+    /// Whether execution ended with a trap.
+    bool trapped;
+    /// Whether function returned a value. Valid only if trapped == false
+    bool has_value;
```
gumb0, line 36: `hasValue`?

include/fizzy/fizzy.h
```diff
@@ -11,7 +11,86 @@
...
+    // TODO function type
+
+    /// Pointer to function.
+    FizzyExternalFn function;
+    /// Opaque pointer to execution context, that will be passed to function.
+    void* context;
+} FizzyExternalFunction;
+
+/// Validate binary module.
+bool FizzyValidate(const uint8_t* wasm_binary, size_t wasm_binary_size);
```
gumb0, line 64: ?
```suggestion
bool FizzyValidate(const uint8_t* wasmBinary, size_t wasmBinarySize);
```

test/unittests/capi_test.cpp
chfast: Counts number of elements, not number of bytes. Here it is the same, so it does not matter.

include/fizzy/fizzy.h
chfast: My preference is `CamelCase` for types, `snake_case` for functions and variables, as in C++.

lib/fizzy/capi.cpp
gumb0: > This is the annoying part, if this function consumes `module`, I think it should consume it in every case. This means `free_module` should be called here.

Another option could be: `instantiate` invalidates `module`, but user is still required to call `fizzy_free_module` in the end.

lib/fizzy/capi.cpp
```diff
@@ -2,10 +2,81 @@
...
+{
+    if (result.trapped)
+        return fizzy::Trap;
+    else if (!result.has_value)
+        return fizzy::Void;
+    else
+        return unwrap(result.value);
+}
+
+inline auto unwrap(FizzyExternalFn func, void* context) noexcept
```
gumb0, line 59: Not sure this is a good idea, because 1. corresponding `wrap` would not be possible or useful; 2. implemetation actualy _wraps_ C function with lambda.

Maybe another name for this?

lib/fizzy/capi.cpp
```diff
@@ -18,4 +89,68 @@ bool fizzy_validate(const uint8_t* wasm_binary, size_t wasm_binary_size)
...
+void fizzy_free_instance(FizzyInstance* instance)
+{
+    delete unwrap(instance);
+}
+
+FizzyExecutionResult fizzy_execute(
+    FizzyInstance* instance, uint32_t func_idx, const FizzyValue* args, size_t args_size, int depth)
+{
+    const auto result =
+        fizzy::execute(*unwrap(instance), func_idx, {unwrap(args), args_size}, depth);
```
gumb0, line 153: Also it would be possible to define 
```
fizzt::span<const Value> unwrap(const FizzyValue* args, size_t size)
```
but corresponding `unwrap` would not be convenient to use.

include/fizzy/fizzy.h
```diff
@@ -11,8 +11,87 @@
...
+
+/// Validate binary module.
 bool fizzy_validate(const uint8_t* wasm_binary, size_t wasm_binary_size);
 
+/// Parse binary module.
+struct FizzyModule* fizzy_parse(const uint8_t* wasm_binary, size_t wasm_binary_size);
+
+/// Free resources associated with the module.
+///
+/// Should be called only unless @p module was passed to fizzy_instantiate.
```
axic, line 71: Opt 1:
```suggestion
/// Should be called unless @p module was passed to fizzy_instantiate.
```

Opt 2:
```suggestion
/// Should be called only if @p module was passed to fizzy_instantiate.
```

include/fizzy/fizzy.h
```diff
@@ -11,8 +11,87 @@
...
+/// Free resources associated with the instance.
+void fizzy_free_instance(struct FizzyInstance* instance);
+
+/// Execute module function.
+///
+/// @param instance     Pointer to module instance.
+/// @param args         Pointer to the argument array.
+/// @param args_size    Size of the argument array.
+/// @param depth        Call stack depth.
+FizzyExecutionResult fizzy_execute(struct FizzyInstance* instance, uint32_t func_idx,
```
axic, line 92: Why is depth signed?

lib/fizzy/capi.cpp
axic: `encapsulate`?

include/fizzy/fizzy.h
gumb0: In C++ we chose it because it's not defined in the spec, `int` is a natural type for value involving some arithmetics. Also UBSan can detect overflows for it in tests.

It might be more natural / user-friendly to make it unsigned ot size_t in C.

include/fizzy/fizzy.h
axic: I'd rather not.

axic:
This needs to add something like `find_function` from `utils/fizzy_engine.cpp`.

include/fizzy/fizzy.h
axic: > My preference is `CamelCase` for types, `snake_case` for functions and variables, as in C++.

Okay with that for now, Rust follows the same too, at least a lot of the code will be consistent.

include/fizzy/fizzy.h
```diff
@@ -11,8 +11,87 @@
...
+    /// Whether function returned a value. Valid only if trapped == false
+    bool has_value;
+    /// Value returned from a function. Valid only if has_value == true
+    union FizzyValue value;
+} FizzyExecutionResult;
+
+
+/// Pointer to external function.
+///
+/// @param context      Opaque pointer to execution context.
```
chfast, line 44:
```suggestion
/// @param context     Opaque pointer to execution context.
```

lib/fizzy/capi.cpp
```diff
@@ -2,10 +2,81 @@
...
+}
+}  // namespace
 
 extern "C" {
+struct FizzyModule
+{
+    fizzy::Module module;
+};
+
+struct FizzyInstance
```
chfast, line 75: This should not be needed.

include/fizzy/fizzy.h
```diff
@@ -11,8 +11,87 @@
...
+/// @param instance     Pointer to module instance.
+/// @param args         Pointer to the argument array.
```
axic, line 46: Can args be null?

include/fizzy/fizzy.h
```diff
@@ -11,8 +11,87 @@
...
+/// @param args_size   Size of the  argument array.
```
axic, line 47:
```suggestion
/// @param args_size    Size of the  argument array.
```

include/fizzy/fizzy.h
```diff
@@ -11,8 +11,87 @@
...
+struct FizzyInstance* fizzy_instantiate(struct FizzyModule* module,
+    const struct FizzyExternalFunction* imported_functions, size_t imported_functions_size);
+
+/// Free resources associated with the instance.
+void fizzy_free_instance(struct FizzyInstance* instance);
+
+/// Execute module function.
+///
+/// @param instance     Pointer to module instance.
+/// @param args         Pointer to the argument array.
```
axic, line 89: Can args be null?

lib/fizzy/capi.cpp
```diff
@@ -2,10 +2,81 @@
...
+        const auto result = func(context, wrap(&instance), wrap(args.data()), args.size(), depth);
+        return unwrap(result);
+    };
+}
+}  // namespace
 
 extern "C" {
+struct FizzyModule
+{
+    fizzy::Module module;
```
gumb0, line 72: I realize now, that this is not going to work for the case when we want to access already instantiated module, i.e. get `FizzyModule*` from `FizzyInstance*` which owns it.

Module has to be dynamically allocated on the C++ side, then on C side module and instance will be represented similarly.

gumb0:
Closing in favor of #576

PR #535, opened 2020-09-20
Improve and expose bit_cast implementation

chfast:
Improve the bit_cast implementation to be as close as the C++20 spec.
Move it to lib/fizzy/cxx20/bit.hpp to match the <bit> header in C++20
and be available in the fizzy library.

### DIFF
--- a/lib/fizzy/CMakeLists.txt
+++ b/lib/fizzy/CMakeLists.txt
@@ -7,6 +7,7 @@ add_library(fizzy::fizzy ALIAS fizzy)
 
 target_sources(
     fizzy PRIVATE
+    cxx20/bit.hpp
     bytes.hpp
     constexpr_vector.hpp
     exceptions.cpp
--- a/lib/fizzy/cxx20/bit.hpp
+++ b/lib/fizzy/cxx20/bit.hpp
@@ -1,0 +1,26 @@
+// Fizzy: A fast WebAssembly interpreter
+// Copyright 2020 The Fizzy Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <type_traits>
+
+namespace fizzy
+{
+/// The non-constexpr implementation of C++20's std::bit_cast.
+/// See https://en.cppreference.com/w/cpp/numeric/bit_cast.
+template <class To, class From>
+[[nodiscard]] inline
+    typename std::enable_if_t<sizeof(To) == sizeof(From) && std::is_trivially_copyable_v<From> &&
+                                  std::is_trivially_copyable_v<To>,
+        To>
+    bit_cast(const From& src) noexcept
+{
+    static_assert(std::is_trivially_constructible_v<To>);  // Additional, non-standard requirement.
+
+    To dst;
+    __builtin_memcpy(&dst, &src, sizeof(To));
+    return dst;
+}
+}  // namespace fizzy
--- a/test/unittests/bit_cast_test.cpp
+++ b/test/unittests/bit_cast_test.cpp
@@ -1,0 +1,46 @@
+// Fizzy: A fast WebAssembly interpreter
+// Copyright 2020 The Fizzy Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "cxx20/bit.hpp"
+#include <gtest/gtest.h>
+#include <array>
+
+using namespace fizzy;
+using namespace testing;
+
+
+TEST(bit_cast, double_to_uint64)
+{
+    // A test case from https://en.cppreference.com/w/cpp/numeric/bit_cast#Example.
+    EXPECT_EQ(bit_cast<uint64_t>(19880124.0), 0x4172f58bc0000000);
+}
+
+TEST(bit_cast, uint64_to_double)
+{
+    // A test case from https://en.cppreference.com/w/cpp/numeric/bit_cast#Example.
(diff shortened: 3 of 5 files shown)


lib/fizzy/cxx20/bit.hpp
```diff
@@ -0,0 +1,26 @@
...
+namespace fizzy
+{
+/// The non-constexpr implementation of C++20's std::bit_cast.
+/// See https://en.cppreference.com/w/cpp/numeric/bit_cast.
+template <class To, class From>
+[[nodiscard]] inline
+    typename std::enable_if_t<sizeof(To) == sizeof(From) && std::is_trivially_copyable_v<From> &&
+                                  std::is_trivially_copyable_v<To>,
+        To>
+    bit_cast(const From& src) noexcept
```
axic, line 18: Are you planning to use this outside of the tests?

chfast: Yes.

chfast: In C API and execute() implementation.

lib/fizzy/CMakeLists.txt
```diff
@@ -7,6 +7,7 @@ add_library(fizzy::fizzy ALIAS fizzy)
 
 target_sources(
     fizzy PRIVATE
+    cxx20/bit.hpp
```
axic, line 10: Are we going to move `span` here too?

test/unittests/bit_cast_test.cpp
```diff
@@ -0,0 +1,21 @@
...
+// Copyright 2020 The Fizzy Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "cxx20/bit.hpp"
+#include <gtest/gtest.h>
+
+using namespace fizzy;
+using namespace testing;
+
+// Test cases from https://en.cppreference.com/w/cpp/numeric/bit_cast#Example.
```
axic, line 11: Is it supposed to work between different signs, e.g uint64<->int64? If so, can you add an example? If not can you add a `static_assert` in the impl?

lib/fizzy/cxx20/bit.hpp
```diff
@@ -0,0 +1,26 @@
...
+{
+    static_assert(std::is_trivially_constructible_v<To>);  // Additional, non-standard requirement.
```
axic, line 20: https://en.cppreference.com/w/cpp/numeric/bit_cast#Example has actually this line, do you mean that the final standard does not require it? I guess it has no issues as long as we keep compiling with <C++20 any such usage will be detected in our codebase, if we want to avoid such usage.

chfast: Yes. The C++20 version only requires the conditions from `std::enable_if`. Personally, I would just use `static_assert` instead of `std::enable_if`, but if C++20 defines it like that let it be. I would also omit the `is_trivially_constructible_v` check, but being more pedantic is ok I guess.

test/unittests/bit_cast_test.cpp
chfast: Yes, it should work for types of different signness. Also for arrays, structs and other C-like types.

lib/fizzy/CMakeLists.txt
chfast: Yes. We briefly discussed that on Telegram with @gumb0 some time ago.

test/unittests/bit_cast_test.cpp
```diff
@@ -0,0 +1,39 @@
...
+    EXPECT_EQ(bit_cast<uint64_t>(19880124.0), 0x4172f58bc0000000);
+}
+
+TEST(bit_cast, uint64_to_double)
+{
+    // A test case from https://en.cppreference.com/w/cpp/numeric/bit_cast#Example.
+    EXPECT_EQ(bit_cast<double>(uint64_t{0x3fe9000000000000}), 0.781250);
+}
+
+TEST(bit_cast, uint32_to_int32)
```
axic, line 25: Can you also add `int32_to_uint32`?

axic (approved):

test/unittests/bit_cast_test.cpp
```diff
@@ -0,0 +1,45 @@
...
+TEST(bit_cast, int32_to_uint32)
+{
+    EXPECT_EQ(bit_cast<uint32_t>(int32_t{-2}), 0xfffffffe);
+    EXPECT_EQ(bit_cast<uint32_t>(int32_t{1}), 1);
+}
+
+TEST(bit_cast, uint32_to_array)
+{
+    std::array<uint8_t, 4> bytes;
+    bytes = bit_cast<decltype(bytes)>(uint32_t{0xaabbbbaa});
```
gumb0, line 40: could be non-symmetric to see the endianness

gumb0 (approved):

test/unittests/bit_cast_test.cpp
chfast: It's symmetric to avoid endianness issues.

lib/fizzy/cxx20/bit.hpp
```diff
@@ -0,0 +1,26 @@
+// Fizzy: A fast WebAssembly interpreter
+// Copyright 2020 The Fizzy Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <type_traits>
+
+namespace fizzy
```
axic, line 9: Oh actually could have added the C++20 check as in `span.hpp`, to test this with a C++20 compiler.

PR #537, opened 2020-09-21
Run shellcheck

axic:
Uses https://github.com/koalaman/shellcheck.

Should consider just installing shellcheck and not using the orb.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -216,6 +216,10 @@ jobs:
     steps:
       - checkout
       - run:
+          name: "Run shellcheck"
+          command: |
+            git ls-files '*.sh' | xargs shellcheck
+      - run:
           name: "Check code format"
           command: |
             clang-format --version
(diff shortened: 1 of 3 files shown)


circle.yml
```diff
@@ -1,5 +1,8 @@
 version: 2.1
 
+orbs:
+  shellcheck: circleci/shellcheck@1.3.16
```
axic, line 4: This is actually just a pip package, we should include it in `cpp-build-env` and make it part of the lint step.

chfast (approved):

circle.yml
chfast: Not a pip package, but added to cpp-build-env:15-lint and cpp-build-env:15-clang-10.

circle.yml
```diff
@@ -215,6 +215,10 @@ jobs:
     executor: linux-clang-latest
     steps:
       - checkout
+      - run:
+          name: "Run shellcheck"
+          command: |
+            git ls-files '*.sh' | xargs shellcheck --external-sources
```
chfast, line 221: Will this fail on error?

axic: Yes, I checked it with removing one quote from `apply_patches.sh`.

chfast (approved):

PR #538, opened 2020-09-21
Rust bindings (validate only)

axic:
Depends on #530. Depends on #533.

- [x] Change bumpversion
- [x] Add CI for `cargo build` and `cargo test`
- [x] Add CI for `cargo fmt`
- [ ] Change readme

### DIFF
--- a/bindings/rust/build.rs
+++ b/bindings/rust/build.rs
@@ -1,0 +1,59 @@
+// Fizzy: A fast WebAssembly interpreter
+// Copyright 2019-2020 The Fizzy Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+extern crate bindgen;
+extern crate cmake;
+
+use cmake::Config;
+
+use std::env;
+use std::path::PathBuf;
+
+fn main() {
+    // This is the root directory.
+    let src = "../../";
+
+    let dst = Config::new(src).define("FIZZY_TESTING", "OFF").build();
+
+    println!("cargo:rustc-link-lib=static=fizzy");
+    println!("cargo:rustc-link-search=native={}/lib", dst.display());
+
+    // We need to link against C++ std lib
+    if let Some(cpp_stdlib) = get_cpp_stdlib() {
+        println!("cargo:rustc-link-lib={}", cpp_stdlib);
+    }
+
+    let bindings = bindgen::Builder::default()
+        .header(format!("{}/include/fizzy/fizzy.h", src))
+        // See https://github.com/rust-lang-nursery/rust-bindgen/issues/947
+        .trust_clang_mangling(false)
+        .generate_comments(true)
+        // https://github.com/rust-lang-nursery/rust-bindgen/issues/947#issuecomment-327100002
+        .layout_tests(false)
+        .whitelist_function("fizzy_.*")
+        // TODO: consider removing this
+        .size_t_is_usize(true)
+        .generate()
+        .expect("Unable to generate bindings");
+
+    let out_path = PathBuf::from(env::var("OUT_DIR").unwrap());
+    bindings
+        .write_to_file(out_path.join("bindings.rs"))
+        .expect("Could not write bindings");
+}
+
+// See https://github.com/alexcrichton/gcc-rs/blob/88ac58e25/src/lib.rs#L1197
+fn get_cpp_stdlib() -> Option<String> {
+    env::var("TARGET").ok().and_then(|target| {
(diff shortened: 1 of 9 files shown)


circle.yml
```diff
@@ -512,6 +512,51 @@ jobs:
...
+    docker:
+      - image: rust:1-buster
+    steps:
+      - checkout
+      - run:
+          name: Update environment
+          command: |
+            echo 'deb http://deb.debian.org/debian buster-backports main' >> /etc/apt/sources.list
+            apt -qq update
+            apt -yq install libclang-dev clang cmake --no-install-recommends
```
axic, line 525: Even though I add buster-backports, this still installs cmake 3.13 from buster, while buster-backports has 3.16: https://packages.debian.org/buster-backports/cmake

axic: Fixed.

bindings/rust/build.rs
```diff
@@ -0,0 +1,64 @@
...
+use std::env;
+use std::path::PathBuf;
+
+fn main() {
+    // This is the root directory.
+    let src = "../../";
+
+    let dst = Config::new(src)
+        .define("FIZZY_TESTING", "OFF")
+        .define("FIZZY_FUZZING", "OFF")
```
chfast, line 19: The `FIZZY_FUZZING` depends on `FIZZY_TESTING` so not needed here.

bindings/rust/src/lib.rs
```diff
@@ -0,0 +1,27 @@
+// Fizzy: A fast WebAssembly interpreter
```
chfast, line 1: Still surprises me that naming a file `lib.rs` is considered good practice.

bindings/rust/src/lib.rs
```diff
@@ -0,0 +1,27 @@
...
+    fn validate_wasm() {
+        assert_eq!(
+            validate(&[0x00, 0x61, 0x73, 0x6d, 0x01, 0x00, 0x00, 0x00]),
+            true
+        );
+        assert_eq!(
+            validate(&[0x00, 0x61, 0x73, 0x6d, 0x01, 0x00, 0x00, 0x01]),
+            false
+        );
+        assert_eq!(validate(&[0x00]), false);
```
chfast, line 25: Add test with empty input.

circle.yml
chfast: Why are you installing cmake twice? Also: alpine 3.12 has cmake 3.17.

circle.yml
```diff
@@ -512,6 +512,53 @@ jobs:
...
+            rustfmt --version
+            cargo fmt --all -- --check
+      - run:
+          name: Build
+          command: cargo build
+      - run:
+          name: Test
+          command: cargo test
+
+  bindings-rust-asan-combined:
```
chfast, line 541: What's "combined" about this?

circle.yml
```diff
@@ -512,6 +512,53 @@ jobs:
...
+      - checkout
+      - run:
+          name: Update environment
+          command: |
+            echo 'deb http://deb.debian.org/debian buster-backports main' >> /etc/apt/sources.list
+            apt -qq update
+            apt -yq -t buster-backports install cmake --no-install-recommends
+            apt -yq install libclang-dev clang cmake --no-install-recommends
+            rustup component add rustfmt
+            rustup update
```
chfast, line 528: Why is update needed?

bindings/rust/src/lib.rs
axic: It is not "good practice", but required. The name is determined by the one in `Cargo.toml` and `lib.rs` is the default entry point.

circle.yml
axic: The name came from evmc, not sure what it had, can remove the combined.

bindings/rust/src/lib.rs
axic: Sure, though this is a copy of the test in `capi_test.cpp`, which does not have that either.

circle.yml
axic: To update any of the dependencies (also those needed by rustfmt). This is mostly a noop if the docker image is up to date (as it should be).

chfast: This seems wrong to be to having docker image being updated and also updating here. But I will not protest.

chfast (approved):

circle.yml
axic: Removed.

chfast (approved):

PR #539, opened 2020-09-21
Fix bounds checking pattern in parser

chfast:
The previous pattern `pos + size < end` was incorrect leading to
possible undefined behavior and potential addition overflow allowing
the check bypass. The correct check is to compute the remaining size
and compare with the declared size. This way invalid pointer comparison
is replaced with correct pointer subtraction (pos and end are from the
same memory buffer) and integer comparison: `(end - pos) < size`.

This will be additionally check with extended ASan options in #469.

### DIFF
--- a/lib/fizzy/parser.hpp
+++ b/lib/fizzy/parser.hpp
@@ -29,8 +29,8 @@ inline parser_result<uint8_t> parse_byte(const uint8_t* pos, const uint8_t* end)
 template <typename T>
 inline parser_result<T> parse_value(const uint8_t* pos, const uint8_t* end)
 {
-    constexpr auto size = sizeof(T);
-    if (pos + size > end)
+    constexpr ptrdiff_t size = sizeof(T);
+    if ((end - pos) < size)
         throw parser_error{"unexpected EOF"};
 
     T value;
--- a/lib/fizzy/utf8.cpp
+++ b/lib/fizzy/utf8.cpp
@@ -105,11 +105,12 @@ bool utf8_validate(const uint8_t* pos, const uint8_t* end) noexcept
         else
             return false;
 
-        // At this point need to read at least one more byte
-        if ((pos + required_bytes - 1) > end)
+
+        // At this point need to read at least one more byte.
+        assert(required_bytes > 1);
+        if ((end - pos) < required_bytes - 1)
             return false;
 
-        assert(required_bytes > 1);
 
         // Byte2 may have exceptional encodings
         const uint8_t byte2 = *pos++;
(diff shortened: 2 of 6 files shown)


chfast:
Good refactoring/optimization opportunity after is to pass `remaining_size` instead of `end` pointer to parsing function.

lib/fizzy/parser.hpp
```diff
@@ -29,8 +29,8 @@ inline parser_result<uint8_t> parse_byte(const uint8_t* pos, const uint8_t* end)
 template <typename T>
 inline parser_result<T> parse_value(const uint8_t* pos, const uint8_t* end)
 {
-    constexpr auto size = sizeof(T);
-    if (pos + size > end)
+    static constexpr int size = sizeof(T);
```
gumb0, line 32: why static?

lib/fizzy/utf8.cpp
```diff
@@ -105,11 +105,12 @@ bool utf8_validate(const uint8_t* pos, const uint8_t* end) noexcept
         else
             return false;
 
+        assert(required_bytes > 1);
```
gumb0, line 108: Could put this line after the comment below

lib/fizzy/parser.hpp
gumb0: And why `int`, isn't it then signed-unsigned comparison below?

gumb0:
> possible undefined behavior

when would it be UB?

lib/fizzy/utf8.cpp
chfast: Oh yeah, good point.

lib/fizzy/parser.hpp
chfast: Pointer subtraction returns signed integer. I will use `ptrdiff_t` for clarity if easily available.
For `static` I don't remember, I will remove it.

chfast:
> when would it be UB?

The adversarial element is `size` (it comes from the outside).

The first issue is in `pos + size` expression. Information about "pointer and integer addition":

> If both the pointer operand and the result point to elements of the same array object, or one past the last element of the array object, the evaluation shall not produce an overﬂow; otherwise, the behavior is undeﬁned.

And later this provokes another issue in pointer comparison `(pos + size) < end`:

> When two pointers are subtracted, both shall point to elements of the same array object, or one past the last element of the array object.

In practice, this is a security bug because in case there is a way to have a wasm module with `size` big enough to cause `pos + size` overflow the bounds check will be bypassed. Maybe not practical on 64-bit systems as I believe `size` is limited to 32-bits. Still worth exploring in context of other wasm binary readers.

gumb0 (approved):

axic (approved):

PR #542, opened 2020-09-22
Use native C++20 feature for bit_cast if available

axic:
Fixes https://github.com/wasmx/fizzy/pull/535/files#r492656891.

### DIFF
--- a/lib/fizzy/cxx20/bit.hpp
+++ b/lib/fizzy/cxx20/bit.hpp
@@ -5,6 +5,19 @@
 #pragma once
 
 #include <cstdint>
+
+#ifdef __cpp_lib_bit_cast
+
+#include <bit>
+
+namespace fizzy
+{
+template <typename To, class From>
+using bit_cast = std::bit_cast<To, From>;
+}
+
+#else
+
 #include <type_traits>
 
 namespace fizzy
@@ -25,6 +38,8 @@ template <class To, class From>
     return dst;
 }
 }  // namespace fizzy
+
+#endif /* __cpp_lib_bit_cast */
 
 #ifdef __cpp_lib_bitops
 


lib/fizzy/cxx20/bit.hpp
```diff
@@ -4,6 +4,18 @@
...
 #pragma once
 
+#if __cplusplus > 201703L
+
+#include <bit>
+
+namespace fizzy
+{
+template <typename To, class From>
+using bit_cast = std::bit_cast<To, From>;
```
axic, line 14: Not sure if this is correct?

chfast:
This is not doable right now because only MSVC has implemented `<bit>` so far. See "std::bit_cast" in https://en.cppreference.com/w/cpp/compiler_support#C.2B.2B20_library_features.

axic:
> This is not doable right now because only MSVC has implemented <bit> so far. See "std::bit_cast" in https://en.cppreference.com/w/cpp/compiler_support#C.2B.2B20_library_features.

Well we could this other macro:
> 2.5. Feature testing
> The __cpp_lib_bit_cast feature test macro should be added.

But since we don't support compilation with MSVC it can't be tested.

lib/fizzy/cxx20/bit.hpp
```diff
@@ -4,6 +4,18 @@
 
 #pragma once
 
+#if __cplusplus > 201703L
```
axic, line 7: Use proper macro here.

chfast:
> This is not doable right now because only MSVC has implemented `<bit>` so far. See "std::bit_cast" in https://en.cppreference.com/w/cpp/compiler_support#C.2B.2B20_library_features.

This still holds.

axic:
> > This is not doable right now because only MSVC has implemented <bit> so far. See "std::bit_cast" in https://en.cppreference.com/w/cpp/compiler_support#C.2B.2B20_library_features.
>
> This still holds.

Totally skipped over this fact.

axic:
This is waiting for GCC+11 now: https://www.gnu.org/software/gcc/gcc-11/changes.html

axic:
GCC 11 and Clang 14 has this now. Since #804 at least gcc is of correct version.

axic:
Tested with introduced a compilation error that the code is actually compiled on gcc 11.

axic:
@chfast @gumb0 please review

chfast (approved):

PR #534, opened 2020-09-18
Update spectests with backports

gumb0:
At this stage, it has 5 spectests failures.

- FIXED: One of the tests that we failed (and reported as invalid), was deleted https://github.com/WebAssembly/spec/pull/1205
- 4 tests are going to be fixed by #546.
- FIXED: 1 failure old failure is "validation-error-before-decoding-error", and it fixed upstream by https://github.com/WebAssembly/spec/pull/1261.
- FIXED: 2 new failures in `func.wast`, the reason is wast2json generates incorrect local index in the module. It still happens in wabt 1.0.19, but is fixed on master. Related issues/PRs:
   - https://github.com/WebAssembly/wabt/issues/1551
   - https://github.com/WebAssembly/spec/pull/1213
   - https://github.com/WebAssembly/wabt/pull/1466
   - https://github.com/WebAssembly/wabt/pull/1494

TODO:
- [x] Rename the wasm-spec branch from `w3c-1.0-backport-tests` to `w3c-1.0-tests-backported`.
- [x] Tag the version we are going to use with this PR as `w3c-1.0-tests-backported-20201113`.
- [x] Backport https://github.com/WebAssembly/spec/pull/1261 if merged.

PR #546, opened 2020-09-22
Change "unexpected memidx" to validation error

chfast:
Depends on #534. 

Fizzy fix for the proposed spectests: https://github.com/WebAssembly/spec/pull/1249.
Element section tests correction: https://github.com/WebAssembly/spec/pull/1251.

I also changed the error message.

> Use validation error message "unknown memory N" for both N != 0 and data segments without any memory in the module. Inspired by messages in wasm reference interpreter.

### DIFF
--- a/lib/fizzy/parser.cpp
+++ b/lib/fizzy/parser.cpp
@@ -410,8 +410,14 @@ inline parser_result<Data> parse(const uint8_t* pos, const uint8_t* end)
 {
     MemIdx memory_index;
     std::tie(memory_index, pos) = leb128u_decode<uint32_t>(pos, end);
+
+    // TODO: The check should be memory_index < num_of_memories (0 or 1),
+    //       but access to the memory section of the module is needed.
     if (memory_index != 0)
-        throw parser_error{"unexpected memidx value " + std::to_string(memory_index)};
+    {
+        throw validation_error{
+            "invalid memory index " + std::to_string(memory_index) + " (only memory 0 is allowed)"};
+    }
 
     ConstantExpression offset;
     // Offset expression is required to have i32 result value
@@ -560,7 +566,10 @@ std::unique_ptr<const Module> parse(bytes_view input)
     }
 
     if (!module->datasec.empty() && !module->has_memory())
-        throw validation_error{"data section encountered without a memory section"};
+    {
+        throw validation_error{
+            "invalid memory index 0 (data section encountered without a memory section)"};
+    }
 
     for (const auto& data : module->datasec)
     {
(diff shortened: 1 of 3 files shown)


gumb0 (approved):

gumb0:
I guess there's the same problem with Element section and table index validation?

chfast:
> I guess there's the same problem with Element section and table index validation?

True. But this also means that these are not covered by spectests so we need to add new tests there. Do you want to work on this?

lib/fizzy/parser.cpp
```diff
@@ -559,7 +571,7 @@ Module parse(bytes_view input)
     }
 
     if (!module.datasec.empty() && !module.has_memory())
-        throw validation_error{"data section encountered without a memory section"};
```
gumb0, line 562: These error messages were more descriptive and user-friendly, though, not sure if it makes sense to always follow the reference interpreter's messages.

chfast: The `unexpected tableidx value 1` is not better.

gumb0 (approved):

lib/fizzy/parser.cpp
```diff
@@ -410,8 +410,14 @@ inline parser_result<Data> parse(const uint8_t* pos, const uint8_t* end)
 {
     MemIdx memory_index;
     std::tie(memory_index, pos) = leb128u_decode<uint32_t>(pos, end);
+
+    // TODO: The check should be memory_index < num_of_memories (0 or 1),
+    //       but access to the memory section of the module is needed.
     if (memory_index != 0)
-        throw validation_error{"unexpected memidx value " + std::to_string(memory_index)};
+    {
```
axic, line 423: Do we need the brackets?

lib/fizzy/parser.cpp
```diff
@@ -560,7 +572,10 @@ std::unique_ptr<const Module> parse(bytes_view input)
     }
 
     if (!module->datasec.empty() && !module->has_memory())
-        throw validation_error{"data section encountered without a memory section"};
+    {
```
axic, line 575: And here?

axic:
@chfast can't we actually merge the first two commits?

lib/fizzy/parser.cpp
chfast: Because they wrap two lines.

chfast:
> @chfast can't we actually merge the first two commits?

Maybe. What's the point?

gumb0 (approved):

chfast:
What is the final git history expected?

axic:
Squash these two if it makes sense:
> Change "unexpected memidx" to validation error
> Change validation messages for unknown memory …

axic (approved):

PR #547, opened 2020-09-23
Asserts for checking expected number of inputs/outputs

chfast:


lib/fizzy/execute.cpp
```diff
@@ -1516,10 +1518,10 @@ ExecutionResult execute(
     }
 
 end:
-    // WebAssembly 1.0 allows at most one return variable.
-    assert(pc == &code.instructions[code.instructions.size()] && stack.size() <= 1);
+    assert(pc == &code.instructions[code.instructions.size()]);  // End of code must be reached.
+    assert(stack.size() == func_type.outputs.size());
 
-    return stack.size() != 0 ? ExecutionResult{stack.pop()} : Void;
+    return !func_type.outputs.empty() ? ExecutionResult{stack.pop()} : Void;
```
chfast, line 1524: Use `func_type`, the `stack.size()` is only used in asserts.

gumb0: I didn't get, why is it better?

It looks you could do assertions like
```
assert(args.size() == instance.module.get_function_type(func_idx).inputs.size());
```
```
assert(stack.size() == instance.module.get_function_type(func_idx).inputs.outputs.size());
```
and then avoid another variable declaration in outer scope and `get_function_type()` call.

I kind of liked the fact that `execute` doesn't need to access called function type, because all type stuff is handled at validation...

chfast: You are right here.

It looks we don't need information about number of outputs. But we will probably need the information about number of arguments. In the end to bootstrap execution we need:
- number of arguments,
- number of locals,
- max stack height.

gumb0 (approved):

axic (approved):

PR #549, opened 2020-09-23
ci: Use git ls-files for inputs to clang-format

chfast:


axic:
How is this better / different?

chfast:
> How is this better / different?

It will not reformat all my 32 build dirs.

axic (approved):

Issue #543, opened 2020-09-22
Make use of C++20 bit manipulation features


axic:
We are currently using compiler builtins:
- `__builtin_clz`
- `__builtin_clzll`
- `__builtin_ctz`
- `__builtin_ctzll`
- `__builtin_popcount`
- `__builtin_popcountll`

These can be replaced with the new C++20 features proposed (see https://en.cppreference.com/w/cpp/numeric).

For an idea how to write such a wrapper, check out `lib/fizzy/cxx20/span.hpp`.

herobank110:
Hello, I have been reading about c++20 features and would like to try solve this issue.

chfast:
Hi @herobank110,

you can start with a simple task as follows:
- move the implementation of `popcount` (https://github.com/wasmx/fizzy/blob/master/lib/fizzy/execute.cpp#L328-L331) to `cxx20/bit.hpp`
- make it's function signature exactly as in C++20: https://en.cppreference.com/w/cpp/numeric/popcount.

chfast:
Correction, start with something other than `popcount` as it is being done in #551.

herobank110:
I made an wrapper of countl_zero. Please can you check it? #570

PR #551, opened 2020-09-24
Use popcount from C++20 if available

axic:
Part of #543.

### DIFF
--- a/lib/fizzy/cxx20/bit.hpp
+++ b/lib/fizzy/cxx20/bit.hpp
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include <cstdint>
 #include <type_traits>
 
 namespace fizzy
@@ -24,3 +25,39 @@ template <class To, class From>
     return dst;
 }
 }  // namespace fizzy
+
+#ifdef __cpp_lib_bitops
+
+#include <bit>
+
+namespace fizzy
+{
+constexpr int popcount(uint32_t x) noexcept
+{
+    return std::popcount(x);
+}
+
+constexpr int popcount(uint64_t x) noexcept
+{
+    return std::popcount(x);
+}
+}  // namespace fizzy
+
+#else
+
+namespace fizzy
+{
+constexpr int popcount(uint32_t x) noexcept
+{
+    static_assert(sizeof(x) == sizeof(unsigned int));
+    return __builtin_popcount(x);
+}
+
+constexpr int popcount(uint64_t x) noexcept
+{
+    static_assert(sizeof(x) == sizeof(unsigned long long));
+    return __builtin_popcountll(x);
+}
+}  // namespace fizzy
+
+#endif /* __cpp_lib_bitops */
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -273,9 +273,9 @@ inline uint32_t ctz32(uint32_t value) noexcept
     return static_cast<uint32_t>(__builtin_ctz(value));
 }
 
-inline uint32_t popcnt32(uint32_t value) noexcept
+constexpr uint32_t popcnt32(uint32_t value) noexcept
 {
-    return static_cast<uint32_t>(__builtin_popcount(value));
+    return static_cast<uint32_t>(popcount(value));
 }
 
 inline uint64_t clz64(uint64_t value) noexcept
@@ -294,9 +294,9 @@ inline uint64_t ctz64(uint64_t value) noexcept
     return static_cast<uint64_t>(__builtin_ctzll(value));
 }
 
-inline uint64_t popcnt64(uint64_t value) noexcept
+constexpr uint64_t popcnt64(uint64_t value) noexcept
 {
-    return static_cast<uint64_t>(__builtin_popcountll(value));
+    return static_cast<uint64_t>(popcount(value));
 }
 
 template <typename T>
--- a/test/unittests/CMakeLists.txt
+++ b/test/unittests/CMakeLists.txt
@@ -10,9 +10,10 @@ target_link_libraries(fizzy-unittests PRIVATE fizzy::fizzy-internal fizzy::test-
 target_sources(
     fizzy-unittests PRIVATE
     api_test.cpp
-    bit_cast_test.cpp
     capi_test.cpp
     constexpr_vector_test.cpp
+    cxx20_bit_test.cpp
+    cxx20_span_test.cpp
     end_to_end_test.cpp
     execute_call_test.cpp
     execute_control_test.cpp
@@ -25,7 +26,6 @@ target_sources(
     module_test.cpp
     parser_expr_test.cpp
     parser_test.cpp
-    span_test.cpp
     stack_test.cpp
     test_utils_test.cpp
     types_test.cpp
--- a/test/unittests/cxx20_bit_test.cpp
+++ b/test/unittests/cxx20_bit_test.cpp
@@ -10,31 +10,31 @@ using namespace fizzy;
 using namespace testing;
 
 
-TEST(bit_cast, double_to_uint64)
+TEST(cxx20_bit, bit_cast_double_to_uint64)
 {
     // A test case from https://en.cppreference.com/w/cpp/numeric/bit_cast#Example.
     EXPECT_EQ(bit_cast<uint64_t>(19880124.0), 0x4172f58bc0000000);
 }
 
-TEST(bit_cast, uint64_to_double)
+TEST(cxx20_bit, bit_cast_uint64_to_double)
 {
     // A test case from https://en.cppreference.com/w/cpp/numeric/bit_cast#Example.
     EXPECT_EQ(bit_cast<double>(uint64_t{0x3fe9000000000000}), 0.781250);
 }
 
-TEST(bit_cast, uint32_to_int32)
+TEST(cxx20_bit, bit_cast_uint32_to_int32)
 {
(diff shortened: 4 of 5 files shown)


lib/fizzy/execute.cpp
```diff
@@ -3,6 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include "execute.hpp"
+#include "cxx20/numeric.hpp"
```
axic, line 6: This should be actually in `bit.hpp`.

lib/fizzy/cxx20/numeric.hpp
```diff
@@ -0,0 +1,37 @@
...
+#if __cplusplus > 201703L
+
+#include <bit>
+
+namespace fizzy
+{
+template <class T>
+constexpr int popcount(T x) noexcept
+{
+    return std::popcount.as<T>(x);
```
chfast, line 16: What is this?

lib/fizzy/cxx20/numeric.hpp
```diff
@@ -0,0 +1,37 @@
...
+#pragma once
+
+#if __cplusplus > 201703L
+
+#include <bit>
+
+namespace fizzy
+{
+template <class T>
+constexpr int popcount(T x) noexcept
```
chfast, line 14: This is not needed and probably not used currently.

lib/fizzy/cxx20/numeric.hpp
```diff
@@ -0,0 +1,37 @@
+// Fizzy: A fast WebAssembly interpreter
+// Copyright 2019-2020 The Fizzy Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#if __cplusplus > 201703L
```
chfast, line 7: This will not work, because you may be building in C++20, but without `<bit>` implemented.
Use [`__cpp_lib_bitops`](https://en.cppreference.com/w/cpp/feature_test#Library_features).

lib/fizzy/cxx20/numeric.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+#else
+
+#include <cstdint>
+
+namespace fizzy
+{
+constexpr int popcount(uint32_t x) noexcept
+{
+    return __builtin_popcount(x);
```
chfast, line 28:
```suggestion
    static_assert(sizeof(x) == sizeof(unsigned int));
    return __builtin_popcount(x);
```

lib/fizzy/cxx20/numeric.hpp
```diff
@@ -0,0 +1,37 @@
...
+}
+
+constexpr int popcount(uint64_t x) noexcept
+{
+    return __builtin_popcountll(x);
```
chfast, line 33:
```suggestion
    static_assert(sizeof(x) == sizeof(unsigned long long));
    return __builtin_popcountll(x);
```

chfast (changes requested):

lib/fizzy/cxx20/numeric.hpp
axic: There is the specific type `uint32_t x`. Why the assert?

lib/fizzy/cxx20/numeric.hpp
axic: The `using` statement didn't seem to work for functions. We still want to "copy" it from `std` to `fizzy`.

chfast: `using` does not work because functions are constant objects, not types. This is good approach. But the `.as<T>` is wrong. And because of we cannot test it now, maybe add `static_assert()` for now?

lib/fizzy/cxx20/numeric.hpp
chfast: To make sure `uint32_t` is `unsigned int` size. The `__buintin_popcount()` takes `unsigned int`.

lib/fizzy/cxx20/numeric.hpp
axic: Should use `__cpp_lib_span` for span too. And likewise https://github.com/wasmx/fizzy/pull/542#issuecomment-696691041 could be merged.

lib/fizzy/cxx20/numeric.hpp
axic: What do you mean?

lib/fizzy/cxx20/numeric.hpp
axic: `static_assert` for what? The two types we support/test?

axic: In that case it may be better not to use a template just the two specialisations.

gumb0:
Looks good, needs some unit tests.

lib/fizzy/cxx20/numeric.hpp
chfast: I propose to this for now:
```
#if __cpp_lib_bitops
static_assert(false, "Add alias to std::popcount");
#endif 
```

As for function alias, only template function is generic function alias. There is also an option to use `constexpr`, but not for overloaded functions: https://stackoverflow.com/questions/9864125/c11-how-to-alias-a-function/49819511#49819511.

lib/fizzy/cxx20/numeric.hpp
chfast: Correct, although the `std::bit_cast` is similarly untestable.

lib/fizzy/cxx20/numeric.hpp
axic: Why don't we add a test run for C++20? There is just no point adding these things otherwise.

chfast: None of the C++ stdlib implementations have `<bit>` support.

axic:
> Looks good, needs some unit tests.

We do have tests for them via the opcodes, but you are right probably better to have unit tests for any of these wrappers.

lib/fizzy/cxx20/bit.hpp
```diff
@@ -24,3 +24,41 @@ template <class To, class From>
...
+
+namespace fizzy
+{
+constexpr int popcount(uint32_t x) noexcept
+{
+    static_assert(sizeof(x) == sizeof(unsigned int));
+    return __builtin_popcount(x);
+}
+
+constexpr int popcount(uint64_t x) noexcept
```
axic, line 57: These should be inline, not consexpr as `__builtin_popcount` isn't constexpr.

chfast: Most builtins are constexpr. https://godbolt.org/z/Pnz4ob.

axic: Cool, the gcc documentation had no mention of either.

axic:
@gumb0 added tests if you want to check

gumb0 (approved):
Looks good to me, needs squashing.

axic:
Will squash, but this is still blocked by clang11 for having a way to tests the C++20 implementation.

test/unittests/CMakeLists.txt
```diff
@@ -10,9 +10,10 @@ target_link_libraries(fizzy-unittests PRIVATE fizzy::fizzy-internal fizzy::test-
 target_sources(
     fizzy-unittests PRIVATE
     api_test.cpp
-    bit_cast_test.cpp
     capi_test.cpp
     constexpr_vector_test.cpp
+    cxx20_bit_test.cpp
+    cxx20_span_test.cpp
```
axic, line 16: @gumb0 @chfast what do you think about this renaming?

gumb0: I think it's fine.

lib/fizzy/cxx20/bit.hpp
axic: Clang actually marks them constexpr: https://clang.llvm.org/docs/LanguageExtensions.html#intrinsics-support-within-constant-expressions

But GCC does not clarify it: https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html

lib/fizzy/execute.cpp
```diff
@@ -296,7 +296,7 @@ inline uint64_t ctz64(uint64_t value) noexcept
 
 inline uint64_t popcnt64(uint64_t value) noexcept
```
axic, line 297: Should make these `constexpr` then since we established the intrinsics are.

lib/fizzy/cxx20/bit.hpp
```diff
@@ -24,3 +24,41 @@ template <class To, class From>
...
+{
+    return std::popcount(x);
+}
+}  // namespace fizzy
+
+#else
+
+#include <cstdint>
```
chfast, line 47: This is also needed for C++20 variant.

axic: Right. Will fix. Though it seems all the places this is used at already includes `cstdint`. See the `f` commit which has a warning to see if the code path is compiled (and it is).

chfast: Just include it on the top of the file.

chfast (changes requested):

lib/fizzy/cxx20/bit.hpp
axic: Done.

chfast (approved):

PR #553, opened 2020-09-24
cmake: Add configuration for Cmake Package

chfast:


lib/fizzy/CMakeLists.txt
```diff
@@ -2,19 +2,14 @@
...
+target_include_directories(fizzy PUBLIC ${FIZZY_INCLUDE_DIR})
 
 target_sources(
-    fizzy
-
-    PUBLIC
-    ${INCLUDE_DIR}/fizzy/fizzy.h
-
+    fizzy PRIVATE
+    ${FIZZY_INCLUDE_DIR}/fizzy/fizzy.h
```
gumb0, line 12: Why is it `PRIVATE` now?

Only single `PRIVATE` keyword is enough now for all sources?

chfast: `PUBLIC` mean something different: public sources from lib A are added to all libs that link to A. This is quite exotic and confusing feature, but not a way to specify public headers to my understanding. `install()` complains if a public source is specified as a local path.

I should move this fix to #530.

Yes, all sources after `PRIVATE` are private. The same as in all `target_*` commands.

gumb0: Don't repeat `PRIVATE` then?

chfast: Removed.

gumb0 (approved):

test/integration/cmake_package/use_fizzy/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+# Fizzy: A fast WebAssembly interpreter
+# Copyright 2020 The Fizzy Authors.
+# SPDX-License-Identifier: Apache-2.0
+
+# This example shows how to use the fizzy library from the fizzy CMake package.
+cmake_minimum_required(VERSION 3.15)
+project(use_fizzy LANGUAGES CXX)
+find_package(fizzy CONFIG REQUIRED)
```
axic, line 8: So this `use_fizzy` directory is a test for `find_package(fizzy)`?

axic (approved):
My understanding is limited, but looks okay-ish.

chfast (commented):
CMake package consist of a number of special files.
- `fizzyConfig.cmake` - the main one; `find_package(fizzy CONFIG)` is looking for it; it is generated from `Config.cmake.in` in https://github.com/wasmx/fizzy/pull/553/files#diff-af3b638bc2a3e6c650974192a53c7291R110.
- `fizzyConfigVersion.cmake` - it contains a script to resolved required version in `find_package(fizzy VERSION a.b.c)`; generated by https://github.com/wasmx/fizzy/pull/553/files#diff-af3b638bc2a3e6c650974192a53c7291R109; we opt for "ExactVersion" compatibility.
- `fizzyTargets.cmake` - contains information about all "installed" targets; generate and installed by `install(EXPORT ...)` in https://github.com/wasmx/fizzy/pull/553/files#diff-af3b638bc2a3e6c650974192a53c7291R114.

CMake docs in subject: https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html

In summary: there is a shit lot of boilerplate here and essential parts well hidden in-between.

PR #557, opened 2020-09-28
C API function type

gumb0:


lib/fizzy/capi.cpp
```diff
@@ -11,6 +11,35 @@
...
+}
+
+inline FizzyFunctionType wrap(const fizzy::FuncType& type) noexcept
+{
+    return {wrap(&type.inputs[0]), type.inputs.size(), wrap(&type.outputs[0]), type.outputs.size()};
+}
+
+inline fizzy::FuncType unwrap(const FizzyFunctionType& type)
+{
+    fizzy::FuncType func_type{std::vector<fizzy::ValType>(type.inputs_size),
```
gumb0, line 31: We could optimize it later on the C++ side to not require vector allocation to pass function type (i.e. replace `vector` with `span` in `fizzy::FuncType`)

include/fizzy/fizzy.h
```diff
@@ -75,6 +97,13 @@ const FizzyModule* fizzy_parse(const uint8_t* wasm_binary, size_t wasm_binary_si
 /// If passed pointer is NULL, has no effect.
 void fizzy_free_module(const FizzyModule* module);
 
+/// Get type of the function defined in the module.
+///
+/// @param module   Pointer to module.
+/// @param func_idx Function index. Can be index of an imported function.
```
axic, line 103: Should mention how the indexes are following eachother (imports vs. module defs).

lib/fizzy/capi.cpp
```diff
@@ -21,6 +21,38 @@ inline const fizzy::Module* unwrap(const FizzyModule* module) noexcept
     return reinterpret_cast<const fizzy::Module*>(module);
 }
 
+static_assert(sizeof(FizzyValueType) == sizeof(fizzy::ValType));
```
axic, line 24: Hm, perhaps there should be such an assert for `FizzyValue` too.

lib/fizzy/types.hpp
```diff
@@ -15,7 +15,7 @@
 namespace fizzy
 {
 // https://webassembly.github.io/spec/core/binary/types.html#binary-valtype
-enum class ValType : uint8_t
```
axic, line 18: Why is this needed?

gumb0: To be able to cast between arrays of `fizzy::ValType` and arrays of `FizzyValueType`

lib/fizzy/capi.cpp
gumb0: There `bit_cast` already does this.

lib/fizzy/capi.cpp
```diff
@@ -108,25 +145,20 @@ void fizzy_free_module(const FizzyModule* module)
     delete unwrap(module);
 }
 
+FizzyFunctionType fizzy_get_function_type(const FizzyModule* module, uint32_t func_idx)
+{
+    return wrap(unwrap(module)->get_function_type(func_idx));
```
gumb0, line 150: Note: `get_function_type` returns reference to `fizzy::FuncType` stored in module, then `wrap` returns pointers to its vector's data.

lib/fizzy/types.hpp
chfast: I don't think we can go with this change.

chfast: Can you try using `uint8_t` on C side for now, and add
```
const uint8_t FizzyI32 = 0x7f;
const uint8_t FizzyI64 = 0x7e;
...
```

gumb0:
> ```
> static const
> ```
> 
> The global `const` means `static const` in C++, but not in C. This will make them equal.

This will create their copies in each translation unit that includes `fizzy.h`, is this what we want?

lib/fizzy/types.hpp
gumb0: Is performance considerations the reason why not?

chfast:
> This will create their copies in each translation unit that includes `fizzy.h`, is this what we want?

They are going to be used as compile-time constants, unlikely to be emitted at all.

axic:
@gumb0 can you please rebase and squash?

gumb0:
> @gumb0 can you please rebase and squash?

Done, squashed the last commit.

include/fizzy/fizzy.h
```diff
@@ -50,11 +50,31 @@ typedef struct FizzyExecutionResult
...
+
+/// Function type.
+typedef struct FizzyFunctionType
+{
+    /// Pointer to input types array.
+    const FizzyValueType* inputs;
+    /// Input types array size.
+    size_t inputs_size;
+    /// Pointer to output types array.
+    const FizzyValueType* outputs;
```
chfast, line 68: This is ok, but also can be greatly simplified by defining `FizzyValueTypeVoid` and using `FizzyValueType output_type`.

gumb0: Good idea, I'll try it.

chfast: Move it to the front.

include/fizzy/fizzy.h
```diff
@@ -50,11 +50,31 @@ typedef struct FizzyExecutionResult
...
     const union FizzyValue* args, size_t args_size, int depth);
 
+/// Value type.
+typedef uint8_t FizzyValueType;
+static const FizzyValueType FizzyValueTypeI32 = 0x7f;
+static const FizzyValueType FizzyValueTypeI64 = 0x7e;
+static const FizzyValueType FizzyValueTypeF32 = 0x7d;
+static const FizzyValueType FizzyValueTypeF64 = 0x7c;
+/// Special value, can be used only as function outuput type.
+static const FizzyValueType FizzyValueTypeVoid = 0;
```
gumb0, line 60: Defining it as `0` allows user to create void type with not params as `{}`, at least in C++, probably in C, too.

include/fizzy/fizzy.h
```diff
@@ -50,11 +50,31 @@ typedef struct FizzyExecutionResult
...
+static const FizzyValueType FizzyValueTypeI64 = 0x7e;
+static const FizzyValueType FizzyValueTypeF32 = 0x7d;
+static const FizzyValueType FizzyValueTypeF64 = 0x7c;
+/// Special value, can be used only as function output type.
+static const FizzyValueType FizzyValueTypeVoid = 0;
+
+/// Function type.
+typedef struct FizzyFunctionType
+{
+    /// Output type, can be FizzyValueTypeVoid, if function has no output.
```
axic, line 65:
```suggestion
    /// Output type, equals to FizzyValueTypeVoid, if function has no output.
```

include/fizzy/fizzy.h
```diff
@@ -75,6 +95,15 @@ const FizzyModule* fizzy_parse(const uint8_t* wasm_binary, size_t wasm_binary_si
...
 void fizzy_free_module(const FizzyModule* module);
 
+/// Get type of the function defined in the module.
+///
+/// @param module   Pointer to module.
+/// @param func_idx Function index. Can be either index of an imported function or of a function
+///                 defined in module.
+///
+/// @note All module function indices are greater than all imported function indices.
+FizzyFunctionType fizzy_get_function_type(const FizzyModule* module, uint32_t func_idx);
```
axic, line 105: What happens on invalid index?

axic:
Needs rebase.

gumb0:
Rebased.

include/fizzy/fizzy.h
gumb0: Changed.

include/fizzy/fizzy.h
gumb0: Added a sentence.

include/fizzy/fizzy.h
```diff
@@ -74,6 +94,16 @@ const FizzyModule* fizzy_parse(const uint8_t* wasm_binary, size_t wasm_binary_si
 /// If passed pointer is NULL, has no effect.
 void fizzy_free_module(const FizzyModule* module);
 
+/// Get type of the function defined in the module.
+///
+/// @param module   Pointer to module.
+/// @param func_idx Function index. Can be either index of an imported function or of a function
+///                 defined in module. Behaviour is undefined, index is not valid according to
```
axic, line 101:
```suggestion
///                 defined in module. Behaviour is undefined if index is not valid according to
```

axic (approved):
Looks okayish to me.

chfast (approved):

PR #559, opened 2020-09-29
ci: Adjust fuzzing config

chfast:


circle.yml
```diff
@@ -454,22 +454,37 @@ jobs:
       - build:
           build_type: RelWithDebInfo
           cmake_options: -DFIZZY_FUZZING=ON -DENABLE_ASSERTIONS=ON
+          target: fizzy-fuzz-parser
```
axic, line 457: What is this doing?

chfast: It only builds `fizzy-fuzz-parser` instead of building everything, so the build takes only ~15 seconds. We don't run tests in this job (although we could, they would be similar to sanitizers-clang job).

axic: It means build target, e.g. `make fizzy-fuzz-parser`?

chfast: Yes.

axic (approved):

PR #560, opened 2020-09-29
Optimize fabs, fneg, fcopysign

### LINKED ISSUE #540: Optimize fN.abs, fN.neg, fN.copysign instructions
The goal is to have the simplest and leanest implementation of the following WebAssembly instructions:
- `f32.abs`, `f64.abs`
- `f32.neg`, `f64.neg`
- `f32.copysign`, `f64.copysign`

Current implementation uses C/C++ standard functions to implement the core of the instructions: `std::fabs`, `std::copysign` and `-x` for negation. Due to compilers optimization pipelines, this does not provide optimal implementation: the generated code sometimes unnecessary uses floating-point or SSE2 instructions.

Change the implementation to use matching unsigned integer types and bit manipulation.

E.g. for `f32.abs` it is enough to zero the most significant bit of the unsigned integer representation of the value.

```cpp
float fabs(float a) noexcept
{
    return bit_cast<float>(bit_cast<uint32_t>(a) & 0x7fffffff);
}
```

Here are some additional tips and hints: https://godbolt.org/z/KaET9c.

Themarek1:
Closes #540.

I created optimized first single case of issue #540 to be sure that i understood a problem.

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -3,6 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include "execute.hpp"
+#include "cxx20/bit.hpp"
 #include "stack.hpp"
 #include "trunc_boundaries.hpp"
 #include "types.hpp"
@@ -18,6 +19,11 @@ namespace
 {
 // code_offset + imm_offset + stack_height
 constexpr auto BranchImmediateSize = 3 * sizeof(uint32_t);
+
+constexpr uint32_t F32AbsMask = 0x7fffffff;
+constexpr uint32_t F32SignMask = ~F32AbsMask;
+constexpr uint64_t F64AbsMask = 0x7fffffffffffffff;
+constexpr uint64_t F64SignMask = ~F64AbsMask;
 
 template <typename T>
 inline T read(const uint8_t*& input) noexcept
@@ -294,6 +300,36 @@ inline uint64_t popcnt64(uint64_t value) noexcept
 }
 
 template <typename T>
+T fabs(T value) noexcept = delete;
+
+template <>
+inline float fabs(float value) noexcept
+{
+    return bit_cast<float>(bit_cast<uint32_t>(value) & F32AbsMask);
+}
+
+template <>
+inline double fabs(double value) noexcept
+{
+    return bit_cast<double>(bit_cast<uint64_t>(value) & F64AbsMask);
+}
+
+template <typename T>
+T fneg(T value) noexcept = delete;
+
+template <>
+inline float fneg(float value) noexcept
+{
+    return bit_cast<float>(bit_cast<uint32_t>(value) ^ F32SignMask);
+}
+
+template <>
+inline double fneg(double value) noexcept
+{
+    return bit_cast<double>(bit_cast<uint64_t>(value) ^ F64SignMask);
+}
+
+template <typename T>
 inline T fceil(T value) noexcept
 {
     static_assert(std::is_floating_point_v<T>);
@@ -390,6 +426,25 @@ inline T fmax(T a, T b) noexcept
         return T{0};
 
     return a < b ? b : a;
+}
+
+template <typename T>
+T fcopysign(T a, T b) noexcept = delete;
+
+template <>
+inline float fcopysign(float a, float b) noexcept
+{
+    const auto a_u = bit_cast<uint32_t>(a);
+    const auto b_u = bit_cast<uint32_t>(b);
+    return bit_cast<float>((a_u & F32AbsMask) | (b_u & F32SignMask));
+}
+
+template <>
+inline double fcopysign(double a, double b) noexcept
+{
+    const auto a_u = bit_cast<uint64_t>(a);
+    const auto b_u = bit_cast<uint64_t>(b);
+    return bit_cast<double>((a_u & F64AbsMask) | (b_u & F64SignMask));
 }
 
 __attribute__((no_sanitize("float-cast-overflow"))) inline constexpr float demote(
@@ -1223,13 +1278,12 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, const Value* args,
 
         case Instr::f32_abs:
         {
-            // TODO: This can be optimized https://godbolt.org/z/aPqvfo
-            unary_op(stack, static_cast<float (*)(float)>(std::fabs));
+            unary_op(stack, fabs<float>);
             break;
         }
         case Instr::f32_neg:
         {
-            unary_op(stack, std::negate<float>{});
+            unary_op(stack, fneg<float>);
             break;
         }
         case Instr::f32_ceil:
@@ -1290,24 +1344,18 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, const Value* args,
         }
         case Instr::f32_copysign:
         {
-            // TODO: This is not optimal implementation. The std::copysign() is inlined, but
-            //       it affects the compiler to still use SSE vectors (probably due to C++ ABI)
-            //       while this can be implemented with just generic registers and integer
-            //       instructions: (a & ABS_MASK) | (b & SIGN_MASK).
-            //       https://godbolt.org/z/aPqvfo
-            binary_op(stack, static_cast<float (*)(float, float)>(std::copysign));
+            binary_op(stack, fcopysign<float>);
             break;
         }
 
         case Instr::f64_abs:
         {
-            // TODO: This can be optimized https://godbolt.org/z/aPqvfo
-            unary_op(stack, static_cast<double (*)(double)>(std::fabs));
+            unary_op(stack, fabs<double>);
             break;
         }
         case Instr::f64_neg:
         {
-            unary_op(stack, std::negate<double>{});
+            unary_op(stack, fneg<double>);
             break;
         }
         case Instr::f64_ceil:
@@ -1368,7 +1416,7 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, const Value* args,
         }
         case Instr::f64_copysign:
         {
-            binary_op(stack, static_cast<double (*)(double, double)>(std::copysign));
+            binary_op(stack, fcopysign<double>);
             break;
         }
 


lib/fizzy/float_handling.hpp
```diff
@@ -0,0 +1,18 @@
...
+#pragma once
+namespace fizzy
+{
+union Num32
+{
+    unsigned int i32;
+    float f32;
+};
+
+inline float abs(float input) noexcept
```
chfast, line 11: No need to place this function is separate file. Move it to `execute.cpp`, next to functions `fceil`, `ffloor`, etc, in the order of the opcodes.

lib/fizzy/execute.cpp
```diff
@@ -1226,7 +1226,7 @@ ExecutionResult execute(
         case Instr::f32_abs:
         {
             // TODO: This can be optimized https://godbolt.org/z/aPqvfo
-            unary_op(stack, static_cast<float (*)(float)>(std::fabs));
+            unary_op(stack, static_cast<float (*)(float)>(abs));
```
chfast, line 1229: The `static_cast` is not needed any more because `abs` is not overloaded.

chfast (changes requested):

lib/fizzy/execute.cpp
```diff
@@ -414,6 +419,71 @@ T fnearest(T value) noexcept
...
+
+union Num64
+{
+    uint64_t i64;
+    double d64;
+};
+
+float fabs(float input) noexcept
+{
+    Num32 result;
```
chfast, line 437: `fizzy::bit_cast` must be used instead of type punning via `union Num32`.

chfast (changes requested):

axic:
Also please squash the commits into a single one.

lib/fizzy/execute.cpp
```diff
@@ -414,6 +419,71 @@ T fnearest(T value) noexcept
...
+
+double dneg(double input) noexcept
+{
+    Num64 result;
+    result.d64 = input;
+    result.i64 ^= SIGN_MASK64;
+    return result.d64;
+}
+
+float copy_sign(float f1, float f2) noexcept
```
axic, line 467: To follow C++ std, this should be called `copysign`.

lib/fizzy/execute.cpp
```diff
@@ -414,6 +419,71 @@ T fnearest(T value) noexcept
...
+{
+    Num32 x, y;
+
+    x.f32 = f1; 
+    y.f32 = f2;
+    x.i32 = (x.i32 & ABS_MASK) | (y.i32 & SIGN_MASK);
+    return x.f32;
+}
+
+double dcopy_sign(double f1, double f2) noexcept
```
axic, line 477: I think you can overload the float-version of this and keep the same name:
```
float copysign(float a, float b) noexcept
double copysign(double a, double b) noexcept
```

lib/fizzy/execute.cpp
axic, line 467: Please use `a` and `b` as parameter names, similar to what the other binary ops are doing.

lib/fizzy/execute.cpp
Themarek1: Hi i am really not sure so i must ask, do you think something like that 
float fabs(float input) noexcept
{
    auto result = bit_cast<unsigned int>(input);
    result &= ABS_MASK;
    return bit_cast<float>(result);
} 
Thanks for your answer.

lib/fizzy/execute.cpp
```diff
@@ -356,6 +362,49 @@ T fnearest(T value) noexcept
         return t;
 }
 
+float fabs(float input) noexcept
```
axic, line 365: Could make these templates to avoid the need for casting in `execute`. See `fmax` as an example.

lib/fizzy/execute.cpp
```diff
@@ -356,6 +362,62 @@ T fnearest(T value) noexcept
         return t;
 }
 
+
+template <typename T>
+T fabs(T) noexcept = delete;
```
chfast, line 367: I allowed myself to use template specialization what is more strict than previous attempt - no other type will be accepted. Please do the same for other functions.

lib/fizzy/execute.cpp
```diff
@@ -356,6 +362,62 @@ T fnearest(T value) noexcept
...
+    else
+    {
+        auto result = bit_cast<uint64_t>(input);
+        result ^= SIGN_MASK64;
+        return bit_cast<T>(result);
+    }
+}
+
+template <typename T>
+T copysign(T a, T b) noexcept
```
chfast, line 403: I noticed that some conversion functions may use `fabs()`, `copysign()` and others. So please move these functions before `fceil()` in this file.

lib/fizzy/execute.cpp
```diff
@@ -19,6 +20,11 @@ namespace
 // code_offset + imm_offset + stack_height
 constexpr auto BranchImmediateSize = 3 * sizeof(uint32_t);
 
+constexpr auto ABS_MASK = 0x7fffffff;
```
chfast, line 23: Please, use the following naming convention. Also specify the type explicitly (the rules what type literals have are confusing).
```suggestion
constexpr uint32_t F32AbsMask = 0x7fffffff;
```

lib/fizzy/execute.cpp
```diff
@@ -19,6 +20,11 @@ namespace
...
+constexpr auto SIGN_MASK = 0x80000000;
```
chfast, line 24: This can be based on the other mask.
```suggestion
constexpr uint32_t F32SignMask = ~F32AbsMask;
```

lib/fizzy/execute.cpp
chfast, line 367: @gumb0, @axic I'm currently in opinion that `abs()` is better name as we don't have `iadd()` and `fadd()`, just `add()`.

lib/fizzy/execute.cpp
chfast: Functions should be marked with `inline`.
```suggestion
inline float fabs(float input) noexcept
```

lib/fizzy/execute.cpp
```diff
@@ -356,6 +362,62 @@ T fnearest(T value) noexcept
...
+    result &= ABS_MASK64;
+    return bit_cast<double>(result);
+}
+
+template <typename T>
+T neg(T input) noexcept
+{
+    if constexpr (std::is_same_v<T, float>)
+    {
+        auto result = bit_cast<unsigned int>(input);
```
chfast, line 390: Use types with precise bit length.
```suggestion
        auto result = bit_cast<uint32_t>(input);
```

chfast (changes requested):

lib/fizzy/execute.cpp
axic: I was actually going to suggest to keep the `f` prefix and also use `fneg`, given these functions are specifically working on floating points. Since there's no `abs` on integers in Wasm, it doesn't come up as an issue. Should there be one in the future, happy to rename it to `abs`.

lib/fizzy/execute.cpp
```diff
@@ -1226,12 +1288,12 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, const Value* args,
         case Instr::f32_abs:
         {
             // TODO: This can be optimized https://godbolt.org/z/aPqvfo
```
axic, line 1290: Please also remove all these TODO comments where the implementation was changed, because the comments are not relevant anymore.

lib/fizzy/execute.cpp
```diff
@@ -293,6 +299,59 @@ inline uint64_t popcnt64(uint64_t value) noexcept
     return static_cast<uint64_t>(__builtin_popcountll(value));
 }
 
+template <typename T>
+inline T fabs(T input) noexcept
+{
+    if constexpr (std::is_same_v<T, float>)
+    {
+        auto result = bit_cast<uint32_t>(input);
```
gumb0, line 307:
```suggestion
        const auto result = bit_cast<uint32_t>(input);
```

lib/fizzy/execute.cpp
```diff
@@ -293,6 +299,59 @@ inline uint64_t popcnt64(uint64_t value) noexcept
...
+        return bit_cast<T>(result);
+    }  
+}
+
+template <typename T>
+inline T copysign(T a, T b) noexcept
+{
+    if constexpr (std::is_same_v<T, float>)
+    {
+        auto x = bit_cast<unsigned int>(a);
```
gumb0, line 341:
```suggestion
        auto x = bit_cast<uint32_t>(a);
```

lib/fizzy/execute.cpp
```diff
@@ -293,6 +299,59 @@ inline uint64_t popcnt64(uint64_t value) noexcept
     return static_cast<uint64_t>(__builtin_popcountll(value));
 }
 
+template <typename T>
+inline T fabs(T input) noexcept
```
gumb0, line 303: Usually called `value` in other functions
```suggestion
inline T fabs(T value) noexcept
```

gumb0 (changes requested):

lib/fizzy/execute.cpp
chfast: Same for `fcopysign` then?

gumb0 (approved):

lib/fizzy/execute.cpp
axic: https://en.cppreference.com/w/cpp/numeric/math/copysign seems to be floating point only

lib/fizzy/execute.cpp
```diff
@@ -19,6 +20,11 @@ namespace
 // code_offset + imm_offset + stack_height
 constexpr auto BranchImmediateSize = 3 * sizeof(uint32_t);
 
+constexpr uint32_t F32AbsMask = 0x7fffffff;
```
axic, line 23: Could go crazy and use `std::numeric_limits<uint32_t>::max() >> 1`, but the constant is nicer.

lib/fizzy/execute.cpp
chfast: Keeping `fcopysign` then.

chfast (approved):

axic (approved):

PR #564, opened 2020-10-01
Invoke refactoring

chfast:


axic:
Rebased on top of #602. Can be benchmarked too.

axic:
@chfast maybe this with #602 would be measurable saving? It removes the passing of `func_type`.

chfast:
> @chfast maybe this with #602 would be measurable saving? It removes the passing of `func_type`.

Yes. This was main goal.

axic:
Rebased. If we want to keep the assertion from #602, then need the last commit, which likely negates all the benefits.

PR #565, opened 2020-10-01
C API instantiate with resolving imported functions

gumb0:


include/fizzy/fizzy.h
```diff
@@ -71,6 +71,9 @@ struct FizzyModule* fizzy_parse(const uint8_t* wasm_binary, size_t wasm_binary_s
 /// Should be called unless @p module was passed to fizzy_instantiate.
 void fizzy_free_module(struct FizzyModule* module);
 
+bool fizzy_find_exported_function(
+    const struct FizzyModule* module, const char* name, uint32_t* out_func_idx);
```
axic, line 75: If this is done on the module that means the caller needs to do this prior to instantiation.

gumb0: I thinks it should be possible to access a module from an instance.

But C++ API has to be changed for that https://github.com/wasmx/fizzy/pull/533#discussion_r498271359

gumb0: Added now `fizzy_get_instance_module`

gumb0: Now I'm not sure again whether it would better to have an instance argument here.

axic:
Can you create a new PR to move on from this branch name?

gumb0:
Replaced by #608

PR #566, opened 2020-10-01
Rust bindings (unsafe execution)

axic:
TODO:
- [x] Cleanup the commits
- [x] Rustdoc
- [x] ~~Fizzy error enum~~ (will do this in a new PR)

### DIFF
--- a/bindings/rust/Cargo.toml
+++ b/bindings/rust/Cargo.toml
@@ -12,6 +12,9 @@ description = "Bindings to Fizzy"
 categories = ["external-ffi-bindings"]
 edition = "2018"
 
+[dev-dependencies]
+hex = "0.4.2"
+
 [build-dependencies]
 bindgen = "0.54.0"
 cmake = "0.1"
--- a/bindings/rust/src/lib.rs
+++ b/bindings/rust/src/lib.rs
@@ -3,8 +3,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
 //! This is a Rust interface to [Fizzy](https://github.com/wasmx/fizzy), a WebAssembly virtual machine.
-//!
-//! Currently it is only possible to validate inputs, but soon execution will be supported too.
 
 mod sys;
 
@@ -72,6 +70,58 @@ impl Module {
     }
 }
 
+// TODO: add a proper wrapper (enum type for Value, and Result<Value, Trap> for ExecutionResult)
+
+/// A WebAssembly value of i32/i64/f32/f64.
+pub type Value = sys::FizzyValue;
+
+/// The result of an execution.
+pub struct ExecutionResult(sys::FizzyExecutionResult);
+
+impl ExecutionResult {
+    /// True if execution has resulted in a trap.
+    pub fn trapped(&self) -> bool {
+        self.0.trapped
+    }
+
+    /// The optional return value. Only a single return value is allowed in WebAssembly 1.0.
+    pub fn value(&self) -> Option<Value> {
+        if self.0.has_value {
+            assert!(!self.0.trapped);
+            Some(self.0.value)
+        } else {
+            None
+        }
+    }
+}
+
+impl From<ExecutionResult> for sys::FizzyExecutionResult {
+    fn from(v: ExecutionResult) -> Self {
+        v.0
+    }
+}
+
+impl Instance {
+    /// Unsafe execution of a given function index `func_idx` with the given values `args`.
+    ///
+    /// An invalid index, invalid inputs, or invalid depth can cause undefined behaviour.
+    ///
+    /// The `depth` argument sets the initial call depth. Should be set to `0`.
+    ///
+    /// # Safety
+    /// This function expects a valid `func_idx`, appropriate number of `args`, and valid `depth`.
+    pub unsafe fn unsafe_execute(
+        &mut self,
+        func_idx: u32,
+        args: &[Value],
+        depth: i32,
+    ) -> ExecutionResult {
+        ExecutionResult {
+            0: sys::fizzy_execute(self.0.as_ptr(), func_idx, args.as_ptr(), depth),
+        }
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
@@ -106,5 +156,46 @@ mod tests {
         assert!(module.is_ok());
         let instance = module.unwrap().instantiate();
         assert!(instance.is_ok());
+    }
+
+    #[test]
+    fn unsafe_execute_wasm() {
+        /* wat2wasm
+          (func)
+          (func (result i32) i32.const 42)
+          (func (param i32 i32) (result i32)
+            (i32.div_u (local.get 0) (local.get 1))
+          )
+          (func unreachable)
+        */
+        let input = hex::decode("0061736d01000000010e036000006000017f60027f7f017f030504000102000a150402000b0400412a0b0700200020016e0b0300000b").unwrap();
+        let module = parse(&input);
+        assert!(module.is_ok());
+        let instance = module.unwrap().instantiate();
+        assert!(instance.is_ok());
+        let mut instance = instance.unwrap();
+
+        let result = unsafe { instance.unsafe_execute(0, &[], 0) };
+        assert!(!result.trapped());
+        assert!(!result.value().is_some());
+
+        let result = unsafe { instance.unsafe_execute(1, &[], 0) };
+        assert!(!result.trapped());
+        assert!(result.value().is_some());
+        unsafe {
+            assert_eq!(result.value().unwrap().i64, 42);
+        } // FIXME: this is actually i32
+
+        let result =
+            unsafe { instance.unsafe_execute(2, &[Value { i64: 42 }, Value { i64: 2 }], 0) };
+        assert!(!result.trapped());
+        assert!(result.value().is_some());
+        unsafe {
+            assert_eq!(result.value().unwrap().i64, 21);
+        } // FIXME: this is actually i32
+
+        let result = unsafe { instance.unsafe_execute(3, &[], 0) };
+        assert!(result.trapped());
+        assert!(!result.value().is_some());
     }
 }
--- a/bindings/rust/src/lib.rs
+++ b/bindings/rust/src/lib.rs
@@ -75,6 +75,63 @@ impl Module {
 /// A WebAssembly value of i32/i64/f32/f64.
 pub type Value = sys::FizzyValue;
 
+// NOTE: the union does not have i32
+impl Value {
+    pub fn as_i32(&self) -> i32 {
+        unsafe { self.i64 as i32 }
+    }
+    pub fn as_u32(&self) -> u32 {
+        unsafe { self.i64 as u32 }
+    }
+    pub fn as_i64(&self) -> i64 {
+        unsafe { self.i64 as i64 }
+    }
+    pub fn as_u64(&self) -> u64 {
+        unsafe { self.i64 }
+    }
+    pub fn as_f32(&self) -> f32 {
+        unsafe { self.f32 }
+    }
+    pub fn as_f64(&self) -> f64 {
+        unsafe { self.f64 }
+    }
+}
+
+impl From<i32> for Value {
+    fn from(v: i32) -> Self {
+        Value { i64: v as u64 }
+    }
+}
+
+impl From<u32> for Value {
+    fn from(v: u32) -> Self {
+        Value { i64: v as u64 }
+    }
+}
+
+impl From<i64> for Value {
+    fn from(v: i64) -> Self {
+        Value { i64: v as u64 }
+    }
+}
+
+impl From<u64> for Value {
+    fn from(v: u64) -> Self {
+        Value { i64: v }
+    }
+}
+impl From<f32> for Value {
+    fn from(v: f32) -> Self {
+        Value { f32: v }
+    }
+}
+
+impl From<f64> for Value {
+    fn from(v: f64) -> Self {
+        Value { f64: v }
+    }
+}
+
 /// The result of an execution.
 pub struct ExecutionResult(sys::FizzyExecutionResult);
 
@@ -182,17 +239,20 @@ mod tests {
         let result = unsafe { instance.unsafe_execute(1, &[], 0) };
         assert!(!result.trapped());
         assert!(result.value().is_some());
-        unsafe {
-            assert_eq!(result.value().unwrap().i64, 42);
-        } // FIXME: this is actually i32
+        assert_eq!(result.value().unwrap().as_i32(), 42);
 
+        // Explicit type specification
         let result =
-            unsafe { instance.unsafe_execute(2, &[Value { i64: 42 }, Value { i64: 2 }], 0) };
+            unsafe { instance.unsafe_execute(2, &[(42 as i32).into(), (2 as i32).into()], 0) };
         assert!(!result.trapped());
         assert!(result.value().is_some());
-        unsafe {
-            assert_eq!(result.value().unwrap().i64, 21);
-        } // FIXME: this is actually i32
+        assert_eq!(result.value().unwrap().as_i32(), 21);
+
+        // Implicit i64 types (even though the code expects i32)
+        let result = unsafe { instance.unsafe_execute(2, &[42.into(), 2.into()], 0) };
+        assert!(!result.trapped());
+        assert!(result.value().is_some());
+        assert_eq!(result.value().unwrap().as_i32(), 21);
 
         let result = unsafe { instance.unsafe_execute(3, &[], 0) };
         assert!(result.trapped());
--- a/bindings/rust/src/lib.rs
+++ b/bindings/rust/src/lib.rs
@@ -34,6 +34,44 @@ pub fn parse<T: AsRef<[u8]>>(input: &T) -> Result<Module, ()> {
     Ok(Module { 0: ptr })
 }
 
+/// An instance of a module.
+pub struct Instance(NonNull<sys::FizzyInstance>);
+
+impl Drop for Instance {
+    fn drop(&mut self) {
+        unsafe { sys::fizzy_free_instance(self.0.as_ptr()) }
+    }
+}
+
+impl Module {
+    /// Create an instance of a module.
+    // TODO: support imported functions
+    pub fn instantiate(self) -> Result<Instance, ()> {
+        if self.0.is_null() {
+            return Err(());
+        }
+        let ptr = unsafe {
+            sys::fizzy_instantiate(
+                self.0,
+                std::ptr::null(),
+                0,
+                std::ptr::null(),
+                std::ptr::null(),
+                std::ptr::null(),
+                0,
+            )
+        };
+        // Forget Module (and avoid calling drop) because it has been consumed by instantiate (even if it failed).
+        core::mem::forget(self);
+        if ptr.is_null() {
+            return Err(());
+        }
+        Ok(Instance {
+            0: unsafe { NonNull::new_unchecked(ptr) },
+        })
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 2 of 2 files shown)


bindings/rust/src/lib.rs
```diff
@@ -4,8 +4,123 @@
...
+}
+impl From<f32> for Value {
+    fn from(v: f32) -> Self {
+        Value { i64: v as u64 }
+    }
+}
+
+impl From<f64> for Value {
+    fn from(v: f64) -> Self {
+        Value { i64: v as u64 }
```
jakelang, line 114: If you just want to interpret the bits rather than do rounding, then using `v.to_bits()` is the correct option.
Also you probably know this but casting this way can go wrong, as it's changing to unsigned.

bindings/rust/src/lib.rs
```diff
@@ -4,8 +4,123 @@
```
jakelang, line 4: The conventional file layout is:
```
<use statements>
<module declarations>
<trait declarations>
<struct declarations>
<struct methods>
<trait impls>
```

bindings/rust/src/lib.rs
```diff
@@ -4,8 +4,123 @@
...
+    }
+}
+
+impl From<f64> for Value {
+    fn from(v: f64) -> Self {
+        Value { i64: v as u64 }
+    }
+}
+
+pub type ExecutionResult = sys::FizzyExecutionResult;
```
jakelang, line 118: Not sure what the contents of `ExecutionResult` are on the C++ end, but newtyping Execution result as `pub struct ExecutionResult(sys::FizzyExecutionResult);` would allow the user to work with this without consulting generated bindings. It would also show up on `cargo doc`.

bindings/rust/src/lib.rs
```diff
@@ -4,8 +4,123 @@
...
+
+impl Instance {
+    pub fn execute(&mut self, func_idx: u32, args: &[Value]) -> ExecutionResult {
```
jakelang, line 121: Ergonomics could be better if `func_idx` was replaced with the name. Users will probably be looking to call an export by name anyway.

bindings/rust/src/lib.rs
```diff
@@ -4,8 +4,123 @@
...
+    ptr: NonNull<sys::FizzyModule>,
+}
+
+impl Drop for Module {
+    fn drop(&mut self) {
+        unsafe { sys::fizzy_free_module(self.ptr.as_ptr()) }
+    }
+}
+
+pub fn parse<T: AsRef<[u8]>>(input: &T) -> Option<Module> {
```
jakelang, line 23: Same here, `Result` makes more sense, especially combined with a dedicated error type.

bindings/rust/src/lib.rs
```diff
@@ -4,8 +4,123 @@
...
+        unsafe { self.i64 as u32 }
+    }
+    pub fn as_i64(&self) -> i64 {
+        unsafe { self.i64 as i64 }
+    }
+    pub fn as_u64(&self) -> u64 {
+        unsafe { self.i64 }
+    }
+    pub fn as_f32(&self) -> f32 {
+        unsafe { self.f32 }
```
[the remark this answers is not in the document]
axic: No, just forgot to use it there. It is a union anyhow so it doesn't matter.

bindings/rust/src/lib.rs
```diff
@@ -4,8 +4,123 @@
...
+
+impl Drop for Instance {
+    fn drop(&mut self) {
+        unsafe { sys::fizzy_free_instance(self.ptr.as_ptr()) }
+    }
+}
+
+impl Module {
+    // TODO: support imported functions{
+    pub fn instantiate(self) -> Option<Instance> {
```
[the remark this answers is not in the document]
axic: Yeah, I was wondering if I could avoid defining error types (given nothing is returned), but the question mark operator is a good point.

bindings/rust/src/lib.rs
axic: There is actually `f64` in the union which I wanted to use.

bindings/rust/src/lib.rs
axic: Will check if that causes and speed regression, but yeah the newtype is a good point.

bindings/rust/src/lib.rs
axic: The C API does not yet offer anything like that and this is a direct translation as of now.

bindings/rust/src/lib.rs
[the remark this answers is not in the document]
axic: Need to have custom error type otherwise question mark operator won't work 😞

bindings/rust/src/lib.rs
```diff
@@ -4,8 +4,144 @@
...
+}
+
+impl From<ExecutionResult> for sys::FizzyExecutionResult {
+    fn from(v: ExecutionResult) -> Self {
+        v.0
+    }
+}
+
+impl Instance {
+    pub fn execute(&mut self, func_idx: u32, args: &[Value]) -> ExecutionResult {
```
axic, line 140: Should rename this `unsafe_execute` and make it unsafe. It can be used to squeeze out speed, at the cost of safety.

bindings/rust/src/lib.rs
```diff
@@ -8,11 +8,144 @@
```
gumb0, line 8: Comment above this line is outdated now

bindings/rust/src/lib.rs
```diff
@@ -8,11 +8,144 @@
...
+}
+
+pub struct ExecutionResult(sys::FizzyExecutionResult);
+
+impl ExecutionResult {
+    pub fn trapped(&self) -> bool {
+        self.0.trapped
+    }
+    pub fn value(&self) -> Option<Value> {
+        if self.0.has_value {
```
gumb0, line 128: Is this possible?
```suggestion
        if self.0.has_value { Some(self.0.value) } else { None }
```

bindings/rust/src/lib.rs
```diff
@@ -34,4 +167,54 @@ mod tests {
...
+    fn instantiate_wasm() {
+        let module = parse(&[0x00, 0x61, 0x73, 0x6d, 0x01, 0x00, 0x00, 0x00]);
+        assert!(module.is_ok());
+        let instance = module.unwrap().instantiate();
+        assert!(instance.is_ok());
+    }
+
+    #[test]
+    fn execute_wasm() {
+        let input = hex::decode("0061736d01000000010e036000006000017f60027f7f017f030504000102000a150402000b0400412a0b0700200020016e0b0300000b").unwrap();
```
gumb0, line 187: Add comment with wasm code for this

bindings/rust/src/lib.rs
```diff
@@ -8,11 +8,144 @@
...
+impl From<ExecutionResult> for sys::FizzyExecutionResult {
+    fn from(v: ExecutionResult) -> Self {
+        v.0
+    }
+}
+
+impl Instance {
+    pub unsafe fn unsafe_execute(&mut self, func_idx: u32, args: &[Value]) -> ExecutionResult {
+        ExecutionResult {
+            0: sys::fizzy_execute(self.0.as_ptr(), func_idx, args.as_ptr(), 0),
```
gumb0, line 144: You really dislike `depth` parameter :smile:

bindings/rust/src/lib.rs
axic: Yes, that works too.

bindings/rust/src/lib.rs
axic: Will add it back 😉

axic: Done.

bindings/rust/src/lib.rs
axic: It should be the same as `capi_test:execute`. Would be nice to have #593 too.

bindings/rust/src/lib.rs
```diff
@@ -8,11 +8,150 @@
...
+}
+
+pub struct ExecutionResult(sys::FizzyExecutionResult);
+
+impl ExecutionResult {
+    pub fn trapped(&self) -> bool {
+        self.0.trapped
+    }
+    pub fn value(&self) -> Option<Value> {
+        if self.0.has_value {
```
gumb0, line 128: `has_value` is valid only if not `trapped`

axic, line 128:
```suggestion
        if !self.0.trapped && self.0.has_value {
```

gumb0 (approved):
Looks ok to me, let's merge it soon.

gumb0:
Out of curiosity, is it possible to create a test for panic?

gumb0:
> Out of curiosity, is it possible to create a test for panic?

Looks like possible somehow https://stackoverflow.com/a/26470361

axic:
> Out of curiosity, is it possible to create a test for panic?

What panic would you like to test?

gumb0:
> > Out of curiosity, is it possible to create a test for panic?
> 
> What panic would you like to test?

Was thinking about passing wrong argument types to execute, not sure such test would make sense.

gumb0:
> > > Out of curiosity, is it possible to create a test for panic?
> > 
> > 
> > What panic would you like to test?
> 
> Was thinking about passing wrong argument types to execute, not sure such test would make sense.

I guess I was confused, because it's actually going to be UB for `unsafe_execute`?

axic:
> > > > Out of curiosity, is it possible to create a test for panic?
> > >
> > > What panic would you like to test?
> >
> > Was thinking about passing wrong argument types to execute, not sure such test would make sense.
>
> I guess I was confused, because it's actually going to be UB for unsafe_execute?

It is possible to test for exceptions (with `catch_unwind`). I don't think it is possible to test for crashes, as that is just a crash in the C code.

bindings/rust/src/lib.rs
axic: Added function to look it up (now that the underlying API has it).

bindings/rust/Cargo.toml
```diff
@@ -12,6 +12,9 @@ description = "Bindings to Fizzy"
 categories = ["external-ffi-bindings"]
 edition = "2018"
 
+[dependencies]
+hex = "0.4.2"
```
chfast, line 16: What kind of dependency is it? This is only required for tests.

axic: Yeah, will need to move it do `dev-dependencies`, forgot to do it last week.

bindings/rust/src/lib.rs
```diff
@@ -3,16 +3,233 @@
...
+        func_idx: u32,
+        args: &[Value],
+        depth: i32,
+    ) -> ExecutionResult {
+        ExecutionResult {
+            0: sys::fizzy_execute(self.0.as_ptr(), func_idx, args.as_ptr(), depth),
+        }
+    }
+
+    /// Find function type for a given index. Must be a valid index otherwise behaviour is undefine.
```
gumb0, line 199:
```suggestion
    /// Find function type for a given index. Must be a valid index otherwise behaviour is undefined.
```

bindings/rust/src/lib.rs
```diff
@@ -3,16 +3,233 @@
...
+        }
+        let func_idx = func_idx.unwrap();
+
+        let func_type = unsafe { self.get_function_type(func_idx) };
+        if func_type.inputs_size != args.len() {
+            return Err(());
+        }
+        // TODO: validate input types too
+
+        Ok(unsafe { self.unsafe_execute(func_idx, &[], depth) })
```
gumb0, line 229:
```suggestion
        Ok(unsafe { self.unsafe_execute(func_idx, args, depth) })
```

axic: Oh wow, need to improve the unit tests.

bindings/rust/src/lib.rs
```diff
@@ -34,4 +251,124 @@ mod tests {
...
+        let result = instance.execute("foo", &[], 0);
+        assert!(result.is_ok());
+        let result = result.unwrap();
+        assert!(!result.trapped());
+        assert!(result.value().is_some());
+        assert_eq!(result.value().unwrap().as_i32(), 42);
+
+        let result = instance.execute("bar", &[], 0);
+        assert!(result.is_err());
+
```
gumb0, line 370: You could add also an attempt to execute non-function export, like `g1`

bindings/rust/Cargo.toml
axic: Done.

bindings/rust/src/lib.rs
axic: This functionality is moved to the next PR.

bindings/rust/src/lib.rs
axic: This functionality is moved to the next PR.

bindings/rust/src/lib.rs
```diff
@@ -8,11 +8,35 @@
...
 mod sys;
 
+use std::ptr::NonNull;
+
 /// Parse and validate the input according to WebAssembly 1.0 rules. Returns true if the supplied input is valid.
 pub fn validate<T: AsRef<[u8]>>(input: T) -> bool {
     unsafe { sys::fizzy_validate(input.as_ref().as_ptr(), input.as_ref().len()) }
 }
 
+/// A parsed and validated WebAssembly 1.0 module.
```
chfast, line 17: I don't think it make sense to mention 1.0 anywhere in the comments.

bindings/rust/src/lib.rs
```diff
@@ -8,11 +8,35 @@
...
+impl Drop for Module {
+    fn drop(&mut self) {
+        unsafe { sys::fizzy_free_module(self.ptr.as_ptr()) }
+    }
+}
+
+/// Parse and validate the input according to WebAssembly 1.0 rules.
+pub fn parse<T: AsRef<[u8]>>(input: &T) -> Result<Module, ()> {
+    let ptr = unsafe { sys::fizzy_parse(input.as_ref().as_ptr(), input.as_ref().len()) };
+    if ptr.is_null() {
```
chfast, line 29: Can you use `NonNull::new()` here instead, and then check the `Option`. Should be on `unsafe` less.

bindings/rust/src/lib.rs
```diff
@@ -159,17 +216,20 @@ mod tests {
...
+        // Explicit type specification
         let result =
-            unsafe { instance.unsafe_execute(2, &[Value { i64: 42 }, Value { i64: 2 }], 0) };
+            unsafe { instance.unsafe_execute(2, &[(42 as i32).into(), (2 as i32).into()], 0) };
+        assert!(!result.trapped);
+        assert!(result.has_value);
+        assert_eq!(result.value.as_i32(), 21);
+
+        // Implicit i64 types (even though the code expects i32)
+        let result = unsafe { instance.unsafe_execute(2, &[42.into(), 2.into()], 0) };
```
chfast, line 295: What is this `.into()` doing?

bindings/rust/src/lib.rs
```diff
@@ -135,7 +135,29 @@ impl From<f64> for Value {
...
+
+impl ExecutionResult {
+    /// True if execution has resulted in a trap.
+    pub fn trapped(&self) -> bool {
+        self.0.trapped
+    }
+
+    /// The optional return value. Only a single return value is allowed in WebAssembly 1.0.
+    pub fn value(&self) -> Option<Value> {
+        if !self.0.trapped && self.0.has_value {
```
chfast, line 146: Documentation says otherwise, but this is too defensive. It is guaranteed that `has_value` if `false` if `trapped` is `true`.

bindings/rust/src/lib.rs
```diff
@@ -14,13 +14,11 @@ pub fn validate<T: AsRef<[u8]>>(input: T) -> bool {
 }
 
 /// A parsed and validated WebAssembly 1.0 module.
-pub struct Module {
-    ptr: NonNull<sys::FizzyModule>,
-}
+pub struct Module(*const sys::FizzyModule);
```
chfast, line 18: Why can't it be done like that from the beginning?

bindings/rust/src/lib.rs
```diff
@@ -213,6 +229,35 @@ mod tests {
         assert!(instance.is_ok());
     }
 
+    #[test]
```
chfast, line 232: Do you think it is good idea to have tests in the same file? We have 80:20 test to implementation ratio and you will have mostly tests here.

bindings/rust/src/lib.rs
axic: `Module` is not using the `NonNull` wrapping.

axic: I'm trying to remember. I think it can use `NonNull` again since I'm using `forget()`.

bindings/rust/src/lib.rs
axic: Explicitly invoking the `from` trait.

bindings/rust/src/lib.rs
axic: I'll squash those commits in, just didn't do it yet.

bindings/rust/src/lib.rs
axic: Well, this is the design pattern in Rust. It is possible to deviate, but this has been the usual standard.

bindings/rust/src/lib.rs
gumb0: Converts integers to `Value`, where `Value` is deduced from function parameter type, I guess

bindings/rust/src/lib.rs
axic: Ok, the reason I'm using `NonNull` for `Module` is because [NonNull requires *mut](https://doc.rust-lang.org/std/ptr/struct.NonNull.html#method.new), while Module is *const.

bindings/rust/src/lib.rs
axic: @gumb0 suggested this. Where is it guaranteed?

bindings/rust/src/lib.rs
axic: We use the "1.0" in the C API too.

bindings/rust/src/lib.rs
axic: The first commit had it with `NonNull` and doing a typecast..

bindings/rust/src/lib.rs
gumb0: > Documentation says otherwise, but this is too defensive. It is guaranteed that `has_value` if `false` if `trapped` is `true`.

Here it's the opposite - when `trapped` is `false`, we have to check `has_value`.

What I suggested was changing `if has_value` -> `if !trapped && has_value`

axic: > What I suggested was changing if has_value -> if !trapped && has_value

This is what the code has now.

bindings/rust/src/lib.rs
chfast: I get that, but in my opinion this looks good only on paper.

bindings/rust/src/lib.rs
chfast: Sounds good.

bindings/rust/src/lib.rs
chfast: The `ExecuteResult` constructor guarantees it. If you want to keep it, swap the checks: `has_value && !trapped`.

bindings/rust/src/lib.rs
chfast: I think this is a mistake. We should not randomly in many places in comments information what version of WebAssembly is supported.

bindings/rust/src/lib.rs
chfast: I added this comment to the first commit, later outdated. So this is resolved.

bindings/rust/src/lib.rs
axic: I can also add the trapped as an assertion, which is only present in debug builds.

axic: > The ExecuteResult constructor guarantees it.

The C API has no constructor guarantee for it though?

gumb0: > > The ExecuteResult constructor guarantees it.
> 
> The C API has no constructor guarantee for it though?

No constructor guarantees but `FizzyExecutionResult wrap(const fizzy::ExecutionResult&)` just copies all three fields always.

bindings/rust/src/lib.rs
axic: I suggest we clean up documentation at once (Rust, C, C++), if we want to remove this "1.0" stuff.

bindings/rust/src/lib.rs
chfast: > I can also add the trapped as an assertion, which is only present in debug builds:

This variant looks good.

bindings/rust/src/lib.rs
chfast: This is not my preference, but ok.

bindings/rust/src/lib.rs
gumb0: I'm ok with changing `trapped` check to assertion.

gumb0 (approved):

axic:
@chfast @gumb0 made the assertion change

chfast (approved):

PR #567, opened 2020-10-01
Mark instruction helpers appropriately as constexpr

axic:
Pulled out of #562.

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -101,7 +101,7 @@ inline T load(bytes_view input, size_t offset) noexcept
 }
 
 template <typename DstT, typename SrcT>
-inline DstT extend(SrcT in) noexcept
+inline constexpr DstT extend(SrcT in) noexcept
 {
     if constexpr (std::is_same_v<SrcT, DstT>)
         return in;
@@ -176,7 +176,7 @@ inline bool load_from_memory(
 }
 
 template <typename DstT>
-inline DstT shrink(Value value) noexcept
+inline constexpr DstT shrink(Value value) noexcept
 {
     if constexpr (std::is_floating_point_v<DstT>)
         return value.as<DstT>();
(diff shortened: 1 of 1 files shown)


lib/fizzy/execute.cpp
```diff
@@ -423,7 +423,7 @@ __attribute__((no_sanitize("float-divide-by-zero"))) inline constexpr T fdiv(T a
 }
 
 template <typename T>
-inline constexpr T fmin(T a, T b) noexcept
+inline T fmin(T a, T b) noexcept
```
axic, line 426: Both `fmin` and `fmax` use `std::signbit` which is not constexpr: https://en.cppreference.com/w/cpp/numeric/math/signbit

axic:
@chfast could you benchmark this?

axic:
I don't see any change in any of the functions touched by this. Some stuff is moved around, so can't tell much with benchmarking.

chfast:
> I don't see any change in any of the functions touched by this. Some stuff is moved around, so can't tell much with benchmarking.

The `execute.cpp.s` is identical in both cases. As it should be.

chfast (approved):

axic:
> > I don't see any change in any of the functions touched by this. Some stuff is moved around, so can't tell much with benchmarking.
>
> The execute.cpp.s is identical in both cases. As it should be.

It wasn't identical for me. Let me check again.

axic:
> > > I don't see any change in any of the functions touched by this. Some stuff is moved around, so can't tell much with benchmarking.
> >
> > The execute.cpp.s is identical in both cases. As it should be.
>
> It wasn't identical for me. Let me check again.
>
> Still not identical, interestingly the resulting .S is the same number of characters, but the content is different.

I totally forgot to consider optimiser settings when comparing the output. With the settings of the release build indeed it is identical.

PR #568, opened 2020-10-02
Centralized call depth control

chfast:
The goal is the remove the duty of bumping depth by host functions. It should be bumped only in execute() or any other suitable place.

This change has a defect that host functions now have call depth limit greater by 1.

### DIFF
--- a/test/unittests/execute_call_test.cpp
+++ b/test/unittests/execute_call_test.cpp
@@ -639,8 +639,8 @@ TEST(execute_call, call_max_depth)
 
     auto instance = instantiate(parse(bin));
 
-    EXPECT_THAT(execute(*instance, 0, {}, MaxDepth), Result(42));
-    EXPECT_THAT(execute(*instance, 1, {}, MaxDepth), Traps());
+    EXPECT_THAT(execute(*instance, 0, {}, MaxDepth - 1), Result(42));
+    EXPECT_THAT(execute(*instance, 1, {}, MaxDepth - 1), Traps());
 }
 
 TEST(execute_call, execute_imported_max_depth)
@@ -662,10 +662,10 @@ TEST(execute_call, execute_imported_max_depth)
 
     auto instance = instantiate(std::move(module), {{host_foo, host_foo_type}});
 
-    EXPECT_THAT(execute(*instance, 0, {}, MaxDepth), Result());
-    EXPECT_THAT(execute(*instance, 1, {}, MaxDepth), Result());
-    EXPECT_THAT(execute(*instance, 0, {}, MaxDepth + 1), Traps());
-    EXPECT_THAT(execute(*instance, 1, {}, MaxDepth + 1), Traps());
+    EXPECT_THAT(execute(*instance, 0, {}, MaxDepth - 1), Result());
+    EXPECT_THAT(execute(*instance, 1, {}, MaxDepth - 1), Result());
+    EXPECT_THAT(execute(*instance, 0, {}, MaxDepth), Traps());
+    EXPECT_THAT(execute(*instance, 1, {}, MaxDepth), Traps());
 }
 
 TEST(execute_call, imported_function_from_another_module_max_depth)
@@ -702,8 +702,8 @@ TEST(execute_call, imported_function_from_another_module_max_depth)
 
     auto instance2 = instantiate(std::move(module2), {{sub, instance1->module->typesec[0]}});
 
-    EXPECT_THAT(execute(*instance2, 2, {}, MaxDepth - 1), Traps());
-    EXPECT_THAT(execute(*instance2, 3, {}, MaxDepth - 1), Result());
+    EXPECT_THAT(execute(*instance2, 2, {}, MaxDepth - 2), Traps());
+    EXPECT_THAT(execute(*instance2, 3, {}, MaxDepth - 2), Result());
 }
 
 // A regression test for incorrect number of arguments passed to a call.
@@ -739,7 +739,7 @@ TEST(execute_call, call_imported_infinite_recursion)
     const auto module = parse(wasm);
     auto host_foo = [](Instance& instance, const Value* args, int depth) -> ExecutionResult {
         EXPECT_LE(depth, MaxDepth);
-        return execute(instance, 0, args, depth + 1);
+        return execute(instance, 0, args, depth);
     };
     const auto host_foo_type = module->typesec[0];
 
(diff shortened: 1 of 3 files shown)


test/unittests/execute_call_test.cpp
```diff
@@ -503,20 +503,19 @@ TEST(execute_call, call_nonempty_stack)
...
     */
-    const auto wasm = from_hex(
-        "0061736d010000000105016000017f020b01036d6f6403666f6f0000030201000a0601040010000b");
+    const auto wasm = from_hex("0061736d010000000105016000017f020b01036d6f6403666f6f0000");
 
     const auto module = parse(wasm);
     auto host_foo = [](Instance& instance, span<const Value>, int depth) -> ExecutionResult {
-        return execute(instance, 0, {}, depth + 1);
+        EXPECT_LE(depth, MaxDepth);
+        return execute(instance, 0, {}, depth);
```
gumb0, line 518: It seems there's no way to make this infinite recursion work correctly without having different meaning of `depth` in host function and in `execute`.

Or unless we introduce a special `execute` to be called only from host functions, that would bump `depth`.

chfast: I managed to unify the meaning, but it is rather "the depth of the next call" or we start with depth -1.

gumb0: Still I don't think it's unified, it's "the depth of the next call" for imported functions, but "current depth" for module functions.
(Note that imported function can be another wasm function from another module, if it has it's own `call`s, their depth would be additionally incremented)

In other words, for a module like
```
    (import "mod" "foo" (func))
    (func (result i32))
```
when calling `execute` with depth 0, first function would receive depth 1, while second would receive depth 0, which might be weird.

axic: If you call the import directly it also receives depth 0. If you call it indirectly via the function, the logically it cannot start at depth 0.

chfast: I can confirm the issue reported by @gumb0 where you call a function imported from other module. Then for the first executed function we go though `execute()` twice and depth is bumped twice. See the "test: Extend imported function call test" commit. This is issue on its own because user is required to create a lambda to link two modules. I think I may have a solution for that on API level.

This also seem artificial issue, because you cannot export imported function, can you?

I fail to understand other reported issues. Can you maybe provide a test cases showing the issue?

gumb0: > This also seem artificial issue, because you cannot export imported function, can you?

You can, for example in this test https://github.com/wasmx/fizzy/blob/e66e20daef2ac96e1697391d5a9bf7ce3e9e76de/test/unittests/api_test.cpp#L255

chfast: Good to know. Anyway, still seems fixable. I'm just not sure I want to mess up with the calling helpers right now.

test/unittests/execute_call_test.cpp
```diff
@@ -533,8 +535,54 @@ TEST(execute_call, call_max_depth)
...
+    (import "mod" "foo" (func))
+    (func)
+    */
+
+    const auto wasm =
+        from_hex("0061736d01000000010401600000020b01036d6f6403666f6f0000030201000a040102000b");
+
+    const auto module = parse(wasm);
+    auto host_foo = [](Instance& /*instance*/, span<const Value>, int depth) -> ExecutionResult {
+        if (depth + 1 > MaxDepth)
```
gumb0, line 574: I would argue that this is logical thing for a host function to do, because that's what `execute` does

axic: But `execute` does this check prior to the host function is executed. Or I'm missing something in the mess we have.

gumb0: Alright, true, this is not needed, disregard this test for now.

test/unittests/execute_call_test.cpp
```diff
@@ -533,10 +535,94 @@ TEST(execute_call, call_max_depth)
...
+    ASSERT_TRUE(func_idx.has_value());
+
+    auto sub = [&instance1, func_idx](
+                   Instance&, span<const Value> args, int depth) -> ExecutionResult {
+        return fizzy::execute(*instance1, *func_idx, args.data(), depth);
+    };
+
+    auto instance2 = instantiate(module2, {{sub, module1.typesec[0]}});
+
+    EXPECT_THAT(execute(*instance2, 2, {}, MaxDepth - 2), Traps());
```
gumb0, line 623: This test shows different result for calling an imported function from another module vs calling a local function

test/unittests/execute_call_test.cpp
gumb0: > Still I don't think it's unified, it's "the depth of the next call" for imported functions, but "current depth" for module functions.

Couldn't demonstrate it in the test yet, but consider this: calling `execute(..., MaxDepth)` will always trap, but `depth == MaxDepth`  is possible inside host function (so depth currently can have two different meanings, I argue)

chfast: > Couldn't demonstrate it in the test yet, but consider this: calling `execute(..., MaxDepth)` will always trap, but `depth == MaxDepth` is possible inside host function (so depth currently can have two different meanings, I argue)

That's correct. To execute on `MaxDepth` you need to pass `execute(..., MaxDepth-1)`. I agree this is confusing part, i.e. what `depth` means in which context.

But where I'm going with this is to replace `depth` with opaque `Context` which you can only create and pass around. So you will not be able to inspect the depth. The PoC of that is in https://github.com/wasmx/fizzy/pull/529/commits/5f08ee277a7959252b9dfd5763ce66dfc6b89b5d. But I'd like to be more prepared to this change.

test/unittests/execute_call_test.cpp
```diff
@@ -533,10 +535,93 @@ TEST(execute_call, call_max_depth)
...
+    (import "mod" "foo" (func))
+    (func)
+    */
+
+    const auto wasm =
+        from_hex("0061736d01000000010401600000020b01036d6f6403666f6f0000030201000a040102000b");
+
+    const auto module = parse(wasm);
+    auto host_foo = [](Instance& /*instance*/, span<const Value>, int depth) -> ExecutionResult {
+        EXPECT_LE(depth, MaxDepth);
```
gumb0, line 574: This shows that `depth` can be `MaxDepth` in host function and the call would return successfully.

PR #571, opened 2020-10-05
test: Add more call tests

chfast:


test/unittests/execute_call_test.cpp
```diff
@@ -58,6 +58,34 @@ TEST(execute_call, call_with_arguments)
...
 }
 
+TEST(execute_call, call_shared_stack_space)
+{
+    /* wat2wasm
+    (module
+      (func $double (param i32) (result i64)
+        local.get 0
+        i64.extend_i32_u
+        local.get 0
```
axic, line 68: Is the point of this test that the same local is used twice?

test/unittests/execute_call_test.cpp
```diff
@@ -205,6 +233,36 @@ TEST(execute_call, call_indirect_uninited_table)
     EXPECT_THAT(execute(module, 3, {4}), Traps());
 }
 
+TEST(execute_call, call_indirect_shared_stack_space)
```
axic, line 236: Seems to be a dup of `imported_functions_call_indirect` ?

chfast: Both are simplified variants of it. They can detect an implementation defect in #572, so I keep them as they are.

test/unittests/execute_call_test.cpp
```diff
@@ -58,6 +58,34 @@ TEST(execute_call, call_with_arguments)
     EXPECT_THAT(execute(parse(wasm), 1, {}), Result(4));
 }
 
+TEST(execute_call, call_shared_stack_space)
+{
+    /* wat2wasm
+    (module
+      (func $double (param i32) (result i64)
```
gumb0, line 65: It's hard to guess what these tests check... Is `i64` result required here? If not, maybe make it `i32` for simplicity and more focused test?

gumb0 (approved):

PR #575, opened 2020-10-05
refactor: Allocate module dynamically

gumb0:
This will simplify C API implementation part, allowing to handle modules and instances in the same manner.

PR #576, opened 2020-10-05
C API execute with dynamically allocated module

gumb0:
Alternative to #533 based on #575

### DIFF
--- a/include/fizzy/fizzy.h
+++ b/include/fizzy/fizzy.h
@@ -11,7 +11,103 @@
 extern "C" {
 #endif
 
+/// The opaque data type representing a module.
+typedef struct FizzyModule FizzyModule;
+
+/// The opaque data type representing an instance (instantiated module).
+typedef struct FizzyInstance FizzyInstance;
+
+/// The data type representing numeric values.
+///
+/// i64 member is used to represent values of both i32 and i64 type.
+union FizzyValue
+{
+    uint64_t i64;
+    float f32;
+    double f64;
+};
+
+/// Result of execution of a function.
+typedef struct FizzyExecutionResult
+{
+    /// Whether execution ended with a trap.
+    bool trapped;
+    /// Whether function returned a value. Valid only if trapped equals false.
+    bool has_value;
+    /// Value returned from a function.
+    /// Valid only if trapped equals false and has_value equals true.
+    union FizzyValue value;
+} FizzyExecutionResult;
+
+
+/// Pointer to external function.
+///
+/// @param context      Opaque pointer to execution context.
+/// @param instance     Pointer to module instance.
+/// @param args         Pointer to the argument array. Can be NULL iff args_size equals 0.
+/// @param args_size    Size of the  argument array.
+/// @param depth        Call stack depth.
+typedef FizzyExecutionResult (*FizzyExternalFn)(void* context, FizzyInstance* instance,
+    const union FizzyValue* args, size_t args_size, int depth);
+
+/// External function.
+typedef struct FizzyExternalFunction
+{
+    // TODO function type
+
+    /// Pointer to function.
+    FizzyExternalFn function;
+    /// Opaque pointer to execution context, that will be passed to function.
+    void* context;
+} FizzyExternalFunction;
+
+/// Validate binary module.
 bool fizzy_validate(const uint8_t* wasm_binary, size_t wasm_binary_size);
+
+/// Parse binary module.
+///
+/// @returns non-NULL pointer to module in case of success, NULL otherwise.
+const FizzyModule* fizzy_parse(const uint8_t* wasm_binary, size_t wasm_binary_size);
+
+/// Free resources associated with the module.
+///
+/// Should be called unless @p module was passed to fizzy_instantiate.
+/// If passed pointer is NULL, has no effect.
+void fizzy_free_module(const FizzyModule* module);
+
+/// Instantiate a module.
+/// Takes ownership of module, i.e. @p module is invalidated after this call.
+///
+/// @param      module                   Pointer to module.
+/// @param      imported_functions       Pointer to the imported function array. Can be NULL iff
+///                                      imported_functions_size equals 0.
+/// @param      imported_functions_size  Size of the imported function array. Can be zero.
+/// @returns    non-NULL pointer to instance in case of success, NULL otherwise.
+///
+/// @note
+/// Function expects @a imported_functions to be in the order of imports defined in the module.
+/// No validation is done on the number of functions passed in, nor on their order.
+/// When number of passed functions or their order is different from the one defined by the
+/// module, behaviour is undefined.
+FizzyInstance* fizzy_instantiate(const FizzyModule* module,
+    const FizzyExternalFunction* imported_functions, size_t imported_functions_size);
+
+/// Free resources associated with the instance.
+/// If passed pointer is NULL, has no effect.
+void fizzy_free_instance(FizzyInstance* instance);
+
+/// Execute module function.
+///
+/// @param instance     Pointer to module instance.
+/// @param args         Pointer to the argument array. Can be NULL if function has 0 inputs.
+/// @param depth        Call stack depth.
+///
+/// @note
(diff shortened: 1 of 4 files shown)


include/fizzy/fizzy.h
```diff
@@ -11,8 +11,86 @@
 extern "C" {
 #endif
 
+/// The opaque data type representing a module.
+struct FizzyModule;
```
chfast, line 15:
```suggestion
typedef struct FizzyModule FizzyModule;
```

I think it will be easier to consistency use C types without `struct` prefix.

include/fizzy/fizzy.h
```diff
@@ -11,8 +11,86 @@
...
+    float f32;
+    double f64;
+};
+
+/// Result of execution of a function.
+typedef struct FizzyExecutionResult
+{
+    /// Whether execution ended with a trap.
+    bool trapped;
+    /// Whether function returned a value. Valid only if trapped == false
```
axic, line 35:
```suggestion
    /// Whether function returned a value. Valid only if trapped == false.
```

include/fizzy/fizzy.h
```diff
@@ -11,8 +11,86 @@
...
+    bool has_value;
+    /// Value returned from a function. Valid only if has_value == true
```
axic, line 37:
```suggestion
    /// Value returned from a function. Valid only if has_value equals true.
```

Can this be true even if trap is true?

include/fizzy/fizzy.h
```diff
@@ -11,8 +11,86 @@
...
+    union FizzyValue value;
+} FizzyExecutionResult;
+
+
+/// Pointer to external function.
+///
+/// @param context      Opaque pointer to execution context.
+/// @param instance     Pointer to module instance.
+/// @param args         Pointer to the argument array. Can be NULL if function has 0 inputs.
```
axic, line 46:
```suggestion
/// @param args         Pointer to the argument array. Can be NULL if args_size equals 0.
```

include/fizzy/fizzy.h
```diff
@@ -11,8 +11,86 @@
...
+///
+/// Should be called unless @p module was passed to fizzy_instantiate.
+void fizzy_free_module(const struct FizzyModule* module);
+
+/// Instantiate a module.
+/// Takes ownership of module, i.e. @p module is invalidated after this call.
+///
+/// @param module                   Pointer to module.
+/// @param imported_functions       Pointer to the imported function array.
+/// @param imported_functions_size  Size of the imported function array.
```
axic, line 79:
```suggestion
/// @param imported_functions_size  Size of the imported function array. Can be zero.
```

include/fizzy/fizzy.h
```diff
@@ -11,8 +11,86 @@
...
+/// Free resources associated with the module.
+///
+/// Should be called unless @p module was passed to fizzy_instantiate.
+void fizzy_free_module(const struct FizzyModule* module);
+
+/// Instantiate a module.
+/// Takes ownership of module, i.e. @p module is invalidated after this call.
+///
+/// @param module                   Pointer to module.
+/// @param imported_functions       Pointer to the imported function array.
```
axic, line 78: Can this be null if `imported_functions_size == 0`?

include/fizzy/fizzy.h
```diff
@@ -11,8 +11,86 @@
...
+struct FizzyInstance* fizzy_instantiate(const struct FizzyModule* module,
+    const struct FizzyExternalFunction* imported_functions, size_t imported_functions_size);
+
+/// Free resources associated with the instance.
+void fizzy_free_instance(struct FizzyInstance* instance);
+
+/// Execute module function.
+///
+/// @param instance     Pointer to module instance.
+/// @param args         Pointer to the argument array. Can be NULL if function has 0 inputs.
```
axic, line 89: Should mention that no validation is done on right amount of args are passed and hence could result in crashes aka unsafe api.

include/fizzy/fizzy.h
axic: > Can this be true even if trap is true?

@gumb0 ?

lib/fizzy/capi.cpp
```diff
@@ -2,8 +2,79 @@
...
+    else
+        return unwrap(result.value);
+}
+
+inline auto unwrap(FizzyExternalFn func, void* context) noexcept
+{
+    return [func, context](fizzy::Instance& instance, fizzy::span<const fizzy::Value> args,
+               int depth) noexcept -> fizzy::ExecutionResult {
+        const auto result = func(context, wrap(&instance), wrap(args.data()), args.size(), depth);
+        return unwrap(result);
```
axic, line 74: This takes `result` as a reference. Will that reference be valid after exiting this lambda?

include/fizzy/fizzy.h
```diff
@@ -11,8 +11,98 @@
...
+    void* context;
+} FizzyExternalFunction;
+
+/// Validate binary module.
 bool fizzy_validate(const uint8_t* wasm_binary, size_t wasm_binary_size);
 
+/// Parse binary module.
+const FizzyModule* fizzy_parse(const uint8_t* wasm_binary, size_t wasm_binary_size);
+
+/// Free resources associated with the module.
```
axic, line 69: Perhaps mention `module` must be non-null.

include/fizzy/fizzy.h
```diff
@@ -11,8 +11,98 @@
...
+    /// Pointer to function.
+    FizzyExternalFn function;
+    /// Opaque pointer to execution context, that will be passed to function.
+    void* context;
+} FizzyExternalFunction;
+
+/// Validate binary module.
 bool fizzy_validate(const uint8_t* wasm_binary, size_t wasm_binary_size);
 
+/// Parse binary module.
```
axic, line 66: Perhaps mention this returns a non-null module pointer on success.

include/fizzy/fizzy.h
axic: And the same comments apply to instantiate.

lib/fizzy/capi.cpp
gumb0: Both `wrap` and `unwrap` for result are doing a copy

axic:
Looks good otherwise.

include/fizzy/fizzy.h
gumb0: This calles `delete`, so is guaranteed to be no-op for null pointer. I'll mention this and add a test.

include/fizzy/fizzy.h
```diff
@@ -64,20 +65,24 @@ typedef struct FizzyExternalFunction
...
 ///
-/// @param module                   Pointer to module.
-/// @param imported_functions       Pointer to the imported function array. Can be NULL iff
-///                                 imported_functions_size equals 0.
-/// @param imported_functions_size  Size of the imported function array. Can be zero.
+/// @param      module                   Pointer to module.
+/// @param      imported_functions       Pointer to the imported function array. Can be NULL iff
+///                                      imported_functions_size equals 0.
+/// @param      imported_functions_size  Size of the imported function array. Can be zero.
+/// @returns    non-null pointer to instance in case of success, NULL otherwise.
```
axic, line 85: Nitpick:
```suggestion
/// @returns    non-NULL pointer to instance in case of success, NULL otherwise.
```

include/fizzy/fizzy.h
```diff
@@ -64,20 +65,24 @@ typedef struct FizzyExternalFunction
 bool fizzy_validate(const uint8_t* wasm_binary, size_t wasm_binary_size);
 
 /// Parse binary module.
+///
+/// @returns non-null pointer to module in case of success, NULL otherwise.
```
axic, line 69: Nitpick:
```suggestion
/// @returns non-NULL pointer to module in case of success, NULL otherwise.
```

include/fizzy/fizzy.h
gumb0: Clarified in comment

axic (approved):
Please squash and add a test for `free_module(nullptr)` and `free_instance(nullptr)`.

chfast (approved):

PR #578, opened 2020-10-05
Instructions implementation cleanup

chfast:


lib/fizzy/execute.cpp
```diff
@@ -1034,39 +1034,36 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, const Value* args,
...
         case Instr::i32_div_s:
         {
-            auto const rhs = stack[0].as<int32_t>();
-            auto const lhs = stack[1].as<int32_t>();
+            const auto rhs = stack.pop().as<int32_t>();
+            const auto lhs = stack.top().as<int32_t>();
             if (rhs == 0 || (lhs == std::numeric_limits<int32_t>::min() && rhs == -1))
                 goto trap;
-            binary_op(stack, div<int32_t>);
+            stack.top() = div(lhs, rhs);
```
gumb0, line 1041: Is this much better?

`binary_op` is doing the same, and also with this change (and similar changes below) we're somewhat incosistent in handling traps: `call` that resulted in a trap leaves arguments on stack, other instructions that trap consume (some of) their arguments.

chfast: This is simpler in a sense the stack items are accessed only once. Compiler can generate better code and the size of `execute()` is smaller.
Secondly, this removes the usage of `stack[]` syntax so `operator[]` overloading can be dropped.

The rest I don't really care about. If this can be implemented in more readable form with the above restrictions I'm open for suggestions.

gumb0 (approved):

lib/fizzy/execute.cpp
chfast: BTW, there are some other `div`/`rem` instructions remaining using `binary_op()`. I wasn't pushing for consistency here, but they can be revisit.

Other approach is to wrap division instruction with a bigger helpers returning `bool` for trap (as we have in other cases).

chfast:
I cannot benchmark it at this point because no matter what is changed in `execute()` the speed is the same all slower.

lib/fizzy/execute.cpp
gumb0: I'm fine with leaving it like this.

chfast:
The benchmarks vary by ±3% (FP cases can be way off, but in other runs the are ±0%).

```
fizzy/execute/blake2b/512_bytes_rounds_1_mean                     -0.0054         -0.0053            77            77            77            77
fizzy/execute/blake2b/512_bytes_rounds_16_mean                    -0.0050         -0.0050          1164          1158          1164          1158
fizzy/execute/ecpairing/onepoint_mean                             -0.0257         -0.0257        384289        374419        384291        374422
fizzy/execute/keccak256/512_bytes_rounds_1_mean                   +0.0301         +0.0301            94            97            94            97
fizzy/execute/keccak256/512_bytes_rounds_16_mean                  +0.0186         +0.0186          1387          1413          1387          1413
fizzy/execute/memset/256_bytes_mean                               +0.0044         +0.0044             6             6             6             6
fizzy/execute/memset/60000_bytes_mean                             +0.0087         +0.0087          1374          1386          1374          1386
fizzy/execute/mul256_opt0/input1_mean                             +0.0031         +0.0031            25            25            25            25
fizzy/execute/ramanujan_pi/33_runs_mean                           +0.0297         +0.0297           118           122           118           122
fizzy/execute/sha1/512_bytes_rounds_1_mean                        -0.0153         -0.0153            84            82            84            82
fizzy/execute/sha1/512_bytes_rounds_16_mean                       -0.0211         -0.0210          1163          1139          1163          1139
fizzy/execute/sha256/512_bytes_rounds_1_mean                      -0.0063         -0.0062            84            84            84            84
fizzy/execute/sha256/512_bytes_rounds_16_mean                     -0.0069         -0.0069          1164          1156          1164          1156
fizzy/execute/taylor_pi/pi_1000000_runs_mean                      +0.2022         +0.2022         40031         48127         40032         48127
fizzy/execute/micro/eli_interpreter/exec105_mean                  +0.0080         +0.0080             4             4             4             4
fizzy/execute/micro/factorial/20_mean                             +0.0273         +0.0273             1             1             1             1
fizzy/execute/micro/fibonacci/24_mean                             -0.0069         -0.0069          4957          4923          4957          4923
fizzy/execute/micro/host_adler32/1_mean                           -0.0028         -0.0028             0             0             0             0
fizzy/execute/micro/host_adler32/1000_mean                        +0.0031         +0.0031            29            29            29            29
fizzy/execute/micro/spinner/1_mean                                -0.0556         -0.0556             0             0             0             0
fizzy/execute/micro/spinner/1000_mean                             -0.0272         -0.0272             9             9             9             9
```

lib/fizzy/execute.cpp
```diff
@@ -830,8 +830,8 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, const Value* args,
         }
         case Instr::i32_eqz:
         {
-            const auto value = stack.pop().as<uint32_t>();
-            stack.push(uint32_t{value == 0});
+            const auto value = stack.top().as<uint32_t>();
+            stack.top() = uint32_t{value == 0};
```
axic, line 834: Why keep `value` as a variable here when in `i64_eqz` it is a single line?

chfast: Because it was like that.

axic: How stuff was before and how it is now didn't stop you here: https://github.com/wasmx/fizzy/pull/578/files#diff-da8232128fc88a65268ab30bf8e8abdcL1409 :wink:

lib/fizzy/execute.cpp
```diff
@@ -1145,42 +1143,41 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, const Value* args,
...
-            auto const lhs = stack[1].as<int64_t>();
+            const auto lhs = stack.top().as<int64_t>();
             if (lhs == std::numeric_limits<int64_t>::min() && rhs == -1)
-            {
-                stack.pop();
                 stack.top() = 0;
-            }
             else
-                binary_op(stack, rem<int64_t>);
+                stack.top() = rem(lhs, rhs);
```
axic, line 1171: Instead of keeping logic here, alternative solution is to introduce `rem_s` and `rem_u` which does the wasm logic and keep using `binary_op`. This would be in-line with how bitops (like clz/ctz) work.

axic: Ah never mind, we kept these separate because they trap.

axic:
I'd merge the last two commits.

axic (approved):
I'd squash the last two commits before merge.

lib/fizzy/execute.cpp
chfast: > Other approach is to wrap division instruction with a bigger helpers returning bool for trap (as we have in other cases).

But I'd leave it for some other time because it is also worth to compare this version with that version.

gumb0 (approved):

PR #580, opened 2020-10-06
Extend call depth tests

chfast:


test/unittests/execute_call_test.cpp
```diff
@@ -557,19 +557,18 @@ TEST(execute_call, call_nonempty_stack)
...
 }
 
+constexpr int MaxDepth = 512;
+
 TEST(execute_call, call_imported_infinite_recursion)
 {
     /* wat2wasm
     (import "mod" "foo" (func (result i32)))
-    (func (result i32)
-      call 0
```
gumb0, line 565: Why didn't you add it as a new test, but modified this one? This was testing `call` instruction and was in `execute_call` suite, now it goes another code path...

test/unittests/execute_call_test.cpp
```diff
@@ -530,8 +530,8 @@ TEST(execute_call, call_max_depth)
     const auto module = parse(bin);
     auto instance = instantiate(module);
 
-    EXPECT_THAT(execute(*instance, 0, {}, 2048), Result(42));
-    EXPECT_THAT(execute(*instance, 1, {}, 2048), Traps());
+    EXPECT_THAT(execute(*instance, 0, {}, 512), Result(42));
```
gumb0, line 533: This could you `MaxDepth` constant from below, too.

test/unittests/execute_call_test.cpp
axic: Actually I think this code can remain, the point of the change was to assert in the host function. Would just need to check for `MaxDepth - 1` there.

axic: Actually this was calling function 0 which is the import. Probably we should duplicate this to have both cases covered.

chfast (commented):
LGTM

test/unittests/execute_call_test.cpp
```diff
@@ -518,6 +518,9 @@ TEST(execute_call, call_indirect_infinite_recursion)
     EXPECT_TRUE(execute(module, 0, {}).trapped);
 }
 
+constexpr int MaxDepth = 2048;
+static_assert(MaxDepth == CallStackLimit);
```
axic, line 522: I think we can remove this now, given the tests should exactly capture it.

axic:
@chfast I think you did not see the `call_via_*` tests.

chfast (commented):
LGTM

axic (approved):

PR #581, opened 2020-10-06
bench: move module and imports in FizzyEngine::instantiate

gumb0:
Clang 10 build with LTO
```
fizzy/instantiate/blake2b_mean                                     -0.0239         -0.0245            54            53            54            52
fizzy/instantiate/ecpairing_mean                                   -0.0485         -0.0403          2838          2701          2839          2724
fizzy/instantiate/keccak256_mean                                   -0.0375         -0.0315            94            90            94            91
fizzy/instantiate/memset_mean                                      -0.0617         -0.0592            18            17            18            17
fizzy/instantiate/mul256_opt0_mean                                 -0.0645         -0.0613            22            20            22            20
fizzy/instantiate/ramanujan_pi_mean                                -0.0103         -0.0079            57            56            57            56
fizzy/instantiate/sha1_mean                                        -0.0218         -0.0205            84            82            84            82
fizzy/instantiate/sha256_mean                                      -0.0218         -0.0209           137           134           137           134
fizzy/instantiate/taylor_pi_mean                                   -0.0674         -0.0671            12            11            12            11
fizzy/instantiate/micro/eli_interpreter_mean                       -0.0433         -0.0424            15            14            15            14
fizzy/instantiate/micro/factorial_mean                             -0.1114         -0.1120             3             2             3             2
fizzy/instantiate/micro/fibonacci_mean                             -0.1179         -0.1188             3             3             3             3
fizzy/instantiate/micro/host_adler32_mean                          -0.1361         -0.1366             8             7             8             7
fizzy/instantiate/micro/spinner_mean                               -0.1040         -0.1041             3             2             3             2
fizzy/instantiate/stress/guido-fuzzer-find-1_mean                  -0.0301         -0.0300           327           317           327           317

```

PR #582, opened 2020-10-06
test: use move semantics for instantiate in most tests (except unit tests)

axic:


test/testfloat/testfloat.cpp
```diff
@@ -401,7 +401,7 @@ int main(int argc, const char* argv[])
 
         const auto func = from_name(function_name);
 
-        auto instance = fizzy::instantiate(fizzy::parse(get_wasm_code()));
+        auto instance = fizzy::instantiate(std::move(fizzy::parse(get_wasm_code())));
```
chfast, line 404: This is not needed. The returned object is rvalue reference already.

test/utils/fizzy_engine.cpp
```diff
@@ -90,7 +90,7 @@ bool FizzyEngine::instantiate(bytes_view wasm_binary)
                         {"env", "adler32", {fizzy::ValType::i32, fizzy::ValType::i32},
                             fizzy::ValType::i32, env_adler32},
                     });
-        m_instance = fizzy::instantiate(module, imports);
+        m_instance = fizzy::instantiate(std::move(module), imports);
```
chfast, line 93: This has no effect because the `module` is const.

test/unittests/execute_floating_point_test.cpp
```diff
@@ -990,7 +990,7 @@ TYPED_TEST(execute_floating_point_types, add_sub_neg_relation)
     constexpr auto sub_fn_idx = 0 + fn_offset;
     constexpr auto addneg_fn_idx = 1 + fn_offset;
 
-    auto instance = instantiate(std::move(module));
+    auto instance = instantiate(module);
```
chfast, line 993: Why is the `std::move` dropped?

chfast (changes requested):

test/utils/fizzy_engine.cpp
gumb0: This is already done in #581 including making module non-const.

axic: I did this at the same time you did #581, and asked since if I can merge the other changes from this into #581, but no response.

test/unittests/execute_floating_point_test.cpp
axic: To keep `unittests` consistently not using move semantics.

chfast: I don't think this is good consistency.

PR #584, opened 2020-10-06
Rust: support packaging and publishing

axic:


circle.yml
```diff
@@ -536,6 +536,11 @@ jobs:
       - run:
           name: Test
           command: cargo test
+      - run:
+          name: Package
+          command: |
```
chfast, line 541: Fancy way of doing this.
```suggestion
          working_dir: bindings/rust
          command: cargo package
```

bindings/rust/build.rs
```diff
@@ -11,8 +11,22 @@ use std::env;
 use std::path::PathBuf;
 
 fn main() {
-    // This is the root directory.
-    let src = "../../";
```
axic, line 15: For some reason `cargo package` copies over symlinks to files with an updated path (that's why this works on evmc, where we only need the header which is symlinked), but symlinking the root does not work (e.g.` bindings/rust/fizzy -> ../../`).

circle.yml
axic: How can I leave a comment in that case? Just as a yaml comment?

chfast: Yes. YAML commands are better IMHO.

axic: Is `~/bindings/rust` the current path? The suggested one does not work.

circle.yml
```diff
@@ -539,7 +539,7 @@ jobs:
       - run:
           name: Package
           # The package must be run within the actual crate and not in the workspace.
-          working_dir: bindings/rust
+          working_dir: ~/bindings/rust
```
chfast, line 542:
```suggestion
          working_dir: ~/project/bindings/rust
```
if absolute path is required.

chfast: Ok, it should have been `working_directory`.

axic: Neither `bindings/rust`, `~/bindings/rust`, or `~/project/bindings/rust` works. Will check `pwd` because this is confusing.

chfast (approved):

gumb0 (approved):

PR #586, opened 2020-10-06
ci: run Rust build in release mode too

axic:
`cargo build` will build Fizzy in Debug profile, while `cargo build --release` will do with Release in cmake, see https://docs.rs/cmake/0.1.44/cmake/struct.Config.html#method.profile. The more specific translation rules can be seen in https://docs.rs/cmake/0.1.44/src/cmake/lib.rs.html#553-558 (it can use `Release`, `RelWithDebInfo`, `MinSizeRel` depending on settings).

We could also decide to force Release mode if we want.

Furthermore it seems PIC is used in both modes, we may want to investigate this (see https://docs.rs/cmake/0.1.44/cmake/struct.Config.html#method.pic).

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -531,10 +531,13 @@ jobs:
             rustfmt --version
             cargo fmt --all -- --check
       - run:
-          name: Build
+          name: Build (debug mode)
           command: cargo build
       - run:
-          name: Test
+          name: Build (release mode)
+          command: cargo build --release
+      - run:
+          name: Test (debug mode)
           command: cargo test
       - run:
           name: Package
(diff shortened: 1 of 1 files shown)


chfast (approved):
> The more specific translation rules can be seen in https://docs.rs/cmake/0.1.44/src/cmake/lib.rs.html#553-558 (it can use `Release`, `RelWithDebInfo`, `MinSizeRel` depending on settings).

These look good.

> Furthermore it seems PIC is used in both modes, we may want to investigate this (see https://docs.rs/cmake/0.1.44/cmake/struct.Config.html#method.pic).

I can benchmark PIC builds some day.

PR #588, opened 2020-10-06
rust: add simple compilation (integration) test

axic:


chfast (commented):
Looks fine, but should rather go to somewhere in `test/`

axic:
> Looks fine, but should rather go to somewhere in test/

I was wondering about that. Right now `test` is strictly C/C++. Once we have Go bindings, would tests of it go there too? Is it feasible with Go? Would any of this conflict with the cmake/gtest stuff?

chfast:
> > Looks fine, but should rather go to somewhere in test/
> 
> I was wondering about that. Right now `test` is strictly C/C++. Once we have Go bindings, would tests of it go there too? Is it feasible with Go? Would any of this conflict with the cmake/gtest stuff?

Unittest should be placed by the given language convention. But this is rather integration test, so I'd rather put in `test/rust_compilation` or `bindings/rust/test/`.

axic:
Apparently can't move to `test/rust-compilation`, because it expects `test` to be a cargo directory for some reason:
```
$ cargo build
error: failed to read `/Users/alex/Projects/fizzy/test/Cargo.toml`

Caused by:
  No such file or directory (os error 2)
```

chfast:
> Apparently can't move to `test/rust-compilation`, because it expects `test` to be a cargo directory for some reason:

So `binndings/rust/test` if that works, otherwise it can stay as it is.

axic:
Replaced by #591. The branch name had a typo here.

PR #590, opened 2020-10-07
Extend ASan checks

chfast:


test/bench_internal/execute_benchmarks.cpp
```diff
@@ -1,80 +0,0 @@
...
-#define ARGS            \
-    Args({0, 0})        \
-        ->Args({2, 4})  \
-        ->Args({2, 38}) \
-        ->Args({3, 4})  \
-        ->Args({3, 8})  \
-        ->Args({3, 13}) \
-        ->Args({5, 30}) \
-        ->Args({10, 100})
-BENCHMARK_TEMPLATE(init_locals, std::vector<uint64_t>, init_locals_1)->ARGS;
```
axic, line 76: These are just messing with ASan and/or are not useful anymore?

axic (approved):

test/bench_internal/execute_benchmarks.cpp
chfast: Yes.

PR #592, opened 2020-10-09
Potential floating point optimizations

axic:


lib/fizzy/execute.cpp
```diff
@@ -299,6 +299,19 @@ inline uint64_t popcnt64(uint64_t value) noexcept
     return static_cast<uint64_t>(__builtin_popcountll(value));
 }
 
+template <typename T>
+T signbit(T value) noexcept = delete;
+
+inline bool signbit(float value) noexcept
+{
+    return (bit_cast<uint32_t>(value) & F32SignMask) != 0;
```
axic, line 307: This generates two very simple instructions with GCC, but clang detects this pattern and replaces it with the SSE instruction.

lib/fizzy/execute.cpp
```diff
@@ -356,7 +388,7 @@ inline T ffloor(T value) noexcept
     // the __builtin_floor() outputs -0 where it should +0.
     // The following workarounds the issue by using the fact that the sign of
     // the output must always match the sign of the input value.
-    return std::copysign(result, value);
+    return fcopysign(result, value);
```
axic, line 391: This and the other cases may be faster or may be slower if the rest of the function already uses SSE registers.

lib/fizzy/execute.cpp
```diff
@@ -408,7 +440,7 @@ inline T fmin(T a, T b) noexcept
     if (std::isnan(a) || std::isnan(b))
         return std::numeric_limits<T>::quiet_NaN();  // Positive canonical NaN.
 
-    if (a == 0 && b == 0 && (std::signbit(a) == 1 || std::signbit(b) == 1))
```
axic, line 411: @chfast [`std::signbit`](https://en.cppreference.com/w/cpp/numeric/math/signbit) actually returns a bool, why did use integer literals here and below?

chfast: I don't remember. Probably it is more clear that signbit is `1`, while the meaning of `true` bit is confusing.

lib/fizzy/execute.cpp
```diff
@@ -408,7 +440,7 @@ inline T fmin(T a, T b) noexcept
...
+    if (a == 0 && b == 0 && (signbit(a) == true || signbit(b) == true))
```
chfast, line 443: Are you using C macro now? It does not return a bool. Besides, it should be
```suggestion
    if (a == 0 && b == 0 && (signbit(a) || signbit(b)))
```

axic: > Are you using C macro now?

It is a template.

axic:
I think this is ready, just needs benchmarking to show which changes are useful.

axic:
Rebased.

lib/fizzy/execute.cpp
```diff
@@ -362,6 +375,25 @@ inline double fneg(double value) noexcept
     return bit_cast<double>(bit_cast<uint64_t>(value) ^ F64SignMask);
 }
 
+template <typename T>
+T fcopysign(T a, T b) noexcept = delete;
+
+template <>
+inline float fcopysign(float a, float b) noexcept
```
chfast, line 382: Here we have small issue with the name: `fcopysign` vs `std::copysign` and `signbit` vs `std::signbit`.

axic: Can rename `fcopysign` to `copysign`, it was an already existing feature and just moved. However if it is 100% compatible with `std::copysign` then should be renamed.

lib/fizzy/execute.cpp
```diff
@@ -362,6 +375,25 @@ inline double fneg(double value) noexcept
     return bit_cast<double>(bit_cast<uint64_t>(value) ^ F64SignMask);
 }
 
+template <typename T>
+T fcopysign(T a, T b) noexcept = delete;
+
+template <>
```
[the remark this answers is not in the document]
axic: Didn't add this, just moved this function. Can remove.

lib/fizzy/execute.cpp
```diff
@@ -362,6 +375,25 @@ inline double fneg(double value) noexcept
     return bit_cast<double>(bit_cast<uint64_t>(value) ^ F64SignMask);
 }
 
+template <typename T>
+T fcopysign(T a, T b) noexcept = delete;
```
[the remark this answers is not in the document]
axic: Why? It should be the same type we're copying between?

lib/fizzy/execute.cpp
axic: Hmm actually https://en.cppreference.com/w/cpp/numeric/math/copysign has some interesting comment:
> float       copysign ( float mag, float sgn );
> 
> If mag is NaN, then NaN with the sign of sgn is returned.
> **If sgn is -0, the result is only negative if the implementation supports the signed zero consistently in arithmetic operations.**

lib/fizzy/execute.cpp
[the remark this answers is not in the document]
chfast: If types are different the template will not match. E.g. `fcopysign(0.0, 0.0f)` will use `fcopysign(double, double)` but we want to be very restrictive to what overloads are provided.

lib/fizzy/execute.cpp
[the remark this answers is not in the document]
chfast: Still good time to improve it.

lib/fizzy/execute.cpp
chfast: I think it is fine to rename to `copysign` and follow C std names (which are bad by themselves).

The C standard does not require IEEE 754, but WebAssembly does. So we must use compilers that support IEEE 754. E.g. GCC supports signed zeros by default, but this can be disabled with `-fno-signed-zeros`. https://gcc.gnu.org/wiki/FloatingPointMath

lib/fizzy/execute.cpp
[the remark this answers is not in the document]
axic: `fdiv`, `fmin`, `fmax` have the same issue, plus all the basic `add`/`mul`/... ones. Should those be changed?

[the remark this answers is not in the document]
axic: Moved these to #826 where we can clean all of them up.

lib/fizzy/execute.cpp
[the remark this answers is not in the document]
chfast: > These should not be template specializations. Function overloading is good enough.

Actually, I've learned this is not enough because of the way this is used. Ignore this request.

chfast (approved):

lib/fizzy/execute.cpp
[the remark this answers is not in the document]
axic: Yes, learned that yesterday too, but was lazy to close the PR 😓

lib/fizzy/execute.cpp
```diff
@@ -332,6 +332,19 @@ inline constexpr T popcnt(T value) noexcept
     return static_cast<T>(popcount(value));
 }
 
+template <typename T>
+T signbit(T value) noexcept = delete;
+
+inline bool signbit(float value) noexcept
```
axic, line 338: We may get a proper `constexpr` `signbit` in C++23: https://open-std.org/JTC1/SC22/WG21/docs/papers/2019/p0533r5.pdf

PR #595, opened 2020-10-09
rust: update description for crates.io

axic:


chfast (approved):

gumb0 (approved):

axic:
This (https://crates.io/crates/lazy_static) should have badges too, but none visible, so probably we should just remove this.

PR #596, opened 2020-10-12
C API Instance memory access

gumb0:


include/fizzy/fizzy.h
```diff
@@ -90,6 +90,14 @@ FizzyInstance* fizzy_instantiate(const FizzyModule* module,
 /// Free resources associated with the instance.
 void fizzy_free_instance(FizzyInstance* instance);
 
+/// Get pointer to memory of an instance.
+/// @returns Pointer to memory data or NULL in case instance doesn't have any memory.
```
axic, line 94: "Any" or "any exported" memory?

test/unittests/capi_test.cpp
```diff
@@ -65,6 +65,57 @@ TEST(capi, instantiate_imported_function)
...
+    auto module = fizzy_parse(wasm.data(), wasm.size());
+    EXPECT_NE(module, nullptr);
+
+    auto instance = fizzy_instantiate(module, nullptr, 0);
+    EXPECT_NE(instance, nullptr);
+
+    uint8_t* memory = fizzy_get_instance_memory_data(instance);
+    EXPECT_NE(memory, nullptr);
+    EXPECT_EQ(memory[0], 0);
+    EXPECT_EQ(memory[1], 0x11);
```
axic, line 90: Should also check for `memory[2] == 0x22`.

include/fizzy/fizzy.h
gumb0: It gives access to any memory

test/unittests/capi_test.cpp
```diff
@@ -65,6 +65,57 @@ TEST(capi, instantiate_imported_function)
...
+
+    memory[0] = 0xaa;
+    memory[1] = 0xbb;
+
+    EXPECT_EQ(fizzy_execute(instance, 0, nullptr, 0).value.i64, 0x22bbaa);
+
+    fizzy_free_instance(instance);
+}
+
+TEST(capi, memory_access_no_memory)
```
axic, line 101: Can you add a test for imported memory? Instantiation should just fail right now, since we lack the API.

include/fizzy/fizzy.h
axic: Can you clarify that in the description?

gumb0: Added a note.

test/unittests/capi_test.cpp
gumb0: Added.

test/unittests/capi_test.cpp
gumb0: Added.

axic (approved):

chfast (approved):

PR #603, opened 2020-10-14
ci: Enhance sanitizer's config

chfast:


CMakeLists.txt
```diff
@@ -94,7 +94,7 @@ if(ENABLE_ASSERTIONS)
 endif()
 
 if(FIZZY_FUZZING)
-    set(fuzzing_flags -fsanitize=fuzzer-no-link,address,pointer-subtract,undefined,nullability,implicit-unsigned-integer-truncation,implicit-signed-integer-truncation)
```
axic, line 97: The commit log says "pointer-subtrack".

chfast: Fixed.

gumb0 (approved):

circle.yml
```diff
@@ -371,16 +372,14 @@ jobs:
...
-      # TODO: Enable detect_stack_use_after_return=1 when https://bugs.llvm.org/show_bug.cgi?id=47626 is fixed.
-      ASAN_OPTIONS: detect_invalid_pointer_pairs=2
+      ASAN_OPTIONS: detect_stack_use_after_return=1:check_initialization_order=1
       UBSAN_OPTIONS: halt_on_error=1
     steps:
       - install_testfloat
       - checkout
       - build:
           build_type: RelWithDebInfo
-          # TODO: pointer-compare produces failure in std::string operator+.
```
axic, line 382: This didn't even use pointer-compare, but I see you removed pointer-subtract -- perhaps that should be removed in the first commit?

circle.yml
```diff
@@ -595,10 +591,10 @@ jobs:
             rustup toolchain install nightly-x86_64-unknown-linux-gnu
       - run:
           name: Build (debug mode)
-          command: RUSTFLAGS="-Z sanitizer=address" ASAN_OPTIONS=detect_leaks=1 cargo +nightly build --target x86_64-unknown-linux-gnu
```
axic, line 598: Ok since this on by default on Linux (not on MacOS): https://clang.llvm.org/docs/AddressSanitizer.html#memory-leak-detection

circle.yml
```diff
@@ -595,10 +591,10 @@ jobs:
...
+          command: RUSTFLAGS="-Z sanitizer=address" cargo +nightly build --target x86_64-unknown-linux-gnu
```
axic, line 594: Does asan does anything at compile time? If not we can remove this step.

axic: Apparently it does some compile-thing things too: https://github.com/rust-lang/rust/issues/45220

Any we should also pass on sanitize mode to cmake: https://users.rust-lang.org/t/sanitizers-in-mixed-rust-and-c-code/48863/4

chfast: You should be building fizzy with `-DSANITIZE=address`.

circle.yml
chfast: Done.

axic: It seem it only removed the comment and not the line below?

axic (approved):

PR #605, opened 2020-10-14
ci: Fix TestFloat compilation

chfast:


circle.yml
```diff
@@ -49,6 +49,9 @@ commands:
           working_directory: /tmp
           command: |
             if [ ! -f testfloat_gen ]; then
+              # GCC compiler is expected, create link from clang if gcc not available.
+              type gxx || (cd $(dirname $(which clang)) && ln -s clang gcc)
```
axic, line 53: It uses `gxx`? I only ever had `gcc` and `g++` and never `gxx`.

chfast: typo

circle.yml
```diff
@@ -49,6 +49,9 @@ commands:
           working_directory: /tmp
           command: |
             if [ ! -f testfloat_gen ]; then
+              # GCC compiler is expected, create link from clang if gcc not available.
+              type gcc || (cd $(dirname $(which clang)) && ln -s clang gcc)
```
axic, line 53: So what problem is this solving if the CI was not breaking? Some future PR?

axic (approved):

circle.yml
```diff
@@ -49,6 +49,9 @@ commands:
...
+
               git clone --depth=1 --single-branch https://github.com/ucb-bar/berkeley-softfloat-3.git
               git clone --depth=1 --single-branch https://github.com/ucb-bar/berkeley-testfloat-3.git
```
axic, line 56: While at it, can we use a specific tag/branch/version of these repos?

chfast: It has no tags.

circle.yml
chfast: This problem is not visible usually, because we take the binary from cache. But cache is disabled for external PRs (like this one) and then jobs having clang compiler only fails. The makefile from TestFloat is using `gcc` for compiler instead of `cc`.

PR #607, opened 2020-10-14
Use proper feature test for C++20 std::span

axic:
See https://en.cppreference.com/w/cpp/utility/feature_test which sets `__cpp_lib_span` as an alias to `201902L`, while we were testing for `201703L`.

### DIFF
--- a/lib/fizzy/cxx20/bit.hpp
+++ b/lib/fizzy/cxx20/bit.hpp
@@ -7,6 +7,10 @@
 #include <cstdint>
 #include <type_traits>
 
+#if __has_include(<version>)
+#include <version>
+#endif
+
 namespace fizzy
 {
 /// The non-constexpr implementation of C++20's std::bit_cast.
--- a/lib/fizzy/cxx20/span.hpp
+++ b/lib/fizzy/cxx20/span.hpp
@@ -4,7 +4,11 @@
 
 #pragma once
 
-#if __cplusplus > 201703L
+#if __has_include(<version>)
+#include <version>
+#endif
+
+#ifdef __cpp_lib_span
 
 #include <span>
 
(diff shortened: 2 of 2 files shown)


lib/fizzy/cxx20/span.hpp
```diff
@@ -4,7 +4,9 @@
 
 #pragma once
 
-#if __cplusplus > 201703L
+#ifdef __cpp_lib_span
+
+#warning "using c++20 span"
```
axic, line 9: @chfast So apparently this feature is not turned on in clang11?

chfast: I don't know.

axic: Should we remove the C++20 implementation then?

chfast: No.

lib/fizzy/cxx20/span.hpp
```diff
@@ -4,7 +4,9 @@
 
 #pragma once
 
-#if __cplusplus > 201703L
+#include <version>
```
chfast, line 7: Unfortunately, the `<version>` must be included conditionally.

axic:
> Conditional includes should be after the normal ones.

The reason I put first in case we'd make the other things conditional based on the `__cpp_lib*` macros, but that is not yet the case so swapped.

lib/fizzy/cxx20/bit.hpp
```diff
@@ -7,6 +7,10 @@
 #include <cstdint>
 #include <type_traits>
 
+#if __cplusplus > 201703L
```
chfast, line 10:
```suggestion
#if __has_include(<version>)
```
https://en.cppreference.com/w/cpp/preprocessor/include

chfast (changes requested):

lib/fizzy/cxx20/bit.hpp
axic: Nice!

gumb0 (approved):

PR #609, opened 2020-10-15
rust: support memory manipulation

axic:


bindings/rust/src/lib.rs
```diff
@@ -139,6 +139,38 @@ impl From<ExecutionResult> for sys::FizzyExecutionResult {
...
+        let size = 0 as usize;
+        offset..offset + size
+    }
+
+    pub fn memory_size(&self) -> usize {
+        unsafe { sys::fizzy_get_instance_memory_size(self.0.as_ptr()) }
+    }
+
+    pub fn memory_get_into(&mut self, offset: u32, target: &mut [u8]) -> Result<(), ()> {
+        let mem = unsafe {
```
axic, line 153: Return err if no memory found (`fizzy_get_instance_memory_data` returns null).

chfast: More errors at the beginning sound good.

chfast (commented):
Copying memory back and forth does not look like good idea. Rust should some kind of mutable slice allowing free access to wasm memory (read/write).

axic:
> Copying memory back and forth does not look like good idea. Rust should some kind of mutable slice allowing free access to wasm memory (read/write).

This PR does mutable slicing.

chfast:
> Right. We can definitely have a separate function exposing the underlying memory as an unsafe API, but this is the safe API. It is modelled after other Rust VMs.

Why exposing the underlying memory is unsafe?

axic:
> > Right. We can definitely have a separate function exposing the underlying memory as an unsafe API, but this is the safe API. It is modelled after other Rust VMs.
>
> Why exposing the underlying memory is unsafe?

Exposing a slice obtained via `std::slice::from_raw_parts` (i.e. the underlying C++ allocated memory) is considered unsafe as ownership is not managed by Rust.

bindings/rust/src/lib.rs
```diff
@@ -289,6 +289,74 @@ impl TypedExecutionResult {
...
 impl Instance {
+    /// Ensure the range is valid according to the currently available memory size.
+    fn checked_memory_range(
+        &self,
+        offset: u32,
+        size: usize,
+    ) -> Result<core::ops::Range<usize>, ()> {
+        let offset = offset as usize;
+        let memory_size = self.memory_size();
+        if size == 0 || memory_size == 0 || offset > memory_size || (offset + size) > memory_size {
```
axic, line 300: - `size == 0` is an arbitrary restriction -- should we allow empty slices?
- `memory_size == 0` is a similarly arbitrary restriction, but at least this allows the nullptr check below to be a debug_assert, otherwise would need to check it

bindings/rust/src/lib.rs
```diff
@@ -289,6 +289,74 @@ impl TypedExecutionResult {
 }
 
 impl Instance {
+    /// Ensure the range is valid according to the currently available memory size.
+    fn checked_memory_range(
+        &self,
+        offset: u32,
+        size: usize,
+    ) -> Result<core::ops::Range<usize>, ()> {
+        let offset = offset as usize;
```
axic, line 298: Questionable. Need to add a safety check for `offset > usize::MAX`.

axic: Should be safe: https://doc.rust-lang.org/std/primitive.usize.html

bindings/rust/src/lib.rs
axic: Decided to allow empty slices.

axic:
Need to do some big squashing, but I think this is ready for review (only look at the final diff).

bindings/rust/src/lib.rs
```diff
@@ -289,6 +289,73 @@ impl TypedExecutionResult {
...
+    /// Ensure the range is valid according to the currently available memory size.
+    fn checked_memory_range(
+        &self,
+        offset: u32,
+        size: usize,
+    ) -> Result<core::ops::Range<usize>, ()> {
+        // This should be safe given usize::BITS >= u32::BITS, see https://doc.rust-lang.org/std/primitive.usize.html.
+        let offset = offset as usize;
+        let memory_size = self.memory_size();
+        if memory_size == 0 || offset > memory_size || (offset + size) > memory_size {
```
gumb0, line 301: This would allow  empty slices `memory_size..memory_size`.

Should this better be
```suggestion
        if offset >= memory_size || (offset + size) > memory_size {
```
(`memory_size == 0` check would be redundant then)

axic: The `memory_size != 0` check is there so that we can have the `debug_assert` for ptr not being null.

axic: > This would allow empty slices memory_size..memory_size.

Also see https://github.com/wasmx/fizzy/pull/609#discussion_r571211243

gumb0: My comment was not about empty slice restriction, but another restriction to allow slices to start only at valid offset.

bindings/rust/src/lib.rs
```diff
@@ -755,4 +822,147 @@ mod tests {
...
+        assert!(module.is_ok());
+        let instance = module.unwrap().instantiate();
+        assert!(instance.is_ok());
+        let mut instance = instance.unwrap();
+
+        assert_eq!(instance.memory_size(), 65536);
+        unsafe {
+            // Allow empty slices.
+            assert!(instance.checked_memory_slice(0, 0).is_ok());
+            assert!(instance.checked_memory_slice_mut(0, 0).is_ok());
```
gumb0, line 877: Also currently should work, but I think it's better to disallow it
```suggestion
            assert!(instance.checked_memory_slice(0, 0).is_ok());
            assert!(instance.checked_memory_slice_mut(0, 0).is_ok());
            assert!(instance.checked_memory_slice(65536, 0).is_ok());
            assert!(instance.checked_memory_slice_mut(65536, 0).is_ok());
```

bindings/rust/src/lib.rs
axic: Well, empty slice at the very end. Well, the `offset + size > memory_size` prevents generating a slice which would do any overread, so `65536, 0` as an input right now is accepted. See tests.

bindings/rust/src/lib.rs
axic: We can duplicate the cases here from `test_memory`, but first let us agree on the cases.

gumb0:
Tests for `memory_get` / `memory_set` with empty slices are needed.

bindings/rust/src/lib.rs
axic: Oh, I should have also used tests with 65535.

bindings/rust/src/lib.rs
```diff
@@ -755,4 +822,147 @@ mod tests {
...
+            let mem = instance
+                .checked_memory_slice_mut(0, 65536)
+                .expect("valid mutable slice");
+            assert_eq!(mem[0], 0);
+            mem[0] = 42;
+        }
+        unsafe {
+            // Check that const slice matches up.
+            let mem = instance.checked_memory_slice(0, 5).expect("valid slice");
+            assert_eq!(mem[0], 42);
```
gumb0, line 919: Maybe also check a couple of adjacent bytes that they're still 0.

bindings/rust/src/lib.rs
```diff
@@ -755,4 +822,147 @@ mod tests {
...
+            result
+                .value()
+                .expect("expected value")
+                .as_u32()
+                .expect("expected u32 result"),
+            42
+        );
+
+        // Set memory via safe helper.
+        assert!(instance.memory_set(0, &[0x11, 0x22, 0x33, 0x44]).is_ok());
```
gumb0, line 935: Here I'd also check
```suggestion
        assert!(instance.memory_set(0, &[]).is_ok());
        assert!(instance.memory_set(65535, &[]).is_ok());
        assert!(instance.memory_set(65537, &[]).is_err());
        assert!(instance.memory_set(65534, &[0x11, 0x22, 0x33, 0x44]).is_err());
        assert!(instance.memory_set(65537, &[0x11, 0x22, 0x33, 0x44]).is_err());
```

bindings/rust/src/lib.rs
```diff
@@ -755,4 +825,229 @@ mod tests {
...
+                0,
+            )
+            .expect("successful execution");
+        assert!(!result.trapped());
+
+        // Read memory via safe helper.
+        let mut dst = Vec::new();
+        dst.resize(65536, 0);
+        // Reading 65536 bytes.
+        assert!(instance.memory_get(0, &mut dst).is_ok());
```
gumb0, line 1049: Need also to add tests for `memory_get` with empty dst slice and with overruns.

gumb0:
```rust
let mut dst_emtpy = Vec::new();
assert!(instance.memory_get(0, &mut dst).is_ok());
assert!(instance.memory_get(65536 + 65535, &mut dst).is_ok());
assert!(instance.memory_get(65536 + 65536, &mut dst).is_err());
assert!(instance.memory_get(65536 + 65537, &mut dst).is_err());

let mut dst_small = Vec::new();
dst_small.resize(4, 0);
assert!(instance.memory_get(0, &mut dst_small).is_ok());
assert!(instance
            .memory_get(65536 + 65532, &mut dst_small)
            .is_ok());
assert!(instance
            .memory_get(65536 + 65533, &mut dst_small)
            .is_err());
assert!(instance
            .memory_get(65536 + 65534, &mut dst_small)
            .is_err());
assert!(instance
            .memory_get(65536 + 65535, &mut dst_small)
            .is_err());
assert!(instance
            .memory_get(65536 + 65536, &mut dst_small)
            .is_err());
assert!(instance
            .memory_get(65536 + 65537, &mut dst_small)
            .is_err());
```

axic: Added, but using slices.

bindings/rust/src/lib.rs
```diff
@@ -289,6 +289,76 @@ impl TypedExecutionResult {
...
+    ) -> Result<core::ops::Range<usize>, ()> {
+        // This should be safe given usize::BITS >= u32::BITS, see https://doc.rust-lang.org/std/primitive.usize.html.
+        let offset = offset as usize;
+        let memory_size = self.memory_size();
+        // Empty slices are allowed, but ensure both starting and ending offsets are valid.
+        if offset >= memory_size || (offset + size) > memory_size {
+            return Err(());
+        }
+        // Assert this too?
+        // debug_assert!(memory_size != 0);
```
gumb0, line 306: can be uncommented

gumb0 (approved):

bindings/rust/src/lib.rs
gumb0: I suspect now that it was better not to listen to me and leave only the check `if (offset + size) > memory_size {err}` (i.e. allow `memory_size..memory_size` empty slices) 

I can't be sure about Rust, but in C/C++ it's allowed to have a range of pointers/iterators with both ends pointing to one-past-last element.

gumb0:
Looks good, but still not sure what's best to allow for slice ranges https://github.com/wasmx/fizzy/pull/609#discussion_r573172168

axic:
Squashed and added https://github.com/wasmx/fizzy/pull/609#discussion_r573172168 as a second commit.

bindings/rust/src/lib.rs
```diff
@@ -289,6 +289,75 @@ impl TypedExecutionResult {
...
+    fn checked_memory_range(
+        &self,
+        offset: u32,
+        size: usize,
+    ) -> Result<core::ops::Range<usize>, ()> {
+        // This should be safe given usize::BITS >= u32::BITS, see https://doc.rust-lang.org/std/primitive.usize.html.
+        let offset = offset as usize;
+        let memory_size = self.memory_size();
+        // Empty slices are allowed, but ensure both starting and ending offsets are valid.
+        if memory_size == 0 || offset > memory_size || (offset + size) > memory_size {
```
gumb0, line 302: `offset > memory_size` condition is redundant then, when it is true `(offset + size) > memory_size` will always be true.
```suggestion
        if memory_size == 0 || (offset + size) > memory_size {
```

Also for the case of `(memory 0)` it would be perhaps logical to allow (0, 0) slice, then the condition should make an error only the absence of memory
```suggestion
        if !sys::fizzy_module_has_memory(self.get_module()) || (offset + size) > memory_size {
```

bindings/rust/src/lib.rs
```diff
@@ -755,4 +824,246 @@ mod tests {
...
+            // Allow empty slices.
+            assert!(instance.checked_memory_slice(65535, 0).is_ok());
+            assert!(instance.checked_memory_slice_mut(65535, 0).is_ok());
+            assert!(instance.checked_memory_slice(65536, 0).is_ok());
+            assert!(instance.checked_memory_slice_mut(65536, 0).is_ok());
+            // Single byte.
+            assert!(instance.checked_memory_slice(65535, 1).is_ok());
+            assert!(instance.checked_memory_slice_mut(65535, 1).is_ok());
+            // Reading over.
+            assert!(instance.checked_memory_slice(65535, 2).is_err());
```
gumb0, line 941: `assert!(instance.checked_memory_slice(65536, 1).is_err());` could be added

bindings/rust/src/lib.rs
axic: > offset > memory_size condition is redundant then, when it is true (offset + size) > memory_size will always be true.

Was a bad attempt for overflow check. Decided to just use `offset.checked_add(size).is_none()` for overflow check.

bindings/rust/src/lib.rs
axic: Done.

bindings/rust/src/lib.rs
axic: Clarified that https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html allows `len=0`, just that `data` has to be valid.

bindings/rust/src/lib.rs
```diff
@@ -289,6 +289,76 @@ impl TypedExecutionResult {
...
 impl Instance {
+    /// Ensure the range is valid according to the currently available memory size.
+    fn checked_memory_range(
+        &self,
+        offset: u32,
+        size: usize,
+    ) -> Result<core::ops::Range<usize>, ()> {
+        // This should be safe given usize::BITS >= u32::BITS, see https://doc.rust-lang.org/std/primitive.usize.html.
+        let offset = offset as usize;
+        let has_memory = unsafe { sys::fizzy_module_has_memory(self.get_module()) };
```
axic, line 300: Actually with solution in #728, I would cache this in the `Instance { has_memory: bool }`.

gumb0 (approved):

bindings/rust/src/lib.rs
```diff
@@ -289,6 +289,76 @@ impl TypedExecutionResult {
 }
 
 impl Instance {
+    /// Ensure the range is valid according to the currently available memory size.
+    fn checked_memory_range(
+        &self,
+        offset: u32,
+        size: usize,
+    ) -> Result<core::ops::Range<usize>, ()> {
+        // This should be safe given usize::BITS >= u32::BITS, see https://doc.rust-lang.org/std/primitive.usize.html.
```
chfast, line 298:
```suggestion
        // This is safe given usize::BITS >= u32::BITS, see https://doc.rust-lang.org/std/primitive.usize.html.
```

chfast (approved):
Can be for now, but I see some design issues with it.

1. The `fizzy_get_instance_memory_size` is called twice.
2. You almost always wants memory size and pointer in the same time. C should return both in single call.

bindings/rust/src/lib.rs

bindings/rust/src/lib.rs
```diff
@@ -292,20 +292,19 @@ impl Instance {
     /// Ensure the range is valid according to the currently available memory size.
     fn checked_memory_range(
         &self,
```
chfast, line 294: Why it needs `self` now?

axic: The `has_memory` call still uses it.

gumb0: > The `has_memory` call still uses it.

Now you don't use it.

chfast (approved):

gumb0 (approved):

PR #610, opened 2020-10-15
Provide constexpr bit_cast if possible

chfast:


axic:
Was this just not worth it, or not working?

chfast:
Too much `#if` to my taste. And we still cannot benefit from it because not available generally.

chfast:
The GCC is lacking `__builtin_bit_cast()` support, only available in GCC11 to be released in 2021Q1.
Here you can track minimal compiler version required, including MSVC: https://godbolt.org/z/Wo7aEx.

PR #612, opened 2020-10-16
Remove dirty stack cases

axic:
Fixes #506.

PR #613, opened 2020-10-19
Add assert to as<int32_t>() for dirty higher bits.

axic:
Missed in #612.


test/unittests/execute_numeric_test.cpp
```diff
@@ -264,7 +264,8 @@ TEST(execute_numeric, i32_mul)
 
 TEST(execute_numeric, i32_div_s)
```
axic, line 265: This is especially annoying as we never had a proper 32-bit tests.

axic:
@chfast @gumb0 oh I forgot to ping y'all for reviews. I'm not sure what's left, but probably we could extend the 32-bit tests if you have some ideas for more edge cases to cover.

gumb0 (approved):

chfast (approved):

chfast (approved):

PR #615, opened 2020-10-20
Get rid of std::function in C++ API

gumb0:


axic:
@gumb0 what is the status of this? Can you rebase and show what is the actually relevant change?

gumb0:
The entire first commit is minimal needed change.

I would close this, because I think it complicates API too much, and currently seems to me not worth to pursue.

axic:
Feel free to close.

PR #619, opened 2020-10-22
test: Use ElementsAre() to check arrays of instructions

chfast:
New checks are more type change tolerant what allows changing the `.instructions` type later.

### DIFF
--- a/test/unittests/parser_expr_test.cpp
+++ b/test/unittests/parser_expr_test.cpp
@@ -3,13 +3,14 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include "parser.hpp"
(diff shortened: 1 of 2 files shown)


axic (approved):
Is better error reporting the reason for this?

gumb0 (approved):

chfast:
> Is better error reporting the reason for this?

No. You can have different types inside `ElementsAre()` and the only requirement is the type of an element is comparable with an element of the instructions array.

chfast:
I just converted some more parser tests to similar form.

axic (approved):

gumb0 (approved):

PR #620, opened 2020-10-22
Convert instructions to vector<uint8_t>

chfast:
This allows to move immediate values to the same array later. That will increase overall performance.

### DIFF
--- a/lib/fizzy/types.hpp
+++ b/lib/fizzy/types.hpp
@@ -362,7 +362,7 @@ struct Code
 
     // The instructions bytecode without immediate values.
     // https://webassembly.github.io/spec/core/binary/instructions.html
-    std::vector<Instr> instructions;
+    std::vector<uint8_t> instructions;
 
     // The decoded instructions' immediate values.
     // These are instruction-type dependent fixed size value in the order of instructions.
--- a/test/utils/asserts.hpp
+++ b/test/utils/asserts.hpp
@@ -53,8 +53,14 @@ MATCHER_P(Result, value, "")  // NOLINT(readability-redundant-string-init)
 
 namespace fizzy
 {
-std::ostream& operator<<(std::ostream& os, ExecutionResult);
+/// Equal operator for Instr and uint8_t. Convenient for unit tests.
+inline constexpr bool operator==(uint8_t a, Instr b) noexcept
+{
+    return a == static_cast<uint8_t>(b);
 }
+
+std::ostream& operator<<(std::ostream& os, ExecutionResult);
+}  // namespace fizzy
 
 namespace fizzy::test
 {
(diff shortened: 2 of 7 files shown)


lib/fizzy/types.hpp
```diff
@@ -362,7 +362,7 @@ struct Code
 
     // The instructions bytecode without immediate values.
     // https://webassembly.github.io/spec/core/binary/instructions.html
-    std::vector<Instr> instructions;
```
axic, line 365: Shouldn't we have a `static_assert(sizeof(Instr) == sizeof(uint8_t));` in this file?

chfast: This is not required because we don't `reinterpret_cast<Instr*>(bytes_ptr)` or the other way.

gumb0 (approved):

test/utils/asserts.hpp
```diff
@@ -53,9 +53,15 @@ MATCHER_P(Result, value, "")  // NOLINT(readability-redundant-string-init)
 
 namespace fizzy
 {
-std::ostream& operator<<(std::ostream& os, ExecutionResult);
+/// Equal operator for Instr and bytes. Convenient for unit tests.
```
axic, line 56:
```suggestion
/// Equal operator for Instr and uint8_t. Convenient for unit tests.
```

test/utils/asserts.hpp
```diff
@@ -53,9 +53,15 @@ MATCHER_P(Result, value, "")  // NOLINT(readability-redundant-string-init)
...
+inline constexpr bool operator==(uint8_t a, Instr b) noexcept
+{
+    return a == static_cast<uint8_t>(b);
 }
 
+std::ostream& operator<<(std::ostream& os, ExecutionResult);
```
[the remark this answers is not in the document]
chfast: This is the prototype.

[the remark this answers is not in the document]
axic: That's why I am asking this was missing?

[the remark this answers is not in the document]
gumb0: I was confused a little bit by this diff, too, but it was not missing, just diff shows like it was moved a few lines down.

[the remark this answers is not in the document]
axic: Ah, the split diff didn't show it nicely, just glossed over it was moved.

lib/fizzy/types.hpp
axic: Yeah though if it is not too much hassle having that extra assert could be helpful down the line, in case we mess it up somewhere.

chfast: The `enum class Instr : uint8_t` is defined in this file.

axic: Oh right, lets hope the `uint8_t` is not removed :)

axic (approved):
Though not sure why oppose the `static_assert`.

PR #623, opened 2020-10-23
test: Move FP conversion tests to a separate file

chfast:
After #618, GCC with sanitizer instrumentation goes OOM when compiling this file.


.clang-format
```diff
@@ -25,16 +25,18 @@ ColumnLimit:     100
 ConstructorInitializerIndentWidth: 2
 IncludeBlocks: Preserve
 IncludeCategories:
+  - Regex:           '^"test/.*'
```
gumb0, line 28: can we sorty these lines by priority?

test/unittests/execute_floating_point_test.hpp
```diff
@@ -0,0 +1,316 @@
+// Fizzy: A fast WebAssembly interpreter
+// Copyright 2020 The Fizzy Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "instructions.hpp"
+#include "types.hpp"
+#include "test/utils/floating_point_utils.hpp"
+#include "test/utils/hex.hpp"
```
gumb0, line 8: was this on purpose? other unit test files use angle bracket for test utils
```suggestion.
#include <test/utils/floating_point_utils.hpp>
#include <test/utils/hex.hpp>
```

test/unittests/execute_floating_point_conversion_test.cpp
```diff
@@ -0,0 +1,559 @@
...
+#include "parser.hpp"
+#include "trunc_boundaries.hpp"
+#include <test/utils/asserts.hpp>
+#include <test/utils/hex.hpp>
+#include <cmath>
+
+using namespace fizzy;
+using namespace fizzy::test;
+
+TEST(execute_floating_point, f64_promote_f32)
```
gumb0, line 17: Maybe make these `execute_floating_point_conversion` suite

.clang-format
chfast: No, because they are first match.

test/unittests/execute_floating_point_test.hpp
chfast: I believe the `""` version is "more correct". The `<>` additionally adds system-level dirs, what is not what we want.

This is the `gcc -v` output:
```
#include "..." search starts here:
#include <...> search starts here:
 /usr/lib/gcc/x86_64-linux-gnu/9/include
 /usr/local/include
 /usr/include/x86_64-linux-gnu
 /usr/include
```

But I can remove this change. I was just curious how would that look like.

chfast:
The relation to #618 is not important any more, but I still think this is useful because the `execute_floating_point_test.cpp` is one of the longest compiling file in the project.

gumb0 (approved):

axic (approved):
Assuming it was really just renaming the file and moving out a big block verbatim. Have not gone through the pain to verify it.

PR #624, opened 2020-10-26
capi: Add imported globals to instantiate

gumb0:


include/fizzy/fizzy.h
```diff
@@ -148,14 +173,23 @@ FizzyInstance* fizzy_instantiate(const FizzyModule* module,
...
+///                                      imported_globals_size equals 0.
+/// @param      imported_globals_size    Size of the imported global array. Can be zero.
 /// @returns    non-NULL pointer to instance in case of success, NULL otherwise.
 ///
 /// @note
 /// Functions in @a imported_functions are allowed to be in any order and allowed to include some
 /// functions not required by the module.
 /// Functions are matched to module's imports based on their module and name strings.
+/// @note
+/// Function expects @a imported_globals to be in the order of imports defined in the module.
```
gumb0, line 186: It's weird how resolve_instantiate currently doesn't resolve globals the way it resolves functions. C++ backend doesn't allow this currently, and I don't know how important it is.

include/fizzy/fizzy.h
```diff
@@ -20,12 +20,12 @@ typedef struct FizzyInstance FizzyInstance;
 /// The data type representing numeric values.
 ///
 /// i64 member is used to represent values of both i32 and i64 type.
-union FizzyValue
+typedef union FizzyValue
```
axic, line 23: Hmm, wonder how the Rust binding will act with this, will check.

include/fizzy/fizzy.h
axic: I agree there should be a function similar to functions to resolve globals in the C++ API. That could be used here.

Probably at that point I'd rename the to instantiates as `fizzy_instantiate` (which does resolution) and `fizzy_unsafe_instantiate` (which doesn't).

axic (approved):
Haven't done a very thorough review of the last two commits, but in general it looks good to me.

include/fizzy/fizzy.h
axic: Seems to be working well.

chfast (approved):

PR #627, opened 2020-10-29
C API Find exported global by name

gumb0:


include/fizzy/fizzy.h
```diff
@@ -213,6 +213,15 @@ uint8_t* fizzy_get_instance_memory_data(FizzyInstance* instance);
 /// @note    Function returns memory size regardless of whether memory is exported or not.
 size_t fizzy_get_instance_memory_size(FizzyInstance* instance);
 
+/// Find exported global by name.
+///
+/// @param  instance        Pointer to instance.
+/// @param  name            The global name. NULL-terminated string. Cannot be NULL.
+/// @param  out_global      Pointer to output struct to store found global. Cannot be NULL.
```
axic, line 220: Is there any better best practice than prefixing it with `out_`?

test/unittests/capi_test.cpp
```diff
@@ -98,6 +98,40 @@ TEST(capi, find_exported_function)
...
+    FizzyExternalGlobal global;
+    ASSERT_TRUE(fizzy_find_exported_global(instance, "g1", &global));
+    EXPECT_EQ(global.type.value_type, FizzyValueTypeI32);
+    EXPECT_FALSE(global.type.is_mutable);
+    EXPECT_EQ(global.value->i64, 42);
+
+    EXPECT_FALSE(fizzy_find_exported_global(instance, "g2", &global));
+    EXPECT_FALSE(fizzy_find_exported_global(instance, "foo", &global));
+    EXPECT_FALSE(fizzy_find_exported_global(instance, "tab", &global));
+    EXPECT_FALSE(fizzy_find_exported_global(instance, "mem", &global));
```
axic, line 130: Why are these failing?

gumb0: first one doesn't exist, and others are not globals.

include/fizzy/fizzy.h
gumb0: I'm not aware of any.

test/unittests/capi_test.cpp
axic: Ah, for a second thought this is called `find_export` and handles all kinds.

axic (approved):

include/fizzy/fizzy.h
chfast: In most case we can come up with a sentinel values. E.g. here you can have 
```
FizzyExternalGlobal fizzy_find_exported_global(
    FizzyInstance* instance, const char* name);
```

And then "not-found" can be represented by `FizzyExternalGlobal{nullptr, FizzyValueTypeVoid}`.

PR #628, opened 2020-10-29
capi: imported and exported tables

gumb0:


include/fizzy/fizzy.h
```diff
@@ -87,6 +87,30 @@ typedef struct FizzyGlobalType
...
+    /// Valid only if has_max equals true.
+    uint32_t max;
+    /// Whether limits has maximum value.
+    bool has_max;
+} FizzyLimits;
+
+/// External table.
+typedef struct FizzyExternalTable
+{
+    // Opaque pointer to table data.
```
chfast, line 108:
```suggestion
    /// Opaque pointer to table data.
```

chfast (approved):

gumb0:
Rebased.

lib/fizzy/capi.cpp
```diff
@@ -168,6 +168,27 @@ inline std::vector<fizzy::ImportedFunction> unwrap(
...
+
+inline fizzy::ExternalTable unwrap(const FizzyExternalTable& external_table) noexcept
+{
+    return {unwrap(external_table.table), unwrap(external_table.limits)};
+}
+
+inline std::vector<fizzy::ExternalTable> unwrap(const FizzyExternalTable* external_table)
+{
+    return external_table ? std::vector<fizzy::ExternalTable>{unwrap(*external_table)} :
+                            std::vector<fizzy::ExternalTable>{};
```
axic, line 204: Cannot just `{}` work here?

include/fizzy/fizzy.h
```diff
@@ -243,6 +243,15 @@ uint8_t* fizzy_get_instance_memory_data(FizzyInstance* instance);
 /// @note    Function returns memory size regardless of whether memory is exported or not.
 size_t fizzy_get_instance_memory_size(FizzyInstance* instance);
 
+/// Find exported table by name.
+///
+/// @param  instance        Pointer to instance.
+/// @param  name            The table name. NULL-terminated string. Cannot be NULL.
+/// @param  out_table       Pointer to output struct to store found table. Cannot be NULL.
+/// @returns                true if table was found, false otherwise.
+bool fizzy_find_exported_table(
```
axic, line 252: I was going to say why isn't this grouped with `find_exported_function`, but realised that is over the module. This still confuses me in our API, both C++ and in C.

test/unittests/capi_test.cpp
```diff
@@ -98,6 +98,41 @@ TEST(capi, find_exported_function)
     fizzy_free_module(module);
 }
 
+TEST(capi, find_exported_table)
+{
+    /* wat2wasm
+    (module
+      (func $f (export "foo") (result i32) (i32.const 42))
+      (global (export "g1") i32 (i32.const 42))
+      (table (export "tab") 10 30 anyfunc)
```
axic, line 107: Perhaps add another table without a maximum?

lib/fizzy/capi.cpp
gumb0: No, `error: initializer list cannot be used on the right hand side of operator '?'`

include/fizzy/fizzy.h
gumb0: Yes, I put them separately, because this one operates on instance.

The change in #633 adds another `fizzy_find_exported_function`, that operates on instance and would be convenient, when you need to get an exported function and import it into another instance. With it there's going to be 4 similar `fizzy_find_exported_*` as in C++ API.
(However it has some complications with managing the lifetime of capture, so I'm not completely sure it's a good idea)
The function operating on module is renamed there to `fizzy_find_exported_function_index`, which better reflects what it does.

test/unittests/capi_test.cpp
gumb0: Added.

axic (approved):

PR #629, opened 2020-11-01
ci: Use libc++ for sanitizers-clang job

chfast:
- ci: Use libc++ for sanitizers-clang job
   This should enable container-overflow checks.
- cmake: Drop _GLIBCXX_SANITIZE_VECTOR instrumentation
   This conflicts with pointer-subtract and pointer-compare ASAn checks. Container-overflow checks are performed by Clang ASan + libc++ builds.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,13 +70,6 @@ if(CMAKE_CXX_COMPILER_ID MATCHES Clang)
     endif()
 endif()
 
-if(SANITIZE MATCHES address)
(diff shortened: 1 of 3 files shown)


axic (approved):
I guess it's alright

chfast:
The blocker is https://gcc.gnu.org/bugzilla/show_bug.cgi?id=97659,
after fixed we can try to reintroduce `_GLIBCXX_SANITIZE_VECTOR`.

Issue #583, opened 2020-10-06
UB in OperandStack


chfast:
There is potential undefined behavior in OperandStack when no locals. Then the stack top pointer is set to `m_bottom - 1`. This is undefined behavior because you are allowed to move pointer within the array or to the element _after_ the array (not before).

### Possible solutions

1. Make the top pointer to always point to the first empty stack slot (init to `m_bottom`). This may negatively affect unary op, because `m_top - 1` must be used instead of `m_top` being used currently.
2. Grow the stack in reverse order. `m_top = m_bottom + max_stack_size` (i.e. the "end")`. Push is `*m_top-- = value`. This will work, but pointer arithmetic will be confusing.
3. Always add unused stack item (aka stack red zone). So `m_bottom[0]` is always allocated but should never be used. Init `m_top = m_bottom`. This works good, but requires more memory.
4. Conditionally add unused stack item when number of locals is zero. The rest works like 3. Requires more memory only in rare case of not having any locals (including params). Requires additional branch in initialization, but can be marked with `[[unlikely]]`.

3 and 4 have additional property that you can assign execution result value unconditionally because you can always safely access `stack.top()`, although in may contain garbage value.

axic:
I'd try to avoid (2) as that makes it worse for people writing host functions. I'd also try to avoid branching, so would vouch for (3).

chfast:
> I'd also try to avoid branching

What is the rationale for that?

axic:
> > I'd also try to avoid branching
>
> What is the rationale for that?

Wait a second, this is only done once in the constructor? If so, fine with branching.

PR #630, opened 2020-11-02
Adjust OperandStack allocation size when no locals

chfast:
Fixes https://github.com/wasmx/fizzy/issues/583.

### DIFF
--- a/lib/fizzy/stack.hpp
+++ b/lib/fizzy/stack.hpp
@@ -100,7 +100,8 @@ public:
     OperandStack(
         const Value* args, size_t num_args, size_t num_local_variables, size_t max_stack_height)
     {
-        const auto storage_size_required = num_args + num_local_variables + max_stack_height;
+        const auto num_locals = num_args + num_local_variables;
+        const auto storage_size_required = num_locals + max_stack_height;
 
         if (storage_size_required <= small_storage_size)
         {
@@ -112,7 +113,7 @@ public:
             m_locals = &m_large_storage[0];
         }
 
-        m_bottom = m_locals + num_args + num_local_variables;
+        m_bottom = m_locals + num_locals;
         m_top = m_bottom - 1;
 
         const auto local_variables = std::copy_n(args, num_args, m_locals);
--- a/lib/fizzy/stack.hpp
+++ b/lib/fizzy/stack.hpp
@@ -101,7 +101,10 @@ public:
         const Value* args, size_t num_args, size_t num_local_variables, size_t max_stack_height)
     {
         const auto num_locals = num_args + num_local_variables;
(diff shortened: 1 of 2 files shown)


test/unittests/stack_test.cpp
```diff
@@ -326,3 +327,17 @@ TEST(operand_stack, to_vector)
...
+{
+    constexpr auto max_height = 33;
+    OperandStack stack(nullptr, 0, 0, max_height);
+    ASSERT_GT(address_diff(&stack, stack.rbegin()), 100) << "not allocated on the heap";
+    EXPECT_EQ(stack.size(), 0);
+    EXPECT_EQ(stack.rbegin(), stack.rend());
+
+    // Even when stack is empty, there exists a single hidden item slot.
+    const_cast<fizzy::Value*>(stack.rbegin())->i64 = 1;
+    EXPECT_EQ(stack.rbegin()->i64, 1);
```
axic, line 341: Hm, I think we should document this behaviour in the header.

chfast: No. This is a logical error to do this.

axic: The documentation (https://en.cppreference.com/w/cpp/iterator/begin) doesn't state anywhere that `begin()` would be UB in case of `size() == 0`.

axic: @chfast any feedback?

chfast: Taking the iterator is fine, but you cannot dereference it when it is the end iterator.

axic: But here you are dereferencing it?

To be honest at this stage I lost what the conversation is about.

axic: Ah okay, we do ensure that `begin() == end()`, just that we know `begin()` is always valid.

axic (approved):

lib/fizzy/stack.hpp
```diff
@@ -99,7 +99,9 @@ class OperandStack
     OperandStack(
         const Value* args, size_t num_args, size_t num_local_variables, size_t max_stack_height)
     {
-        const auto storage_size_required = num_args + num_local_variables + max_stack_height;
+        const auto num_locals = num_args + num_local_variables;
+        const auto num_locals_adjusted = num_locals + (num_locals == 0);  // Bump to 1 if 0.
```
gumb0, line 103: Would be nice to have a comment explaining why it's needed.

gumb0 (approved):

axic:
@chfast can this be merged?

chfast:
If you really want to.

PR #631, opened 2020-11-02
capi: imported and exported memory

gumb0:


include/fizzy/fizzy.h
```diff
@@ -190,8 +204,8 @@ bool fizzy_find_exported_function(
 /// module, behaviour is undefined.
 FizzyInstance* fizzy_instantiate(const FizzyModule* module,
     const FizzyExternalFunction* imported_functions, size_t imported_functions_size,
-    const FizzyExternalTable* imported_table, const FizzyExternalGlobal* imported_globals,
-    size_t imported_globals_size);
+    const FizzyExternalTable* imported_table, const FizzyExternalMemory* imported_memory,
```
axic, line 207: The C++ API supports a vector of tables and memories, while the C API only supports a singular for both. While both can be at most one in wasm 1.0, there is a discrepancy between the two APIs.

Perhaps should also comment about this, as it may be confused to those familiar with the wasm spec.

In short: I think we should support a single table/memory in the C API as we can make a breaking API release if we do start supporting newer wasm versions. I would be inclined to say the same restriction would make sense in C++. Perhaps just open a tracking issue about this?

include/fizzy/fizzy.h
```diff
@@ -268,6 +268,15 @@ size_t fizzy_get_instance_memory_size(FizzyInstance* instance);
 bool fizzy_find_exported_table(
     FizzyInstance* instance, const char* name, FizzyExternalTable* out_table);
 
+/// Find exported memory by name.
+///
+/// @param  instance        Pointer to instance.
+/// @param  name            The table name. NULL-terminated string. Cannot be NULL.
+/// @param  out_memory      Pointer to output struct to store found memory. Cannot be NULL.
+/// @returns                true if memory was found, false otherwise.
+
```
axic, line 277: Should leave a comment this support a singular exported memory as per wasm 1.0.

gumb0: Well this function would support several memories just fine, because it finds them by name.

include/fizzy/fizzy.h
gumb0: #634

include/fizzy/fizzy.h
```diff
@@ -111,6 +111,18 @@ typedef struct FizzyExternalTable
     FizzyLimits limits;
 } FizzyExternalTable;
 
+/// The opaque data type representing a memory.
+typedef struct FizzyMemory FizzyMemory;
+
+/// External memory.
+typedef struct FizzyExternalMemory
+{
+    // Opaque pointer to memory data.
```
axic, line 120:
```suggestion
    /// Opaque pointer to memory data.
```

include/fizzy/fizzy.h
```diff
@@ -111,6 +111,18 @@ typedef struct FizzyExternalTable
...
+    FizzyMemory* memory;
+    // Memory limits.
```
axic, line 122:
```suggestion
    /// Memory limits.
```

test/unittests/capi_test.cpp
```diff
@@ -133,6 +133,41 @@ TEST(capi, find_exported_table)
...
+    ASSERT_NE(module, nullptr);
+
+    auto instance = fizzy_instantiate(module, nullptr, 0, nullptr, nullptr, nullptr, 0);
+    ASSERT_NE(instance, nullptr);
+
+    FizzyExternalMemory memory;
+    ASSERT_TRUE(fizzy_find_exported_memory(instance, "mem", &memory));
+    EXPECT_NE(memory.memory, nullptr);
+    EXPECT_EQ(memory.limits.min, 1);
+    EXPECT_TRUE(memory.limits.has_max);
```
axic, line 160: Can you add a test which has no max?

gumb0: Added.

axic (approved):

include/fizzy/fizzy.h
chfast: I agree on having single table/memory.

chfast (approved):

PR #633, opened 2020-11-03
capi: find_exported_function

gumb0:
I'm not completely sure it's a good idea to include this, because complications with context can be confusing, but the motivation is 
1. to make it simpler to import exported function into another module - see `capi.imported_function_from_anothe_module` test change.
2. symmetry with C++ API `fizzy::find_exported_*` functions.

### DIFF
--- a/include/fizzy/fizzy.h
+++ b/include/fizzy/fizzy.h
@@ -275,6 +275,29 @@ uint8_t* fizzy_get_instance_memory_data(FizzyInstance* instance);
 /// @note    Function returns memory size regardless of whether memory is exported or not.
 size_t fizzy_get_instance_memory_size(FizzyInstance* instance);
 
+/// Find exported function by name.
+///
+/// @param  instance        Pointer to instance.
+/// @param  name            The function name. NULL-terminated string. Cannot be NULL.
+/// @param  out_function    Pointer to output struct to store the found function. Cannot be NULL.
+///                         If function is found, associated context is allocated, which must exist
+///                         as long as the function can be called by some other instance, and should
+///                         be destroyed with fizzy_free_exported_function afterwards.
+///                         When function is not found (false returned), this out_function is not
+///                         modified, and fizzy_free_exported_function must not be called.
+/// @returns                true if function was found, false otherwise.
+bool fizzy_find_exported_function(
+    FizzyInstance* instance, const char* name, FizzyExternalFunction* out_function);
+
+/// Free resources associated with exported function.
+///
+/// @param  external_function   Pointer to external function struct filled by
+///                             fizzy_find_exported_function. Cannot be NULL.
+///
+/// @note   This function may not be called with external function, which was not returned from
+///         fizzy_find_exported_function.
+void fizzy_free_exported_function(FizzyExternalFunction* external_function);
+
 /// Find exported table by name.
 ///
 /// @param  instance        Pointer to instance.
--- a/include/fizzy/fizzy.h
+++ b/include/fizzy/fizzy.h
@@ -159,7 +159,7 @@ void fizzy_free_module(const FizzyModule* module);
 
 /// Get type of the function defined in the module.
 ///
-/// @param module   Pointer to module.
+/// @param module   Pointer to module. Cannot be NULL.
 /// @param func_idx Function index. Can be either index of an imported function or of a function
 ///                 defined in module. Behaviour is undefined, if index is not valid according to
 ///                 module definition.
@@ -169,7 +169,7 @@ FizzyFunctionType fizzy_get_function_type(const FizzyModule* module, uint32_t fu
 
 /// Find index of exported function by name.
 ///
-/// @param  module          Pointer to module.
+/// @param  module          Pointer to module. Cannot be NULL.
 /// @param  name            The function name. NULL-terminated string. Cannot be NULL.
 /// @param  out_func_idx    Pointer to output where function index will be stored. Cannot be NULL.
 /// @returns                true if function was found, false otherwise.
@@ -184,7 +184,7 @@ bool fizzy_find_exported_function_index(
 /// than once with the same module results in undefined behaviour), but after fizzy_instantiate
 /// functions querying module info can still be called with @p module.
 ///
-/// @param      module                   Pointer to module.
+/// @param      module                   Pointer to module. Cannot be NULL.
 /// @param      imported_functions       Pointer to the imported function array. Can be NULL iff
 ///                                      imported_functions_size equals 0.
 /// @param      imported_functions_size  Size of the imported function array. Can be zero.
@@ -223,7 +223,7 @@ FizzyInstance* fizzy_instantiate(const FizzyModule* module,
 /// than once with the same module results in undefined behaviour), but after fizzy_instantiate
 /// functions querying module info can still be called with @p module.
 ///
-/// @param      module                   Pointer to module.
+/// @param      module                   Pointer to module. Cannot be NULL.
 /// @param      imported_functions       Pointer to the imported function array. Can be NULL iff
 ///                                      imported_functions_size equals 0.
 /// @param      imported_functions_size  Size of the imported function array. Can be zero.
@@ -277,7 +277,7 @@ size_t fizzy_get_instance_memory_size(FizzyInstance* instance);
 
 /// Find exported function by name.
 ///
-/// @param  instance        Pointer to instance.
+/// @param  instance        Pointer to instance. Cannot be NULL.
 /// @param  name            The function name. NULL-terminated string. Cannot be NULL.
 /// @param  out_function    Pointer to output struct to store the found function. Cannot be NULL.
 ///                         If function is found, associated context is allocated, which must exist
@@ -300,7 +300,7 @@ void fizzy_free_exported_function(FizzyExternalFunction* external_function);
 
 /// Find exported table by name.
 ///
-/// @param  instance        Pointer to instance.
+/// @param  instance        Pointer to instance. Cannot be NULL.
 /// @param  name            The table name. NULL-terminated string. Cannot be NULL.
 /// @param  out_table       Pointer to output struct to store found table. Cannot be NULL.
 /// @returns                true if table was found, false otherwise.
@@ -311,7 +311,7 @@ bool fizzy_find_exported_table(
 
 /// Find exported memory by name.
 ///
-/// @param  instance        Pointer to instance.
+/// @param  instance        Pointer to instance. Cannot be NULL.
 /// @param  name            The table name. NULL-terminated string. Cannot be NULL.
 /// @param  out_memory      Pointer to output struct to store found memory. Cannot be NULL.
 /// @returns                true if memory was found, false otherwise.
@@ -322,7 +322,7 @@ bool fizzy_find_exported_memory(
 
 /// Find exported global by name.
 ///
-/// @param  instance        Pointer to instance.
+/// @param  instance        Pointer to instance. Cannot be NULL.
 /// @param  name            The global name. NULL-terminated string. Cannot be NULL.
 /// @param  out_global      Pointer to output struct to store found global. Cannot be NULL.
 /// @returns                true if global was found, false otherwise.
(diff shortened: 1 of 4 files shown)


axic:
I think this would also require a benchmark to see if the overhead changed.

gumb0:
> I think this would also require a benchmark to see if the overhead changed.

Well the old option to access exported function via index stays, and `FizzyCEngine` still uses it as before.  So what kind of benchmark do you have  in mind?

chfast (commented):
Seems ok, but there is always a question if this can be done without allocating new object on heap.

include/fizzy/fizzy.h
```diff
@@ -275,6 +278,24 @@ uint8_t* fizzy_get_instance_memory_data(FizzyInstance* instance);
 /// @note    Function returns memory size regardless of whether memory is exported or not.
 size_t fizzy_get_instance_memory_size(FizzyInstance* instance);
 
+/// Find exported function by name.
+///
+/// @param  instance        Pointer to instance.
+/// @param  name            The table name. NULL-terminated string. Cannot be NULL.
```
axic, line 284:
```suggestion
/// @param  name            The function name. NULL-terminated string. Cannot be NULL.
```

include/fizzy/fizzy.h
```diff
@@ -275,6 +278,24 @@ uint8_t* fizzy_get_instance_memory_data(FizzyInstance* instance);
...
+/// @param  out_function    Pointer to output struct to store found table. Cannot be NULL.
```
axic, line 285:
```suggestion
/// @param  out_function    Pointer to output struct to store the found function. Cannot be NULL.
```

include/fizzy/fizzy.h
```diff
@@ -275,6 +278,24 @@ uint8_t* fizzy_get_instance_memory_data(FizzyInstance* instance);
 /// @note    Function returns memory size regardless of whether memory is exported or not.
 size_t fizzy_get_instance_memory_size(FizzyInstance* instance);
 
+/// Find exported function by name.
+///
+/// @param  instance        Pointer to instance.
+/// @param  name            The function name. NULL-terminated string. Cannot be NULL.
+/// @param  out_function    Pointer to output struct to store the found function. Cannot be NULL.
+/// @param  out_context_ptr Pointer to opaque output pointer to a context associated with this
```
axic, line 286: Should also mention that in the case of failure (`returns false`), what will be the value of this (whether undefined or set to null) and whether that needs to be freed, or not.

gumb0:
> Seems ok, but there is always a question if this can be done without allocating new object on heap.

I think only if it were allocated on the heap already on C++ side. Or if we made some other, yet more complicated change in C++ to get rid of `std::function`.

test/unittests/capi_test.cpp
```diff
@@ -98,6 +98,45 @@ TEST(capi, find_exported_function_index)
...
+
+    auto instance = fizzy_instantiate(module, nullptr, 0, nullptr, nullptr, nullptr, 0);
+    ASSERT_NE(instance, nullptr);
+
+    FizzyExternalFunction function;
+    FizzyExportedFunctionContext* context;
+    ASSERT_TRUE(fizzy_find_exported_function(instance, "foo", &function, &context));
+    EXPECT_EQ(function.type.inputs_size, 0);
+    EXPECT_EQ(function.type.output, FizzyValueTypeI32);
+    EXPECT_EQ(function.context, context);
```
axic, line 126: Hmm, why do we need to return `FizzyExportedFunctionContext` if the context is available here as well?

gumb0: Yeah, we return it twice - as `void*` opaque external function context and as `FizzyExportedFunctionContext` for convenience.

Also not completely sure it's a good idea, but without this we either require the user to do the cast to `FizzyExportedFunctionContext*` themselves, or make "free" function less type-safe like `fizzy_free_exported_function_context(void*)`.

include/fizzy/fizzy.h
gumb0: Added a sentence.

test/unittests/capi_test.cpp
axic: But why not make the function expect a `void*` so that no cast is needed, and introduce `fizzy_free_exported_function(FizzyExternalFunction)`?

This was the `FizzyExternalFunction` is fully symmetric whether it is defined as a host function or acquired as an import. Can be called the same way, can be passed the same way.

The only safety thing the user has to observer is that `fizzy_free_exported_function()` is UB if they call it on something not acquired via `fizzy_find_exported_function()`.

What do you think @gumb0 @chfast ?

gumb0: > Also not completely sure it's a good idea, but without this we either require the user to do the cast to `FizzyExportedFunctionContext*` themselves, or make "free" function less type-safe like `fizzy_free_exported_function_context(void*)`.

Actually in this variant cast would be implicit, so it's either
`fizzy_free_exported_function_context(FizzyExportedFunctionContext*)` and user would pass `void*` there
or `fizzy_free_exported_function_context(void*)`

gumb0: I like the idea with `fizzy_free_exported_function(FizzyExternalFunction*)`, it's simpler.

Only need to clarify in documentation that it should be used with exported functions only.

include/fizzy/fizzy.h
```diff
@@ -275,6 +275,28 @@ uint8_t* fizzy_get_instance_memory_data(FizzyInstance* instance);
...
+bool fizzy_find_exported_function(
+    FizzyInstance* instance, const char* name, FizzyExternalFunction* out_function);
+
+/// Free resources associated with exported function.
+///
+/// If passed pointer is NULL, has no effect.
+///
+/// @note   This function must be called only with external functions returned from
+///         fizzy_find_exported_function.
+void fizzy_free_exported_function(const FizzyExternalFunction* external_function);
```
axic, line 298: I wonder if we should actually make this work on a copy (`fizzy_free_exported_function(FizzyExternalFunction)`), given the only thing we're freeing is a pointer in the struct.

From an API PoV I think the current one makes more sense, but it may not be fully intuitive for the user?

include/fizzy/fizzy.h
```diff
@@ -275,6 +275,28 @@ uint8_t* fizzy_get_instance_memory_data(FizzyInstance* instance);
...
+///                         be destroyed with fizzy_free_exported_function afterwards.
+///                         When function is not found (false returned), this out_function is not
+///                         modified, and fizzy_free_exported_function must not be called.
+/// @returns                true if function was found, false otherwise.
+bool fizzy_find_exported_function(
+    FizzyInstance* instance, const char* name, FizzyExternalFunction* out_function);
+
+/// Free resources associated with exported function.
+///
+/// If passed pointer is NULL, has no effect.
```
axic, line 294: This is a bit confusing since we never return `FizzyExternalFunction` as a pointer.

gumb0: Well we pass it by non-const pointer in `fizzy_find_exported_function`

include/fizzy/fizzy.h
gumb0: Not sure, it's a function operating on a struct, so we pass it by const pointer to avoid unnecessaty copying, seems usual for C ?

axic: I mean the fact we don't obtain it as a pointer, but rather allocate it on the stack in the caller.

include/fizzy/fizzy.h
axic: I understand that. This comment is connected to the one above.

include/fizzy/fizzy.h
gumb0: Yeah I see, I'm not sure what's better.

Another option is to call it with `_context` anyway: `fizzy_free_exported_function_context(FizzyExternalFunction func)` - this might be less weird than passing an instance copy to a function that's supposed to "free function".

lib/fizzy/capi.cpp
```diff
@@ -120,6 +120,20 @@ inline auto unwrap(FizzyExternalFn func, void* context) noexcept
     };
 }
 
+inline FizzyExternalFunction wrap(fizzy::ExternalFunction external_func)
+{
+    FizzyExternalFn c_function = [](void* context, FizzyInstance* instance, const FizzyValue* args,
```
chfast, line 125:
```suggestion
    static constexpr FizzyExternalFn c_function = [](void* context, FizzyInstance* instance, const FizzyValue* args,
```

lib/fizzy/capi.cpp
```diff
@@ -120,6 +120,20 @@ inline auto unwrap(FizzyExternalFn func, void* context) noexcept
...
+                                     int depth) -> FizzyExecutionResult {
+        auto* func = static_cast<fizzy::ExternalFunction*>(context);
+        return wrap((func->function)(*unwrap(instance), unwrap(args), depth));
+    };
+
+    auto context = std::make_unique<fizzy::ExternalFunction>(std::move(external_func));
```
chfast, line 131: This is not great, but it can stay until we remove usage of `std::function`.

include/fizzy/fizzy.h
chfast, line 298:
```suggestion
void fizzy_free_exported_function(FizzyExternalFunction* external_function);
```
Although technically possible, this should be passed without `const` because you are logically modifying the object.

gumb0: My motivation was to show exactly that the object itself is not modifed, to make it look a bit less weird when we call it with an instance on stack.
```
FizzyExternalFunction function;
...
fizzy_free_exported_function(&function);
```
But I don't insist, removed `const`  now.

axic:
What is the status here?

gumb0:
> What is the status here?

It's ready from my side.

You suggested passing argument by value to `fizzy_free_exported_function`, but I prefer for it to stay a pointer.

include/fizzy/fizzy.h
gumb0: According to comment https://github.com/wasmx/fizzy/pull/633#discussion_r521299948 @chfast seems to prefer `FizzyExternalFunction*` argument

axic:
I think neither of the options are good, but it is still better than not having it. Probably with the Rust implementation we'll be informed if we should revisit this.

axic (approved):

test/unittests/capi_test.cpp
```diff
@@ -87,17 +87,55 @@ TEST(capi, find_exported_function)
...
+
+    FizzyExternalFunction function;
+    ASSERT_TRUE(fizzy_find_exported_function(instance, "foo", &function));
+    EXPECT_EQ(function.type.inputs_size, 0);
+    EXPECT_EQ(function.type.output, FizzyValueTypeI32);
+    EXPECT_NE(function.context, nullptr);
+    ASSERT_NE(function.function, nullptr);
+    EXPECT_THAT(function.function(function.context, instance, nullptr, 0), CResult(42));
+
+    fizzy_free_exported_function(&function);
```
axic, line 129: Since you state `fizzy_free_exported_function(nullptr)` works, can you add it somewhere?

include/fizzy/fizzy.h
```diff
@@ -275,6 +275,28 @@ uint8_t* fizzy_get_instance_memory_data(FizzyInstance* instance);
 /// @note    Function returns memory size regardless of whether memory is exported or not.
 size_t fizzy_get_instance_memory_size(FizzyInstance* instance);
 
+/// Find exported function by name.
+///
+/// @param  instance        Pointer to instance.
```
chfast, line 280:
```suggestion
/// @param  instance        Pointer to instance. Cannot be NULL?
```

chfast (approved):

test/unittests/capi_test.cpp
gumb0: It was a mistake, it's not allowed, will change the comment.

gumb0: Changed.

include/fizzy/fizzy.h
gumb0: Added here and in other functions.

axic (approved):

PR #637, opened 2020-11-10
api: resolving imported globals

gumb0:
~~This extends `resolve_imported_functions` function into supporting all kinds of imports.~~

I decided to leave here only support for imported globals. Resolving imported table and memory doesn't seem very useful, becasue there's always at most one of each. I suggest to postpone that till we see there's real need to resolve them.

C API part will be in another PR.

### DIFF
--- a/lib/fizzy/instantiate.cpp
+++ b/lib/fizzy/instantiate.cpp
@@ -375,7 +375,7 @@ std::unique_ptr<Instance> instantiate(std::unique_ptr<const Module> module,
 }
 
 std::vector<ExternalFunction> resolve_imported_functions(
-    const Module& module, std::vector<ImportedFunction> imported_functions)
+    const Module& module, const std::vector<ImportedFunction>& imported_functions)
 {
     std::vector<ExternalFunction> external_functions;
     for (const auto& import : module.importsec)
--- a/lib/fizzy/instantiate.cpp
+++ b/lib/fizzy/instantiate.cpp
@@ -233,6 +233,37 @@ Value eval_constant_expression(ConstantExpression expr,
         return globals[global_idx - imported_globals.size()];
 }
 
+ExternalFunction find_imported_function(const std::string& module, const std::string& name,
+    const FuncType& module_func_type, const std::vector<ImportedFunction>& imported_functions)
+{
+    const auto it = std::find_if(imported_functions.begin(), imported_functions.end(),
+        [module, name](const auto& func) { return module == func.module && name == func.name; });
+
+    if (it == imported_functions.end())
+    {
+        throw instantiate_error{"imported function " + module + "." + name + " is required"};
+    }
+
+    if (module_func_type.inputs != it->inputs)
+    {
+        throw instantiate_error{"function " + module + "." + name +
+                                " input types don't match imported function in module"};
+    }
+    if (module_func_type.outputs.empty() && it->output.has_value())
+    {
+        throw instantiate_error{
+            "function " + module + "." + name + " has output but is defined void in module"};
+    }
+    if (!module_func_type.outputs.empty() &&
+        (!it->output.has_value() || module_func_type.outputs[0] != *it->output))
+    {
+        throw instantiate_error{"function " + module + "." + name +
+                                " output type doesn't match imported function in module"};
+    }
+
+    return {it->function, module_func_type};
+}
+
 std::optional<uint32_t> find_export(const Module& module, ExternalKind kind, std::string_view name)
 {
     const auto it = std::find_if(module.exportsec.begin(), module.exportsec.end(),
@@ -383,40 +414,12 @@ std::vector<ExternalFunction> resolve_imported_functions(
         if (import.kind != ExternalKind::Function)
             continue;
 
-        const auto it = std::find_if(
-            imported_functions.begin(), imported_functions.end(), [&import](const auto& func) {
-                return import.module == func.module && import.name == func.name;
-            });
-
-        if (it == imported_functions.end())
-        {
-            throw instantiate_error{
-                "imported function " + import.module + "." + import.name + " is required"};
-        }
-
         assert(import.desc.function_type_index < module.typesec.size());
         const auto& module_func_type = module.typesec[import.desc.function_type_index];
 
-        if (module_func_type.inputs != it->inputs)
-        {
-            throw instantiate_error{"function " + import.module + "." + import.name +
-                                    " input types don't match imported function in module"};
-        }
-        if (module_func_type.outputs.empty() && it->output.has_value())
-        {
-            throw instantiate_error{"function " + import.module + "." + import.name +
-                                    " has output but is defined void in module"};
-        }
-        if (!module_func_type.outputs.empty() &&
-            (!it->output.has_value() || module_func_type.outputs[0] != *it->output))
-        {
-            throw instantiate_error{"function " + import.module + "." + import.name +
-                                    " output type doesn't match imported function in module"};
-        }
-
-        external_functions.emplace_back(ExternalFunction{it->function, module_func_type});
+        external_functions.emplace_back(find_imported_function(
+            import.module, import.name, module_func_type, imported_functions));
     }
-
     return external_functions;
 }
 
--- a/lib/fizzy/instantiate.cpp
+++ b/lib/fizzy/instantiate.cpp
@@ -264,6 +264,33 @@ ExternalFunction find_imported_function(const std::string& module, const std::st
     return {it->function, module_func_type};
 }
 
+ExternalGlobal find_imported_global(const std::string& module, const std::string& name,
+    GlobalType module_global_type, const std::vector<ImportedGlobal>& imported_globals)
+{
+    const auto it = std::find_if(imported_globals.begin(), imported_globals.end(),
+        [module, name](const auto& func) { return module == func.module && name == func.name; });
+
+    if (it == imported_globals.end())
+    {
+        throw instantiate_error{"imported global " + module + "." + name + " is required"};
+    }
+
+    if (module_global_type.value_type != it->type)
+    {
+        throw instantiate_error{"global " + module + "." + name +
+                                " value type doesn't match imported global in module"};
+    }
+    if (module_global_type.is_mutable != it->is_mutable)
+    {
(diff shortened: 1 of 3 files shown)


gumb0:
@chfast @axic Please check that it's an ok direction for resolve API.

I will add tests next.

lib/fizzy/instantiate.hpp
```diff
@@ -117,11 +117,47 @@ struct ImportedFunction
...
+struct ImportedGlobal
+{
+    std::string module;
+    std::string name;
+    Value* value = nullptr;
+    ValType type = ValType::i32;
+    bool is_mutable = false;
+};
+
+struct Externals
```
gumb0, line 147: `instantiate` can also be changed to use `Externals` struct as input, but it can be another PR.

chfast:
> @chfast @axic Please check that it's an ok direction for resolve API.

Add description first.

gumb0:
> > @chfast @axic Please check that it's an ok direction for resolve API.
> 
> Add description first.

Added.

lib/fizzy/instantiate.hpp
```diff
@@ -117,11 +117,47 @@ struct ImportedFunction
...
+    std::vector<ExternalFunction> functions;
+    std::vector<ExternalTable> tables;
+    std::vector<ExternalMemory> memories;
+    std::vector<ExternalGlobal> globals;
+};
+
+// Create vectors of External* ready to be passed to instantiate.
+// imported_* may be in any order,
+// but must contain imports for all of the import names defined in the module.
+Externals resolve_imports(const Module& module, std::vector<ImportedFunction> imported_functions,
```
chfast, line 158: The `std::span` should be used as the parameters type, but I'm not sure how good our implementation is.

chfast (commented):
Looks ok, but I'm confused how this was done previously.

lib/fizzy/instantiate.hpp
axic, line 158: I think it would be nice to structure this differently, from a consumer side:
```
Externals externals(module);
externals.resolve_global(ExternalGlobal);
externals.resolve_globals(...); // with a span
```

gumb0: At what point would it check that all required imports are provided? In a separate `bool is_complete()` method?

gumb0: I thinks I'm still more in favor of simpler functional API like `vector<External*> resolve_imported_*(Module, span<Imported*>)` because the rest of the API is mostly functional.

(Otherwise it seems better to make this a full-fledged builder object with `instantiate` method in the end, but then there'll probably be some complications around `module` and externals being consumed by `instantiate`...)

I will split `resolve_imports` into 4 functions for now and will try to use `span`.

chfast: I'm not convinced this is good design, but I don't think this is important right now. Good API is hard to do on paper, so I will accept whatever forks for Rust currently.

Better is to observe how this works in some real-life examples. We have two of these right now: `fizzy-wasi` and `FizzyEngine`.

gumb0: In general I'm not content either with any of the options for resolve proposed so far, and would just pick something good enough. 

(`fizzy-wasi` and `FizzyEngine` are quite simplistic use cases, as they import only functions. I think any of the options would look fine in them)

chfast: My point is we should start with something "good enough" and the tune it based on example integrations, not on intuition nor unit tests.

gumb0: I tried it, and `span` doesn't look suitable here, because it's supposed to be created over existing container (it doesn't have an `initializer_list` constructor, because `initializer_list` is not a container, but just a temporary), so the `vector` would need to be explicitly created on the caller side. Whereas now it's possible to pass just an initializer list.

Now:
```
        auto imports = fizzy::resolve_imported_functions(
            *module, {
                     });
```

With `span`:
```
        auto imports = fizzy::resolve_imported_functions(
            *module, std::vector<ImportedFunction>{
                     });
```

gumb0: I think I'll leave it as vector const reference.

lib/fizzy/instantiate.cpp
```diff
@@ -227,6 +227,62 @@ Value eval_constant_expression(ConstantExpression expr,
...
+    {
+        throw instantiate_error{"global " + module + "." + name +
+                                " value type doesn't match imported global in module"};
+    }
+    if (module_global_type.is_mutable != it->is_mutable)
+    {
+        throw instantiate_error{"global " + module + "." + name +
+                                " mutability doesn't match imported global in module"};
+    }
+
```
gumb0, line 282: Should we check here also that `it->value != nullptr` ?
(note that it's checked in `instantiate` anyway)

chfast (approved):

gumb0:
Rebased.

lib/fizzy/instantiate.cpp
```diff
@@ -375,51 +431,38 @@ std::unique_ptr<Instance> instantiate(std::unique_ptr<const Module> module,
...
-        }
-        if (!module_func_type.outputs.empty() &&
-            (!it->output.has_value() || module_func_type.outputs[0] != *it->output))
-        {
-            throw instantiate_error{"function " + import.module + "." + import.name +
-                                    " output type doesn't match imported function in module"};
-        }
-
-        external_functions.emplace_back(ExternalFunction{it->function, module_func_type});
+        external_functions.emplace_back(find_imported_function(
```
axic, line 445: Is this refactoring changing behaviour? i.e. it would have stopped with `instantiate_error` in certain cases, but now it will catch an assert?

lib/fizzy/instantiate.hpp
```diff
@@ -135,7 +135,23 @@ struct ImportedFunction
...
 std::vector<ExternalFunction> resolve_imported_functions(
-    const Module& module, std::vector<ImportedFunction> imported_functions);
+    const Module& module, const std::vector<ImportedFunction>& imported_functions);
+
+// Global that should be used by instantiate as import, identified by module and global name.
+struct ImportedGlobal
+{
+    std::string module;
+    std::string name;
+    Value* value = nullptr;
```
axic, line 145: Hmm, `module`/`name`/`value` does not seem to be used ye?

gumb0: `find_imported_global` puts it into returned `ExternalGlobal`

lib/fizzy/instantiate.cpp
gumb0: Now, it just changes the order, the checks for `instantiate_error` are moved into `find_imported_function`.

Assertion is guaranteed to be true by parser.

axic: Yes I see the checks are moved to `find_imported_function`, but now they are executed post assertions.

lib/fizzy/instantiate.hpp
axic: Ah yes I missed the return value.

lib/fizzy/instantiate.cpp
axic: I think it is fine as is. Perhaps a comment here could be useful to say that `instantiate` will validate the presence of a value.

test/unittests/api_test.cpp
```diff
@@ -212,6 +212,150 @@ TEST(api, resolve_imported_function_duplicate)
...
+        {"mod1", "g1", &global1, ValType::i64, false},
+        {"mod1", "g2", &global2, ValType::i32, true},
+        {"mod2", "g1", &global3, ValType::i64, false},
+        {"mod2", "g2", &global4, ValType::i64, true},
+    };
+
+    EXPECT_THROW_MESSAGE(resolve_imported_globals(*module, imported_globals_invalid_type1),
+        instantiate_error, "global mod1.g1 value type doesn't match imported global in module");
+
+    const std::vector<ImportedGlobal> imported_globals_invalid_type2 = {
```
axic, line 320:
```suggestion
    const std::vector<ImportedGlobal> imported_globals_invalid_mutability = {
```

test/unittests/api_test.cpp
```diff
@@ -212,6 +212,150 @@ TEST(api, resolve_imported_function_duplicate)
...
+    const std::vector<ImportedGlobal> imported_globals_missing = {
+        {"mod2", "g1", &global3, ValType::i64, false},
+        {"mod1", "g2", &global2, ValType::i32, true},
+        {"mod1", "g1", &global1, ValType::i32, false},
+    };
+
+    EXPECT_THROW_MESSAGE(resolve_imported_globals(*module, imported_globals_missing),
+        instantiate_error, "imported global mod2.g2 is required");
+
+    const std::vector<ImportedGlobal> imported_globals_invalid_type1 = {
```
axic, line 310:
```suggestion
    const std::vector<ImportedGlobal> imported_globals_invalid_type = {

```

lib/fizzy/instantiate.cpp
gumb0: They are independent - ifs are checking `imported_functions` argument, and assertion checks that `module` is not malformed.

lib/fizzy/instantiate.cpp
gumb0: Added a comment.

test/unittests/api_test.cpp
gumb0: Changed.

test/unittests/api_test.cpp
gumb0: Changed.

lib/fizzy/instantiate.hpp
```diff
@@ -137,6 +137,22 @@ struct ImportedFunction
...
+    ValType type = ValType::i32;
+    bool is_mutable = false;
+};
+
+// Create vector of ExternalGlobals ready to be passed to instantiate.
+// imported_globals may be in any order, but must contain globals for all of the imported global
```
axic, line 151: Just realised these comments are no doxygen, neither for `resolve_imported_functions`. If we want to doxygenize them, lets do it in a new PR.

axic (approved):

PR #639, opened 2020-11-11
Fix resolving imported functions when importing the same function more than once

gumb0:


lib/fizzy/instantiate.cpp
```diff
@@ -408,8 +408,7 @@ std::vector<ExternalFunction> resolve_imported_functions(
                                     " output type doesn't match imported function in module"};
         }
 
-        external_functions.emplace_back(
-            ExternalFunction{std::move(it->function), module_func_type});
+        external_functions.emplace_back(ExternalFunction{it->function, module_func_type});
```
gumb0, line 411: Now input `imported_functions` could be passed by const reference, not sure it's worth to change.

chfast (approved):

test/unittests/api_test.cpp
```diff
@@ -193,6 +193,31 @@ TEST(api, resolve_imported_functions)
         "function mod1.foo2 output type doesn't match imported function in module");
 }
 
+TEST(api, resolve_imported_function_duplicate)
+{
+    /* wat2wasm
+      (func (import "mod1" "foo1") (param i32) (result i32))
+      (func (import "mod1" "foo1") (param i32) (result i32))
```
axic, line 200: Hmm, are there such test cases in spectests?

axic: Also can such a test be added to the C++ part as well?

gumb0: > Hmm, are there such test cases in spectests?

I found at least this one case of importing the same function twice:
https://github.com/wasmx/wasm-spec/blob/7526564b56c30250b66504fe795e9c1e88a938af/test/core/imports.wast#L30-L33

Our spec test runner doesn't use resolve function, so it wouldn't be affected by this.

gumb0: Also there's this curious case of importing overloaded functions with different types, which is expected to be valid but fail to instantiate:
https://github.com/wasmx/wasm-spec/blob/7526564b56c30250b66504fe795e9c1e88a938af/test/core/imports.wast#L568

Perhaps `resolve` should allow several functions with the same name and different types, too. I would fix this in another PR.

gumb0: > Also can such a test be added to the C++ part as well?

I assume you mean to C API part, because this is C++ API test.

axic: > > Also can such a test be added to the C++ part as well?
>
> I assume you mean to C API part, because this is C++ API test.

Ah you do seem to have it in both.

axic: > Our spec test runner doesn't use resolve function, so it wouldn't be affected by this.

Hm, should it?

gumb0: > > Our spec test runner doesn't use resolve function, so it wouldn't be affected by this.
> 
> Hm, should it?

I think maybe not necessary, and in current form it would be really inconvenient. 

To me it seems that resolve API suits more to import host functions - because of convenience of flattened `ImportedFunction` struct and other `Imported*` from #637. Whereas importing from one module to another requires `External*` structs - which are returned from `find_exported_*` functions - that's what spectest runner uses.

gumb0: Added an issue https://github.com/wasmx/fizzy/issues/645

gumb0: > Also there's this curious case of importing overloaded functions with different types, which is expected to be valid but fail to instantiate:
> https://github.com/wasmx/wasm-spec/blob/7526564b56c30250b66504fe795e9c1e88a938af/test/core/imports.wast#L568
> 
> Perhaps `resolve` should allow several functions with the same name and different types, too. I would fix this in another PR.

Also an issue for this https://github.com/wasmx/fizzy/issues/647

axic (approved):

PR #641, opened 2020-11-16
test: Add more tests of calls to functions without results

chfast:
Extracted from #642.


gumb0 (approved):

chfast:
Addressing @axic comment: I don't know if this is covered by the spectests, but this is motivated by "future Fizzy-specific implementation edge case". I don't this this is something that must be in the spectests.

axic (approved):

PR #642, opened 2020-11-16
Optimize internal calls by passing only stack end for arguments

chfast:
Instead of passing the reference to the call arguments (a span on the top of the stack) just pass the pointer to the stack end.
Also, don't get the result by `ExecutionResult`, just expect the result being set to `args[0]` and the new stack end pointer to be returned.

By this change, the caller does not need to know what the type of the callee function is. See `invoke_function()` simplification. It only passes the stack pointer and updates the stack pointer after the call.

1. This is a prototype, implementation is messy.
2. The host functions are not updated to the new calling convention. So there is even more `execute()` wrapping currently. But they may and probably should be updated.
3. I used "stack end", but also "stack top" can be used. The "stack end" looked easier and safer for prototype.
4. Returning `nullptr` means trap, but this may be a trap for call without arguments where `nullptr` could be passed as the "no arguments" value. Currently the wrapper handles that nicely (see `fake_arg`). But we can consider some sentinel address values.
5. It would be better to handle the imported function calls in the wrapper or in `call` implementation, not in the `execute()`. I realized this to late to change the design in the prototype.


test/unittests/execute_call_test.cpp
```diff
@@ -58,6 +58,69 @@ TEST(execute_call, call_with_arguments)
...
+    )
+    */
+    const auto wasm = from_hex(
+        "0061736d010000000108026000006000017f03030200010606017f01417f0b0a0f020600410124000b06001000"
+        "23000b");
+
+    EXPECT_THAT(execute(parse(wasm), 1, {}), Result(1));
+}
+
+TEST(execute_call, call_void_with_one_argument)
```
axic, line 82: @chfast I think these tests could be merged anyway.

chfast: #641.

PR #643, opened 2020-11-16
refactor: ExecuteFunction instead of std::function

gumb0:


lib/fizzy/instantiate.cpp
```diff
@@ -420,18 +420,29 @@ std::optional<FuncIdx> find_exported_function(const Module& module, std::string_
     return find_export(module, ExternalKind::Function, name);
 }
 
+ExecuteFunction::ExecuteFunction(Instance& instance, FuncIdx func_idx)
+  : m_function{[](void* context, Instance&, const Value* args, int depth) {
+        auto [func_idx, instance] = *static_cast<std::pair<FuncIdx, Instance*>*>(context);
+        return execute(*instance, func_idx, args, depth);
+    }},
+    m_context{new std::pair<FuncIdx, Instance*>(func_idx, &instance),
```
chfast, line 428: I don't have full picture but this does not look good.

I want to thing about the "ExecuteFunction" as a variant of `(this instance, func_idx)`, `(other instance, func_idx)`, `(host function pointer, host function context pointer)`. So you need to store two pointers at most + enum to know the kind. You can start with `std::variant<WasmFunction, HostFunction>` as a start, but it might be difficult to convert it to C.

gumb0: ( `(this instance, func_idx)` is not needed, because this now is used only for exported/imported functions)

Yeah, on the C side this is going to be a `union`, which might be ugly, but maybe we could add some helpers to simplify work with it... I'll think about it.

Anyway C API is the point of most complications with these refactorings...

chfast: This should be union on C++ side too, the `std::variant` is just a wrapper for union. So you can use the C type (and add convenient wrapping by inheritance) if you expect `reinterpret_cast` is convenient.

The version with union can be like this:
```
struct F {
  Instance* instance;
  union {
    FuncIdx func_idx;
    HostFn host_fn;
  }
  void* host_context;
}
```
When `instance != nullptr` you use `func_idx`, otherwise this is a host function and you use `host_fn`.

Dropping `union{}` for now will also work, just the size is +33%.

gumb0: Main complications are around converting from C function pointer to C++ function pointer (different signature) and back.

E.g. when providing host imported function from C side, both context and C function pointer have to be saved somewhere, function pointer can't be just casted. (perhaps having a second `void*` in `ExecuteFunction` might solve it)
When returning re-exported C++ host function to C side, orignal context and original C++ host function pointer and also arrays with input parameter types have to be stored somewhere (perhaps having a second `void*` context in `FizzyExternalFunction` might solve it; parameter types should just point to original C++ module, that's another refactoring....)

gumb0:
Current status:

1. Union apporach is generally working, with some caveats.

2. #668 is prerequisite

3. Main complication is around converting between C API host function pointer and C++ API host function pointer.

4. When importing host function from the C side, the original C function ponter has to be saved together the original `void*` context. Without dynamical allocation I achieve this by defining C++ function pointer as having two `void*` context parameters, where second context is used only by C API implementation, passing pointer to original C function in it. Normal C++ API is supposed to set second context to `nullptr` always (but I don't see how this can be enforced elegantly (one ugly idea would involve leaking details of C API into C++ part))

This can also be expressed as a union with three members (wasm func, normal host func with single context, C host func with 2 contexts)

5. (this can be a separate PR)
When C++ host function is returned to the C side from fizzy_find_exported_function (to be imported into another module), the original C++ function pointer has to be saved together with original context. It can be addressed with dynamic allocation (similar to what happens currently), otherwise we need a complexity of a union in C API. This is also working now, see last commit.

6. Validity of `reinterpret_cast`ing from a function pointer to `void*` and back is implementation-dependent (as I understand it might not work on some weird system where pointer to function is not a pointer in regular memory, can have different width etc.), but is guaranteed on POSIX systems at least
image

lib/fizzy/instantiate.hpp
```diff
@@ -22,9 +22,42 @@ namespace fizzy
...
+    {}
+
+    ExecutionResult operator()(Instance& instance, const Value* args, int depth);
+
+    union
+    {
+        std::tuple<ExecuteFunctionExtendedPtr, void*, void*> host_func;
+        std::tuple<Instance*, FuncIdx> wasm_func;
+    };
+    bool m_is_host_func = false;
```
gumb0, line 55: This perhaps better could be

```suggestion
    union
    {
        std::tuple<ExecuteFunctionPtr, void*> host_func; // C++ host func
        std::tuple<ExecuteFunctionExtendedPtr, void*, void*> c_host_func; // C host func
        std::tuple<Instance*, FuncIdx> wasm_func;
    };
    ExecuteFunctionKind kind;
```

gumb0:
Note on current usability

7. C++ API: new `void*` argument for host function signatures + additional curly braces required for some reason:
image


8. C API: ...

axic:
What is the status of this PR?

gumb0:
> What is the status of this PR?

I described the status in the comments above, but also it can be seen as three variants of solutions:

1) 1st commit - the simplest one, C API is not changed at all, we can make execution `noexcept`, but we don't get rid of dynamic allocation.
2) 2nd and 3d commits - we get rid of dynamic allocation on the C++ side by making `ExecuteFunction` less general, APIs are still relatively simple, and there is still dynamic allocation in the C API.
3) 4th commit - get rid of dynamic allocation on the C API side by introducing union, C API gets hairy.

lib/fizzy/instantiate.hpp
```diff
@@ -21,23 +21,47 @@ namespace fizzy
...
+      : m_function{f}, m_context{context, context_deleter}
+    {}
+
+    template <class F>
+    ExecuteFunction(F f) : m_function{f}, m_context{nullptr, [](void*) noexcept {}}
+    {}
+
+    ExecuteFunction(Instance& instance, FuncIdx func_idx);
+
+    ExecutionResult operator()(Instance& instance, const Value* args, int depth);
```
axic, line 48: Wouldn't this be also `noexcept`? Or `m_context.get` is not.

gumb0: It  will be `noexcept`.
I'm thinking next to try extracting minimal needed changes from this PR and applying `noexcept` change. It will look much simpler than this PR, but will have extra dynamic allocations, whiche "does not look good" according to @chfast

gumb0:
The approach in #716 hid a lot of ugliness here behind `std::any`, so this is mostly not needed now.
(The final part here provided some additional benefit - getting rid of dynamic allocation in C API, but it would need to be somehow reworked over `std::any` approach, if it would be needed)

Issue #644, opened 2020-11-17
Improve code coverage


chfast:
- [ ] Add coverage for Rust. #706
- [x] Add GCC coverage for spectests. #649
- [x] Improve LLVM coverage for spectests - do it only for `fizzy-spectests` binary. #646
- [x] Upload spectests coverage to codecov with separate tag. #649

PR #649, opened 2020-11-18
CI: Coverage flags

chfast:
- Upload coverage from internal tests with "unittests" flag.
- Upload coverage from spectests with "spectests" flag.
- Show the coverage for each flag in the Codecov Report comment.
See #644.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -170,6 +170,9 @@ commands:
 
   upload_coverage:
     description: "Upload coverage data"
+    parameters:
+      flags:
+        type: string
     steps:
       - run:
           name: "Upgrade codecov"
@@ -181,7 +184,7 @@ commands:
             sed -i 's|$(pwd)/||' ~/build/coverage.lcov
 
             counter=1
-            until codecov --required --file ~/build/coverage.lcov -X gcov || [ $counter = 5 ]; do
+            until codecov --flags <<parameters.flags>> --required --file ~/build/coverage.lcov -X gcov || [ $counter = 5 ]; do
               counter=$((counter+1))
               sleep 1
               echo "Try #$counter..."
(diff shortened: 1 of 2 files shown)


circle.yml
```diff
@@ -614,30 +632,30 @@ workflows:
...
-          requires:
-            - release-native-linux
-      - release-native-linux
-      - release-macos:
-          requires:
-            - release-native-macos
-      - release-native-macos:
-          requires:
-            - sanitizers-macos
+#      - lint
```
chfast, line 635: I will drop this commit before the merge. This PR only affects the "coverage-gcc" job.

codecov.yml
```diff
@@ -1,10 +1,14 @@
...
+    project: false
+    patch: false
 
 comment:
-  layout: "diff"
+  layout: "diff, flags, files"
+
+flags:
+  spectests:
+    joined: false
```
chfast, line 14: The "spectests" coverage is disjoined with the "master" coverage. We want to have the 100% with internal tests only.

circle.yml
```diff
@@ -138,6 +138,18 @@ commands:
           working_directory: ~/build
           command: ctest -E 'unittests|smoketests' -j4 --schedule-random --output-on-failure
 
+  collect_coverage_gcc:
```
chfast, line 141: This is refactoring. This step runs twice now. It additionally also rests the coverage counters and the HTML report so the next test suite executed is counted from 0.

circle.yml
```diff
@@ -138,6 +138,18 @@ commands:
...
+    description: "Collect coverage data (GCC)"
+    steps:
+      - run:
+          name: "Collect coverage data (GCC)"
+          working_directory: ~/build
+          command: |
+            lcov --capture --directory . --output-file coverage.lcov --exclude='/usr/include/*' --exclude="$HOME/.hunter/*" --exclude="$PWD/_deps/*"
```
chfast, line 148: Not sure what do you think about it, but I try to use full option names to make it more readable for people not familiar with the tools used.

axic (approved):

PR #651, opened 2020-11-18
test: Refactor utils for binop/unop tests

chfast:


axic:
I think some of these changes are in conflict with (or duplicated in) #655, lets wait for that.

test/unittests/execute_numeric_test.cpp
```diff
@@ -170,11 +170,19 @@ TEST(execute_numeric, i32_ge_u)
...
-    EXPECT_THAT(execute_unary_operation(Instr::i64_eqz, 1_u64), Result(0));
-    // 64-bit value on the stack
-    EXPECT_THAT(execute_unary_operation(fizzy::Instr::i64_eqz, 0xff00000000_u64), Result(0));
-    EXPECT_THAT(execute_unary_operation(fizzy::Instr::i64_eqz, 0xff00000001_u64), Result(0));
+    const auto i64_eqz = create_unary_operation_executor(Instr::i64_eqz);
+
+    EXPECT_THAT(i64_eqz(0_u64), Result(1_u32));
+    EXPECT_THAT(i64_eqz(1_u64), Result(0_u32));
+
+    // 64-bit value on the stack, type violation.
```
axic, line 178: Where is the type violation?

test/unittests/execute_numeric_test.cpp
```diff
@@ -49,7 +49,21 @@ TypedExecutionResult execute_binary_operation(Instr instr, TypedValue lhs, Typed
...
+    auto instance = instantiate(std::move(module));
+
+    return [instance = std::move(instance)](TypedValue lhs, TypedValue rhs) {
+        return execute(*instance, 0, {lhs, rhs});
+    };
+}
+
+TypedExecutionResult execute_unary_operation(Instr instr, TypedValue arg)
+{
+    return create_unary_operation_executor(instr)(arg);
```
axic, line 61: Will these create helpers used later in other PRs?

chfast: The `i64_eqz` is using it.

test/unittests/execute_numeric_test.cpp
chfast: Obsolete comment. Fixed.

test/unittests/execute_numeric_test.cpp
axic: The point is to avoid type check or why is it needed?

chfast: It is not needed, but it is lame to create module+instance for every test case. Here we create a lambda capturing the instance and we can execute multiple test cases.

axic: Well, the fresh instance gives the benefit, that nothing is dirty anywhere in the instance. However this way perhaps we could catch other bugs/problems caused by some uncleaned state.

axic (approved):

test/unittests/execute_numeric_test.cpp
gumb0: > It is not needed, but it is lame to create module+instance for every test case. Here we create a lambda capturing the instance and we can execute multiple test cases.

But there are many tests still using `execute_unary_operation` / `execute_binary_operation`, why leave it inconsistent?

test/unittests/execute_numeric_test.cpp
```diff
@@ -16,12 +16,11 @@ using namespace fizzy::test;
 
 namespace
 {
-TypedExecutionResult execute_unary_operation(Instr instr, TypedValue arg)
+auto create_unary_operation_executor(Instr instr)
 {
     const auto& instr_type = get_instruction_type_table()[static_cast<uint8_t>(instr)];
     EXPECT_EQ(instr_type.inputs.size(), 1);
     EXPECT_EQ(instr_type.outputs.size(), 1);
-    EXPECT_EQ(instr_type.inputs[0], arg.type);
```
gumb0, line 24: Should this check go inside lambda now?

test/unittests/execute_numeric_test.cpp
chfast: This PR mostly enables you to do the change. I needed it for more `eqz` test cases. For other instructions I prefer to go one-by-one later and add more test cases (something more than `42`). But I can syntactically replace the code right now if you prefer all at once (will not add more cases though).

test/unittests/execute_numeric_test.cpp
chfast: The `test::execute()` checks types inside.

test/unittests/execute_numeric_test.cpp
gumb0: I would prefer to replace all of them here, if it's easy.

gumb0 (approved):

chfast:
1. I moved `i64_eqz` enhancement to https://github.com/wasmx/fizzy/pull/679.
2. Here we have one old commit which reworks the helpers + 2 commits updating the syntax of test cases. There can be squashed in various combinations if any preferences.
3. The last WIP commit merges tests of the same instruction. I think it is fine doing it this way, but I don't have to proceed.

gumb0 (approved):

axic (approved):

PR #652, opened 2020-11-18
rust: implement safer execution wrapper

axic:


bindings/rust/src/lib.rs
```diff
@@ -193,6 +198,45 @@ impl Instance {
             0: sys::fizzy_execute(self.0.as_ptr(), func_idx, args.as_ptr(), depth),
         }
     }
+
+    /// Find function type for a given index. Must be a valid index otherwise behaviour is undefine.
```
chfast, line 202:
```suggestion
    /// Find function type for a given index. Must be a valid index otherwise behaviour is undefined.
```

bindings/rust/src/lib.rs
```diff
@@ -160,6 +160,11 @@ impl From<ExecutionResult> for sys::FizzyExecutionResult {
 }
 
 impl Instance {
+    /// Get a read-only pointer to the module.
+    unsafe fn get_instance_module(&self) -> *const sys::FizzyModule {
```
chfast, line 164:
```suggestion
    unsafe fn get_module(&self) -> *const sys::FizzyModule {
```

chfast (approved):

bindings/rust/src/lib.rs
```diff
@@ -193,6 +198,45 @@ impl Instance {
...
+        let func_idx = func_idx.unwrap();
+
+        let func_type = unsafe { self.get_function_type(func_idx) };
+        if func_type.inputs_size != args.len() {
+            return Err(());
+        }
+        // TODO: validate input types too
+
+        let ret = unsafe { self.unsafe_execute(func_idx, args, depth) };
+        // Fail if output is expected, but none was returned.
```
gumb0, line 233: It shouldn't happy unless there're bugs in execution, so maybe this should be an assertion?

gumb0: The opposite could also be asserted: function type is void, but something is returned.

bindings/rust/src/lib.rs
axic: Good point. I tried to mirror the underlying C API names, but probably it makes no sense.

bindings/rust/src/lib.rs
axic: We should use assertions when there is an API bug, and errors when it is a user validation question. I guess in this case assertion is fine as we get the `func_idx` from the name.

axic: I think this should do it?
```
        if !ret.trapped()
        {
           assert!((func_type.output == sys::FizzyValueTypeVoid) == !ret.value().is_some());
        }
```

bindings/rust/src/lib.rs

bindings/rust/src/lib.rs
```diff
@@ -193,6 +198,44 @@ impl Instance {
...
+    ///
+    /// The `depth` argument sets the initial call depth. Should be set to `0`.
+    pub fn execute(
+        &mut self,
+        name: &str,
+        args: &[Value],
+        depth: i32,
+    ) -> Result<ExecutionResult, ()> {
+        let func_idx = self.find_exported_function_index(&name);
+        if func_idx.is_none() {
```
axic, line 221: If get #677 merged, these could be replaced with `let func_idx = self.find_exported_function_index(&name)?;` where the question mark operator means errors are bubbled up.

I would however go with this PR as is first, and then merge the error one showing the improvement.

bindings/rust/build.rs
```diff
@@ -29,6 +29,8 @@ fn main() {
         // https://github.com/rust-lang-nursery/rust-bindgen/issues/947#issuecomment-327100002
         .layout_tests(false)
         .whitelist_function("fizzy_.*")
+        .whitelist_var("Fizzy.*")
```
axic, line 32: Needed for `FizzyValueTypeVoid`.

bindings/rust/src/lib.rs
```diff
@@ -193,6 +198,44 @@ impl Instance {
...
+            return Err(());
+        }
+        let func_idx = func_idx.unwrap();
+
+        let func_type = unsafe { self.get_function_type(func_idx) };
+        if func_type.inputs_size != args.len() {
+            return Err(());
+        }
+        // TODO: validate input types too
```
axic, line 230: For type validation need to revamp the type of `Value` here to include type information. Maybe better separately as it is quite large.

gumb0 (approved):
Looks good, for completeness you could add a test with a function having some params and calling it with 0 or incorrect number of arguments.

axic:
> Looks good, for completeness you could add a test with a function having some params and calling it with 0 or incorrect number of arguments.

The last line with `foo` does the "more args than required" case. Will add a "less args than required" case too.

bindings/rust/src/lib.rs
```diff
@@ -303,4 +346,49 @@ mod tests {
...
+        (module
+          (func $f (export "foo") (result i32) (i32.const 42))
+          (func $f (export "bar") (param i32) (result i32) (i32.const 42))
+          (global (export "g1") i32 (i32.const 0))
+          (table (export "tab") 0 anyfunc)
+          (memory (export "mem") 1 2)
+        )
+        */
+        let input = hex::decode(
+        "0061736d010000000105016000017f030201000404017000000504010101020606017f0041000b07180403666f6f00000267310300037461620100036d656d02000a06010400412a0b").unwrap();
```
axic, line 362: Need to update this hex too, but was lazy.

bindings/rust/src/lib.rs
axic: See #705.

gumb0 (approved):

PR #655, opened 2020-11-20
test: Check the types of execution arguments

chfast:
After previously introducing `TypedValue` in #658, this PR adds type checking for execution arguments in unit tests. The type checking is done in `fizzy::test::execute()` so can be bypassed by using `fizzy::execute()` directly. Still most of the tests uses the former.

Later we will add type checking to the execution result and the `Result()` matcher. We will be also able to see the type-checking bypassing cases and correct these if needed.

### DIFF
--- a/test/unittests/api_test.cpp
+++ b/test/unittests/api_test.cpp
@@ -102,7 +102,7 @@ TEST(api, resolve_imported_functions)
     EXPECT_THAT(execute(*instance, 0, {}), Result(0));
     EXPECT_THAT(execute(*instance, 1, {0_u32}), Result(1));
     EXPECT_THAT(execute(*instance, 2, {0_u32}), Result(2));
-    EXPECT_THAT(execute(*instance, 3, {0, 0}), Result());
+    EXPECT_THAT(execute(*instance, 3, {0_u64, 0_u32}), Result());
 
 
     std::vector<ImportedFunction> imported_functions_reordered = {
@@ -122,7 +122,7 @@ TEST(api, resolve_imported_functions)
     EXPECT_THAT(execute(*instance_reordered, 0, {}), Result(0));
     EXPECT_THAT(execute(*instance_reordered, 1, {0_u32}), Result(1));
     EXPECT_THAT(execute(*instance_reordered, 2, {0_u32}), Result(2));
-    EXPECT_THAT(execute(*instance_reordered, 3, {0, 0}), Result());
+    EXPECT_THAT(execute(*instance_reordered, 3, {0_u64, 0_u32}), Result());
 
 
     std::vector<ImportedFunction> imported_functions_extra = {
@@ -144,7 +144,7 @@ TEST(api, resolve_imported_functions)
     EXPECT_THAT(execute(*instance_extra, 0, {}), Result(0));
     EXPECT_THAT(execute(*instance_extra, 1, {0_u32}), Result(1));
     EXPECT_THAT(execute(*instance_extra, 2, {0_u32}), Result(2));
-    EXPECT_THAT(execute(*instance_extra, 3, {0, 0}), Result());
+    EXPECT_THAT(execute(*instance_extra, 3, {0_u64, 0_u32}), Result());
 
 
     std::vector<ImportedFunction> imported_functions_missing = {
--- a/test/unittests/api_test.cpp
+++ b/test/unittests/api_test.cpp
@@ -100,8 +100,8 @@ TEST(api, resolve_imported_functions)
         *module, external_functions, {}, {}, std::vector<ExternalGlobal>(external_globals));
 
     EXPECT_THAT(execute(*instance, 0, {}), Result(0));
-    EXPECT_THAT(execute(*instance, 1, {Value{0}}), Result(1));
-    EXPECT_THAT(execute(*instance, 2, {Value{0}}), Result(2));
+    EXPECT_THAT(execute(*instance, 1, {0_u32}), Result(1));
+    EXPECT_THAT(execute(*instance, 2, {0_u32}), Result(2));
     EXPECT_THAT(execute(*instance, 3, {0, 0}), Result());
 
 
@@ -120,8 +120,8 @@ TEST(api, resolve_imported_functions)
         std::vector<ExternalGlobal>(external_globals));
 
     EXPECT_THAT(execute(*instance_reordered, 0, {}), Result(0));
-    EXPECT_THAT(execute(*instance_reordered, 1, {Value{0}}), Result(1));
-    EXPECT_THAT(execute(*instance_reordered, 2, {Value{0}}), Result(2));
+    EXPECT_THAT(execute(*instance_reordered, 1, {0_u32}), Result(1));
+    EXPECT_THAT(execute(*instance_reordered, 2, {0_u32}), Result(2));
     EXPECT_THAT(execute(*instance_reordered, 3, {0, 0}), Result());
 
 
@@ -142,8 +142,8 @@ TEST(api, resolve_imported_functions)
         *module, external_functions_extra, {}, {}, std::vector<ExternalGlobal>(external_globals));
 
     EXPECT_THAT(execute(*instance_extra, 0, {}), Result(0));
-    EXPECT_THAT(execute(*instance_extra, 1, {Value{0}}), Result(1));
-    EXPECT_THAT(execute(*instance_extra, 2, {Value{0}}), Result(2));
+    EXPECT_THAT(execute(*instance_extra, 1, {0_u32}), Result(1));
+    EXPECT_THAT(execute(*instance_extra, 2, {0_u32}), Result(2));
     EXPECT_THAT(execute(*instance_extra, 3, {0, 0}), Result());
 
 
@@ -214,8 +214,8 @@ TEST(api, resolve_imported_function_duplicate)
 
     auto instance = instantiate(*module, external_functions, {}, {}, {});
 
-    EXPECT_THAT(execute(*instance, 0, {Value{0}}), Result(42));
-    EXPECT_THAT(execute(*instance, 1, {Value{0}}), Result(42));
+    EXPECT_THAT(execute(*instance, 0, {0_u32}), Result(42));
+    EXPECT_THAT(execute(*instance, 1, {0_u32}), Result(42));
 }
 
 TEST(api, find_exported_function_index)
--- a/test/unittests/execute_numeric_test.cpp
+++ b/test/unittests/execute_numeric_test.cpp
@@ -33,11 +33,13 @@ ExecutionResult execute_unary_operation(Instr instr, TypedValue arg)
     return execute(*instantiate(std::move(module)), 0, {arg});
 }
 
-ExecutionResult execute_binary_operation(Instr instr, uint64_t lhs, uint64_t rhs)
+ExecutionResult execute_binary_operation(Instr instr, TypedValue lhs, TypedValue rhs)
 {
     const auto& instr_type = get_instruction_type_table()[static_cast<uint8_t>(instr)];
     EXPECT_EQ(instr_type.inputs.size(), 2);
     EXPECT_EQ(instr_type.outputs.size(), 1);
+    EXPECT_EQ(instr_type.inputs[0], lhs.type);
+    EXPECT_EQ(instr_type.inputs[1], rhs.type);
 
     auto module{std::make_unique<Module>()};
     module->typesec.emplace_back(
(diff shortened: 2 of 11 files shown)


test/unittests/execute_numeric_test.cpp
```diff
@@ -3,22 +3,28 @@
...
 using namespace fizzy;
 using namespace fizzy::test;
 
 namespace
 {
-ExecutionResult execute_unary_operation(Instr instr, uint64_t arg)
+ExecutionResult execute_unary_operation(Instr instr, TypedValue arg)
 {
+    const auto& instr_type = get_instruction_type_table()[static_cast<size_t>(instr)];
+    EXPECT_EQ(instr_type.inputs.size(), 1);
```
gumb0, line 22: size checks should be better `ASSERT`s, I think

chfast: `ASSERT`s does not work here.

test/unittests/typed_value_test.cpp
```diff
@@ -0,0 +1,29 @@
...
+using namespace fizzy::test;
+
+TEST(typed_value, u32_literal)
+{
+    static_assert(std::is_same_v<decltype(0_u32), uint32_t>);
+    EXPECT_EQ(0_u32, uint32_t{0});
+    EXPECT_EQ(1_u32, uint32_t{1});
+    EXPECT_EQ(0xffffffff_u32, uint32_t{0xffffffff});
+
+    unsigned long long input = 0x100000000;  // NOLINT
```
gumb0, line 17: Why NOLINT is needed?

gumb0 (approved):

test/unittests/typed_value_test.cpp
chfast: CLion says I should be using `uint64_t`.

test/unittests/execute_numeric_test.cpp
chfast: `ASSERT` returns void. It should be possible to use an `ASSERT` if converted to a proper test fixture. But I don't want to mix such changes. Refactoring is in #651.

test/unittests/constexpr_vector_test.cpp
```diff
@@ -27,6 +27,14 @@ TEST(constexpr_vector, size)
     EXPECT_EQ(*v3.data(), 1);
 }
 
+TEST(constexpr_vector, subscript)
+{
+    constexpr constexpr_vector<int, 4> v = {1, 2, 3};
```
axic, line 32: Does `<int, 4>` mean there are 4 members, the last being implicitly 0?

chfast: No. The `4` is max capacity. Size is still 3 here. See https://github.com/wasmx/fizzy/pull/656.

test/unittests/api_test.cpp
```diff
@@ -100,9 +100,9 @@ TEST(api, resolve_imported_functions)
         *module, external_functions, {}, {}, std::vector<ExternalGlobal>(external_globals));
 
     EXPECT_THAT(execute(*instance, 0, {}), Result(0));
-    EXPECT_THAT(execute(*instance, 1, {Value{0}}), Result(1));
-    EXPECT_THAT(execute(*instance, 2, {Value{0}}), Result(2));
-    EXPECT_THAT(execute(*instance, 3, {0, 0}), Result());
+    EXPECT_THAT(execute(*instance, 1, {0_u32}), Result(1));
+    EXPECT_THAT(execute(*instance, 2, {0_u32}), Result(2));
+    EXPECT_THAT(execute(*instance, 3, {0_u64, 0}), Result());
```
gumb0, line 105:
```suggestion
    EXPECT_THAT(execute(*instance, 3, {0_u64, 0_u32}), Result());
```

test/unittests/api_test.cpp
```diff
@@ -120,9 +120,9 @@ TEST(api, resolve_imported_functions)
         std::vector<ExternalGlobal>(external_globals));
 
     EXPECT_THAT(execute(*instance_reordered, 0, {}), Result(0));
-    EXPECT_THAT(execute(*instance_reordered, 1, {Value{0}}), Result(1));
-    EXPECT_THAT(execute(*instance_reordered, 2, {Value{0}}), Result(2));
-    EXPECT_THAT(execute(*instance_reordered, 3, {0, 0}), Result());
+    EXPECT_THAT(execute(*instance_reordered, 1, {0_u32}), Result(1));
+    EXPECT_THAT(execute(*instance_reordered, 2, {0_u32}), Result(2));
+    EXPECT_THAT(execute(*instance_reordered, 3, {0_u64, 0}), Result());
```
gumb0, line 125:
```suggestion
    EXPECT_THAT(execute(*instance_reordered, 3, {0_u64, 0_u32}), Result());
```

test/unittests/api_test.cpp
```diff
@@ -142,9 +142,9 @@ TEST(api, resolve_imported_functions)
         *module, external_functions_extra, {}, {}, std::vector<ExternalGlobal>(external_globals));
 
     EXPECT_THAT(execute(*instance_extra, 0, {}), Result(0));
-    EXPECT_THAT(execute(*instance_extra, 1, {Value{0}}), Result(1));
-    EXPECT_THAT(execute(*instance_extra, 2, {Value{0}}), Result(2));
-    EXPECT_THAT(execute(*instance_extra, 3, {0, 0}), Result());
+    EXPECT_THAT(execute(*instance_extra, 1, {0_u32}), Result(1));
+    EXPECT_THAT(execute(*instance_extra, 2, {0_u32}), Result(2));
+    EXPECT_THAT(execute(*instance_extra, 3, {0_u64, 0}), Result());
```
gumb0, line 147:
```suggestion
    EXPECT_THAT(execute(*instance_extra, 3, {0_u64, 0_u32}), Result());
```

test/utils/execute_helpers.hpp
```diff
@@ -6,21 +6,34 @@
...
 {
-    assert(args.size() == instance.module->get_function_type(func_idx).inputs.size());
-    return fizzy::execute(instance, func_idx, args.begin());
+    const auto& func_type = instance.module->get_function_type(func_idx);
+    if (typed_args.size() != func_type.inputs.size())
+        throw std::invalid_argument{"invalid number of arguments"};
+
+    std::vector<fizzy::Value> args(typed_args.size());
+    size_t i = 0;
+    for (const auto& typed_arg : typed_args)
```
gumb0, line 23: this could be `std::transform`

test/unittests/api_test.cpp
chfast: Can be, but I wanted to stick with minimal set of changes here.

gumb0: Not sure I understand this motivation.

What is the type of the second argument passed? Presumably it's i32, because test passed. But it's some implicit conversion involved, it might be platform-dependent in general case and, it just looks weird.

chfast:
I only changed the values required to compile and pass the test.

The `0` work when `decltype(0) === int32_t`. It will also compile (and work incorrectly) if `decltype(0) === int64_t`. Otherwise this will fail to compile.

This variant is "middle-ground" for convenience and type safety - it implements constructors for `int32_t, uint32_t, int64_t, uint64_t`. This requires to bump the value to `u64` when `i64` type is required.

The less strict variant is to conditionally at runtime zero extend the `uint32_t` and `int32_t` values to `i64` when they are not negative (e.g. `1` and `-1u` are fine, but `-1` is not). I'm not sure this is easy to implement here, but I did it for `Result()` in #659.

The more strict variant is to only implement constructors for `uint32_t` and `uint64_t`. This is what I tried first. It is more annoying. It also encourages things like `-1_u32`.

That's why I stayed with minimal changes to focus on design discussion rather than esthetics.

My personal opinions is that this variant is good enough and definitely good for stating point to experiment with other variants on top of it. I also want to consider the option to promote this API to public type safe API. I prefer having one common type-safe API.

gumb0:
The set of constructors looks good and logical. It shouldn't be less strict, because then tests wouldn't show what types are really expected. It shouldn't be more  more strict, because signed->unsigned conversion is already handled by `Value`.

But I still don't get why it's better to rely on implicit conversion (which might theoretically be different for different platform).
And why is it not everywhere but only for second argument?
(I mean you do use `_u32` when it's the single argument
```
EXPECT_THAT(execute(*instance_reordered, 1, {0_u32}), Result(1));
EXPECT_THAT(execute(*instance_reordered, 3, {0_u64, 0}), Result());
```
)

Is it just to minimize changes in this PR? What is you preference for the new tests that will execute functions with two parameters?

chfast:
The `{0_u32}` (previously `{Value{0}}`) handles other famous C++ issue where `0` binds to a pointer type instead of `Value` because the former is built-in conversion and later is user-provided conversion. Somehow the `uint32_t` does not. Except for this case, I have not introduced `0_u32` otherwise (at least not intentionally).

Keep in mind that `_u32` only creates the `uint32_t` objects, not `Value`s.

test/utils/execute_helpers.hpp
chfast: See the new version.

test/utils/execute_helpers.hpp
```diff
@@ -6,21 +6,38 @@
...
+        throw std::invalid_argument{"invalid type of the argument " +
+                                    std::to_string(typed_arg_it - std::cbegin(typed_args))};
+    else if (typed_arg_it != std::cend(typed_args))
+        throw std::invalid_argument{"too many arguments"};
+    else if (type_it != std::cend(func_type.inputs))
+        throw std::invalid_argument{"too few arguments"};
+
+    std::vector<fizzy::Value> args;
+    std::transform(std::cbegin(typed_args), std::cend(typed_args), std::back_inserter(args),
+        [](const auto& typed_arg) { return typed_arg.value; });
```
gumb0, line 32:
```suggestion
    std::vector<fizzy::Value> args(typed_args.size());
    std::transform(std::cbegin(typed_args), std::cend(typed_args), std::begin(args),
        [](const auto& typed_arg) { return typed_arg.value; });
```

chfast: I don't think efficiency like this matters here.

gumb0: To me it's a simplification.

gumb0 (approved):

test/unittests/typed_value_test.cpp
```diff
@@ -73,3 +73,23 @@ TEST(typed_value, construct)
...
+}
+
+TEST(typed_value, u64_literal)
+{
+    static_assert(std::is_same_v<decltype(0_u64), uint64_t>);
+    EXPECT_EQ(0_u64, uint64_t{0});
+    EXPECT_EQ(1_u64, uint64_t{1});
+    EXPECT_EQ(0xffffffff_u64, uint64_t{0xffffffff});
+    EXPECT_EQ(0x100000000_u64, uint64_t{0x100000000});
+    EXPECT_EQ(0xffffffffffffffff_u64, uint64_t{0xffffffffffffffff});
```
axic, line 94: What happens to literals larger than 64-bit? Are 128-bit literals supported where `__uint128_t` is supported?

test/unittests/typed_value_test.cpp
```diff
@@ -73,3 +73,23 @@ TEST(typed_value, construct)
     EXPECT_EQ(f64.type, ValType::f64);
     EXPECT_EQ(f64.value.f64, -2.2250738585072014e-308);
 }
+
+TEST(typed_value, u32_literal)
+{
+    static_assert(std::is_same_v<decltype(0_u32), uint32_t>);
+    EXPECT_EQ(0_u32, uint32_t{0});
+    EXPECT_EQ(1_u32, uint32_t{1});
```
axic, line 81: What happens with negative literals?

test/unittests/execute_numeric_test.cpp
```diff
@@ -16,9 +17,12 @@ namespace
 {
 ExecutionResult execute_unary_operation(Instr instr, uint64_t arg)
 {
+    const auto& instr_type = get_instruction_type_table()[static_cast<size_t>(instr)];
```
axic, line 21: Why `size_t` and not `uint8_t`, given `Instr` is `uint8_t`?

test/unittests/execute_numeric_test.cpp
```diff
@@ -444,10 +446,10 @@ TEST(execute_numeric, i32_rotr)
 TEST(execute_numeric, i32_wrap_i64)
 {
     // <=32-bits set
-    EXPECT_THAT(execute_unary_operation(Instr::i32_wrap_i64, 0xffffffff), Result(0xffffffff));
+    EXPECT_THAT(execute_unary_operation(Instr::i32_wrap_i64, 0xffffffff_u64), Result(0xffffffff));
     // >32-bits set
     EXPECT_THAT(
-        execute_unary_operation(Instr::i32_wrap_i64, 0xffffffffffffffff), Result(0xffffffff));
+        execute_unary_operation(Instr::i32_wrap_i64, 0xffffffffffffffff_u64), Result(0xffffffff));
```
axic, line 454: Hmm, perhaps result could be also using the `_u32`/`_u64` literals.

test/unittests/execute_numeric_test.cpp
```diff
@@ -163,76 +165,76 @@ TEST(execute_numeric, i64_eqz)
...
+    EXPECT_THAT(execute_binary_operation(Instr::i64_ne, 22_u64, 22_u64), Result(0));
 }
 
 TEST(execute_numeric, i64_lt_s)
 {
-    EXPECT_THAT(execute_binary_operation(Instr::i64_lt_s, 22, 20), Result(0));
-    EXPECT_THAT(execute_binary_operation(Instr::i64_lt_s, 20, 22), Result(1));
+    EXPECT_THAT(execute_binary_operation(Instr::i64_lt_s, 22_u64, 20_u64), Result(0));
+    EXPECT_THAT(execute_binary_operation(Instr::i64_lt_s, 20_u64, 22_u64), Result(1));
     EXPECT_THAT(execute_binary_operation(Instr::i64_lt_s, uint64_t(-41), uint64_t(-42)), Result(0));
```
axic, line 182: If negative literals are supported, these should be something like `-41_i64`? I agree keeping the explicit `unit64_t()` may be nicer, however confusing it looks to have both ways.

test/unittests/api_test.cpp
axic: Agree with @gumb0 here. Would be more consistent to use `0_u64, 0_u32`.

test/unittests/api_test.cpp
axic: Same here.

test/unittests/api_test.cpp
axic: And here.

test/unittests/execute_test.cpp
```diff
@@ -617,10 +617,10 @@ TEST(execute, i64_store_all_variants)
         *store_instr = static_cast<uint8_t>(instr);
         auto instance = instantiate(*module);
         std::fill_n(instance->memory->begin(), 10, uint8_t{0xcc});
-        EXPECT_THAT(execute(*instance, 0, {0xb7b6b5b4b3b2b1b0, 1}), Result());
+        EXPECT_THAT(execute(*instance, 0, {0xb7b6b5b4b3b2b1b0_u64, 1}), Result());
         EXPECT_EQ(instance->memory->substr(0, 10), expected);
 
-        EXPECT_THAT(execute(*instance, 0, {0xb7b6b5b4b3b2b1b0, 65537}), Traps());
+        EXPECT_THAT(execute(*instance, 0, {0xb7b6b5b4b3b2b1b0_u64, 65537}), Traps());
```
axic, line 623: Perhaps use `_u32` here and above too, since these lines are being changed.

test/unittests/execute_numeric_test.cpp
chfast: This is separate issue being handled in https://github.com/wasmx/fizzy/pull/659.

test/unittests/typed_value_test.cpp
chfast: There are no negative literals. The `-1` is `"-" + literal "1"`.

test/unittests/typed_value_test.cpp
chfast: There are no builtin integer literals bigger than `unsigned long long`. The `__int128` does not have its own literal.

test/unittests/execute_numeric_test.cpp
chfast: The `-41_u64` is correct and can be used already. This is `-uint64_t{41}`. Although this usually case MSVC complain about applying `-` operator to an unsigned type.

Using signed types `int32_t` and `int64_t` is also an option, but I don't recall that is the current status of this support. Definitely does not look like signed types were supported when these tests were created.

There is no change here because the provided values have the correct types. And making tests look beautiful and consistent is not the goal of this PR.

test/unittests/execute_numeric_test.cpp
axic: I guessed it is for the indexing type, but `uint8_t` would be more clear I guess. It fits `size_t` anyhow.

test/unittests/execute_numeric_test.cpp
axic: It is not about beauty, just removing confusion.

test/unittests/execute_test.cpp
```diff
@@ -62,7 +62,7 @@ TEST(execute, local_get)
     )
     */
     const auto wasm = from_hex("0061736d0100000001060160017e017e030201000a0601040020000b");
-    EXPECT_THAT(execute(parse(wasm), 0, {42}), Result(42));
+    EXPECT_THAT(execute(parse(wasm), 0_u64, {42_u64}), Result(42));
```
axic, line 65:
```suggestion
    EXPECT_THAT(execute(parse(wasm), 0, {42_u64}), Result(42));
```

test/unittests/execute_test.cpp
```diff
@@ -900,8 +900,8 @@ TEST(execute, imported_two_functions_different_type)
         instantiate(*module, {{host_foo1, module->typesec[0]}, {host_foo2, module->typesec[1]}});
 
     EXPECT_THAT(execute(*instance, 0, {20, 22}), Result(42));
```
axic, line 902:
```suggestion
    EXPECT_THAT(execute(*instance, 0, {20_u32, 22_u32}), Result(42));
```

axic (approved):

PR #657, opened 2020-11-20
Move execute(initializer_list) to fizzy::test

chfast:


lib/fizzy/execute.hpp
```diff
@@ -38,11 +38,4 @@ constexpr ExecutionResult Trap{false};
 
 // Execute a function on an instance.
 ExecutionResult execute(Instance& instance, FuncIdx func_idx, const Value* args, int depth = 0);
-
-inline ExecutionResult execute(
-    Instance& instance, FuncIdx func_idx, std::initializer_list<Value> args)
```
axic, line 43: Why move this though?

chfast: 1. The `std::initializer_list<Value>` cannot be used in public API like that. This is not what it is designed for. We can use `std::span` instead.
2. I prefer to change it to `std::initializer_list<TypedValue>` later instead of having a ton of overloads spread accross many places.

axic: > The std::initializer_list<Value> cannot be used in public API like that. This is not what it is designed for. We can use std::span instead.

Not arguing against, but can you explain the reason for this?

chfast: This usage is probably fine, but there are some doubts about the lifetime of the `{}` object. The `std::initializer_list` is to provide fancy syntax to containers. Somehow in C++20 had been decided that `std::span` cannot reference `std::initializer_list`.

In practice this is mostly useful in cases when you want to provide some fixed number of arguments, i.e. mostly tests and examples. None of the other tools need it.

axic (approved):

PR #658, opened 2020-11-20
Introduce TypedValue to fizzy::test

chfast:


test/utils/typed_value.hpp
```diff
@@ -0,0 +1,27 @@
...
+// Copyright 2020 The Fizzy Authors.
+// SPDX-License-Identifier: Apache-2.0
+#pragma once
+
+#include "types.hpp"
+#include "value.hpp"
+
+namespace fizzy::test
+{
+struct TypedValue
```
axic, line 11: Can you add some basic tests for this? I know... testing testing tools.

chfast: Added.

test/utils/typed_value.hpp
```diff
@@ -0,0 +1,29 @@
...
+{
+    Value value;
+    ValType type;
+
+    /// Assign type to the typeless Value.
+    /// No validation is possible, so the correctness is on the user only.
+    constexpr TypedValue(ValType ty, Value v) noexcept : value{v}, type{ty} {}
+
+    constexpr TypedValue(int32_t v) noexcept : TypedValue{ValType::i32, v} {}
```
chfast, line 20: This is in my options minimal set of constructors. Problem here is that on macOS `uint64_t` is defined differently than on Linux. On macOS the literals of `long` type are ambiguous, e.g. `TypedValue(0x100000000)` and it must be explicitly converted to `uint64_t` as `TypedValue(0x100000000_u64)`.

test/unittests/typed_value_test.cpp
```diff
@@ -0,0 +1,75 @@
...
+using namespace fizzy;
+using namespace fizzy::test;
+
+TEST(typed_value, construct_contexpr)
+{
+    constexpr TypedValue i32{int32_t{-1}};
+    static_assert(i32.type == ValType::i32);
+    static_assert(i32.value.i64 == uint32_t(-1));
+    EXPECT_EQ(i32.type, ValType::i32);
+    EXPECT_EQ(i32.value.i64, uint32_t(-1));
```
axic, line 17: Why have both `sattic_assert` and `EXPECT_EQ`?

axic (approved):

test/unittests/typed_value_test.cpp
chfast: Performing check in constexpr and non-contexpr context. This first one does not give code coverage.

test/utils/typed_value.hpp
```diff
@@ -0,0 +1,29 @@
...
+    constexpr TypedValue(uint32_t v) noexcept : TypedValue{ValType::i32, v} {}
+    constexpr TypedValue(int64_t v) noexcept : TypedValue{ValType::i64, v} {}
+    constexpr TypedValue(uint64_t v) noexcept : TypedValue{ValType::i64, v} {}
+    constexpr TypedValue(float v) noexcept : TypedValue{ValType::f32, v} {}
+    constexpr TypedValue(double v) noexcept : TypedValue{ValType::f64, v} {}
+
+    constexpr operator Value() const noexcept { return value; }
```
gumb0, line 27: Where is this used?

chfast: I don't know any more. Some future PRs propose to drop it.

chfast: It was used ones, I dropped it. It is always easy to drop the type information by `.value`.

gumb0: Yeah, I was concerned that with it can be misleading e.g.
```
TypedValue v{1_u64};
if (v == Value{1_i32}) // true
```

test/unittests/typed_value_test.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+#include "test/utils/typed_value.hpp"
+#include <gtest/gtest.h>
+
+using namespace fizzy;
+using namespace fizzy::test;
+
+TEST(typed_value, construct_contexpr)
+{
+    constexpr TypedValue i32{int32_t{-1}};
```
gumb0, line 13: Nit: some value literals are defined with `{}` others with `()`.

Are `_i32` literals merged yet? Can't we use them here?

gumb0: Ah I got it, it was intentional, because somewhere conversion is happening.

gumb0 (approved):

PR #659, opened 2020-11-23
test: Add TypedExecutionResult, check result type

chfast:
- [x] Introduce `TypedExecutionResult` which takes the result type from the function type.
- [x] Adjust `Result()` matcher to expect `TypedExecutionResult` and perform type-safe value checks.
- [x] The `Result()` is more liberal comparing with `TypedValue`: it allows `u64` - `u32` and `u64` - `i32` if not negative compares.
- [x] Add `Result()` unit tests.

### DIFF
--- a/test/unittests/test_utils_test.cpp
+++ b/test/unittests/test_utils_test.cpp
@@ -34,11 +34,11 @@ TEST(test_utils, from_hex)
     EXPECT_THROW_MESSAGE(from_hex("FG"), std::out_of_range, "not a hex digit");
 }
 
-TEST(test_utils, result_signed_int)
+TEST(test_utils, result_signed_int_typed)
 {
-    EXPECT_THAT(ExecutionResult{Value{-1}}, Result(-1));
+    EXPECT_THAT(TypedExecutionResult(Value{-1}, ValType::i32), Result(-1));
     constexpr auto v = std::numeric_limits<int32_t>::min();
-    EXPECT_THAT(ExecutionResult{Value{v}}, Result(v));
+    EXPECT_THAT(TypedExecutionResult(Value{v}, ValType::i32), Result(v));
 }
 
 TEST(test_utils, print_execution_result)
@@ -69,4 +69,50 @@ TEST(test_utils, print_c_execution_result)
     std::stringstream str_value;
     str_value << FizzyExecutionResult{false, true, {42}};
     EXPECT_EQ(str_value.str(), "result(42 [0x2a])");
+}
+
+TEST(test_utils, print_typed_execution_result)
+{
+    std::stringstream str_trap;
+    str_trap << TypedExecutionResult{Trap, {}};
+    EXPECT_EQ(str_trap.str(), "trapped");
+
+    std::stringstream str_void;
+    str_void << TypedExecutionResult{Void, {}};
+    EXPECT_EQ(str_void.str(), "result()");
+
+    std::stringstream str_value_i32;
+    str_value_i32 << TypedExecutionResult{ExecutionResult{Value{42_u32}}, ValType::i32};
+    EXPECT_EQ(str_value_i32.str(), "result(42 [0x2a] (i32))");
+    str_value_i32.str({});
+    str_value_i32 << TypedExecutionResult{ExecutionResult{Value{0x80000000_u32}}, ValType::i32};
+    EXPECT_EQ(str_value_i32.str(), "result(2147483648 [0x80000000] (i32))");
+    str_value_i32.str({});
+    str_value_i32 << TypedExecutionResult{ExecutionResult{Value{-2_u32}}, ValType::i32};
+    EXPECT_EQ(str_value_i32.str(), "result(4294967294 [0xfffffffe] (i32))");
+
+
+    std::stringstream str_value_i64;
+    str_value_i64 << TypedExecutionResult{ExecutionResult{Value{42_u64}}, ValType::i64};
+    EXPECT_EQ(str_value_i64.str(), "result(42 [0x2a] (i64))");
+    str_value_i64.str({});
+    str_value_i64 << TypedExecutionResult{ExecutionResult{Value{0x100000000_u64}}, ValType::i64};
+    EXPECT_EQ(str_value_i64.str(), "result(4294967296 [0x100000000] (i64))");
+    str_value_i64.str({});
+    str_value_i64 << TypedExecutionResult{ExecutionResult{Value{-3_u64}}, ValType::i64};
+    EXPECT_EQ(str_value_i64.str(), "result(18446744073709551613 [0xfffffffffffffffd] (i64))");
+
+    std::stringstream str_value_f32;
+    str_value_f32 << TypedExecutionResult{ExecutionResult{Value{1.125f}}, ValType::f32};
+    EXPECT_EQ(str_value_f32.str(), "result(1.125 (f32))");
+    str_value_f32.str({});
+    str_value_f32 << TypedExecutionResult{ExecutionResult{Value{-1.125f}}, ValType::f32};
+    EXPECT_EQ(str_value_f32.str(), "result(-1.125 (f32))");
+
+    std::stringstream str_value_f64;
+    str_value_f64 << TypedExecutionResult{ExecutionResult{Value{1.125}}, ValType::f64};
+    EXPECT_EQ(str_value_f64.str(), "result(1.125 (f64))");
+    str_value_f64.str({});
+    str_value_f64 << TypedExecutionResult{ExecutionResult{Value{-1.125}}, ValType::f64};
+    EXPECT_EQ(str_value_f64.str(), "result(-1.125 (f64))");
 }
--- a/test/unittests/test_utils_test.cpp
+++ b/test/unittests/test_utils_test.cpp
@@ -116,3 +116,67 @@ TEST(test_utils, print_typed_execution_result)
     str_value_f64 << TypedExecutionResult{ExecutionResult{Value{-1.125}}, ValType::f64};
     EXPECT_EQ(str_value_f64.str(), "result(-1.125 (f64))");
 }
+
+TEST(test_utils, result_value_matcher)
+{
+    // Exercise every check in Result(value) implementation.
+    // The implementation and checks below are organized by the value's type in Result(value).
+    using testing::Not;
+
+    // TypedExecutionResult is required to be matched against Result(value).
+    EXPECT_THAT(ExecutionResult(Value{1_u64}), Not(Result(1_u64)));
+
+    EXPECT_THAT(TypedExecutionResult(fizzy::Void, {}), Not(Result(0)));
+    EXPECT_THAT(TypedExecutionResult(fizzy::Trap, {}), Not(Result(0)));
+
+    EXPECT_THAT(TypedExecutionResult(Value{0.0f}, ValType::f32), Result(0.0f));
+    EXPECT_THAT(TypedExecutionResult(Value{0.0}, ValType::f64), Not(Result(0.0f)));
+
+    EXPECT_THAT(TypedExecutionResult(Value{0.0}, ValType::f64), Result(0.0));
+    EXPECT_THAT(TypedExecutionResult(Value{0.0f}, ValType::f32), Not(Result(0.0)));
+
+    EXPECT_THAT(TypedExecutionResult(Value{0_u64}, ValType::i64), Result(0_u64));
+    EXPECT_THAT(TypedExecutionResult(Value{0_u32}, ValType::i32), Not(Result(0_u64)));
+
+    EXPECT_THAT(TypedExecutionResult(Value{0_u32}, ValType::i32), Result(0_u32));
+
+    // For non-negative values zero-extension is conveniently allowed.
+    EXPECT_THAT(TypedExecutionResult(Value{0_u64}, ValType::i64), Result(0));
+    EXPECT_THAT(TypedExecutionResult(Value{0_u64}, ValType::i64), Result(0_u32));
+
+    EXPECT_THAT(TypedExecutionResult(Value{-1_u32}, ValType::i32), Result(-1));
+    EXPECT_THAT(TypedExecutionResult(Value{-1_u32}, ValType::i32), Result(-1_u32));
+    EXPECT_THAT(TypedExecutionResult(Value{-1_u32}, ValType::i32), Not(Result(-1_u64)));
+
+    EXPECT_THAT(TypedExecutionResult(Value{-1_u64}, ValType::i64), Result(-1_u64));
+    EXPECT_THAT(TypedExecutionResult(Value{-1_u64}, ValType::i64), Not(Result(-1)));
+    EXPECT_THAT(TypedExecutionResult(Value{-1_u64}, ValType::i64), Not(Result(-1_u32)));
+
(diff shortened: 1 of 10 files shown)


test/utils/asserts.cpp
```diff
@@ -20,4 +20,19 @@ std::ostream& operator<<(std::ostream& os, ExecutionResult result)
...
+
+namespace test
+{
+std::ostream& operator<<(std::ostream& os, const TypedExecutionResult& result)
+{
+    const auto format_flags = os.flags();
+
+    os << ExecutionResult{result};
+
+    os << " (type " << std::hex << static_cast<int>(result.type) << ")";
```
gumb0, line 32: `operator<<` for `ValType` would be nice

gumb0:
Please rebase.

chfast:
> Please rebase.

I'm not touching this PR until #655 is merged.

gumb0:
> > Please rebase.
> 
> I'm not touching this PR until #655 is merged.

Then I'm not reviewing it till then.

axic:
Needs rebase, and possibly merging #655 first.

chfast:
Rebased, without checking the implementation.

test/utils/asserts.cpp
chfast: Not needed any more.

test/utils/asserts.hpp
```diff
@@ -23,45 +23,56 @@ MATCHER(Result, "empty result")
...
     // location of the error.
     using result_type = std::remove_cv_t<std::remove_reference_t<arg_type>>;
     if constexpr (!std::is_same_v<result_type, test::TypedExecutionResult>)
+    {
+        *result_listener << "TypedExecutionResult expected";
         return false;
+    }
 
     if (arg.trapped || !arg.has_value)
         return false;
```
axic, line 41: Perhaps display a reason here too if you do that above?

chfast: This is not needed because "trapped" execution is printed as "trapped", etc.

Random example:
```
Value of: execute(*instance, 0, {65534})
Expected: result 1
  Actual: trapped (of type fizzy::test::TypedExecutionResult)
```

test/utils/asserts.cpp
```diff
@@ -38,4 +38,39 @@ std::ostream& operator<<(std::ostream& os, ExecutionResult result)
...
+{
+    const auto format_flags = os.flags();
+
+    if (result.has_value)
+    {
+        os << "result(";
+        switch (result.type)
+        {
+        case ValType::i32:
+            os << result.value.as<uint32_t>() << " [0x" << std::hex << result.value.as<uint32_t>()
```
axic, line 54: Switch back to `dec`?

axic: And/or switch to `dec` explicitly for the first part?

test/utils/asserts.cpp
```diff
@@ -38,4 +38,39 @@ std::ostream& operator<<(std::ostream& os, ExecutionResult result)
...
+        case ValType::f64:
+            os << result.value.f64 << " (f64)";
+            break;
+        }
+        os << ")";
+    }
+    else
+        os << ExecutionResult{result};
+
+    os.flags(format_flags);
```
axic, line 72: Ah this is the reset. So is decimal mode on by default according to the std?

test/unittests/test_utils_test.cpp
```diff
@@ -70,3 +77,30 @@ TEST(test_utils, print_c_execution_result)
...
+    std::stringstream str_void;
+    str_void << TypedExecutionResult{Void, {}};
+    EXPECT_EQ(str_void.str(), "result()");
+
+    std::stringstream str_value_i32;
+    str_value_i32 << TypedExecutionResult{ExecutionResult{Value{42}}, ValType::i32};
+    EXPECT_EQ(str_value_i32.str(), "result(42 [0x2a] (i32))");
+
+    std::stringstream str_value_i64;
+    str_value_i64 << TypedExecutionResult{ExecutionResult{Value{42_u64}}, ValType::i64};
```
axic, line 89: Perhaps have a second test with `>2^32-1` value.

test/unittests/test_utils_test.cpp
```diff
@@ -70,3 +77,30 @@ TEST(test_utils, print_c_execution_result)
...
+    std::stringstream str_trap;
+    str_trap << TypedExecutionResult{Trap, {}};
+    EXPECT_EQ(str_trap.str(), "trapped");
+
+    std::stringstream str_void;
+    str_void << TypedExecutionResult{Void, {}};
+    EXPECT_EQ(str_void.str(), "result()");
+
+    std::stringstream str_value_i32;
+    str_value_i32 << TypedExecutionResult{ExecutionResult{Value{42}}, ValType::i32};
```
axic, line 85: Does this constructor verify that the input is `<=2^32-`? If not, can you put a test with a large value?

test/unittests/test_utils_test.cpp
```diff
@@ -70,3 +77,30 @@ TEST(test_utils, print_c_execution_result)
...
+    EXPECT_EQ(str_value_i32.str(), "result(42 [0x2a] (i32))");
+
+    std::stringstream str_value_i64;
+    str_value_i64 << TypedExecutionResult{ExecutionResult{Value{42_u64}}, ValType::i64};
+    EXPECT_EQ(str_value_i64.str(), "result(42 [0x2a] (i64))");
+
+    std::stringstream str_value_f32;
+    str_value_f32 << TypedExecutionResult{ExecutionResult{Value{1.125f}}, ValType::f32};
+    EXPECT_EQ(str_value_f32.str(), "result(1.125 (f32))");
```
axic, line 94: Can you add a test with negative floating point values?

test/unittests/test_utils_test.cpp
```diff
@@ -70,3 +77,30 @@ TEST(test_utils, print_c_execution_result)
...
+    str_void << TypedExecutionResult{Void, {}};
+    EXPECT_EQ(str_void.str(), "result()");
+
+    std::stringstream str_value_i32;
+    str_value_i32 << TypedExecutionResult{ExecutionResult{Value{42}}, ValType::i32};
+    EXPECT_EQ(str_value_i32.str(), "result(42 [0x2a] (i32))");
+
+    std::stringstream str_value_i64;
+    str_value_i64 << TypedExecutionResult{ExecutionResult{Value{42_u64}}, ValType::i64};
+    EXPECT_EQ(str_value_i64.str(), "result(42 [0x2a] (i64))");
```
axic, line 90: Can you add a test with negative integer values?

test/utils/execute_helpers.hpp
```diff
@@ -31,13 +31,15 @@ inline ExecutionResult execute(
     std::transform(std::cbegin(typed_args), std::cend(typed_args), std::begin(args),
         [](const auto& typed_arg) { return typed_arg.value; });
 
-    return fizzy::execute(instance, func_idx, args.data());
+    const auto result = fizzy::execute(instance, func_idx, args.data(), depth);
+    const auto result_type = func_type.outputs.empty() ? ValType{} : func_type.outputs[0];
```
axic, line 35: Perhaps an assertion that `outputs.size() <= 1`?

test/unittests/test_utils_test.cpp
```diff
@@ -34,13 +34,6 @@ TEST(test_utils, from_hex)
     EXPECT_THROW_MESSAGE(from_hex("FG"), std::out_of_range, "not a hex digit");
 }
 
-TEST(test_utils, DISABLED_result_signed_int)
```
axic: Why not just remove this in the previous commit?

test/utils/asserts.hpp
```diff
@@ -23,45 +23,56 @@ MATCHER(Result, "empty result")
...
 
-    static_assert(is_any_of<value_type, uint32_t, int32_t, uint64_t, int64_t, float, double>);
-
     // Require the arg to be TypedExecutionResult.
     // This can be a static_assert, but just returning false and failing a test provides better
     // location of the error.
     using result_type = std::remove_cv_t<std::remove_reference_t<arg_type>>;
     if constexpr (!std::is_same_v<result_type, test::TypedExecutionResult>)
+    {
+        *result_listener << "TypedExecutionResult expected";
```
axic, line 36: Can the test for texts include this?

test/utils/asserts.hpp
```diff
@@ -23,45 +23,56 @@ MATCHER(Result, "empty result")
...
 }
 
-template <typename T, typename... Ts>
-constexpr bool is_any_of = std::disjunction_v<std::is_same<T, Ts>...>;
-
 MATCHER_P(Result, value, "")  // NOLINT(readability-redundant-string-init)
 {
     using namespace fizzy;
 
-    static_assert(is_any_of<value_type, uint32_t, int32_t, uint64_t, int64_t, float, double>);
```
axic: Why remove this?

test/unittests/test_utils_test.cpp
```diff
@@ -104,3 +97,34 @@ TEST(test_utils, print_typed_execution_result)
...
+    EXPECT_THAT(TypedExecutionResult(Value{0_u64}, ValType::i64), Result(0_u64));
+    EXPECT_THAT(TypedExecutionResult(Value{0_u32}, ValType::i32), Not(Result(0_u64)));
+
+    EXPECT_THAT(TypedExecutionResult(Value{0_u32}, ValType::i32), Result(0_u32));
+
+    // For non-negative values zero-extension is conveniently allowed.
+    EXPECT_THAT(TypedExecutionResult(Value{0_u64}, ValType::i64), Result(0));
+    EXPECT_THAT(TypedExecutionResult(Value{0_u64}, ValType::i64), Result(0_u32));
+
+    EXPECT_THAT(TypedExecutionResult(Value{-1_u64}, ValType::i64), Not(Result(-1)));
```
axic, line 128: Missing test for `-1_u32`?

test/unittests/test_utils_test.cpp
chfast: Here you mean the `Value{}` constructor. It currently has overloads for many integer types and maps them to wasm types. No additional check is required.

test/utils/asserts.cpp
chfast: No. It resets to what state it was before. So explicit `dec` is needed to the first part.

test/utils/asserts.hpp
chfast: Not strictly needed now and the Result() will just return false.

axic (approved):

test/utils/asserts.hpp
```diff
@@ -23,13 +25,57 @@ MATCHER(Result, "empty result")
...
 {
+    using namespace fizzy;
+
+    // Require the arg to be TypedExecutionResult.
+    // This can be a static_assert, but just returning false and failing a test provides better
+    // location of the error.
+    using result_type = std::remove_cv_t<std::remove_reference_t<arg_type>>;
+    if constexpr (!std::is_same_v<result_type, test::TypedExecutionResult>)
+    {
+        if (result_listener->IsInterested())
```
gumb0, line 36: Where is this part of google.test documented?

test/utils/asserts.hpp
```diff
@@ -23,13 +25,57 @@ MATCHER(Result, "empty result")
...
+            *result_listener << "TypedExecutionResult expected";
+        return false;
+    }
```
gumb0, line 39: to avoid repeating `std::is_same_v<result_type, test::TypedExecutionResult>` below
```suggestion
    }
    else
    {
        if (arg.trapped || !arg.has_value)
            return false;
            
        if constexpr (std::is_same_v<value_type, float>)
        {
        ...
    }
```

test/utils/asserts.hpp
chfast: I found this in implementation of builtin matchers. This seems not to be strictly needed, but this is false e.q. when `Not()` is used.

test/utils/asserts.hpp
```diff
@@ -23,13 +25,57 @@ MATCHER(Result, "empty result")
...
+            return arg.type == ValType::i64 && arg.value.i64 == static_cast<uint64_t>(value);
+        }
+
+        if constexpr (std::is_integral_v<value_type> && sizeof(value_type) == sizeof(uint32_t))
+        {
+            if (arg.type == ValType::i32)
+            {
+                return arg.value.i64 == static_cast<uint32_t>(value);
+            }
+            else if (arg.type == ValType::i64 && value >= 0)
```
gumb0, line 67: Do you plan to remove this and change the tests at some point?

chfast: Not sure. Probably make sense to always use `Result(0_u32)` or `Result(0_u64)` only.

test/utils/asserts.hpp
gumb0: Seems to be undocumented feature. The only mention of `result_listener` I found uses only `operator<<` https://github.com/google/googletest/blob/master/googlemock/docs/cheat_sheet.md#defining-matchers

test/unittests/test_utils_test.cpp
```diff
@@ -70,3 +70,109 @@ TEST(test_utils, print_c_execution_result)
...
+}
+
+TEST(test_utils, result_value_matcher)
+{
+    // Exercise every check in Result(value) implementation.
+    // The implementation and checks below are organized by the value's type in Result(value).
+    using testing::Not;
+
+    // TypedExecutionResult is required to be matched against Result(value).
+    EXPECT_THAT(ExecutionResult(Value{1_u64}), Not(Result(1_u64)));
```
gumb0, line 127: Add a test for non-wasm type in `Result`?

test/unittests/test_utils_test.cpp
```diff
@@ -70,3 +70,109 @@ TEST(test_utils, print_c_execution_result)
...
+    EXPECT_THAT(TypedExecutionResult(Value{-1_u32}, ValType::i32), Not(Result(-1_u64)));
+
+    EXPECT_THAT(TypedExecutionResult(Value{-1_u64}, ValType::i64), Result(-1_u64));
+    EXPECT_THAT(TypedExecutionResult(Value{-1_u64}, ValType::i64), Not(Result(-1)));
+    EXPECT_THAT(TypedExecutionResult(Value{-1_u64}, ValType::i64), Not(Result(-1_u32)));
+}
+
+TEST(test_utils, result_value_matcher_explain_missing_result_type)
+{
+    const auto result = testing::internal::MakePredicateFormatterFromMatcher(Result(1_u64))(
```
gumb0, line 158: Well this is clearly something internal to google.mock, I'm not sure it's a good idea to use it. For coverage tests with `Not` above should be sufficient.

gumb0: Ah for coverage you mentioned that tests wth `Not` don't print output... Ok fine, we can leave this.

gumb0 (approved):
Looks good, one minor test suggestion https://github.com/wasmx/fizzy/pull/659#discussion_r551893461

test/unittests/test_utils_test.cpp
chfast: This is to test the message output. Using `Not()` will get us coverage (if `IsInterested()` is not used) but we cannot test the message. This was the previous version. I can roll back to that one if preferred.

PR #661, opened 2020-11-23
test: properly print FizzyExecutionResult when CResult matcher fails

gumb0:


test/utils/asserts.cpp
```diff
@@ -5,19 +5,34 @@
 #include "asserts.hpp"
 #include <ostream>
 
-namespace fizzy
+namespace
 {
-std::ostream& operator<<(std::ostream& os, ExecutionResult result)
+template <typename ResultType>
```
chfast, line 10:
```suggestion
template <typename ResultT>
```

chfast:
Please add unit testes for all functions in this file. Code coverage is 0.

chfast (approved):

test/utils/asserts.hpp
```diff
@@ -81,6 +81,8 @@ MATCHER_P(CResult, value, "")  // NOLINT(readability-redundant-string-init)
     }                                                                                        \
     (void)0
 
+std::ostream& operator<<(std::ostream& os, FizzyExecutionResult);
```
axic, line 84: Should this be moved closer to the other definition or left here as it is on the top-level namespace?

axic (approved):

test/utils/asserts.hpp
chfast: Because the intention of these is print tests failures only, it is ok to have these defined in `asserts.hpp`. And the definition is in global namespace as the `FizzyExecutionResult`.

axic: I mean closer to the other prototype 5 lines below it.

gumb0: They have to be in different namespaces.

axic: What I meant is moving this prototype **after** the other prototype so they are easy to see. Anyhow it is a minor detail.

PR #662, opened 2020-11-24
test: Add locals init tests

chfast:


test/unittests/execute_test.cpp
```diff
@@ -39,9 +39,9 @@ TEST(execute, select)
 {
     /* wat2wasm
     (func (param i64 i64 i32) (result i64)
-      get_local 0
-      get_local 1
-      get_local 2
+      local.get 0
```
axic, line 42: Both of these are valid, are we using this form now consistently?

axic (approved):

test/unittests/execute_test.cpp
chfast: I'm using `local.xxx` consistently for some time :) Not all instructions have been converted to the newer form.

chfast: There are still other occurrences of `{set,get}_local` in other files.

axic: But is this first commit making it uniform, or there are still places using `get_local`?

chfast: There are other files having `get_local`.

gumb0: I always forget what is the preference.

gumb0 (approved):

test/unittests/execute_test.cpp
chfast: The ones with the `.`.

axic: > There are other files having get_local.

Is it too many to change them here and now?

gumb0 (approved):

PR #664, opened 2020-11-24
capi: Documentation cleanup

gumb0:


gumb0:
Question: do we need to mention that it supports only ASCII-subset of UTF-8 for names, or is it obvious from the face that the string is `const char*` ?

include/fizzy/fizzy.h
```diff
@@ -184,20 +184,21 @@ bool fizzy_find_exported_function_index(
...
-///                                      doesn't import a table. Not an array, because Wasm 1.0
-///                                      doesn't support more than one table in a module.
-/// @param      imported_memory          Pointer to the imported memory. Can be NULL iff module
-///                                      doesn't import a memory. Not an array, because Wasm 1.0
-///                                      doesn't support more than one memory in a module.
-/// @param      imported_globals         Pointer to the imported globals array. Can be NULL iff
-///                                      imported_globals_size equals 0.
-/// @param      imported_globals_size    Size of the imported global array. Can be zero.
-/// @returns    non-NULL pointer to instance in case of success, NULL otherwise.
+/// @param      module                     Pointer to module. Cannot be NULL.
```
axic, line 187: Are we also keeping some specific width between `@param` and the name? We could say that and do it in the other PR too (#528).

include/fizzy/fizzy.h
```diff
@@ -32,10 +32,10 @@ typedef struct FizzyExecutionResult
 {
     /// Whether execution ended with a trap.
     bool trapped;
-    /// Whether function returned a value. Valid only if trapped equals false.
+    /// Whether function returned a value.
     bool has_value;
```
axic, line 36: So if we make this ambiguous here, why don't we state that if `trapped = false`, the other fields and undefined behaviour.

include/fizzy/fizzy.h
```diff
@@ -159,10 +159,10 @@ void fizzy_free_module(const FizzyModule* module);
 
 /// Get type of the function defined in the module.
 ///
-/// @param module   Pointer to module. Cannot be NULL.
-/// @param func_idx Function index. Can be either index of an imported function or of a function
-///                 defined in module. Behaviour is undefined, if index is not valid according to
-///                 module definition.
+/// @param module      Pointer to module. Cannot be NULL.
+/// @param func_idx    Function index. Can be either index of an imported function or of a function
```
chfast, line 163: IMHO it is fine to just keep <=2 spaces between param name and description.

include/fizzy/fizzy.h
chfast: I don't see this ambiguous. This is `false` if `trapped == true`. You can add this in the description.

include/fizzy/fizzy.h
axic: I agree, vote for 2 spaces.

Also https://github.com/wasmx/fizzy/pull/664#discussion_r529725849 ^^

include/fizzy/fizzy.h
chfast: I keep one space because this looks like a "command" `@param name` and this is always aligned.

include/fizzy/fizzy.h
gumb0: Added a note.

include/fizzy/fizzy.h
gumb0: Changed both widths.

axic: Oh sorry, I meant the 2 spaces for between `@param` and `module`, and 4 spaces between `module `and `Pointer to module` 😬

chfast: What is the point of additional fucking spaces?

axic: It was the style @gumb0 used half time and I found it rather readable.

gumb0: I think it looks more readable when it's more spaced out.

(what I originally used was starting comments on tab positions)

chfast (approved):

axic (approved):

PR #665, opened 2020-11-25
test: improve capi.imported_memory_access test

gumb0:
This test was added before imporing memory was supported in C API. Now I remembered that it should actually test something.

### DIFF
--- a/test/unittests/capi_test.cpp
+++ b/test/unittests/capi_test.cpp
@@ -594,14 +594,55 @@ TEST(capi, memory_access)
 TEST(capi, imported_memory_access)
 {
     /* wat2wasm
-      (memory (import "mod" "mem") 1)
+      (memory (export "mem") 1)
+      (data (i32.const 1) "\11\22")
+      (func (result i32)
+        i32.const 0
+        i32.load
+      )
     */
-    const auto wasm = from_hex("0061736d01000000020c01036d6f64036d656d020001");
-    auto module = fizzy_parse(wasm.data(), wasm.size());
+    const auto wasm_memory = from_hex(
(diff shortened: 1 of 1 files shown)


test/unittests/capi_test.cpp
```diff
@@ -600,15 +600,50 @@ TEST(capi, memory_access)
...
+
+    EXPECT_EQ(fizzy_get_instance_memory_size(instance), 65536);
+
+    uint8_t* memory_data = fizzy_get_instance_memory_data(instance);
+    ASSERT_NE(memory_data, nullptr);
+
+    memory_data[0] = 0xaa;
+    memory_data[1] = 0xbb;
+
+    EXPECT_EQ(fizzy_execute(instance, 0, nullptr, 0).value.i64, 0x22bbaa);
```
chfast, line 643: Also check memory in `instance_memory`.

test/unittests/capi_test.cpp
```diff
@@ -600,15 +600,50 @@ TEST(capi, memory_access)
...
 {
+    /* wat2wasm
+      (memory (export "mem") 1)
+      (data (i32.const 1) "\11\22")
+    */
+    const auto wasm_memory =
+        from_hex("0061736d010000000503010001070701036d656d02000b08010041010b021122");
+    auto module_memory = fizzy_parse(wasm_memory.data(), wasm_memory.size());
+    ASSERT_NE(module_memory, nullptr);
+    auto instance_memory =
```
chfast, line 611: For readability, you can indicate which objects are pointers by `auto*`.

chfast (approved):

PR #666, opened 2020-11-25
test: Add more tests for call depth

chfast:


test/unittests/execute_call_test.cpp
```diff
@@ -622,6 +643,100 @@ TEST(execute_call, call_max_depth)
...
+    const auto wasm1 = from_hex(
+        "0061736d01000000010401600000020f01026d3007636f756e746572037e0103020100070c0108696e66696e69"
+        "746500000a0d010b00230042017c240010000b");
+
+    Value counter;
+
+    auto i1 = instantiate(parse(wasm1), {}, {}, {}, {{&counter, {ValType::i64, true}}});
+    counter.i64 = 0;
+    EXPECT_THAT(execute(*i1, 0, {}), Traps());
+    EXPECT_EQ(counter.i64, MaxDepth + 1);  // FIXME: Should be MaxDepth.
```
gumb0, line 727: Why should it be? Function is executed `MaxDepth + 1` times, because depth starts with 0 and ends with `MaxDepth` inclusive.

chfast: Oh, this is correct, but I prefer there to be 2048 allowed calls, not 2049. This is changed in #669.

test/unittests/execute_call_test.cpp
```diff
@@ -676,16 +802,48 @@ TEST(execute_call, call_via_imported_infinite_recursion)
...
+    const auto host_foo_type = module->typesec[0];
+
+    auto instance = instantiate(*module, {{host_foo, host_foo_type}});
+
+    counter = 0;
+    EXPECT_THAT(execute(*instance, 0, {}), Traps());
+    EXPECT_EQ(counter, MaxDepth / 2 + 1);
+}
+
+TEST(execute_call, call_imported_interleaved_infinite_recursion_start_with_wasm)
```
gumb0, line 821: I think this could have been just an additional execution in the previous test?

chfast: Merged.

test/unittests/execute_call_test.cpp
chfast: I removed the FIXME comment as we have #669 to discuss it.

gumb0 (approved):

PR #668, opened 2020-11-26
refactor: Use span instead of vectors in ExternalFunction

gumb0:
This allows to avoid extra vector allocation when going from C `FizzyExternalFunction` to C++ `fizzy::ExternalFunction` and also when C++ API returns from `find_exported_function()`.

Also I it helps with `std::function` refactoring in #643 

The downside is that user has to ensure that arrays pointed to by spans live long enough (usually these arrays are in `Instance`, and you need this `Instance` to run `ExternalFunction` anyway).

This leaves `ImportedFunction` still using `vector<ValType>` for input types, I think it's convenient to leave it like this, otherwise it wouldn't be possible to pass input types as inline initializer list like this:
```cpp
        auto imports = fizzy::resolve_imported_functions(
            *module, {
                     });
```
neither like this
```cpp
    std::vector<ImportedFunction> imported_functions = {
        {"mod1", "foo1", {}, ValType::i32, function_returning_value(0)},
        {"mod1", "foo2", {ValType::i32}, ValType::i32, function_returning_value(1)},
        {"mod2", "foo1", {ValType::i32}, ValType::i32, function_returning_value(2)},
        {"mod2", "foo2", {ValType::i64, ValType::i32}, std::nullopt, function_returning_void},
    };
```
(Spans have to point to some container, so explicit vector declarations would be required on the user side) 

Also note that to make above use cases work, `resolve_imported_functions()` produces `ExternalFunction`s with type `span`s pointing to the type vector inside `Module`, not to the passed vector. Not sure whether it's confusing from the API standpoint, but it works.

### DIFF
--- a/lib/fizzy/cxx20/span.hpp
+++ b/lib/fizzy/cxx20/span.hpp
@@ -59,6 +59,8 @@ public:
     constexpr T* data() const noexcept { return m_begin; }
     constexpr std::size_t size() const noexcept { return m_size; }
 
+    constexpr bool empty() const noexcept { return m_size == 0; }
+
     constexpr iterator begin() const noexcept { return m_begin; }
     constexpr iterator end() const noexcept { return m_begin + m_size; }
 
--- a/lib/fizzy/instantiate.hpp
+++ b/lib/fizzy/instantiate.hpp
@@ -21,9 +21,14 @@ namespace fizzy
 struct ExecutionResult;
 struct Instance;
 
+/// Function representing WebAssembly or host function execution.
+using execute_function = std::function<ExecutionResult(Instance&, const Value*, int depth)>;
+
+/// Function with associated input/output types,
+/// used to represent both imported and exported functions.
 struct ExternalFunction
 {
-    std::function<ExecutionResult(Instance&, const Value*, int depth)> function;
+    execute_function function;
     FuncType type;
 };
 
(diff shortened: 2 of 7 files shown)


chfast (approved):
Looks ok, but this even more complicates the function-related types like "imported function", "exported function", "execute function", "execute fn", etc.

gumb0:
> Looks ok, but this even more complicates the function-related types like "imported function", "exported function", "execute function", "execute fn", etc.

Added comments to `execute_function` and `ExternalFunction`, not sure if it helps with clarification.

lib/fizzy/cxx20/span.hpp
```diff
@@ -59,6 +59,8 @@ class span
     constexpr T* data() const noexcept { return m_begin; }
     constexpr std::size_t size() const noexcept { return m_size; }
 
+    constexpr bool empty() const noexcept { return m_size == 0; }
```
axic, line 62: Matches std definition: https://en.cppreference.com/w/cpp/container/span/empty

lib/fizzy/instantiate.hpp
```diff
@@ -22,9 +22,14 @@ namespace fizzy
 struct ExecutionResult;
 struct Instance;
 
+// Function representing WebAssembly or host function execution.
```
axic, line 25: Can you make these comments doxygen?

gumb0: Changed.

axic (approved):

PR #669, opened 2020-11-27
Correct max call depth

chfast:
Previously there were 2049 number of recursive calls allowed (depth levels from 0 to 2048). With this change the number is reduced to 2048 (depth levels from 0 to 2047).

### DIFF
--- a/test/unittests/execute_call_test.cpp
+++ b/test/unittests/execute_call_test.cpp
@@ -605,8 +605,8 @@ TEST(execute_call, call_indirect_infinite_recursion)
     EXPECT_TRUE(execute(module, 0, {}).trapped);
 }
 
-constexpr int MaxDepth = 2048;
-static_assert(MaxDepth == CallStackLimit);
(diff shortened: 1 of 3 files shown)


gumb0 (approved):
Looks fine, but consider merging two tests here or in #666 
https://github.com/wasmx/fizzy/pull/666#discussion_r535285879

gumb0 (approved):

test/unittests/execute_call_test.cpp
```diff
@@ -605,8 +605,7 @@ TEST(execute_call, call_indirect_infinite_recursion)
     EXPECT_TRUE(execute(module, 0, {}).trapped);
 }
 
-constexpr int MaxDepth = 2048;
-static_assert(MaxDepth == CallStackLimit);
```
axic, line 609: The point of this duplicated declaration was that it would be detected in the tests if we adjust this limit.

axic (approved):

PR #670, opened 2020-11-27
spectests: Refactor value reading and checking

chfast:


test/spectests/spectests.cpp
```diff
@@ -525,7 +522,7 @@ class test_runner
             return false;
         }
 
-        const auto expected_value = read_floating_point_value<T>(expected);
+        const auto expected_value = read_value(expected)->value;  // TODO: Type ignored.
```
gumb0, line 525: Is this comment still relevant? (second commits removes the same comment for integers, but the code is quite similar)

test/spectests/spectests.cpp
```diff
@@ -542,10 +539,14 @@ class test_runner
 
     bool check_result(fizzy::Value actual_value, const json& expected)
     {
+        // TODO: Check types here.
+        // TODO: Use read_value() here too if it can handle "nan:..." strings.
```
gumb0, line 543: Well `Value` can't represent expectation like "arithmetic NaN expected"

test/spectests/spectests.cpp
```diff
@@ -406,40 +407,33 @@ class test_runner
...
-    std::optional<fizzy::Value> read_value(const json& v)
-    {
-        const auto type = v.at("type").get<std::string>();
+        const uint64_t raw_value = std::stoull(v.at("value").get<std::string>());
 
-        if (type == "i32" || type == "i64")
-            return read_integer_value(v);
+        if (type == "i32")
+        {
+            assert(static_cast<uint32_t>(raw_value) == raw_value && "overflow if i32 test value");
```
gumb0, line 419:
```suggestion
            assert(static_cast<uint32_t>(raw_value) == raw_value && "overflow in i32 test value");
```

test/spectests/spectests.cpp
chfast: True. But you can extend the Value to handle that or handle `nan:...` first using strings and then `read_value()`.

test/spectests/spectests.cpp
chfast: Yes. read_value() produces type and value, but type is discarded here.

test/spectests/spectests.cpp
```diff
@@ -406,40 +407,33 @@ class test_runner
...
+            return static_cast<uint32_t>(raw_value);
+        }
+        else if (type == "i64")
+        {
+            return raw_value;
+        }
         else if (type == "f32")
-            return read_floating_point_value<float>(v);
+        {
+            assert(static_cast<uint32_t>(raw_value) == raw_value && "overflow if f32 test value");
```
axic, line 428:
```suggestion
            assert(static_cast<uint32_t>(raw_value) == raw_value && "overflow in f32 test value");
```

test/spectests/spectests.cpp
axic: What exactly is the string part doing? It i just a true so no op. Why not leave it as a comment instead?

axic (approved):
Looks good apart from those smaller comments.

gumb0 (approved):

test/spectests/spectests.cpp
gumb0: It's a comment, that will be in the output when assertion fails, I think. Could be a regular commit, but I don't mind merging it like this.

axic: Ah right, it will be "literalized". Yeah, could keep it this way.

PR #672, opened 2020-12-11
test: Log error message for spectests expecting a failure

gumb0:
This can be helpful, when you'd like to check whether the invalid case fails for the right reason.
(It is displayed only with `--show-passed` option)

### DIFF
--- a/test/smoketests/spectests/CMakeLists.txt
+++ b/test/smoketests/spectests/CMakeLists.txt
@@ -43,6 +43,16 @@ set_tests_properties(
 )
 
 add_test(
+        NAME fizzy/smoketests/spectests/showpassed-error-message
+        COMMAND fizzy-spectests ${CMAKE_CURRENT_LIST_DIR}/default --show-passed
+)
+set_tests_properties(
+        fizzy/smoketests/spectests/showpassed-error-message
+        PROPERTIES
+        PASS_REGULAR_EXPRESSION "assert_malformed PASSED invalid wasm module prefix"
+)
+
+add_test(
         NAME fizzy/smoketests/spectests/showskipped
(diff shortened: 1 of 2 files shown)


test/spectests/spectests.cpp
```diff
@@ -293,15 +293,15 @@ class test_runner
                 catch (fizzy::parser_error const& ex)
                 {
                     if (type == "assert_malformed")
-                        pass();
+                        pass(ex.what());
```
chfast, line 296: It would be better to also include the expected error kind, e.g.

```suggestion
                        pass("malformed: " + ex.what());
```

gumb0: Each line already includes the kind of assertion it tests:
```
Line 29: assert_invalid PASSED invalid index of an exported function
```
```
Line 6: assert_malformed PASSED invalid wasm module prefix
```

chfast: Ah ok. This is not visible in the added CTest test.

test/smoketests/spectests/CMakeLists.txt
```diff
@@ -42,6 +42,16 @@ set_tests_properties(
...
 )
 
+add_test(
+        NAME fizzy/smoketests/spectests/showpassed-error-message
+        COMMAND fizzy-spectests ${CMAKE_CURRENT_LIST_DIR}/default --show-passed
+)
+set_tests_properties(
+        fizzy/smoketests/spectests/showpassed-error-message
+        PROPERTIES
+        PASS_REGULAR_EXPRESSION "PASSED invalid wasm module prefix"
```
chfast, line 52:
```suggestion
        PASS_REGULAR_EXPRESSION "assert_invalid PASSED invalid wasm module prefix"
```

chfast (changes requested):

chfast (approved):

PR #673, opened 2020-12-15
cmake: Remove redundant target_compile_features(... cxx_std_17)

gumb0:
These are not needed, because `fizzy` library includes it, as pointed out in https://github.com/wasmx/fizzy/pull/329#discussion_r461691537


axic (approved):

chfast (approved):
It has it here and is `PUBLIC`: https://github.com/wasmx/fizzy/blob/master/lib/fizzy/CMakeLists.txt#L7

PR #674, opened 2020-12-15
capi: Cloning a module

gumb0:


include/fizzy/fizzy.h
```diff
@@ -157,6 +157,15 @@ const FizzyModule* fizzy_parse(const uint8_t* wasm_binary, size_t wasm_binary_si
 /// If passed pointer is NULL, has no effect.
 void fizzy_free_module(const FizzyModule* module);
 
+/// Make a copy of a module.
+///
+/// @param  module          Pointer to module. Cannot be NULL.
+/// @returns                Pointer to a newly allocated module, identical to @a module.
+///
+/// @note Creating a copy is required if more than single instance of a module is required, because
+/// instantiation takes ownetship of a module, and the same module can't be instantiated twice.
```
axic, line 166:
```suggestion
/// instantiation takes ownership of a module, and the same module can not be instantiated twice.
```

include/fizzy/fizzy.h
```diff
@@ -157,6 +157,15 @@ const FizzyModule* fizzy_parse(const uint8_t* wasm_binary, size_t wasm_binary_si
 /// If passed pointer is NULL, has no effect.
 void fizzy_free_module(const FizzyModule* module);
 
+/// Make a copy of a module.
+///
+/// @param  module          Pointer to module. Cannot be NULL.
+/// @returns                Pointer to a newly allocated module, identical to @a module.
+///
+/// @note Creating a copy is required if more than single instance of a module is required, because
```
axic, line 165:
```suggestion
/// @note Creating a copy is required if more than single instance of a module is needed, because
```
or
```suggestion
/// @note Creating a copy is needed if more than single instance of a module is required, because
```

include/fizzy/fizzy.h
```diff
@@ -157,6 +157,15 @@ const FizzyModule* fizzy_parse(const uint8_t* wasm_binary, size_t wasm_binary_si
 /// If passed pointer is NULL, has no effect.
 void fizzy_free_module(const FizzyModule* module);
 
+/// Make a copy of a module.
+///
+/// @param  module          Pointer to module. Cannot be NULL.
+/// @returns                Pointer to a newly allocated module, identical to @a module.
```
axic, line 163: Need to explain in what case null is returned. Also need to be clear what happens to the input, in case of a failure (I assume there is no case where it would be undefined content).

include/fizzy/fizzy.h
```diff
@@ -157,6 +157,17 @@ const FizzyModule* fizzy_parse(const uint8_t* wasm_binary, size_t wasm_binary_si
...
 void fizzy_free_module(const FizzyModule* module);
 
+/// Make a copy of a module.
+///
+/// @param  module          Pointer to module. Cannot be NULL.
+/// @returns                Pointer to a newly allocated module, identical to @a module,
+///                         or NULL in case memory for a module couldn't be allocated.
+///
+/// @note Creating a copy is needed if more than single instance of a module is required, because
+/// instantiation takes ownership of a module, and the same module can't be instantiated twice.
```
axic, line 167:
```suggestion
/// instantiation takes ownership of a module, and the same module cannot be instantiated twice.
```

We have not been using contractions in docs.

gumb0: Changed.

test/unittests/capi_test.cpp
```diff
@@ -35,6 +35,29 @@ TEST(capi, free_module_null)
...
+    /* wat2wasm
+      (func (param i32 i32) (result i32) (i32.const 0))
+    */
+    const auto wasm = from_hex("0061736d0100000001070160027f7f017f030201000a0601040041000b");
+    const auto* module1 = fizzy_parse(wasm.data(), wasm.size());
+    EXPECT_NE(module1, nullptr);
+
+    const auto* module2 = fizzy_clone_module(module1);
+    EXPECT_NE(module2, nullptr);
+    EXPECT_NE(module1, module2);
```
axic, line 49: I guess we can't do a "deep equal" without some new API :wink:

axic (approved):

chfast (approved):

PR #677, opened 2020-12-29
rust: add error class

axic:


bindings/rust/src/lib.rs
```diff
@@ -9,6 +9,33 @@ mod sys;
 use std::ffi::CString;
 use std::ptr::NonNull;
 
+#[derive(Eq, PartialEq, Debug, Clone)]
+pub enum Error2 {
+    ParsingFailure,
```
axic, line 14: In this PR the main question is whether we should go with error codes like this, or free text.

axic: If we have the enum, it is much easier to compare outcome in tests, but we need to add display trait implementations so those codes can be displayed with a text counterpart during printing (such as `println!("{:#}", err)`).

chfast: I'd got with string first.

axic:
Pushed the most minimalistic version, where we just return `String` as an error. This allows for callers to figure what went wrong.

The main reason for going beyond `String`, to a newtype (i.e. `struct Error(String)`) or an enum, because that allows conversion between different error types. Such a case is that if we call some other API which returns an error, we can add a conversion trait between that error class and the "fizzy error class". We do not have anything like that yet (the only exception is that `CString` thing, but I think that is more like an unrecoverable error).

At the minimum I would vote for the type alias version (`type Error = String`) with the expectation that we'll have a proper error class.

axic:
I also tried to use `err()` (see https://doc.rust-lang.org/std/result/enum.Result.html#method.err) in the tests, but for some reason it does not work.

chfast:
Using strings as the first step looks fine to me.

axic:
#743 did the first step of error texts, #781 does the proper final step. This PR is not needed anymore.

PR #682, opened 2021-01-07
Mention public API in readme

gumb0:


README.md
```diff
@@ -181,6 +180,9 @@ Licensed under the [Apache License, Version 2.0].
 [@chfast]: https://github.com/chfast
 [@gumb0]: https://github.com/gumb0
 
+[C API]: https://github.com/wasmx/fizzy/tree/master/include/fizzy
```
axic, line 183: I think `./include/fizzy/fizzy.h` would work well.

README.md
```diff
@@ -42,8 +42,7 @@ $ cmake ..
 $ cmake --build .
 ```
 
-This will build Fizzy as a library and since there is no public API
-(the so called *embedder API* in WebAssembly) yet, this is not very useful.
+This will build Fizzy as a library. [C API] is provided for embedding Fizzy engine in applications, as well as [Rust binding].
```
axic, line 45: I think for Rust we should give an actual example using the published crate, and perhaps link to `crates.io` and/or `docs.rs` (for documentation).

- https://crates.io/crates/fizzy
- https://docs.rs/fizzy

axic, line 45: For C perhaps having a simple example would be useful, just to give an indication. Something which loads a wasm file and executes ta function called "main"?

axic: We also have the how to include fizzy in cmake. We have a test for that.

README.md
```diff
@@ -42,8 +42,38 @@ $ cmake ..
...
 
-This will build Fizzy as a library and since there is no public API
-(the so called *embedder API* in WebAssembly) yet, this is not very useful.
+This will build Fizzy as a library. [C API] is provided for embedding Fizzy engine in applications, as well as 
+[Rust binding].
+
+A small example of using C API, that loads a wasm binary and executes a function:
+
+```c
+#include <fizzy/fizzy/h>
```
axic, line 51:
```suggestion
#include <fizzy/fizzy.h>
```

chfast (approved):

axic:
@gumb0 do you want to drop the rust part? I can create a PR and deal with some rust description later.

gumb0:
> @gumb0 do you want to drop the rust part? I can create a PR and deal with some rust description later.

Yes, would be good.

README.md
axic: @gumb0 perhaps this cmake example would be useful. See `test/integration/cmake_package/use_fizzy`.

gumb0: I'm not sure yet how to describe it other than mentioning `find_package(fizzy CONFIG REQUIRED)`. Can look into it later, or @chfast perhaps could help.

gumb0: Added a sentence anyway.

README.md
```diff
@@ -42,8 +42,40 @@ $ cmake ..
 $ cmake --build .
 ```
 
-This will build Fizzy as a library and since there is no public API
-(the so called *embedder API* in WebAssembly) yet, this is not very useful.
+This will build Fizzy as a library. [C API] is provided for embedding Fizzy engine in applications.
+
+A small example of using C API, that loads a wasm binary and executes a function:
```
axic, line 47:
```suggestion
A small example of using the C API, which loads a wasm binary and executes a function named `main`:
```

README.md
```diff
@@ -42,8 +42,40 @@ $ cmake ..
...
+    const FizzyExecutionResult result = fizzy_execute(instance, main_fn_index, NULL, 0);
+    if (result.trapped)
+        return false;
+
+    fizzy_free_instance(instance);
+    return true;
+}
+```
+
+Fizzy also provides CMake package for easy integration in the projects that use it:
```
axic, line 75:
```suggestion
Fizzy also provides a CMake package for easy integration in projects that use it:
```

README.md
```diff
@@ -42,8 +42,42 @@ $ cmake ..
...
-This will build Fizzy as a library and since there is no public API
-(the so called *embedder API* in WebAssembly) yet, this is not very useful.
+This will build Fizzy as a library. [C API] is provided for embedding Fizzy engine in applications.
+
+A small example of using the C API, which loads a wasm binary and executes a function named `main`:
+
+```c
+#include <fizzy/fizzy.h>
+
+bool execute_main(const uint8_t* wasm_binary, size_t wasm_binary_size)
```
chfast, line 52: It would be nice to create matching integration test at some point to make sure this example continue to work.

chfast (approved):

axic (approved):
But can you rename the commit log to not include rust?

PR #675, opened 2020-12-17
capi: Type and global inspection functions

gumb0:

PR #683, opened 2021-01-07
capi: Import inspection functions

gumb0:
Split from #675

### DIFF
--- a/include/fizzy/fizzy.h
+++ b/include/fizzy/fizzy.h
@@ -140,6 +140,38 @@ typedef struct FizzyExternalGlobal
     FizzyGlobalType type;
 } FizzyExternalGlobal;
 
+/// External kind.
+typedef enum FizzyExternalKind
+{
+    FizzyExternalKindFunction,
+    FizzyExternalKindTable,
+    FizzyExternalKindMemory,
+    FizzyExternalKindGlobal
+} FizzyExternalKind;
+
+/// Import description.
+///
+/// @note Only one member of #desc union corresponding to #kind is defined for each import.
+/// @note For the lifetime of the #module and #kind fields, please refer to the description provided
+/// by the function used to obtain this structure.
+typedef struct FizzyImportDescription
+{
+    /// Import's module name.
+    const char* module;
+    /// Import name.
+    const char* name;
+    /// Import kind.
+    FizzyExternalKind kind;
+    /// Import type definition.
+    union
+    {
+        FizzyFunctionType function_type;
+        FizzyLimits memory_limits;
+        FizzyLimits table_limits;
+        FizzyGlobalType global_type;
+    } desc;
+} FizzyImportDescription;
+
 /// Imported function.
 typedef struct FizzyImportedFunction
 {
@@ -197,6 +229,23 @@ uint32_t fizzy_get_type_count(const FizzyModule* module);
 ///                     to module definition.
 /// @return             Type corresponding to the index.
 FizzyFunctionType fizzy_get_type(const FizzyModule* module, uint32_t type_idx);
+
+/// Get number of imports defined in the module.
+///
+/// @param  module    Pointer to module. Cannot be NULL.
+/// @return           Number of imports in the module.
+uint32_t fizzy_get_import_count(const FizzyModule* module);
+
+/// Get the import description defined in the module.
+///
+/// @param  module        Pointer to module. Cannot be NULL.
+/// @param  import_idx    Import index. Behaviour is undefined if index is not valid according
+///                       to module definition.
+/// @return               Type of the import corresponding to the index.
+///                       FizzyImportDescription::module and FizzyImportDescription::name fields
+///                       point to the string stored inside the module and are valid as long as
+///                       module is alive (including after successful instantiation.)
+FizzyImportDescription fizzy_get_import_description(const FizzyModule* module, uint32_t import_idx);
 
 /// Get type of the function defined in the module.
 ///
--- a/lib/fizzy/capi.cpp
+++ b/lib/fizzy/capi.cpp
@@ -282,6 +282,37 @@ inline std::vector<fizzy::ExternalGlobal> unwrap(
         external_globals.begin(), unwrap_external_global_fn);
     return external_globals;
 }
+
+inline FizzyExternalKind wrap(fizzy::ExternalKind kind) noexcept
+{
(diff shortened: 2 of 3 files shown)


include/fizzy/fizzy.h
```diff
@@ -133,6 +133,33 @@ typedef struct FizzyExternalGlobal
     FizzyGlobalType type;
 } FizzyExternalGlobal;
 
+/// External kind.
+typedef enum FizzyExternalKind
```
chfast, line 137: Please compare with https://github.com/WebAssembly/wasm-c-api/blob/master/include/wasm.h#L254-L260.

gumb0: What do you mean? They're quite similar except that I use full word for "function" and list them in the order of the spec.

chfast: They use `uint8_t` storage type. So it could be used in C++ directly. Or wrap it in C++ for better control.

```cpp
typedef uint8_t wasm_externkind_t;
enum wasm_externkind_enum {
  WASM_EXTERN_FUNC,
  WASM_EXTERN_GLOBAL,
  WASM_EXTERN_TABLE,
  WASM_EXTERN_MEMORY,
};

// C++

enum class ExternalKind : wasm_externkind_t {
  func = WASM_EXTERN_FUNC,
  global = WASM_EXTERN_GLOBAL,
  table = WASM_EXTERN_TABLE,
  memory = WASM_EXTERN_MEMORY,
};
```

This may be done other day, but I'm suggesting exploring the design where some declarations are imported from C to C++ instead of only declaring C equivalents of C++ ones.

gumb0: Hmm, this could have been done for `FizzyValueType`, too :thinking:

chfast: True. We may go old-school in this PR and then consider conversion. This probably requires off-site discussion.

axic: > I use full word for "function"

I like this better.

> list them in the order of the spec.

So is the wasm-c-api not in order with the spec?

include/fizzy/fizzy.h
```diff
@@ -133,6 +133,33 @@ typedef struct FizzyExternalGlobal
...
+    FizzyExternalKindTable,
+    FizzyExternalKindMemory,
+    FizzyExternalKindGlobal
+} FizzyExternalKind;
+
+/// Import description.
+///
+/// @note Only one member of FizzyImportDescription::desc union corresponding to
+/// FizzyImportDescription::kind is defined for each import.
+typedef struct FizzyImportDescription
```
axic, line 149: Do we need to mention the lifetime of module/name here? It is explained in `fizzy_get_import_description`.

Perhaps what we could note here is that the user should refer to the documentation from the function returning this value about lifetimes:
```suggestion
typedef struct FizzyImportDescription
//// For the lifetime of the #module and #kind fields, please refer to the description provided by the function used to obtain this structure.
```

Is `#module` the correct way?

include/fizzy/fizzy.h
gumb0: > > list them in the order of the spec.
> 
> So is the wasm-c-api not in order with the spec?

Yeah, there it's func-global-table-memory.

test/unittests/capi_test.cpp
```diff
@@ -1195,6 +1195,127 @@ TEST(capi, get_type)
...
+    EXPECT_STREQ(import5.name, "mem");
+    EXPECT_EQ(import5.kind, FizzyExternalKindMemory);
+    EXPECT_EQ(import5.desc.memory_limits.min, 1);
+    EXPECT_TRUE(import5.desc.memory_limits.has_max);
+    EXPECT_EQ(import5.desc.memory_limits.max, 4);
+
+    fizzy_free_module(module);
+}
+
+TEST(capi, import_name_after_instantiate)
```
axic, line 1293: Is this testing the lifetime of the name/namespace pointers?

test/unittests/capi_test.cpp
```diff
@@ -1195,6 +1195,127 @@ TEST(capi, get_type)
...
+
+    EXPECT_EQ(fizzy_get_import_count(module), 4);
+    fizzy_free_module(module);
+}
+
+TEST(capi, get_import_description)
+{
+    /* wat2wasm
+      (func (import "m" "f1") (result i32))
+      (func (import "m" "f2") (param i64))
```
axic, line 1232: Would it make any difference in coverage having a case with multiple params and a result at the same time? Don't want to overcomplicate tests either.

axic (approved):
Looks good in general, but please squash the renaming commit. Added some smaller remarks.

include/fizzy/fizzy.h
```diff
@@ -133,6 +133,33 @@ typedef struct FizzyExternalGlobal
...
+{
+    const char* module;
```
axic, line 151: These fields are documented in #686. It should be consistent, either document here too, or not in #686.

test/unittests/capi_test.cpp
gumb0: Yes, checks that it's alive after instantiation, as the documentation claims.

include/fizzy/fizzy.h
gumb0: Yes, `#module` and `#name` worked, but I needed to add individual documentation to the members.

test/unittests/capi_test.cpp
gumb0: I'll add it for completeness.

lib/fizzy/capi.cpp
```diff
@@ -282,6 +282,37 @@ inline std::vector<fizzy::ExternalGlobal> unwrap(
...
+            wrap(module.typesec[import.desc.function_type_index]);
+        break;
+    case FizzyExternalKindTable:
+        c_import_description.desc.table_limits = wrap(import.desc.table.limits);
+        break;
+    case FizzyExternalKindMemory:
+        c_import_description.desc.memory_limits = wrap(import.desc.memory.limits);
+        break;
+    case FizzyExternalKindGlobal:
+        c_import_description.desc.global_type = wrap(import.desc.global);
```
axic, line 311: Hmm, shouldn't this be `wrap(import.desc.global.type)`?

axic (approved):

PR #687, opened 2021-01-08
Type hardening in tests

chfast:


test/unittests/api_test.cpp
```diff
@@ -44,10 +44,10 @@ TEST(api, execution_result_void)
 
 TEST(api, execution_result_value)
 {
-    const ExecutionResult result = Value{1234};
+    const ExecutionResult result = Value{1234_u32};
```
axic, line 47: If this i32 now, shouldn't there be another test for i64/f32/f64 (pick one or more?)?

test/unittests/execute_numeric_test.cpp
```diff
@@ -510,8 +510,7 @@ TEST(execute_numeric, i64_extend_i32_u_2)
         "0061736d0100000001060160017f017e030201000a1201100042effdb6f5fdddefd65e1a2000ad0b");
 
     auto instance = instantiate(parse(wasm));
-    const auto r = execute(*instance, 0, {0xff000000});
-    EXPECT_THAT(r, Result(uint64_t{0x00000000ff000000}));
+    EXPECT_THAT(execute(*instance, 0, {0xff000000}), Result(0x00000000ff000000_u64));
```
axic, line 513: Is this the last such instance or why pick on this one?

chfast: The last case, at least in files using `_u64`.

test/unittests/api_test.cpp
chfast: Not sure what this test is about. Seems to testing implicit Value -> ExecutionResult conversion.

axic: @gumb0 ?

gumb0: Yeah this tests `ExecutionResult` constructor taking `Value`

gumb0: > Yeah this tests `ExecutionResult` constructor taking `Value`

Right, I guess implicit one, because there's also a test for explicit constuction below.

gumb0 (approved):

test/utils/asserts.hpp
```diff
@@ -102,8 +102,21 @@ MATCHER_P(CResult, value, "")  // NOLINT(readability-redundant-string-init)
...
+        return arg.value.i64 == uint64_t{value};
+    }
+    else if constexpr (std::is_same_v<value_type, uint64_t>)
+    {
+        return arg.value.i64 == value;
+    }
+    else
+    {
+        if (result_listener->IsInterested())
+            *result_listener << "unsupported type";
```
axic, line 117: Do we want a test for this?

axic (approved):

PR #689, opened 2021-01-08
Use countl_zero/countr_zero from C++20 if available

### LINKED ISSUE #543: Make use of C++20 bit manipulation features

axic:
Closes #543.

### DIFF
--- a/lib/fizzy/cxx20/bit.hpp
+++ b/lib/fizzy/cxx20/bit.hpp
@@ -36,31 +36,61 @@ template <class To, class From>
 
 namespace fizzy
 {
-constexpr int popcount(uint32_t x) noexcept
-{
-    return std::popcount(x);
-}
-
-constexpr int popcount(uint64_t x) noexcept
-{
-    return std::popcount(x);
-}
+using std::countl_zero;
+using std::countr_zero;
+using std::popcount;
 }  // namespace fizzy
 
 #else
 
 namespace fizzy
 {
-constexpr int popcount(uint32_t x) noexcept
+inline constexpr int popcount(uint32_t x) noexcept
 {
     static_assert(sizeof(x) == sizeof(unsigned int));
     return __builtin_popcount(x);
 }
 
-constexpr int popcount(uint64_t x) noexcept
+inline constexpr int popcount(uint64_t x) noexcept
 {
     static_assert(sizeof(x) == sizeof(unsigned long long));
     return __builtin_popcountll(x);
+}
+
+inline constexpr int countl_zero(uint32_t x) noexcept
+{
+    static_assert(sizeof(x) == sizeof(unsigned int));
+    // NOTE: C++20 specifies this case, but the intrinsic leaves it as undefined.
+    if (x == 0)
+        return 32;
+    return __builtin_clz(x);
+}
+
+inline constexpr int countl_zero(uint64_t x) noexcept
+{
+    static_assert(sizeof(x) == sizeof(unsigned long long));
+    // NOTE: C++20 specifies this case, but the intrinsic leaves it as undefined.
+    if (x == 0)
+        return 64;
+    return __builtin_clzll(x);
+}
+
+inline constexpr int countr_zero(uint32_t x) noexcept
+{
+    static_assert(sizeof(x) == sizeof(unsigned int));
+    // NOTE: C++20 specifies this case, but the intrinsic leaves it as undefined.
+    if (x == 0)
+        return 32;
+    return __builtin_ctz(x);
+}
+
+inline constexpr int countr_zero(uint64_t x) noexcept
+{
+    static_assert(sizeof(x) == sizeof(unsigned long long));
+    // NOTE: C++20 specifies this case, but the intrinsic leaves it as undefined.
+    if (x == 0)
+        return 64;
+    return __builtin_ctzll(x);
 }
 }  // namespace fizzy
 
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -310,46 +310,22 @@ inline constexpr T rotr(T lhs, T rhs) noexcept
     return (lhs >> k) | (lhs << (num_bits - k));
 }
 
-inline uint32_t clz32(uint32_t value) noexcept
+template <typename T>
+inline constexpr T clz(T value) noexcept
 {
-    // NOTE: Wasm specifies this case, but C/C++ intrinsic leaves it as undefined.
-    if (value == 0)
-        return 32;
-    return static_cast<uint32_t>(__builtin_clz(value));
+    return static_cast<T>(countl_zero(value));
 }
 
-inline uint32_t ctz32(uint32_t value) noexcept
+template <typename T>
+inline constexpr T ctz(T value) noexcept
 {
-    // NOTE: Wasm specifies this case, but C/C++ intrinsic leaves it as undefined.
-    if (value == 0)
-        return 32;
-    return static_cast<uint32_t>(__builtin_ctz(value));
+    return static_cast<T>(countr_zero(value));
 }
 
-constexpr uint32_t popcnt32(uint32_t value) noexcept
+template <typename T>
+inline constexpr T popcnt(T value) noexcept
 {
-    return static_cast<uint32_t>(popcount(value));
-}
-
-inline uint64_t clz64(uint64_t value) noexcept
-{
-    // NOTE: Wasm specifies this case, but C/C++ intrinsic leaves it as undefined.
-    if (value == 0)
-        return 64;
-    return static_cast<uint64_t>(__builtin_clzll(value));
-}
-
-inline uint64_t ctz64(uint64_t value) noexcept
-{
-    // NOTE: Wasm specifies this case, but C/C++ intrinsic leaves it as undefined.
-    if (value == 0)
-        return 64;
-    return static_cast<uint64_t>(__builtin_ctzll(value));
-}
-
-constexpr uint64_t popcnt64(uint64_t value) noexcept
-{
-    return static_cast<uint64_t>(popcount(value));
+    return static_cast<T>(popcount(value));
 }
 
 template <typename T>
@@ -1081,17 +1057,17 @@ ExecutionResult execute(
 
         case Instr::i32_clz:
         {
-            unary_op(stack, clz32);
+            unary_op(stack, clz<uint32_t>);
             break;
         }
         case Instr::i32_ctz:
         {
-            unary_op(stack, ctz32);
+            unary_op(stack, ctz<uint32_t>);
             break;
         }
         case Instr::i32_popcnt:
         {
-            unary_op(stack, popcnt32);
+            unary_op(stack, popcnt<uint32_t>);
             break;
         }
         case Instr::i32_add:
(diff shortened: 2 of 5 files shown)


test/unittests/execute_numeric_test.cpp
```diff
@@ -519,15 +519,15 @@ TEST(execute_numeric, i64_extend_i32_u_2)
...
+        EXPECT_THAT(i64_clz(input), Result(expected)) << input;
 }
 
 TEST(execute_numeric, i64_ctz)
 {
     const auto i64_ctz = create_unary_operation_executor(Instr::i64_ctz);
-    EXPECT_THAT(i64_ctz(0_u64), Result(64));
-    EXPECT_THAT(i64_ctz(0x80_u64), Result(7));
+    for (const auto& [input, expected] : countr_zero64_test_cases)
+        EXPECT_THAT(i64_ctz(input), Result(expected)) << input;
```
axic, line 530: Perhaps this is the correct one, as `i64.ctz` returns `u32`?

```suggestion
        EXPECT_THAT(i64_ctz(input), Result(static_cast<uint32_t>(expected))) << input;
```

axic: Or we change all the test cases to always return `uint32_t`, as even though the c++ bit implementations are defined as `T countl_zero(T)`, they never return anything big?

axic: Missed that it is actually `int countl_zero(T)`, only wasm does otherwise. So changed all to use `int` properly.

lib/fizzy/execute.cpp
```diff
@@ -254,41 +254,29 @@ inline constexpr T rotr(T lhs, T rhs) noexcept
     return (lhs >> k) | (lhs << (num_bits - k));
 }
 
-inline uint32_t clz32(uint32_t value) noexcept
+constexpr uint32_t clz32(uint32_t value) noexcept
```
axic, line 257: We can replace these with a template and use them as `clz<uint32_t>` below. Would cat this code duplication into half.

axic: Maybe we dont even need this wrapper layer, the only thing it does is explicit casting of the return value.

lib/fizzy/cxx20/bit.hpp
```diff
@@ -45,6 +45,27 @@ constexpr int popcount(uint64_t x) noexcept
 {
     return std::popcount(x);
 }
+
+constexpr int countl_zero(uint32_t x) noexcept
```
chfast, line 49:
```suggestion
using std::countl_zero;
```

lib/fizzy/cxx20/bit.hpp
```diff
@@ -45,6 +45,27 @@ constexpr int popcount(uint64_t x) noexcept
...
+{
+    return std::countl_zero(x);
+}
+
+constexpr int countl_zero(uint64_t x) noexcept
+{
+    return std::countl_zero(x);
+}
+
+constexpr int countr_zero(uint32_t x) noexcept
```
chfast, line 59:
```suggestion
using std::countr_zero;
```

lib/fizzy/cxx20/bit.hpp
```diff
@@ -62,6 +83,38 @@ constexpr int popcount(uint64_t x) noexcept
     static_assert(sizeof(x) == sizeof(unsigned long long));
     return __builtin_popcountll(x);
 }
+
+constexpr int countl_zero(uint32_t x) noexcept
```
chfast, line 87:
```suggestion
inline constexpr int countl_zero(uint32_t x) noexcept
```

lib/fizzy/cxx20/bit.hpp
```diff
@@ -62,6 +83,38 @@ constexpr int popcount(uint64_t x) noexcept
...
+{
+    static_assert(sizeof(x) == sizeof(unsigned int));
+    if (x == 0)
+        return 32;
+    return __builtin_clz(x);
+}
+
+constexpr int countl_zero(uint64_t x) noexcept
```
chfast, line 95:
```suggestion
inline constexpr int countl_zero(uint64_t x) noexcept
```

lib/fizzy/cxx20/bit.hpp
```diff
@@ -62,6 +83,38 @@ constexpr int popcount(uint64_t x) noexcept
...
+{
+    static_assert(sizeof(x) == sizeof(unsigned long long));
+    if (x == 0)
+        return 64;
+    return __builtin_clzll(x);
+}
+
+constexpr int countr_zero(uint32_t x) noexcept
```
chfast, line 103:
```suggestion
inline constexpr int countr_zero(uint32_t x) noexcept
```

lib/fizzy/cxx20/bit.hpp
```diff
@@ -62,6 +83,38 @@ constexpr int popcount(uint64_t x) noexcept
...
+{
+    static_assert(sizeof(x) == sizeof(unsigned int));
+    if (x == 0)
+        return 32;
+    return __builtin_ctz(x);
+}
+
+constexpr int countr_zero(uint64_t x) noexcept
```
chfast, line 111:
```suggestion
inline constexpr int countr_zero(uint64_t x) noexcept
```

lib/fizzy/execute.cpp
chfast, line 257:
```suggestion
inline constexpr uint32_t clz32(uint32_t value) noexcept
```

lib/fizzy/execute.cpp
```diff
@@ -254,41 +254,29 @@ inline constexpr T rotr(T lhs, T rhs) noexcept
...
 {
-    // NOTE: Wasm specifies this case, but C/C++ intrinsic leaves it as undefined.
-    if (value == 0)
-        return 32;
-    return static_cast<uint32_t>(__builtin_clz(value));
+    return static_cast<uint32_t>(countl_zero(value));
 }
 
-inline uint32_t ctz32(uint32_t value) noexcept
+constexpr uint32_t ctz32(uint32_t value) noexcept
```
chfast, line 262:
```suggestion
inline constexpr uint32_t ctz32(uint32_t value) noexcept
```

lib/fizzy/execute.cpp
```diff
@@ -254,41 +254,29 @@ inline constexpr T rotr(T lhs, T rhs) noexcept
...
+    return static_cast<uint32_t>(countr_zero(value));
 }
 
 constexpr uint32_t popcnt32(uint32_t value) noexcept
 {
     return static_cast<uint32_t>(popcount(value));
 }
 
-inline uint64_t clz64(uint64_t value) noexcept
+constexpr uint64_t clz64(uint64_t value) noexcept
```
chfast, line 272:
```suggestion
inline constexpr uint64_t clz64(uint64_t value) noexcept
```

lib/fizzy/execute.cpp
```diff
@@ -254,41 +254,29 @@ inline constexpr T rotr(T lhs, T rhs) noexcept
...
 {
-    // NOTE: Wasm specifies this case, but C/C++ intrinsic leaves it as undefined.
-    if (value == 0)
-        return 64;
-    return static_cast<uint64_t>(__builtin_clzll(value));
+    return static_cast<uint64_t>(countl_zero(value));
 }
 
-inline uint64_t ctz64(uint64_t value) noexcept
+constexpr uint64_t ctz64(uint64_t value) noexcept
```
chfast, line 277:
```suggestion
inline constexpr uint64_t ctz64(uint64_t value) noexcept
```

test/unittests/bitcount_test_cases.hpp
```diff
@@ -9,7 +9,7 @@
 
 namespace fizzy::test
 {
-constexpr std::pair<uint32_t, uint32_t> popcount32_test_cases[]{
+constexpr std::pair<uint32_t, int> popcount32_test_cases[]{
```
chfast, line 12: Duplicating inputs does not make sense. Create a struct with results for all instructions.
```
template <T>
struct SomeGoodName
{
   T input;
   int popcount;
   int countl_zero;
   int countr_zero;
};

constexpr SomeGoodName<uint32_t> bit_counting32_test_cases[]{
    {0, 0, 32, 32},
    {0x80, 1, 24, 7},
    ...
};

**Review by chfast — CHANGES_REQUESTED** [review]

File: lib/fizzy/execute.cpp (hunk 9 above)
Comment (axic) [reply lib/fizzy/execute.cpp:262]: https://github.com/wasmx/fizzy/pull/689#discussion_r555032201

Actually just removing these wrappers seems to be okay?

Comment (axic) [reply lib/fizzy/execute.cpp]: So how about the template proposed in https://github.com/wasmx/fizzy/pull/689#discussion_r555031222 ?

Comment (chfast) [reply lib/fizzy/execute.cpp]: Template is fine.

File: test/unittests/bitcount_test_cases.hpp (hunk 12 above)
Comment (axic) [reply test/unittests/bitcount_test_cases.hpp]: I will keep this a separate commit even for merging, as not sure it is more readable alltogether.

File: CMakeLists.txt [hunk 13]
```diff
@@ -43,6 +43,7 @@ add_compile_options(
     -Wmissing-declarations
     $<$<COMPILE_LANGUAGE:CXX>:-Wextra-semi>
     $<$<COMPILE_LANGUAGE:CXX>:-Wold-style-cast>
+    -Wno-invalid-constexpr
```
axic, line 46: Stray, will remove before merge.

lib/fizzy/execute.cpp
```diff
@@ -254,46 +254,22 @@ inline constexpr T rotr(T lhs, T rhs) noexcept
     return (lhs >> k) | (lhs << (num_bits - k));
 }
 
-inline uint32_t clz32(uint32_t value) noexcept
-{
-    // NOTE: Wasm specifies this case, but C/C++ intrinsic leaves it as undefined.
```
gumb0, line 259: This comment could be useful in new implementation, too.

axic: You mean to say that C++20 defines that? Yeah, can do.

axic: Adding this to all:
```
// NOTE: C++ specifies this case, but intrinsic leaves it as undefined.
```

lib/fizzy/execute.cpp
chfast: > This mean you now need to construct a unit test what will reproduce this bug.

This is still missing. It requires a special test where you pollute the stack first and then assign result of `i64.clz`. If the result is `int` some garbage should stay in the value. One case should be enough. See `TEST(execute_numeric, i64_extend_i32_u)`.

I can also handle this later.

chfast (approved):

gumb0 (approved):
Looks good, don't forget to remove the change in CMakeLists.txt

lib/fizzy/execute.cpp
chfast: It looks it is not easy to omit the wrapper so I will skip creating tests for this.

chfast (approved):

chfast (approved):

PR #691, opened 2021-01-13
test: Drop as_uint32()

chfast:
Depends on #517.

### DIFF
--- a/test/unittests/execute_call_test.cpp
+++ b/test/unittests/execute_call_test.cpp
@@ -390,7 +390,7 @@ TEST(execute_call, imported_function_call_with_arguments)
 
     const auto module = parse(wasm);
 
-    auto host_foo = [](Instance&, const Value* args, int) { return Value{as_uint32(args[0]) * 2}; };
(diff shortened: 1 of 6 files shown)


test/unittests/execute_call_test.cpp
```diff
@@ -433,12 +433,10 @@ TEST(execute_call, imported_functions_call_indirect)
     ASSERT_EQ(module->codesec.size(), 2);
 
     constexpr auto sqr = [](Instance&, const Value* args, int) {
-        const auto x = as_uint32(args[0]);
-        return Value{uint64_t{x} * uint64_t{x}};
+        return Value{uint64_t{(args[0].i32)} * uint64_t{(args[0].i32)}};
```
axic, line 436: Why are the extra parentheses needed?

axic (approved):

test/unittests/execute_call_test.cpp
chfast: automatic refactoring... There is "inline variable" command in CLion which placed `x` in both places where it was used.

axic (approved):

PR #692, opened 2021-01-13
ci: Build and store doxygen documentation

chfast:


include/fizzy/fizzy.h
```diff
@@ -1,6 +1,8 @@
-// Fizzy: A fast WebAssembly interpreter
-// Copyright 2020 The Fizzy Authors.
-// SPDX-License-Identifier: Apache-2.0
+/// @file
```
gumb0, line 1: Without this file is considered private by default and documentation for free functions is not generated.

axic: Are you sure it can't just be added in the doxygen config?

Doxyfile
```diff
@@ -0,0 +1,2531 @@
...
+#---------------------------------------------------------------------------
+
+# The INPUT tag is used to specify the files and/or directories that contain
+# documented source files. You may enter file names like myfile.cpp or
+# directories like /usr/src/myproject. Separate the files or directories with
+# spaces. See also FILE_PATTERNS and EXTENSION_MAPPING
+# Note: If this tag is empty the current directory is searched.
+
+INPUT                  = include/fizzy/ \
+                         lib/fizzy/ \
```
gumb0, line 817: Do we want to include these internal files?

include/fizzy/fizzy.h
gumb0: Well adding the file explicitly in `INPUT` doesn't help (it's included anyway)

The only other way to fix it that I found is to set `EXTRACT_ALL = YES`, but it might include more stuff than needed, EVMC has it at `NO`.

gumb0: https://www.doxygen.nl/manual/commands.html#cmdfile

> The documentation of global functions, variables, typedefs, and enums will only be included in the output if the file they are in is documented as well or if EXTRACT_ALL is set to YES.

Doxyfile
axic: Do we want to use this doxygen generated file for the external documentation only or as a project documentation?

gumb0: Perhaps we can decide later, and for now merge it like this, to have something working.

gumb0 (approved):

Doxyfile
chfast: I'd go with generating docs for everything as a start. There may be better way of separating public and internal API.

chfast: E.g. there is `@internal` tag. https://stackoverflow.com/questions/11025971/separate-internal-from-external-documentation-in-doxygen

include/fizzy/fizzy.h
chfast: This is done wrongly. We don't want to the "copyright header" to be included in the documentation text.

It rather should be
```
// Copyright...

/// Fizzy public C API.
/// @file
```

gumb0: Changed. I just hate doxygen :angry:

chfast: `clang-doc` maybe if you want to contribute to LLVM.

chfast (commented):
Good for me.

gumb0 (approved):

Doxyfile
```diff
@@ -0,0 +1,2531 @@
...
+# quick idea about the purpose of the project. Keep the description short.
+
+PROJECT_BRIEF          =
+
+# With the PROJECT_LOGO tag one can specify a logo or an icon that is included
+# in the documentation. The maximum height of the logo should not exceed 55
+# pixels and the maximum width should not exceed 200 pixels. Doxygen will copy
+# the logo to the output directory.
+
+PROJECT_LOGO           =
```
axic, line 54: Hmm, can this accept svg?

Doxyfile
```diff
@@ -0,0 +1,2531 @@
...
+# When TYPEDEF_HIDES_STRUCT tag is enabled, a typedef of a struct, union, or
+# enum is documented as struct, union, or enum with the name of the typedef. So
+# typedef struct TypeS {} TypeT, will appear in the documentation as a struct
+# with name TypeT. When disabled the typedef will appear as a member of a file,
+# namespace, or class. And the struct will be named TypeS. This can typically be
+# useful for C code in case the coding convention dictates that all compound
+# types are typedef'ed and only the typedef is referenced, never the tag name.
+# The default value is: NO.
+
+TYPEDEF_HIDES_STRUCT   = NO
```
axic, line 427: Can y'all check the description above if it applies to us? I don't think we do too much typedefing.

axic (approved):
Ok, but would maybe we could do a pass on the doxygen config. I only did a brief scroll. We can do it after this is merged too.

Doxyfile
chfast: This should be `YES`.

Doxyfile
gumb0: I tried locally, and it didn't work for me.

gumb0: Actually, it does work, when I resize it to 200x200.

chfast:
I split Doxyfile into two commits:
- first generates the default config,
- second has Fizzy related changes. Some of these comes from EVMC, others are mine. I reviewed the options up to the HTML config (I don't care about exact visuals right now).

Doxyfile
```diff
@@ -2204,7 +2156,7 @@ INCLUDE_FILE_PATTERNS  =
 # recursively expanded use the := operator instead of the = operator.
 # This tag requires that the tag ENABLE_PREPROCESSING is set to YES.
 
-PREDEFINED             =
+PREDEFINED             = EVMC_DOCUMENTATION=1
```
axic, line 2159: EVMC? :)

chfast: Fixed this and other around.

axic (approved):

gumb0 (approved):

PR #694, opened 2021-01-20
Experiment on reusing C API definitions in C++ API

gumb0:


lib/fizzy/capi.cpp
```diff
@@ -45,14 +45,16 @@ inline fizzy::ValType unwrap(FizzyValueType value_type) noexcept
 
 inline FizzyFunctionType wrap(const fizzy::FuncType& type) noexcept
 {
-    return {(type.outputs.empty() ? FizzyValueTypeVoid : wrap(type.outputs[0])),
+    return {(type.outputs.empty() ? static_cast<FizzyValueType>(FizzyValueTypeVoid) :
```
chfast, line 48:
```suggestion
    return {(type.outputs.empty() ? FizzyValueType{FizzyValueTypeVoid} :
```
Nicer conversion may be possible here. https://godbolt.org/z/o9Y4jK

chfast (commented):
Looks good to be, except the `fizzy::ValType` deserves extended documentation now.

axic:
This causes quite a few issues on the Rust binding side. Not sure this PR is worth it. Created #821 as an alternative approach.

PR #695, opened 2021-01-21
test: Update spectest module definition

gumb0:
Spectest predefined module now has a proper spec: https://github.com/WebAssembly/spec/pull/1280

There were a couple of members missing which are not used anyway, so this change is only cosmetic.

### DIFF
--- a/test/spectests/spectests.cpp
+++ b/test/spectests/spectests.cpp
@@ -28,6 +28,15 @@ constexpr unsigned TestMemoryPagesLimit = (4 * 1024 * 1024 * 1024ULL) / fizzy::P
 // https://github.com/WebAssembly/spec/blob/99564b7eaa3452c2633b623c92fc286db2823f39/interpreter/README.md#spectest-host-module
 /* wat2wasm
 (module
+  (global (export "global_i32") i32 (i32.const 666))
+  (global (export "global_i64") i64 (i64.const 666))
+  (global (export "global_f32") f32 (f32.const 666))
+  (global (export "global_f64") f64 (f64.const 666))
+
+  (table (export "table") 10 20 anyfunc)
+
+  (memory (export "memory") 1 2)
+
   (func (export "print"))
   (func (export "print_i32") (param i32))
   (func (export "print_i64") (param i64))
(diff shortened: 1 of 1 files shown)


test/spectests/spectests.cpp
```diff
@@ -24,32 +24,33 @@ constexpr auto JsonExtension = ".json";
 constexpr auto UnnamedModule = "_unnamed";
 constexpr unsigned TestMemoryPagesLimit = (4 * 1024 * 1024 * 1024ULL) / fizzy::PageSize;  // 4 Gb
 
-// spectest module details:
-// https://github.com/WebAssembly/spec/issues/1111
-// https://github.com/WebAssembly/spec/blob/e3a3c12f51221daaa901691e674d7ff990029122/interpreter/host/spectest.ml#L33
-// https://github.com/WebAssembly/spec/blob/f85f6e6b70fff9ee0f6ba21ebcdad2183bd67874/test/harness/sync_index.js#L79
+// spectest module definition:
+// https://github.com/WebAssembly/spec/blob/master/interpreter/README.md#spectest-host-module
```
axic, line 28: Should we link to a commit hash or a tag? I think it is fine probably as is.

test/spectests/spectests.cpp
```diff
@@ -24,32 +24,36 @@ constexpr auto JsonExtension = ".json";
...
 /* wat2wasm
 (module
+  (global (export "global_i32") i32 (i32.const 666))
```
axic, line 31: Do you think this new ordering is dependent on? And if so do you think wabt produced the correct expected output?

If not, is this just to show it is a copy from a spec?

gumb0: It's just to show correspondence to the spec. Ordering of sections in the module is strict, all members are referred by names only in spectests, and it's not even required for module to be Wasm.

test/spectests/spectests.cpp
gumb0: Changed to commit hash

test/spectests/spectests.cpp
axic: > Ordering of sections in the module is strict

The sections are strictly ordered, but the wat->wasm translator could decide how it lays out members within those, e.g. the ordering of functions.

axic (approved):

test/spectests/spectests.cpp
gumb0: Well many out unit tests assume that wabt puts them in the order as defined in wat.

PR #697, opened 2021-01-27
api: Doxygenize comments for resolve functions

gumb0:


lib/fizzy/instantiate.hpp
```diff
@@ -131,13 +131,13 @@ struct ImportedFunction
...
-/// Create vector of ExternalFunctions ready to be passed to instantiate.
+/// Create vector of fizzy::ExternalFunction ready to be passed to instantiate().
 /// @a imported_functions may be in any order, but must contain functions for all of the imported
 /// function names defined in the module.
 std::vector<ExternalFunction> resolve_imported_functions(
     const Module& module, const std::vector<ImportedFunction>& imported_functions);
 
-// Global that should be used by instantiate as import, identified by module and global name.
+/// Global that should be used by instantiate as import, identified by module and global name.
 struct ImportedGlobal
```
axic, line 141: Do we want to also document the fields of `ImportedFunction`, `ImportedGlobal`, `Instance`? Maybe even if we do, we do not need to document very clear fields as this is mostly internal documentation.

axic (approved):

PR #660, opened 2020-11-23
capi: resolve imported globals

gumb0:
Depends on #637.

PR #699, opened 2021-01-27
capi: Random minor simplifications

gumb0:
One commit from #660 and one commit from #694

### DIFF
--- a/lib/fizzy/capi.cpp
+++ b/lib/fizzy/capi.cpp
@@ -162,9 +162,8 @@ inline std::vector<fizzy::ExternalFunction> unwrap(
 inline fizzy::ImportedFunction unwrap(const FizzyImportedFunction& c_imported_func)
 {
     fizzy::ImportedFunction imported_func;
-    imported_func.module =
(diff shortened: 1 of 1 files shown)


lib/fizzy/capi.cpp
```diff
@@ -162,9 +162,8 @@ inline std::vector<fizzy::ExternalFunction> unwrap(
 inline fizzy::ImportedFunction unwrap(const FizzyImportedFunction& c_imported_func)
 {
     fizzy::ImportedFunction imported_func;
-    imported_func.module =
-        c_imported_func.module ? std::string{c_imported_func.module} : std::string{};
-    imported_func.name = c_imported_func.name ? std::string{c_imported_func.name} : std::string{};
+    imported_func.module = c_imported_func.module;
```
axic, line 164: https://stackoverflow.com/a/10771938

But we require `module` and `name` to be non-null, right?

gumb0: Right, that's why it was unnecessary check.

axic (approved):

PR #702, opened 2021-01-28
rust: use i32 type in Value

axic:
Depends on #517.

### DIFF
--- a/bindings/rust/src/lib.rs
+++ b/bindings/rust/src/lib.rs
@@ -200,6 +200,41 @@ mod tests {
     use super::*;
 
     #[test]
+    fn value_conversions() {
+        // NOTE: since the underlying type is a union, a conversion or access to other members is undefined
+
+        let v: Value = i32::MIN.into();
+        assert_eq!(v.as_i32(), i32::MIN);
+        let v: Value = i32::MAX.into();
+        assert_eq!(v.as_i32(), i32::MAX);
+
+        let v: Value = u32::MIN.into();
+        assert_eq!(v.as_u32(), u32::MIN);
+        let v: Value = u32::MAX.into();
+        assert_eq!(v.as_u32(), u32::MAX);
+
+        let v: Value = i64::MIN.into();
+        assert_eq!(v.as_i64(), i64::MIN);
+        let v: Value = i64::MAX.into();
+        assert_eq!(v.as_i64(), i64::MAX);
+
+        let v: Value = u64::MIN.into();
+        assert_eq!(v.as_u64(), u64::MIN);
+        let v: Value = u64::MAX.into();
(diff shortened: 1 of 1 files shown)


bindings/rust/src/lib.rs
```diff
@@ -79,10 +79,10 @@ pub type Value = sys::FizzyValue;
 // NOTE: the union does not have i32
 impl Value {
     pub fn as_i32(&self) -> i32 {
-        unsafe { self.i64 as i32 }
+        unsafe { self.i32 as i32 }
     }
     pub fn as_u32(&self) -> u32 {
-        unsafe { self.i64 as u32 }
+        unsafe { self.i32 }
```
axic, line 85: The confusing fact is that even though it is called `i32`, the type is `u32` in the bindgen translated fizzy.h.

gumb0 (commented):
It looks like no unit tests use `i64` now (neither floating point types)

Consider adding tests for `Value` conversions.

axic:
Sure, but what test exactly? Neither `Value` or `sys::FizzyValue` is typed so not sure what is an appropriate test.

axic:
This is the kind of test we could have (this passes, at least on my CPU/FPU):
```
    #[test]
    fn value_conversions() {
        let v:Value = i32::MIN.into();
        assert_eq!(v.as_i32(), i32::MIN);
        assert_eq!(v.as_u32(), 2147483648);
        assert_eq!(v.as_i64(), 2147483648);
        assert_eq!(v.as_u64(), 2147483648);
        assert_eq!(v.as_f32(), -0.0);
        assert_eq!(v.as_f64(), 0.000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000010609978955);
```

bindings/rust/src/lib.rs
```diff
@@ -199,6 +199,41 @@ impl Instance {
...
+    fn value_conversions() {
+        // NOTE: since the underlying type is a union, a conversion or access to other members is undefined
+
+        let v: Value = i32::MIN.into();
+        assert_eq!(v.as_i32(), i32::MIN);
+        let v: Value = i32::MAX.into();
+        assert_eq!(v.as_i32(), i32::MAX);
+
+        let v: Value = u32::MIN.into();
+        assert_eq!(v.as_i32(), 0);
```
gumb0, line 212:
```suggestion
        assert_eq!(v.as_u32(), MIN);
```

axic: Good catch, fixed!

gumb0 (approved):

PR #703, opened 2021-01-28
ci: Don't run doxygen installation in background

gumb0:
Trying to fix random lint job failures.


axic:
@gumb0 do you think this is the cause for the issues on `master`?

I'll leave this one for @chfast.

gumb0:
I'm not sure, it was suggested by @chfast and it seems to help.

axic (approved):

PR #704, opened 2021-01-28
Support Rust files in wat2wasm4cpp.py script

### LINKED ISSUE #593: Extend wast2wasm4cpp to also work with Rust sources
[`wast2wasm4cpp`](https://github.com/wasmx/fizzy/blob/master/wat2wasm4cpp.py) is a tool proccessing `.cpp` files, which looks for a specific comment containing wasm source code and updates the binary code below it, e.g.
```
    /* wat2wasm
    (func (param i64) (result i64)
      (local i64)
      get_local 0
      set_local 1
      get_local 1
    )
    */
    const auto wasm =
        from_hex("0061736d0100000001060160017e017e030201000a0c010a01017e2000210120010b");
```

Here the source are the lines below `/* wat2wasm` and, the tool fills in the binary hex string within the `from_hex()` call.

Change this tool to emit Rust-compatible output if the filename is `.rs`:
```
    /* wat2wasm
    (func (param i64) (result i64)
      (local i64)
      get_local 0
      set_local 1
      get_local 1
    )
    */
    let input = hex::decode("0061736d0100000001060160017e017e030201000a0c010a01017e2000210120010b").unwrap();
```

Please note it must also run `cargo fmt` after the replacement, just like it runs `clang-format` for C/C++ sources.

gumb0:
Closes #593.

I used `rustfmt` becaust it takes a single file as argument.
However, it doesn't seem much useful, e.g. long hex strings are not broken down (this option is unstable and off by default
https://rust-lang.github.io/rustfmt/?version=v1.4.33&search=#format_strings
https://github.com/rust-lang/rustfmt/issues/3353)

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -35,7 +35,7 @@ commands:
             else
               [[ $OSTYPE = darwin* ]] && os=macos || os=ubuntu
               cd /usr/local
-              curl -L https://github.com/WebAssembly/wabt/releases/download/1.0.20/wabt-1.0.20-$os.tar.gz | sudo tar xz --strip 1
+              curl -L https://github.com/WebAssembly/wabt/releases/download/1.0.20/wabt-1.0.20-$os.tar.gz | (sudo tar xz --strip 1 || tar xz --strip 1)
             fi
 
   install_testfloat:
@@ -586,6 +586,13 @@ jobs:
           command: |
             rustfmt --version
             cargo fmt --all -- --check
+      - install_wabt
+      - run:
+          name: Check wat2wasm4tests
+          command: |
+            rustfmt --version
+            find bindings/rust -name '*.rs' -print -exec ./wat2wasm4tests.py {} \;
+            git diff --color --exit-code
       - run:
           name: Build (debug mode)
           command: cargo build
--- a/circle.yml
+++ b/circle.yml
@@ -287,27 +287,10 @@ jobs:
             codespell --quiet-level=4 -I .codespell-whitelist
       - install_wabt
       - run:
-          name: "Check wat2wasm4cpp"
+          name: "Check wat2wasm4tests"
           command: |
             export PATH=$PATH:~/bin
-            ./wat2wasm4cpp.py test/unittests/api_test.cpp
-            ./wat2wasm4cpp.py test/unittests/capi_test.cpp
-            ./wat2wasm4cpp.py test/unittests/end_to_end_test.cpp
-            ./wat2wasm4cpp.py test/unittests/execute_call_test.cpp
-            ./wat2wasm4cpp.py test/unittests/execute_control_test.cpp
-            ./wat2wasm4cpp.py test/unittests/execute_floating_point_test.cpp
-            ./wat2wasm4cpp.py test/unittests/execute_invalid_test.cpp
-            ./wat2wasm4cpp.py test/unittests/execute_numeric_test.cpp
-            ./wat2wasm4cpp.py test/unittests/execute_test.cpp
-            ./wat2wasm4cpp.py test/unittests/instantiate_test.cpp
-            ./wat2wasm4cpp.py test/unittests/module_test.cpp
-            ./wat2wasm4cpp.py test/unittests/parser_test.cpp
-            ./wat2wasm4cpp.py test/unittests/validation_stack_test.cpp
-            ./wat2wasm4cpp.py test/unittests/validation_stack_type_test.cpp
-            ./wat2wasm4cpp.py test/unittests/validation_test.cpp
-            ./wat2wasm4cpp.py test/unittests/wasm_engine_test.cpp
-            ./wat2wasm4cpp.py test/spectests/spectests.cpp
-            ./wat2wasm4cpp.py test/testfloat/testfloat.cpp
+            find test -name '*.cpp' -print -exec ./wat2wasm4tests.py {} \;
             git diff --color --exit-code
       - build:
           configuration_name: "Tidy"
--- a/wat2wasm4tests.py
+++ b/wat2wasm4tests.py
@@ -1,6 +1,6 @@
 #!/usr/bin/python3
 
-"""wat2wasm4cpp
+"""wat2wasm4tests
 
 This script converts C++ or Rust comments containing WebAssembly text format (WAT)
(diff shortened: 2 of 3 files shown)


axic:
> However, it doesn't seem much useful, e.g. long hex strings are not broken down

Yes that is a bit annoying, but still better to have this automated.

circle.yml
```diff
@@ -287,27 +287,27 @@ jobs:
...
+            ./wat2wasm4tests.py test/unittests/instantiate_test.cpp
+            ./wat2wasm4tests.py test/unittests/module_test.cpp
+            ./wat2wasm4tests.py test/unittests/parser_test.cpp
+            ./wat2wasm4tests.py test/unittests/validation_stack_test.cpp
+            ./wat2wasm4tests.py test/unittests/validation_stack_type_test.cpp
+            ./wat2wasm4tests.py test/unittests/validation_test.cpp
+            ./wat2wasm4tests.py test/unittests/wasm_engine_test.cpp
+            ./wat2wasm4tests.py test/spectests/spectests.cpp
+            ./wat2wasm4tests.py test/testfloat/testfloat.cpp
             git diff --color --exit-code
```
axic, line 311: Should also add `bindings/rust/src/lib.rs` to the CI.

wat2wasm4tests.py
```diff
@@ -31,10 +31,16 @@
 WAT2WASM_TOOL = 'wat2wasm'
 WAT2WASM_DEFAULT_OPTIONS = ['--disable-saturating-float-to-int',
                             '--disable-sign-extension', '--disable-multi-value']
-FORMAT_TOOL = 'clang-format'
+CPP_FORMAT_TOOL = ['clang-format', '-i']
+RS_FORMAT_TOOL = ['rustfmt']
```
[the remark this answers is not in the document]
gumb0: It works https://app.circleci.com/pipelines/github/wasmx/fizzy/6030/workflows/f00e5270-7b86-447c-82e8-2ddddfaadba8/jobs/50042/parallel-runs/0/steps/0-106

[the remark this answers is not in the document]
gumb0: Actually it would ignore failures if rustfmt were not found, let me check it then.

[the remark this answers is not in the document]
gumb0: You were right, it didn't work. I fixed it by moving into `bindings-rust` job, where it's already installed (but then wabt needs to be installed there, too)

circle.yml
```diff
@@ -35,7 +35,7 @@ commands:
             else
               [[ $OSTYPE = darwin* ]] && os=macos || os=ubuntu
               cd /usr/local
-              curl -L https://github.com/WebAssembly/wabt/releases/download/1.0.20/wabt-1.0.20-$os.tar.gz | sudo tar xz --strip 1
+              curl -L https://github.com/WebAssembly/wabt/releases/download/1.0.20/wabt-1.0.20-$os.tar.gz | (sudo tar xz --strip 1 || tar xz --strip 1)
```
gumb0, line 38: This solves the problem of `sudo` not working in Rust docker (user is already root there), probably there are more elegant ways to solve it, but this works.

circle.yml
```diff
@@ -287,27 +287,27 @@ jobs:
...
-            ./wat2wasm4cpp.py test/unittests/instantiate_test.cpp
-            ./wat2wasm4cpp.py test/unittests/module_test.cpp
-            ./wat2wasm4cpp.py test/unittests/parser_test.cpp
-            ./wat2wasm4cpp.py test/unittests/validation_stack_test.cpp
-            ./wat2wasm4cpp.py test/unittests/validation_stack_type_test.cpp
-            ./wat2wasm4cpp.py test/unittests/validation_test.cpp
-            ./wat2wasm4cpp.py test/unittests/wasm_engine_test.cpp
-            ./wat2wasm4cpp.py test/spectests/spectests.cpp
-            ./wat2wasm4cpp.py test/testfloat/testfloat.cpp
+            ./wat2wasm4tests.py test/unittests/api_test.cpp
```
chfast, line 293: At this point you want want to pass all `_test.cpp` files to the tool using `find`.

circle.yml
```diff
@@ -287,27 +287,10 @@ jobs:
...
-            ./wat2wasm4cpp.py test/unittests/instantiate_test.cpp
-            ./wat2wasm4cpp.py test/unittests/module_test.cpp
-            ./wat2wasm4cpp.py test/unittests/parser_test.cpp
-            ./wat2wasm4cpp.py test/unittests/validation_stack_test.cpp
-            ./wat2wasm4cpp.py test/unittests/validation_stack_type_test.cpp
-            ./wat2wasm4cpp.py test/unittests/validation_test.cpp
-            ./wat2wasm4cpp.py test/unittests/wasm_engine_test.cpp
-            ./wat2wasm4cpp.py test/spectests/spectests.cpp
-            ./wat2wasm4cpp.py test/testfloat/testfloat.cpp
+            find test -name '*.cpp' -printf '%p\n' -exec ./wat2wasm4tests.py {} \;
```
chfast, line 293:
```suggestion
            find test -name '*.cpp' -print -exec ./wat2wasm4tests.py {} \;
```

chfast (approved):

circle.yml
```diff
@@ -586,6 +586,13 @@ jobs:
           command: |
             rustfmt --version
             cargo fmt --all -- --check
+      - install_wabt
+      - run:
+          name: Check wat2wasm4tests
+          command: |
+            rustfmt --version
+            ./wat2wasm4tests.py bindings/rust/src/lib.rs
```
axic, line 594: Why not use `find` here too (like for C++)?

axic (approved):

circle.yml
chfast: That would be good to.

circle.yml
axic: @gumb0 @chfast Since #706 also runs into this problem with `sudo pip3`, how about we just install sudo on the rust tasks and leave these helpers intact?

chfast: It might be better to migrate to [cimg/rust](https://circleci.com/developer/images/image/cimg/rust) where build are probably not executed as root and there may be more CI related benefits.

PR #705, opened 2021-01-28
rust: add typed value and execution result

axic:
Depends on #652.

### DIFF
--- a/bindings/rust/src/lib.rs
+++ b/bindings/rust/src/lib.rs
@@ -69,12 +69,11 @@ impl Module {
     }
 }
 
-// TODO: add a proper wrapper (enum type for Value, and Result<Value, Trap> for ExecutionResult)
+// TODO: add a proper wrapper (Result<Value, Trap> for ExecutionResult)
 
 /// A WebAssembly value of i32/i64/f32/f64.
 pub type Value = sys::FizzyValue;
 
-// NOTE: the union does not have i32
 impl Value {
     pub fn as_i32(&self) -> i32 {
         unsafe { self.i32 as i32 }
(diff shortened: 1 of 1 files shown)


bindings/rust/src/lib.rs
```diff
@@ -133,6 +133,75 @@ impl From<f64> for Value {
...
 
+/// A WebAssembly value i32/i64/f32/f64 with its type specified.
+pub enum TypedValue {
+    I32(i32),
+    U32(u32),
+    I64(i64),
+    U64(u64),
+    F32(f32),
+    F64(f64),
+    Void,
```
axic, line 144: I think we don't need Void as we should use `Option<>` instead.

bindings/rust/src/lib.rs
```diff
@@ -131,6 +131,84 @@ impl From<f64> for Value {
     }
 }
 
+/// A WebAssembly value i32/i64/f32/f64 with its type specified.
+pub enum TypedValue {
+    U32(u32),
```
axic, line 136: Since `sys::FizzyValueType` only supports an "unsigned" type for 32/64-bit integers, we are matching that here. The naming discrepancy is that I wanted to make clear U32/U64 matches the unsigned Rust type `u32`/`u64`.

bindings/rust/src/lib.rs
```diff
@@ -131,6 +131,84 @@ impl From<f64> for Value {
...
+            TypedValue::U32(_) => sys::FizzyValueTypeI32,
+            TypedValue::U64(_) => sys::FizzyValueTypeI64,
+            TypedValue::F32(_) => sys::FizzyValueTypeF32,
+            TypedValue::F64(_) => sys::FizzyValueTypeF64,
+        }
+    }
+
+    pub fn as_i32(&self) -> Option<i32> {
+        match self {
+            TypedValue::U32(v) => Some(*v as i32),
```
axic, line 154: This is still a question, whether we want to do this, or use Rust specific helpers for casting between u32 -> i32.

bindings/rust/src/lib.rs
```diff
@@ -151,6 +229,35 @@ impl ExecutionResult {
...
+    /// True if execution has resulted in a trap.
+    pub fn trapped(&self) -> bool {
+        self.result.trapped
+    }
+
+    /// The optional return value. Only a single return value is allowed in WebAssembly 1.0.
+    pub fn value(&self) -> Option<TypedValue> {
+        if self.result.has_value {
+            assert!(!self.result.trapped);
+            assert!(self.value_type != sys::FizzyValueTypeVoid);
```
axic, line 247: Can merge these assertions.

bindings/rust/src/lib.rs
chfast: I don't know Rust style, but seems Void will be useful in C/C++ in some cases.

bindings/rust/src/lib.rs
chfast: Do you mean the alternative being `x.as_u32() as i32`? I think this is fine then.

bindings/rust/src/lib.rs
chfast: These can stay as they are.

bindings/rust/src/lib.rs
```diff
@@ -414,7 +631,7 @@ mod tests {
         let result = result.unwrap();
         assert!(!result.trapped());
         assert!(result.value().is_some());
-        assert_eq!(result.value().unwrap().as_i32(), 42);
+        assert_eq!(result.value().unwrap().as_u32().unwrap(), 42);
```
axic, line 634: Hmm, perhaps could consider a TryInto trait implementation instead, maybe more canonical.

bindings/rust/src/lib.rs
axic: Using Result/Option is more natural than having a special enum value in this case, it seems.

bindings/rust/src/lib.rs
```diff
@@ -269,6 +380,113 @@ mod tests {
         assert_eq!(v.as_f64(), f64::MAX);
     }
 
+    #[test]
+    fn typed_value_conversion() {
```
chfast, line 384: You need tests for wrong type.

axic: Wrong type is a compilation error.

chfast: I mean the case when `.as_u32()` is `None`.

axic: I pushed that 10 minutes ago, just github is slow today showing up changes.

chfast (approved):

bindings/rust/src/lib.rs
```diff
@@ -151,6 +218,36 @@ impl ExecutionResult {
     }
 }
 
+/// The result of an execution.
+pub struct TypedExecutionResult {
+    result: sys::FizzyExecutionResult,
```
gumb0, line 223: Would this be more convenient as `ExecutionResult` type?

axic: No, because it hides the very details we are checking. Basically TypedExecutionResult and ExecutionResult is a bit of a duplication. I think in the future we may want to review this and perhaps get rid of the unsafe execution, once we have benchmarked the overheads.

gumb0 (approved):

PR #711, opened 2021-02-01
rust: Add missing test coverage

axic:


bindings/rust/src/lib.rs
```diff
@@ -46,9 +46,7 @@ impl Module {
     /// Create an instance of a module.
     // TODO: support imported functions
     pub fn instantiate(self) -> Result<Instance, ()> {
-        if self.0.is_null() {
-            return Err(());
-        }
+        assert!(!self.0.is_null());
```
axic, line 49: Having null here is an error of using the Rust API and not a proper error emitted by the C code.

bindings/rust/src/lib.rs
```diff
@@ -153,12 +151,6 @@ impl ExecutionResult {
     }
 }
 
-impl From<ExecutionResult> for sys::FizzyExecutionResult {
```
axic, line 156: This has not been used and probably will not be used in this form, rather as a `Box<>`. Will add it back when needed.

gumb0 (approved):

chfast (approved):

PR #712, opened 2021-02-01
test: Unit test for CResult matcher

gumb0:
This improves coverage in `asserts.hpp`

### DIFF
--- a/test/unittests/test_utils_test.cpp
+++ b/test/unittests/test_utils_test.cpp
@@ -159,6 +159,53 @@ TEST(test_utils, result_value_matcher)
     EXPECT_THAT(TypedExecutionResult(Value{1_u64}, ValType::i64), Not(Result(uint8_t{1})));
 }
 
+TEST(test_utils, cresult_value_matcher)
+{
+    using testing::Not;
+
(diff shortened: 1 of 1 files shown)


test/unittests/test_utils_test.cpp
```diff
@@ -159,6 +159,53 @@ TEST(test_utils, result_value_matcher)
...
+    FizzyExecutionResult result_void{false, false, FizzyValue{}};
+    EXPECT_THAT(result_void, CResult());
+    EXPECT_THAT(result_void, Not(CTraps()));
+
+    // Each result type is checked only against values with equal type, because for other types
+    // CResult matcher would access non-active union members of result's value, which presents UB.
+
+    EXPECT_THAT(trap, Not(CResult(0_u32)));
+    EXPECT_THAT(result_void, Not(CResult(0_u32)));
+    FizzyValue value_i32{0};
```
axic, line 179: Would we be better not using 0 but actual values in these tests?

gumb0: Not sure, but maybe we would, I'll change it.

chfast (approved):

axic (approved):

PR #713, opened 2021-02-01
WASI testing

### LINKED ISSUE #708: Split up wasi implementation into unit testable pieces
Suggested [here](https://github.com/wasmx/fizzy/pull/329#issuecomment-701475005). It would be possible to place the wasi implementation into a new namespace (e.g. `fizzy::wasi`), in a separate file, and have it unit tested. Then the frontend (`fizzy-wasi`) would use those features and retain the integration testing (the cli tests) it has.

chfast:
Closes #708 

TODO
- [x] Separate wasm file loading from execution.
- [x] Consider separate unittest executable: `fizzy-wasi-unittests`. This relaxes `WASI` and `TESTING` CMake options dependencies. (resolution: It is fine for now; less CMake option combinations is good for development and nobody reported issues with compiling libuvwasi so far; the CMake option can always be added later).
- [x] Consider different location for `fizzy::wasi` library (resolution: it is fine where it is).
- [x] ~~Create interface for WASI implementation / uvwasi. This allows mocking the implementation.~~ (To be done in a separate PR later)

### DIFF
--- a/tools/wasi/CMakeLists.txt
+++ b/tools/wasi/CMakeLists.txt
@@ -6,9 +6,8 @@ include(ProjectUVWASI)
 
 add_library(wasi)
 add_library(fizzy::wasi ALIAS wasi)
-target_compile_features(wasi PUBLIC cxx_std_17)
 target_include_directories(wasi PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
-target_link_libraries(wasi PRIVATE fizzy::fizzy-internal uvwasi::uvwasi)
+target_link_libraries(wasi PUBLIC fizzy::fizzy-internal PRIVATE uvwasi::uvwasi)
 target_sources(wasi PRIVATE wasi.cpp wasi.hpp)
 
 add_executable(fizzy-wasi)
--- a/tools/wasi/CMakeLists.txt
+++ b/tools/wasi/CMakeLists.txt
@@ -1,14 +1,12 @@
 # Fizzy: A fast WebAssembly interpreter
-# Copyright 2019-2020 The Fizzy Authors.
+# Copyright 2020 The Fizzy Authors.
 # SPDX-License-Identifier: Apache-2.0
 
 include(ProjectUVWASI)
 
-set(fizzy_include_dir ${PROJECT_SOURCE_DIR}/lib/fizzy)
-
-add_executable(fizzy-wasi wasi.cpp)
+add_executable(fizzy-wasi)
 target_link_libraries(fizzy-wasi PRIVATE fizzy::fizzy-internal uvwasi::uvwasi)
-target_include_directories(fizzy-wasi PRIVATE ${fizzy_include_dir})
+target_sources(fizzy-wasi PRIVATE main.cpp wasi.cpp wasi.hpp)
 
 if(UNIX AND NOT APPLE)
     # For libstdc++ up to version 8 (included) this is needed for proper <filesystem> support.
--- a/tools/wasi/CMakeLists.txt
+++ b/tools/wasi/CMakeLists.txt
@@ -4,9 +4,16 @@
 
 include(ProjectUVWASI)
 
+add_library(wasi)
+add_library(fizzy::wasi ALIAS wasi)
+target_compile_features(wasi PUBLIC cxx_std_17)
+target_include_directories(wasi PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
+target_link_libraries(wasi PRIVATE fizzy::fizzy-internal uvwasi::uvwasi)
+target_sources(wasi PRIVATE wasi.cpp wasi.hpp)
+
 add_executable(fizzy-wasi)
-target_link_libraries(fizzy-wasi PRIVATE fizzy::fizzy-internal uvwasi::uvwasi)
-target_sources(fizzy-wasi PRIVATE main.cpp wasi.cpp wasi.hpp)
+target_link_libraries(fizzy-wasi PRIVATE fizzy::wasi)
+target_sources(fizzy-wasi PRIVATE main.cpp)
 
 if(UNIX AND NOT APPLE)
     # For libstdc++ up to version 8 (included) this is needed for proper <filesystem> support.
--- a/tools/wasi/wasi.cpp
+++ b/tools/wasi/wasi.cpp
@@ -10,7 +10,7 @@
 #include <cassert>
 #include <filesystem>
 #include <fstream>
-#include <iostream>
+#include <ostream>
 
 namespace fizzy::wasi
 {
@@ -99,7 +99,7 @@ ExecutionResult environ_sizes_get(Instance& instance, const Value* args, int)
 }
 }  // namespace
 
-bool run(int argc, const char** argv)
+bool run(int argc, const char** argv, std::ostream& err)
 {
     constexpr auto ns = "wasi_snapshot_preview1";
     const std::vector<ImportedFunction> wasi_functions = {
@@ -126,24 +126,24 @@ bool run(int argc, const char** argv)
         nullptr,  // NOTE: no special allocator
     };
 
-    const uvwasi_errno_t err = uvwasi_init(&state, &options);
-    if (err != UVWASI_ESUCCESS)
+    const uvwasi_errno_t uvwasi_err = uvwasi_init(&state, &options);
+    if (uvwasi_err != UVWASI_ESUCCESS)
     {
-        std::cerr << "Failed to initialise UVWASI: " << uvwasi_embedder_err_code_to_string(err)
-                  << "\n";
+        err << "Failed to initialise UVWASI: " << uvwasi_embedder_err_code_to_string(uvwasi_err)
+            << "\n";
         return false;
     }
 
     if (!std::filesystem::is_regular_file(argv[0]))
     {
-        std::cerr << "File does not exists or is irregular: " << argv[0] << "\n";
+        err << "File does not exists or is irregular: " << argv[0] << "\n";
         return false;
     }
 
     std::ifstream wasm_file{argv[0]};
     if (!wasm_file)
     {
-        std::cerr << "Failed to open file: " << argv[0] << "\n";
+        err << "Failed to open file: " << argv[0] << "\n";
         return false;
     }
     const auto wasm_binary =
@@ -158,7 +158,7 @@ bool run(int argc, const char** argv)
     const auto start_function = find_exported_function(*instance->module, "_start");
     if (!start_function.has_value())
     {
-        std::cerr << "File is not WASI compatible (_start not found)\n";
+        err << "File is not WASI compatible (_start not found)\n";
         return false;
     }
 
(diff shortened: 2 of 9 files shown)


tools/wasi/CMakeLists.txt
```diff
@@ -1,14 +1,21 @@
...
-# Copyright 2019-2020 The Fizzy Authors.
+# Copyright 2020 The Fizzy Authors.
 # SPDX-License-Identifier: Apache-2.0
 
 include(ProjectUVWASI)
 
-set(fizzy_include_dir ${PROJECT_SOURCE_DIR}/lib/fizzy)
+add_library(wasi)
+add_library(fizzy::wasi ALIAS wasi)
+target_compile_features(wasi PUBLIC cxx_std_17)
```
gumb0, line 9: Not needed, `fizzy:fizzy` has it.

chfast: The `fizzy::fizzy-internal` is private dependency: the `fizzy::wasi` will be built with `-std=c++17`, but users of `fizzy::wasi` (e.g. `fizzy-wasi` tool) will not have this requirement.

Here it make sense to declare `fizzy::wasi` with C++17 because it uses a C++17 feature in the main header.

tools/wasi/wasi.cpp
```diff
@@ -82,40 +85,56 @@ fizzy::ExecutionResult wasi_fd_prestat_get(fizzy::Instance& instance, const fizz
...
     // TODO: implement properly (only returns 0 now)
     uvwasi_serdes_write_uint32_t(instance.memory->data(), environc, 0);
     uvwasi_serdes_write_uint32_t(instance.memory->data(), environ_buf_size, 0);
-    return fizzy::Value{uint32_t{UVWASI_ESUCCESS}};
+    return Value{uint32_t{UVWASI_ESUCCESS}};
 }
+}  // namespace
 
-bool run(int argc, const char** argv)
+std::optional<bytes> load_wasm_binary(const std::filesystem::path& path, std::ostream& err)
```
chfast, line 102: This function has pretty bad API, but I'm not sure how exactly improve it right now.
Also current error reporting is in some middle ground: it provides some additional but not complete set of diagnostics. And finally, the `std::ifstream` can sometimes be in "fail" state but it can also throw an exception in other cases...

chfast (commented):
Update: all necessary changes are done to allow writing "top-down" unit tests - you can provide input as wat2wasm comment and check result and error string. For full testability we need WASI interface abstraction, but this should be done in a separate PR.

Without WASI abstraction we still have some integration tests in CTest. These are also nice so we know that the basic features of libuvwasi are working fine.

Finally, some unit tests truly accessing filesystem can be done in GTest provided it has some helpers creating/deleting temporary files.

Also see updated description.

chfast:
Lastly, this has triggered a know bug in stdlibc++'s `std::ostringstream` implementation: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=97415.

I think we should disable GCC sanitizer for GCC 10 and wait for GCC 11 release (or use experimental Debian package).

tools/wasi/CMakeLists.txt
chfast: This is not relevant any more because `fizzy::wasi` now links `fizzy::fizzy-internal` publicly.

chfast:
> I think we should disable GCC sanitizer for GCC 10 and wait for GCC 11 release (or use experimental Debian package).

One ASan runtime option is disabled for current GCC runs.

tools/wasi/wasi.cpp
gumb0: Not sure how `filesystem::path` works but maybe this function should take a string (provided by user) and create `filesytem::path` inside, handling the errors like invalid symbols in path.

tools/wasi/wasi.hpp
```diff
@@ -0,0 +1,25 @@
...
+
+/// Executes WASI main function with given CLI arguments.
+///
+/// @param  argc    Number of CLI arguments.
+/// @param  argv    Array of CLI arguments.
+/// @param  err     Error output stream.
+/// @return         False in case of error.
+///
+/// @todo Make noexcept.
+bool run(int argc, const char* argv[], std::ostream& err);
```
gumb0, line 22: Please add a note that path to a binary is expected as first argument. 

Also maybe it should be named `load_and_run`.

tools/wasi/wasi.cpp
```diff
@@ -129,85 +148,57 @@ bool run(int argc, const char** argv)
         nullptr,  // NOTE: no special allocator
     };
 
-    const uvwasi_errno_t err = uvwasi_init(&state, &options);
-    if (err != UVWASI_ESUCCESS)
+    const uvwasi_errno_t e = uvwasi_init(&state, &options);
```
gumb0, line 151:
```suggestion
    const uvwasi_errno_t uvwasi_err = uvwasi_init(&state, &options);
```

tools/wasi/CMakeLists.txt
gumb0: > This is not relevant any more because `fizzy::wasi` now links `fizzy::fizzy-internal` publicly.

Why does it?

chfast: Because now Fizzy is included in the `wasi.hpp`.

tools/wasi/wasi.cpp
chfast: I considered the following options and used the one which is the easiest to use at the moment.
1. `const char*` - only literals and `argv[]` work really,
2. `const std::string&` - specific string type used, conversion from `const char*` is needed.
3. `std::string_view` - does not work with `std::ifstream` - conversion to `std::string` is needed.
4. `std::filesystem::path` - works with everything, emphasize that this is path.

I agree that invalid path format should be also handled inside. In the some time, there are other error cases that are not handled explicitly.

So switch to some other type?

gumb0: (I imagined it like taking `string_view` then constucting `path` out of it, and then `ifstream` out of this `path`)

I'm fine to leave it like this for now. At some point exceptions thrown by `path` constructor will need to be handled (when we address `@todo Make noexcept.`)

chfast: Done.

gumb0 (approved):
Looks good!

tools/wasi/wasi.cpp
```diff
@@ -150,15 +145,15 @@ bool run(int argc, const char** argv)
...
     }
     const auto wasm_binary =
-        fizzy::bytes(std::istreambuf_iterator<char>{wasm_file}, std::istreambuf_iterator<char>{});
+        bytes(std::istreambuf_iterator<char>{wasm_file}, std::istreambuf_iterator<char>{});
 
-    auto module = fizzy::parse(wasm_binary);
-    auto imports = fizzy::resolve_imported_functions(*module, wasi_functions);
-    auto instance = fizzy::instantiate(
-        std::move(module), std::move(imports), {}, {}, {}, fizzy::MemoryPagesValidationLimit);
+    auto module = parse(wasm_binary);
```
axic, line 170: I actually liked having the explicit fizzy prefix on things here, but I'm fine either way.

tools/wasi/wasi.cpp
```diff
@@ -1,7 +1,8 @@
 // Fizzy: A fast WebAssembly interpreter
-// Copyright 2019-2020 The Fizzy Authors.
+// Copyright 2020 The Fizzy Authors.
 // SPDX-License-Identifier: Apache-2.0
 
+#include "wasi.hpp"
```
axic, line 5: Is this lint-safe ordering?

tools/wasi/wasi.cpp
```diff
@@ -1,7 +1,8 @@
...
 #include "execute.hpp"
 #include "limits.hpp"
 #include "parser.hpp"
```
axic, line 8: Is the `<iostream>` header still needed here (since `std::cerr` is moved to `main.cpp`)?

tools/wasi/wasi.cpp
```diff
@@ -99,7 +99,7 @@ ExecutionResult environ_sizes_get(Instance& instance, const Value* args, int)
 }
 }  // namespace
 
-bool run(int argc, const char** argv)
+bool run(int argc, const char** argv, std::ostream& err)
```
axic: I know you say to make this noexcept, but if not noexcept we could consider just outputting error text as part of an exception? Or we want to pass stdout/stdin here too?

tools/wasi/CMakeLists.txt
```diff
@@ -4,11 +4,18 @@
...
+target_sources(wasi PRIVATE
+    wasi.cpp wasi.hpp)
+
 add_executable(fizzy-wasi)
-target_link_libraries(fizzy-wasi PRIVATE fizzy::fizzy-internal uvwasi::uvwasi)
+target_link_libraries(fizzy-wasi PRIVATE fizzy::wasi)
 target_sources(fizzy-wasi PRIVATE
-    main.cpp
-    wasi.cpp wasi.hpp)
+    main.cpp)
```
axic, line 17: Why not fold this and the above into the same target_sources line?

tools/wasi/CMakeLists.txt
```diff
@@ -6,9 +6,8 @@ include(ProjectUVWASI)
 
 add_library(wasi)
 add_library(fizzy::wasi ALIAS wasi)
-target_compile_features(wasi PUBLIC cxx_std_17)
 target_include_directories(wasi PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
-target_link_libraries(wasi PRIVATE fizzy::fizzy-internal uvwasi::uvwasi)
+target_link_libraries(wasi PUBLIC fizzy::fizzy-internal PRIVATE uvwasi::uvwasi)
```
axic, line 10: Does this change below to the testing commit?

axic (approved):
You may want to address some of these comments, but I am also okay if not.

tools/wasi/wasi.cpp
chfast: It does not make sense when we are in `fizzy::wasi`.

tools/wasi/wasi.cpp
chfast: This is to be able to inspect the error messages, one step from using `std::cerr` implicitly inside. Further modifications are possible, but I'm not sure what's best.

tools/wasi/wasi.cpp
```diff
@@ -99,6 +99,39 @@ ExecutionResult environ_sizes_get(Instance& instance, const Value* args, int)
...
+        {
+            err << "Failed to open file: " << path << "\n";
+            return std::nullopt;
+        }
+        return bytes(std::istreambuf_iterator<char>{wasm_file}, std::istreambuf_iterator<char>{});
+    }
+    catch (...)
+    {
+        // Generic error message due to other exceptions.
+        err << "Failed to load: " << file << "\n";
```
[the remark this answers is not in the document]
chfast: We want to catch the also the exceptions from `path` constructor. @gumb0 pointed this out.

axic:
We forgot one thing: the `uvwasi_state` must be moved into some context and can't stay as a static global now that this is librarized.

chfast:
This will be done with mocking.

PR #714, opened 2021-02-02
Build and testing for arm64

### LINKED ISSUE #710: Create a CI job for little endian ARM
Ideally we would want to [support Apple Silicon](https://github.com/wasmx/fizzy/issues/709), it seems CircleCI may [not support it anytime soon](https://support.circleci.com/hc/en-us/articles/360047734172-Can-I-build-Apple-Silicon-Universal-apps-). Hence the next best option we have is to run a CI job with Linux and qemu for little endian ARM.

chfast:
Closes #710.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -445,6 +445,22 @@ jobs:
           min_time: "0.01"
       - spectest
 
+  arm64-linux:
+    executor: linux-gcc-latest
+    steps:
+      - run:
+          name: "Install arm64 toolchain"
+          command: |
+            sudo apt-get -q update
+            sudo apt-get -qy install --no-install-recommends g++-aarch64-linux-gnu qemu-user-static
+      - install_testfloat
+      - checkout
+      - build:
+          build_type: Debug
+          cmake_options: -DCMAKE_TOOLCHAIN_FILE=~/project/cmake/toolchains/arm64-linux.cmake
+      - test
+      # TODO: Enable spectests
+
   benchmark:
     machine:
       image: ubuntu-1604:201903-01
@@ -702,6 +718,9 @@ workflows:
       - spectest:
           requires:
             - coverage-clang
+      - arm64-linux:
+          requires:
+            - coverage-gcc
       - bindings-rust
       - bindings-rust-asan:
           requires:
--- a/test/testfloat/CMakeLists.txt
+++ b/test/testfloat/CMakeLists.txt
@@ -72,12 +72,14 @@ if(TESTFLOAT_GEN)
         "f64_trunc\;-rminMag f64_roundToInt"
         "f64_nearest\;-rnear_even f64_roundToInt")
 
+    # Convert ;-separated CROSSCOMPILING_EMULATOR arguments to space-separated command.
+    string(REPLACE ";" " " EMULATOR "${CMAKE_CROSSCOMPILING_EMULATOR}")
 
     macro(add_testfloat_test NAME TESTFLOAT_GEN_ARGS FIZZY_TESTFLOAT_ARGS)
         if(NOT ${NAME} IN_LIST IGNORE_LIST)
             add_test(
                 NAME fizzy/testfloat/${NAME}
-                COMMAND bash -c "set -o pipefail; ${TESTFLOAT_GEN} -seed ${TESTFLOAT_SEED} -level ${TESTFLOAT_LEVEL} ${TESTFLOAT_GEN_ARGS} | $<TARGET_FILE:fizzy-testfloat> ${FIZZY_TESTFLOAT_ARGS}"
+                COMMAND bash -c "set -o pipefail; ${TESTFLOAT_GEN} -seed ${TESTFLOAT_SEED} -level ${TESTFLOAT_LEVEL} ${TESTFLOAT_GEN_ARGS} | ${EMULATOR} $<TARGET_FILE:fizzy-testfloat> ${FIZZY_TESTFLOAT_ARGS}"
             )
             set_tests_properties(
                 fizzy/testfloat/${NAME}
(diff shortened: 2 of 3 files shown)


test/testfloat/CMakeLists.txt
```diff
@@ -26,6 +26,50 @@ if(TESTFLOAT_GEN)
     if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
         # TODO: Clang produces -0 for input 0, see https://bugs.llvm.org/show_bug.cgi?id=47393.
         list(APPEND IGNORE_LIST ui64_to_f64/min)
+    elseif(CMAKE_CROSSCOMPILING AND CMAKE_SYSTEM_PROCESSOR STREQUAL arm64)
+        # These tests fails on arm64 emulation. TODO: Check if NaNs are the reason.
```
axic, line 30: @chfast you mentioned these tests are more strict than wasm requires them, but okay on x86. Perhaps mention that here so it is not "misleading" that arm is just failing expected things.

test/testfloat/CMakeLists.txt
```diff
@@ -124,10 +125,10 @@ endif()
 
 function(add_testfloat_selftest NAME ARGS INPUT EXPECTED_ERROR)
     if(INPUT)
-        set(command sh -c "echo ${INPUT} | $<TARGET_FILE:fizzy-testfloat> ${ARGS}")
+        set(command sh -c "echo ${INPUT} | ${EMULATOR} $<TARGET_FILE:fizzy-testfloat> ${ARGS}")
     else()
         separate_arguments(ARGS)
-        set(command $<TARGET_FILE:fizzy-testfloat> ${ARGS})
+        set(command ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:fizzy-testfloat> ${ARGS})
```
axic, line 175: Why not `EMULATOR` here?

circle.yml
```diff
@@ -702,6 +718,9 @@ workflows:
       - spectest:
           requires:
             - coverage-clang
+      - arm64:
+          requires:
+            - coverage-gcc
```
axic, line 723: Why coverage-gcc? That is the only gcc job we have?

test/testfloat/CMakeLists.txt
chfast: I disabled NaN checking for these.

test/testfloat/CMakeLists.txt
chfast: It depends if arguments must be separated with `;` or not.

circle.yml
chfast: Arbitrary ordering. Both use GCC compiler.

test/testfloat/CMakeLists.txt
```diff
@@ -22,63 +22,72 @@ if(TESTFLOAT_GEN)
...
         "f32_trunc\;-rminMag f32_roundToInt"
         "f32_nearest\;-rnear_even f32_roundToInt"
         "f64_ceil\;-rmax f64_roundToInt"
         "f64_floor\;-rmin f64_roundToInt"
         "f64_trunc\;-rminMag f64_roundToInt"
-        "f64_nearest\;-rnear_even f64_roundToInt"
-    )
+        "f64_nearest\;-rnear_even f64_roundToInt")
 
+    string(REPLACE ";" " " EMULATOR "${CMAKE_CROSSCOMPILING_EMULATOR}")
```
gumb0, line 84: Can you explain why is this replacing needed?

chfast: Explained.

test/testfloat/CMakeLists.txt
```diff
@@ -22,10 +22,19 @@ if(TESTFLOAT_GEN)
     set(TESTFLOAT_SEED ${TESTFLOAT_SEED} CACHE INTERNAL "testfloat_get seed")  # Save in cache for lookup.
     message(VERBOSE "testfloat_gen seed: ${TESTFLOAT_SEED}")
 
+    # List of test names to ignore.
     set(IGNORE_LIST)
+
+    # Ignore NaN values in _some_ tests. The NaN checks in testfloat are move restrictive
```
gumb0, line 28:
```suggestion
    # Ignore NaN values in _some_ tests. The NaN checks in testfloat are more restrictive
```

test/testfloat/CMakeLists.txt
gumb0: But why can't it be with spaces from the beginning?

cmake/toolchains/arm64.cmake
```diff
@@ -0,0 +1,19 @@
...
+# Copyright 2020 The Fizzy Authors.
+# SPDX-License-Identifier: Apache-2.0
+
+# The toolchain file for arm64 cross-compilation.
+# It selects default version of GCC targeting aarch64-linux-gnu, tested with GCC 10.
+
+set(CMAKE_SYSTEM_PROCESSOR arm64)
+set(CMAKE_SYSTEM_NAME Linux)
+set(CMAKE_C_COMPILER aarch64-linux-gnu-gcc)
+set(CMAKE_CXX_COMPILER aarch64-linux-gnu-g++)
```
axic, line 11: Shouldn't this file be named `aarch64-linux` perhaps?

axic (approved):
Apart from those two minor comments I think it looks good.

cmake/toolchains/arm64.cmake
chfast: Can be `arm64-linux` if you prefer.

axic: I think it is more correct that way.

gumb0 (approved):

test/testfloat/CMakeLists.txt
chfast: It must be "a semicolon-separated list, then the first value is the command and remaining values are its arguments."
https://cmake.org/cmake/help/latest/prop_tgt/CROSSCOMPILING_EMULATOR.html

cmake/toolchains/arm64.cmake
chfast: For the other case, I prefer `arm64` over `aarch64`. It is more informative. Kinda similar issue to `x86-64` vs `amd64`.

PR #715, opened 2021-02-02
Rework testfloat test definitions in CMake

chfast:


axic (approved):
Don't fully understand the second commit, but the first seems to be okay and no test is lost.

PR #716, opened 2021-02-02
refactor: ExecuteFunction without std::function

gumb0:
Simpler alternative to #643.

### DIFF
--- a/lib/fizzy/capi.cpp
+++ b/lib/fizzy/capi.cpp
@@ -110,13 +110,16 @@ inline fizzy::ExecutionResult unwrap(const FizzyExecutionResult& result) noexcep
         return unwrap(result.value);
 }
 
-inline auto unwrap(FizzyExternalFn func, void* context) noexcept
+inline fizzy::ExecuteFunction unwrap(FizzyExternalFn c_function, void* c_host_context)
 {
-    return [func, context](fizzy::Instance& instance, const fizzy::Value* args,
-               int depth) noexcept -> fizzy::ExecutionResult {
-        const auto result = func(context, wrap(&instance), wrap(args), depth);
-        return unwrap(result);
-    };
+    static constexpr fizzy::HostFunctionPtr function =
+        [](std::any& host_context, fizzy::Instance& instance, const fizzy::Value* args, int depth) {
+            const auto [c_func, c_host_ctx] =
+                std::any_cast<std::pair<FizzyExternalFn, void*>>(host_context);
+            return unwrap(c_func(c_host_ctx, wrap(&instance), wrap(args), depth));
+        };
+
+    return {function, std::make_any<std::pair<FizzyExternalFn, void*>>(c_function, c_host_context)};
 }
 
 inline FizzyExternalFunction wrap(fizzy::ExternalFunction external_func)
--- a/lib/fizzy/capi.cpp
+++ b/lib/fizzy/capi.cpp
@@ -163,32 +163,34 @@ inline std::vector<fizzy::ExternalFunction> unwrap(
 
 inline fizzy::ImportedFunction unwrap(const FizzyImportedFunction& c_imported_func)
 {
-    fizzy::ImportedFunction imported_func;
-    imported_func.module = c_imported_func.module;
-    imported_func.name = c_imported_func.name;
-
     const auto& c_type = c_imported_func.external_function.type;
-    imported_func.inputs.resize(c_type.inputs_size);
-    fizzy::ValType (*unwrap_valtype_fn)(FizzyValueType value) = &unwrap;
-    std::transform(c_type.inputs, c_type.inputs + c_type.inputs_size, imported_func.inputs.begin(),
-        unwrap_valtype_fn);
-    imported_func.output = c_type.output == FizzyValueTypeVoid ?
-                               std::nullopt :
-                               std::make_optional(unwrap(c_type.output));
 
-    imported_func.function = unwrap(
+    std::vector<fizzy::ValType> inputs(c_type.inputs_size);
+    fizzy::ValType (*unwrap_valtype_fn)(FizzyValueType value) = &unwrap;
+    std::transform(
+        c_type.inputs, c_type.inputs + c_type.inputs_size, inputs.begin(), unwrap_valtype_fn);
+
+    const std::optional<fizzy::ValType> output(c_type.output == FizzyValueTypeVoid ?
+                                                   std::nullopt :
+                                                   std::make_optional(unwrap(c_type.output)));
+
+    auto function = unwrap(
         c_imported_func.external_function.function, c_imported_func.external_function.context);
 
-    return imported_func;
+    return {c_imported_func.module, c_imported_func.name, std::move(inputs), output,
+        std::move(function)};
 }
 
 inline std::vector<fizzy::ImportedFunction> unwrap(
     const FizzyImportedFunction* c_imported_functions, size_t imported_functions_size)
 {
-    std::vector<fizzy::ImportedFunction> imported_functions(imported_functions_size);
+    std::vector<fizzy::ImportedFunction> imported_functions;
+    imported_functions.reserve(imported_functions_size);
+
     fizzy::ImportedFunction (*unwrap_imported_func_fn)(const FizzyImportedFunction&) = &unwrap;
     std::transform(c_imported_functions, c_imported_functions + imported_functions_size,
-        imported_functions.begin(), unwrap_imported_func_fn);
+        std::back_inserter(imported_functions), unwrap_imported_func_fn);
+
     return imported_functions;
 }
 
--- a/lib/fizzy/instantiate.cpp
+++ b/lib/fizzy/instantiate.cpp
@@ -271,6 +271,7 @@ ExternalFunction find_imported_function(const std::string& module, const std::st
                                 " output type doesn't match imported function in module"};
     }
 
+    // Note: it->function is copied here.
     return {it->function, module_func_type};
 }
 
@@ -310,6 +311,14 @@ std::optional<uint32_t> find_export(const Module& module, ExternalKind kind, std
 }
 
 }  // namespace
+
+ExecutionResult ExecuteFunction::operator()(Instance& instance, const Value* args, int depth)
+{
+    if (m_instance)
+        return execute(*m_instance, m_func_idx, args, depth);
+    else
+        return m_host_function(m_host_context, instance, args, depth);
+}
 
 std::unique_ptr<Instance> instantiate(std::unique_ptr<const Module> module,
     std::vector<ExternalFunction> imported_functions, std::vector<ExternalTable> imported_tables,
@@ -487,11 +496,8 @@ std::optional<ExternalFunction> find_exported_function(Instance& instance, std::
         return std::nullopt;
 
     const auto idx = *opt_index;
-    auto func = [idx, &instance](fizzy::Instance&, const Value* args, int depth) {
-        return execute(instance, idx, args, depth);
-    };
-
-    return ExternalFunction{std::move(func), instance.module->get_function_type(idx)};
+    return ExternalFunction{
+        ExecuteFunction(instance, idx), instance.module->get_function_type(idx)};
 }
 
 std::optional<ExternalGlobal> find_exported_global(Instance& instance, std::string_view name)
--- a/lib/fizzy/instantiate.hpp
+++ b/lib/fizzy/instantiate.hpp
@@ -46,9 +46,6 @@ class ExecuteFunction
     std::any m_host_context;
 
 public:
-    /// Default function constructor.
-    ExecuteFunction() noexcept = default;
-
     // WebAssembly function constructor.
     ExecuteFunction(Instance& instance, FuncIdx func_idx) noexcept
       : m_instance{&instance}, m_func_idx{func_idx}
--- a/lib/fizzy/instantiate.hpp
+++ b/lib/fizzy/instantiate.hpp
@@ -10,6 +10,7 @@
 #include "module.hpp"
 #include "types.hpp"
 #include "value.hpp"
+#include <any>
 #include <cstdint>
 #include <functional>
 #include <memory>
@@ -21,23 +22,70 @@ namespace fizzy
 struct ExecutionResult;
 struct Instance;
 
-/// Function representing WebAssembly or host function execution.
-using execute_function = std::function<ExecutionResult(Instance&, const Value*, int depth)>;
+/// Function pointer to the execution function.
+using HostFunctionPtr = ExecutionResult (*)(
+    std::any& host_context, Instance&, const Value* args, int depth);
+
+/// Function representing either WebAssembly or host function execution.
+class ExecuteFunction
+{
+    /// Pointer to WebAssembly function instance.
+    /// Equals nullptr in case this ExecuteFunction represents host function.
+    Instance* m_instance = nullptr;
+
+    /// Index of WebAssembly function.
+    /// Equals 0 in case this ExecuteFunction represents host function.
+    FuncIdx m_func_idx = 0;
+
+    /// Pointer to a host function.
+    /// Equals nullptr in case this ExecuteFunction represents WebAssembly function.
+    HostFunctionPtr m_host_function = nullptr;
+
+    /// Opaque context of host function execution, which is passed to it as host_context parameter.
+    /// Doesn't have value in case this ExecuteFunction represents WebAssembly function.
+    std::any m_host_context;
+
+public:
+    /// Default function constructor.
+    ExecuteFunction() noexcept = default;
+
+    // WebAssembly function constructor.
+    ExecuteFunction(Instance& instance, FuncIdx func_idx) noexcept
+      : m_instance{&instance}, m_func_idx{func_idx}
+    {}
+
+    /// Host function constructor without context.
+    /// The function will always be called with empty host_context argument.
+    ExecuteFunction(HostFunctionPtr f) noexcept : m_host_function{f} {}
+
+    /// Host function constructor with context.
+    /// The function will be called with a reference to @a host_context.
+    /// Copies of the function will have their own copy of @a host_context.
+    ExecuteFunction(HostFunctionPtr f, std::any host_context)
+      : m_host_function{f}, m_host_context{std::move(host_context)}
+    {}
+
+    /// Function call operator.
+    ExecutionResult operator()(Instance& instance, const Value* args, int depth);
+
+    /// Function pointer stored inside this object.
+    HostFunctionPtr get_host_function() const noexcept { return m_host_function; }
+};
 
 /// Function with associated input/output types,
 /// used to represent both imported and exported functions.
 struct ExternalFunction
 {
-    execute_function function;
+    ExecuteFunction function;
     span<const ValType> input_types;
     span<const ValType> output_types;
 
-    ExternalFunction(execute_function _function, span<const ValType> _input_types,
+    ExternalFunction(ExecuteFunction _function, span<const ValType> _input_types,
         span<const ValType> _output_types)
       : function(std::move(_function)), input_types(_input_types), output_types(_output_types)
     {}
 
-    ExternalFunction(execute_function _function, const FuncType& type)
+    ExternalFunction(ExecuteFunction _function, const FuncType& type)
       : function(std::move(_function)), input_types(type.inputs), output_types(type.outputs)
     {}
 };
@@ -154,7 +202,7 @@ struct ImportedFunction
     std::optional<ValType> output;
 
     /// Function object, which will be called to execute the function.
-    execute_function function;
+    ExecuteFunction function;
 };
 
 /// Create vector of fizzy::ExternalFunction ready to be passed to instantiate().
--- a/test/unittests/api_test.cpp
+++ b/test/unittests/api_test.cpp
@@ -217,6 +217,40 @@ TEST(api, resolve_imported_function_duplicate)
     EXPECT_THAT(execute(*instance, 1, {0_u32}), Result(42));
 }
 
+TEST(api, resolve_imported_function_duplicate_with_context)
+{
+    /* wat2wasm
+      (func (import "mod1" "foo1"))
+      (func (import "mod1" "foo1"))
+    */
+    const auto wasm = from_hex(
+        "0061736d01000000010401600000021902046d6f643104666f6f310000046d6f643104666f6f310000");
+    const auto module = parse(wasm);
+
+    constexpr auto host_func = [](std::any& context, Instance&, const Value*, int) {
+        auto* counter = std::any_cast<int*>(context);
+        ++(*counter);
+        return Void;
+    };
+
+    int counter = 0;
+    auto context = std::make_any<int*>(&counter);
+
+    std::vector<ImportedFunction> imported_functions = {
+        {"mod1", "foo1", {}, std::nullopt, {host_func, std::move(context)}},
+    };
+
+    const auto external_functions = resolve_imported_functions(*module, imported_functions);
+
+    EXPECT_EQ(external_functions.size(), 2);
+
+    auto instance = instantiate(*module, external_functions, {}, {}, {});
+
+    EXPECT_THAT(execute(*instance, 0, nullptr), Result());
+    EXPECT_THAT(execute(*instance, 1, nullptr), Result());
+    EXPECT_EQ(counter, 2);
+}
+
 TEST(api, resolve_imported_globals)
 {
     /* wat2wasm
--- a/test/unittests/api_test.cpp
+++ b/test/unittests/api_test.cpp
@@ -15,12 +15,17 @@ using namespace fizzy::test;
 
 namespace
 {
-auto function_returning_value(Value value) noexcept
+ExecuteFunction function_returning_value(Value value)
 {
-    return [value](Instance&, const Value*, int) { return value; };
+    static constexpr HostFunctionPtr func = [](std::any& host_context, Instance&, const Value*,
+                                                int) -> ExecutionResult {
+        return std::any_cast<Value>(host_context);
+    };
+
+    return {func, std::make_any<Value>(value)};
 }
 
-ExecutionResult function_returning_void(Instance&, const Value*, int) noexcept
+ExecutionResult function_returning_void(std::any&, Instance&, const Value*, int) noexcept
 {
     return Void;
 }
@@ -427,11 +432,13 @@ TEST(api, find_exported_function)
         "0061736d010000000105016000017f021001087370656374657374036261720000040401700000050401010102"
         "0606017f0041000b07170403666f6f000001670300037461620100036d656d0200");
 
-    auto bar = [](Instance&, const Value*, int) { return Value{42}; };
+    constexpr auto bar = [](std::any&, Instance&, const Value*, int) -> ExecutionResult {
+        return Value{42};
+    };
     const auto bar_type = FuncType{{}, {ValType::i32}};
 
     auto instance_reexported_function =
-        instantiate(parse(wasm_reexported_function), {{bar, bar_type.inputs, bar_type.outputs}});
+        instantiate(parse(wasm_reexported_function), {{{bar}, bar_type}});
 
     auto opt_reexported_function = find_exported_function(*instance_reexported_function, "foo");
     ASSERT_TRUE(opt_reexported_function);
--- a/test/unittests/execute_call_test.cpp
+++ b/test/unittests/execute_call_test.cpp
@@ -340,10 +340,12 @@ TEST(execute_call, imported_function_call)
 
     const auto module = parse(wasm);
 
-    constexpr auto host_foo = [](Instance&, const Value*, int) { return Value{42}; };
-    const auto host_foo_type = module->typesec[0];
+    constexpr auto host_foo = [](std::any&, Instance&, const Value*, int) -> ExecutionResult {
+        return Value{42};
+    };
+    const auto& host_foo_type = module->typesec[0];
 
-    auto instance = instantiate(*module, {{host_foo, host_foo_type}});
+    auto instance = instantiate(*module, {{{host_foo}, host_foo_type}});
 
     EXPECT_THAT(execute(*instance, 1, {}), Result(42));
 }
@@ -361,14 +363,14 @@ TEST(execute_call, imported_function_call_void)
 
     const auto module = parse(wasm);
 
-    bool called = false;
-    const auto host_foo = [&called](Instance&, const Value*, int) {
+    static bool called = false;
+    constexpr auto host_foo = [](std::any&, Instance&, const Value*, int) {
         called = true;
         return Void;
     };
     const auto host_foo_type = module->typesec[0];
 
-    auto instance = instantiate(*module, {{host_foo, host_foo_type}});
+    auto instance = instantiate(*module, {{{host_foo}, host_foo_type}});
     execute(*instance, 1, {});
     EXPECT_TRUE(called);
 }
@@ -390,10 +392,12 @@ TEST(execute_call, imported_function_call_with_arguments)
 
     const auto module = parse(wasm);
 
-    auto host_foo = [](Instance&, const Value* args, int) { return Value{args[0].i32 * 2}; };
+    constexpr auto host_foo = [](std::any&, Instance&, const Value* args, int) -> ExecutionResult {
+        return Value{args[0].i32 * 2};
+    };
     const auto host_foo_type = module->typesec[0];
 
-    auto instance = instantiate(*module, {{host_foo, host_foo_type}});
+    auto instance = instantiate(*module, {{{host_foo}, host_foo_type}});
 
     EXPECT_THAT(execute(*instance, 1, {20}), Result(42));
 }
@@ -432,14 +436,17 @@ TEST(execute_call, imported_functions_call_indirect)
     ASSERT_EQ(module->importsec.size(), 2);
     ASSERT_EQ(module->codesec.size(), 2);
 
-    constexpr auto sqr = [](Instance&, const Value* args, int) {
-        return Value{uint64_t{args[0].i32} * uint64_t{args[0].i32}};
+    constexpr auto sqr = [](std::any&, Instance&, const Value* args, int) -> ExecutionResult {
+        const auto x = args[0].i32;
+        return Value{uint64_t{x} * uint64_t{x}};
     };
-    constexpr auto isqrt = [](Instance&, const Value* args, int) {
-        return Value{(11 + uint64_t{args[0].i32} / 11) / 2};
+    constexpr auto isqrt = [](std::any&, Instance&, const Value* args, int) -> ExecutionResult {
+        const auto x = args[0].i32;
+        return Value{(11 + uint64_t{x} / 11) / 2};
     };
 
-    auto instance = instantiate(*module, {{sqr, module->typesec[0]}, {isqrt, module->typesec[0]}});
+    auto instance =
+        instantiate(*module, {{{sqr}, module->typesec[0]}, {{isqrt}, module->typesec[0]}});
     EXPECT_THAT(execute(*instance, 3, {0, 10}), Result(20));  // double(10)
     EXPECT_THAT(execute(*instance, 3, {1, 9}), Result(81));   // sqr(9)
     EXPECT_THAT(execute(*instance, 3, {2, 50}), Result(7));   // isqrt(50)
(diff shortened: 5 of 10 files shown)


lib/fizzy/execute.cpp
```diff
@@ -840,7 +840,7 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, const Value* args,
                     throw std::bad_alloc();
                 memory->resize(new_pages * PageSize);
             }
-            catch (std::bad_alloc const&)
+            catch (...)
```
gumb0, line 843: Clang-tidy complains it can't be `noexcept` without this. I guess it assumes that `resize` call above can throw any exception.

axic: Isn't it that we are not catching `std::length_error` of https://en.cppreference.com/w/cpp/string/basic_string/resize ?

gumb0: Looks like catching `bad_alloc` and `length_error` is enough to silence the warning.

Logically `std::bad_array_new_length` (which would mean trying to allocate more than max size_t memory) can't happen both because one element of string is 1 byte and number of elements is given as size_t, but also because `std::length_error` is supposed to be thrown if new length is more than `string::max_size()`.

gumb0: It think it's correct only to return `-1` because `memory.grow` can't trap according to the spec.

axic: Btw, what does `resize` do if the old and new size equal? Does it have an early exit?

axic: Asking because querying the current size can be done via `memory.grow 0` in wasm, so that may be a "common" pattern.

gumb0: It would reallocate only if new size is greater than `capacity()`, so yeah, should return quickly.

test/unittests/execute_call_test.cpp
```diff
@@ -694,14 +702,13 @@ TEST(execute_call, imported_function_from_another_module_max_depth)
     const auto func_idx = fizzy::find_exported_function(*instance1->module, "f");
     ASSERT_TRUE(func_idx.has_value());
 
-    auto sub = [&instance1, func_idx](Instance&, const Value* args, int depth) -> ExecutionResult {
-        return fizzy::execute(*instance1, *func_idx, args, depth + 1);
-    };
-
-    auto instance2 = instantiate(std::move(module2), {{sub, instance1->module->typesec[0]}});
+    auto instance2 =
+        instantiate(std::move(module2), {{{*instance1, *func_idx}, instance1->module->typesec[0]}});
```
gumb0, line 706: This actually changes what the test is doing (importing wasm function instead of host function), and hence expected result. I should change it back to using lambda.

chfast:
I pushed a proposed implementation simplification.
There should be also an option to eliminate shared_ptr, but I will leave it for some other time - this requires wrapping C with C++ not the other way around.

gumb0:
> I pushed a proposed implementation simplification.
> There should be also an option to eliminate shared_ptr, but I will leave it for some other time - this requires wrapping C with C++ not the other way around.

Have you seen https://github.com/wasmx/fizzy/pull/643#issuecomment-737927318? 
shared_ptr is got rid of there, suggestion are welcome on how to simplify it.

chfast:
Ok, but is it still worth to add memory allocation for wasm function case? Even if temporarily?

gumb0:
> Ok, but is it still worth to add memory allocation for wasm function case? Even if temporarily?

Yeah, agree it might be a good optimization here.

test/unittests/execute_call_test.cpp
```diff
@@ -615,13 +620,14 @@ TEST(execute_call, call_initial_depth)
     const auto wasm = from_hex("0061736d01000000010401600000020b01036d6f6403666f6f0000");
 
     auto module = parse(wasm);
-    auto host_foo = [](Instance& /*instance*/, const Value*, int depth) -> ExecutionResult {
+    constexpr auto host_foo = [](void*, Instance&, const Value*,
+                                  int depth) noexcept -> ExecutionResult {
```
chfast, line 624:
```suggestion
                                  int depth) noexcept {
```
You don't need to specify the return type in many places.

lib/fizzy/instantiate.hpp
```diff
@@ -21,23 +21,81 @@ namespace fizzy
...
+
+    // WebAssembly function constructor.
+    ExecuteFunction(Instance& instance, FuncIdx func_idx) noexcept
+      : m_instance{&instance}, m_func_idx{func_idx}
+    {}
+
+    /// Host function constructor without taking ownership of context.
+    /// Context lifetime is assumed to be managed by the caller.
+    ExecuteFunction(ExecuteFunctionPtr f, void* context = nullptr)
+      : m_host_function{f}, m_host_context{context, [](void*) noexcept {}}
```
gumb0, line 68: It might be not a great idea - I think this shared pointer will still have an overhead of reference counting. Maybe better solution would be to have a union of raw pointer and `shared_pointer` and use raw pointer here.

But also this contstructor is currently used only with `nullptr` contexr, I just added it at some point as a general version because it seemed it would be convenient for a user (still seems so). I think I'll leave this feature for a later PR.

gumb0:
It looks like the approach in the commit c678a1b858bff9b5937ddc76257d9e1896ef963d `Simplify context lifetime management` is not valid, because type is erased and and regular `shared_ptr` would delete `void*` then, which is UB (But none of sanitizers complained about this)

image
https://en.cppreference.com/w/cpp/language/delete
https://stackoverflow.com/a/12761979

lib/fizzy/instantiate.hpp
```diff
@@ -21,23 +21,79 @@ namespace fizzy
...
+    /// Opaque context of host function execution, which is passed to it as a context parameter.
+    /// `shared_ptr` is used because ExecuteFunction can be copied.
+    /// Empty in case this ExecuteFunction represents WebAssembly function.
+    std::shared_ptr<void> m_host_context;
+
+public:
+    /// Default function constructor.
+    ExecuteFunction() noexcept = default;
+    /// Copy constructor.
+    ExecuteFunction(const ExecuteFunction&) noexcept = default;
```
chfast, line 52: None of these 4 declarations are strictly needed. They will be added anyway if you don't declare a single one of them nor destructor.

gumb0: This is correct, I wanted to make them explicit, because they are used (not all of them though.)

chfast: See [Rule of zero](https://en.cppreference.com/w/cpp/language/rule_of_three#Rule_of_zero).

axic:
> because type is erased and and regular shared_ptr would delete void* then, which is UB (But none of sanitizers complained about this)

Perhaps you want to consider opening an issue on clang?

lib/fizzy/capi.cpp
```diff
@@ -110,13 +110,17 @@ inline fizzy::ExecutionResult unwrap(const FizzyExecutionResult& result) noexcep
...
 {
-    return [func, context](fizzy::Instance& instance, const fizzy::Value* args,
-               int depth) noexcept -> fizzy::ExecutionResult {
-        const auto result = func(context, wrap(&instance), wrap(args), depth);
-        return unwrap(result);
-    };
+    auto context = std::make_any<std::pair<FizzyExternalFn, void*>>(c_function, c_context);
+
+    static constexpr fizzy::ExecuteFunctionPtr function =
+        [](std::any cntx, fizzy::Instance& instance, const fizzy::Value* args, int depth) noexcept {
```
chfast, line 118:
```suggestion
        [](std::any host_ctx, fizzy::Instance& instance, const fizzy::Value* args, int depth) noexcept {
```

I propose to explicitly named these params "host context" to differentiate from "execute context" introduced later.

gumb0:
GCC failures around `nlohmann_json` look to be known issue: https://github.com/nlohmann/json/issues/2146

lib/fizzy/instantiate.hpp
```diff
@@ -21,23 +22,67 @@ namespace fizzy
...
+    /// Index of WebAssembly function.
+    /// Equal 0 in case this ExecuteFunction represents host function.
+    FuncIdx m_func_idx = 0;
+
+    /// Pointer to a host function.
+    /// Equals nullptr in case this ExecuteFunction represents WebAssembly function.
+    ExecuteFunctionPtr m_host_function = nullptr;
+
+    /// Opaque context of host function execution, which is passed to it as a context parameter.
+    /// `shared_ptr` is used because ExecuteFunction can be copied.
```
axic, line 45: Outdated comment.

lib/fizzy/capi.cpp
```diff
@@ -110,13 +110,20 @@ inline fizzy::ExecutionResult unwrap(const FizzyExecutionResult& result) noexcep
...
-    return [func, context](fizzy::Instance& instance, const fizzy::Value* args,
-               int depth) noexcept -> fizzy::ExecutionResult {
-        const auto result = func(context, wrap(&instance), wrap(args), depth);
-        return unwrap(result);
-    };
+    auto host_context =
+        std::make_any<std::pair<FizzyExternalFn, void*>>(c_function, c_host_context);
+
+    static constexpr fizzy::ExecuteFunctionPtr function =
+        [](std::any host_cntx, fizzy::Instance& instance, const fizzy::Value* args,
```
axic, line 119: This uses `host_cntx`. I think throughout the project we use `context` or `ctx`, it would be nice to stay consistent.

gumb0: See https://github.com/wasmx/fizzy/pull/716#discussion_r576314628

lib/fizzy/execute.cpp
```diff
@@ -837,10 +837,10 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, const Value* args,
             try
             {
                 if (new_pages > instance.memory_pages_limit)
-                    throw std::bad_alloc();
+                    throw std::length_error("");
```
axic, line 840: Even though we do not use `bad_alloc` here, the underlying resize can still emit it according to the docs.

lib/fizzy/capi.cpp
axic: Still, it can named `host_ctx` since the other is `host_context`. But perhaps using the wrapped/unwrapped terms in the context variable name may be more clear.

lib/fizzy/execute.cpp
gumb0: Yes, it's better to add it back, but it will be difficult to cover it with tests.

lib/fizzy/capi.cpp
gumb0: What's the exact suggestion here?

I can actually call them both `host_context` if I move the outer definition after lambda definition and avoid GCC shadowing warning (I did this in some tests)

axic: There's an extra `n`. Suggestion is to not use `cntx` but `ctx` or `context` and/or use the `wrapped` or `unwrapped` prefixes.

lib/fizzy/instantiate.cpp
```diff
@@ -470,18 +471,28 @@ std::optional<FuncIdx> find_exported_function(const Module& module, std::string_
     return find_export(module, ExternalKind::Function, name);
 }
 
+ExecuteFunction::ExecuteFunction(Instance& instance, FuncIdx func_idx)
+  : m_function([](std::any host_context, Instance&, const Value* args, int depth) {
+        auto [idx, inst] = std::any_cast<std::pair<FuncIdx, Instance*>>(host_context);
+        return execute(*inst, idx, args, depth);
+    }),
+    m_host_context(std::make_any<std::pair<FuncIdx, Instance*>>(func_idx, &instance))
```
gumb0: If I understand libstd++ code correctly, small object optimization happens only for objects smaller than pointer size, so putting this `pair` into `any` will still do heap allocation, so I guess Pawel's optimization should make sense, I will squash it.
(too bad we don't have benchmarks that import wasm functions)

gumb0:
I moved `noexcept` change to #738

lib/fizzy/instantiate.hpp
```diff
@@ -21,23 +22,66 @@ namespace fizzy
 struct ExecutionResult;
 struct Instance;
 
-/// Function representing WebAssembly or host function execution.
-using execute_function = std::function<ExecutionResult(Instance&, const Value*, int depth)>;
+/// Function pointer to the execution function.
+using ExecuteFunctionPtr = ExecutionResult (*)(
```
chfast, line 26: This is only for host functions.

chfast (commented):
It may be good to allow users omit host function context if they not need it: we would need `HostFunctionWithContextPtr` and `HostFunctionPtr` and separate `ExecuteFunction` constructors.

gumb0:
> It may be good to allow users omit host function context if they not need it: we would need `HostFunctionWithContextPtr` and `HostFunctionPtr` and separate `ExecuteFunction` constructors.

I'm not sure it's worth the complication - currently they can already use constructor with only function pointer and ignore conext parameter in function implementation - which most of the tests and `FizzyEngine` and wasi do.

I would leave this for another PR.

gumb0:
GCC 10.2 with LTO
```
fizzy/instantiate/blake2b_mean                                     -0.0037         -0.0040            68            68            68            68
fizzy/execute/blake2b/512_bytes_rounds_1_mean                      +0.0629         +0.0626           159           169           159           169
fizzy/execute/blake2b/512_bytes_rounds_16_mean                     +0.0549         +0.0547          2383          2514          2384          2514
fizzy/instantiate/ecpairing_mean                                   -0.0277         -0.0279          3483          3386          3483          3386
fizzy/execute/ecpairing/onepoint_mean                              +0.0289         +0.0287        810348        833729        810489        833741
fizzy/instantiate/keccak256_mean                                   -0.0298         -0.0300           116           112           116           112
fizzy/execute/keccak256/512_bytes_rounds_1_mean                    +0.2151         +0.2150           170           206           170           206
fizzy/execute/keccak256/512_bytes_rounds_16_mean                   +0.1997         +0.1996          2454          2944          2455          2944
fizzy/instantiate/memset_mean                                      -0.0170         -0.0172            20            20            20            20
fizzy/execute/memset/256_bytes_mean                                -0.0006         -0.0008            15            15            15            15
fizzy/execute/memset/60000_bytes_mean                              -0.0110         -0.0111          3247          3211          3248          3211
fizzy/instantiate/mul256_opt0_mean                                 -0.0182         -0.0183            25            25            25            25
fizzy/execute/mul256_opt0/input1_mean                              +0.0066         +0.0065            55            55            55            55
fizzy/instantiate/ramanujan_pi_mean                                -0.0291         -0.0293            67            65            67            65
fizzy/execute/ramanujan_pi/33_runs_mean                            +0.0002         +0.0001           239           239           239           239
fizzy/instantiate/sha1_mean                                        -0.0295         -0.0296           105           102           105           102
fizzy/execute/sha1/512_bytes_rounds_1_mean                         -0.0268         -0.0201           190           185           188           185
fizzy/execute/sha1/512_bytes_rounds_16_mean                        -0.0733         -0.0697          2771          2567          2760          2568
fizzy/instantiate/sha256_mean                                      -0.0339         -0.0322           170           164           170           164
fizzy/execute/sha256/512_bytes_rounds_1_mean                       +0.0197         +0.0208           168           171           167           171
fizzy/execute/sha256/512_bytes_rounds_16_mean                      +0.0192         +0.0198          2290          2334          2289          2334
fizzy/instantiate/taylor_pi_mean                                   -0.0107         -0.0103            12            12            12            12
fizzy/execute/taylor_pi/pi_1000000_runs_mean                       +0.0002         +0.0006         89879         89900         89841         89898
fizzy/instantiate/micro/eli_interpreter_mean                       -0.0042         -0.0037            15            15            15            15
fizzy/execute/micro/eli_interpreter/exec105_mean                   +0.0016         +0.0021            10            10            10            10
fizzy/instantiate/micro/factorial_mean                             -0.0077         -0.0075             2             2             2             2
fizzy/execute/micro/factorial/20_mean                              -0.0379         -0.0378             1             1             1             1
fizzy/instantiate/micro/fibonacci_mean                             +0.0090         +0.0091             3             3             3             3
fizzy/execute/micro/fibonacci/24_mean                              -0.0160         -0.0159         10961         10785         10955         10781
fizzy/instantiate/micro/host_adler32_mean                          -0.0235         -0.0261             7             7             7             7
fizzy/execute/micro/host_adler32/1_mean                            +0.0228         +0.0197             0             0             0             0
fizzy/execute/micro/host_adler32/1000_mean                         +0.0471         +0.0449            64            67            64            67
fizzy/instantiate/micro/icall_hash_mean                            -0.0125         -0.0139            13            13            13            13
fizzy/execute/micro/icall_hash/1000_steps_mean                     +0.0847         +0.0836           143           155           143           155
fizzy/instantiate/micro/spinner_mean                               -0.0040         -0.0045             2             2             2             2
fizzy/execute/micro/spinner/1_mean                                 -0.0367         -0.0371             0             0             0             0
fizzy/execute/micro/spinner/1000_mean                              -0.0693         -0.0697            21            20            21            20
fizzy/instantiate/stress/guido-fuzzer-find-1_mean                  -0.0071         -0.0075           356           354           356           353
```

lib/fizzy/instantiate.hpp
chfast:
```suggestion
using HostFunctionPtr = ExecutionResult (*)(
```

chfast:
> I'm not sure it's worth the complication - currently they can already use constructor with only function pointer and ignore conext parameter in function implementation - which most of the tests and `FizzyEngine` and wasi do.
> 
> I would leave this for another PR.

Yes, but you need to add another unused `std::any` param. My main issue is that this introduces a lot of noise and I will need to repeat changes there related to ExecutionContext.

gumb0:
> Yes, but you need to add another unused `std::any` param. My main issue is that this introduces a lot of noise and I will need to repeat changes there related to ExecutionContext.

Is this a good reason to introduce noise into API?

gumb0:
But I can imagine achieving this with a template constuctor of `ExecuteFunction`, it would take any callable (can be lambda, can be lambda with capture or can be a function), then your `HostFunctionWithContextPtr` won't be needed.

But still it seems it could be confusing to have host functions with diffirenet signatures.

chfast:
I agree it is better to explore this later.

chfast:
Clang11:
```
fizzy/execute/blake2b/512_bytes_rounds_1_mean                     +0.0182         +0.0182            81            82            81            82
fizzy/execute/blake2b/512_bytes_rounds_16_mean                    +0.0159         +0.0159          1212          1231          1212          1231
fizzy/execute/ecpairing/onepoint_mean                             -0.0080         -0.0080        410335        407053        410338        407056
fizzy/execute/keccak256/512_bytes_rounds_1_mean                   +0.0434         +0.0434           102           107           102           107
fizzy/execute/keccak256/512_bytes_rounds_16_mean                  +0.0451         +0.0451          1495          1562          1495          1562
fizzy/execute/memset/256_bytes_mean                               +0.0159         +0.0159             7             7             7             7
fizzy/execute/memset/60000_bytes_mean                             +0.0187         +0.0187          1495          1523          1495          1523
fizzy/execute/mul256_opt0/input1_mean                             +0.0336         +0.0336            29            30            29            30
fizzy/execute/ramanujan_pi/33_runs_mean                           +0.0169         +0.0169           110           112           110           112
fizzy/execute/sha1/512_bytes_rounds_1_mean                        +0.0197         +0.0197            88            89            88            89
fizzy/execute/sha1/512_bytes_rounds_16_mean                       +0.0162         +0.0162          1224          1244          1224          1244
fizzy/execute/sha256/512_bytes_rounds_1_mean                      +0.0220         +0.0220            81            83            81            83
fizzy/execute/sha256/512_bytes_rounds_16_mean                     +0.0272         +0.0272          1109          1139          1109          1139
fizzy/execute/taylor_pi/pi_1000000_runs_mean                      -0.0006         -0.0006         40681         40655         40682         40656
fizzy/execute/micro/eli_interpreter/exec105_mean                  +0.0225         +0.0225             4             5             4             5
fizzy/execute/micro/factorial/20_mean                             -0.0112         -0.0112             1             1             1             1
fizzy/execute/micro/fibonacci/24_mean                             -0.0045         -0.0045          5211          5187          5211          5187
fizzy/execute/micro/host_adler32/1_mean                           -0.0046         -0.0046             0             0             0             0
fizzy/execute/micro/host_adler32/1000_mean                        -0.0146         -0.0146            33            32            33            32
fizzy/execute/micro/icall_hash/1000_steps_mean                    +0.0102         +0.0102            70            70            70            70
fizzy/execute/micro/spinner/1_mean                                +0.0177         +0.0177             0             0             0             0
fizzy/execute/micro/spinner/1000_mean                             +0.0138         +0.0138             9             9             9             9
```

I'm not sure why calls got slower.

chfast (approved):

gumb0:
My clang results:
```
fizzy/execute/blake2b/512_bytes_rounds_1_mean                     +0.0130         +0.0132           180           183           180           183
fizzy/execute/blake2b/512_bytes_rounds_16_mean                    +0.0132         +0.0125          2719          2755          2721          2755
fizzy/execute/ecpairing/onepoint_mean                             +0.0790         +0.0784        883452        953241        883945        953207
fizzy/execute/keccak256/512_bytes_rounds_1_mean                   -0.0259         -0.0263           229           223           229           223
fizzy/execute/keccak256/512_bytes_rounds_16_mean                  -0.0425         -0.0428          3411          3266          3412          3266
fizzy/execute/memset/256_bytes_mean                               +0.0026         +0.0023            15            15            15            15
fizzy/execute/memset/60000_bytes_mean                             -0.0011         -0.0012          3372          3368          3373          3368
fizzy/execute/mul256_opt0/input1_mean                             -0.0471         -0.0472            66            63            66            63
fizzy/execute/ramanujan_pi/33_runs_mean                           -0.0187         -0.0187           250           246           250           246
fizzy/execute/sha1/512_bytes_rounds_1_mean                        -0.0021         -0.0021           193           192           193           192
fizzy/execute/sha1/512_bytes_rounds_16_mean                       -0.0030         -0.0030          2679          2671          2679          2671
fizzy/execute/sha256/512_bytes_rounds_1_mean                      +0.0094         +0.0094           175           177           175           177
fizzy/execute/sha256/512_bytes_rounds_16_mean                     +0.0044         +0.0044          2403          2413          2403          2414
fizzy/execute/taylor_pi/pi_1000000_runs_mean                      -0.0005         -0.0005         87713         87670         87710         87670
fizzy/execute/micro/eli_interpreter/exec105_mean                  +0.0006         +0.0006            10            10            10            10
fizzy/execute/micro/factorial/20_mean                             +0.0058         +0.0058             1             1             1             1
fizzy/execute/micro/fibonacci/24_mean                             +0.0061         +0.0062         11540         11611         11540         11611
fizzy/execute/micro/host_adler32/1_mean                           +0.0161         +0.0156             0             0             0             0
fizzy/execute/micro/host_adler32/1000_mean                        +0.0340         +0.0334            66            68            66            68
fizzy/execute/micro/icall_hash/1000_steps_mean                    +0.0244         +0.0239           144           147           144           147
fizzy/execute/micro/spinner/1_mean                                +0.0516         +0.0511             0             0             0             0
fizzy/execute/micro/spinner/1000_mean                             -0.0017         -0.0021            19            19            19            19
```

gumb0:
> I think passing it by const reference makes it actually unusable as the generic host context which should be mutable by the host functions.

I think I agree, it would simpler for the user to pass it as non-const.

(Counter-argument could be that it adds more opportunity to shoot yourself in the foot, and passing by const reference is like making context it immutable by default, but it can be made mutable if desired if they put a pointer into it.)

chfast:
I expect `std::any` will not be user facing later, so this can stay I guess.

chfast:
Clang11, pass by value vs pass by ref:
```
fizzy/execute/blake2b/512_bytes_rounds_1_mean                     -0.0147         -0.0147            82            81            82            81
fizzy/execute/blake2b/512_bytes_rounds_16_mean                    -0.0125         -0.0125          1231          1216          1231          1216
fizzy/execute/ecpairing/onepoint_mean                             +0.0025         +0.0025        407053        408088        407056        408090
fizzy/execute/keccak256/512_bytes_rounds_1_mean                   -0.0408         -0.0407           107           102           107           102
fizzy/execute/keccak256/512_bytes_rounds_16_mean                  -0.0444         -0.0444          1562          1493          1562          1493
fizzy/execute/memset/256_bytes_mean                               +0.0065         +0.0065             7             7             7             7
fizzy/execute/memset/60000_bytes_mean                             +0.0071         +0.0071          1523          1534          1523          1534
fizzy/execute/mul256_opt0/input1_mean                             -0.0038         -0.0038            30            29            30            29
fizzy/execute/ramanujan_pi/33_runs_mean                           -0.0106         -0.0106           112           111           112           111
fizzy/execute/sha1/512_bytes_rounds_1_mean                        -0.0129         -0.0129            89            88            89            88
fizzy/execute/sha1/512_bytes_rounds_16_mean                       -0.0107         -0.0107          1244          1231          1244          1231
fizzy/execute/sha256/512_bytes_rounds_1_mean                      -0.0255         -0.0255            83            81            83            81
fizzy/execute/sha256/512_bytes_rounds_16_mean                     -0.0090         -0.0090          1139          1129          1139          1129
fizzy/execute/taylor_pi/pi_1000000_runs_mean                      +0.0018         +0.0018         40655         40728         40656         40729
fizzy/execute/micro/eli_interpreter/exec105_mean                  +0.0012         +0.0012             5             5             5             5
fizzy/execute/micro/factorial/20_mean                             -0.0043         -0.0043             1             1             1             1
fizzy/execute/micro/fibonacci/24_mean                             +0.0020         +0.0020          5187          5198          5187          5198
fizzy/execute/micro/host_adler32/1_mean                           -0.0427         -0.0427             0             0             0             0
fizzy/execute/micro/host_adler32/1000_mean                        -0.0524         -0.0524            32            31            32            31
fizzy/execute/micro/icall_hash/1000_steps_mean                    +0.0051         +0.0051            70            71            70            71
fizzy/execute/micro/spinner/1_mean                                -0.0004         -0.0004             0             0             0             0
fizzy/execute/micro/spinner/1000_mean                             -0.0171         -0.0171             9             9             9             9
```

test/unittests/execute_call_test.cpp
```diff
@@ -361,14 +363,14 @@ TEST(execute_call, imported_function_call_void)
 
     const auto module = parse(wasm);
 
-    bool called = false;
-    const auto host_foo = [&called](Instance&, const Value*, int) {
+    static bool called = false;
+    constexpr auto host_foo = [](std::any&, Instance&, const Value*, int) {
```
chfast, line 367: Here it would be better example to pass the `bool called` by `std::any`.

test/unittests/execute_call_test.cpp
```diff
@@ -837,16 +891,16 @@ TEST(execute_call, call_imported_interleaved_infinite_recursion)
         "0061736d010000000105016000017f020b01036d6f6403666f6f0000030201000a0601040010000b");
 
     const auto module = parse(wasm);
-    int counter = 0;
-    auto host_foo = [&counter](Instance& instance, const Value* args, int depth) {
+    static int counter = 0;
+    constexpr auto host_foo = [](std::any&, Instance& instance, const Value* args, int depth) {
```
chfast, line 895: Same here.

test/unittests/execute_call_test.cpp
gumb0: I think it's better to have separate tests for explicitly checking context mechanism, and leave the rest of the tests in the simplest form with statics.

So I'll rather add a new test.

chfast: ok

chfast (approved):

lib/fizzy/execute.cpp
axic: Fixed in #737.

lib/fizzy/instantiate.cpp
```diff
@@ -470,18 +471,23 @@ std::optional<FuncIdx> find_exported_function(const Module& module, std::string_
     return find_export(module, ExternalKind::Function, name);
 }
 
+ExecutionResult ExecuteFunction::operator()(Instance& instance, const Value* args, int depth)
```
axic, line 474: Nit: this feels like a weird place for this. Aren't also these functions within an anonymous namesapce?

test/unittests/api_test.cpp
```diff
@@ -427,11 +432,13 @@ TEST(api, find_exported_function)
...
         "0606017f0041000b07170403666f6f000001670300037461620100036d656d0200");
 
-    auto bar = [](Instance&, const Value*, int) { return Value{42}; };
+    constexpr auto bar = [](std::any&, Instance&, const Value*, int) -> ExecutionResult {
+        return Value{42};
+    };
     const auto bar_type = FuncType{{}, {ValType::i32}};
 
     auto instance_reexported_function =
-        instantiate(parse(wasm_reexported_function), {{bar, bar_type.inputs, bar_type.outputs}});
```
axic, line 434: The inputs/outputs seems to be like a stray change, but looking at other cases below we do seem to have a constructor for it.

test/unittests/execute_call_test.cpp
```diff
@@ -361,14 +363,14 @@ TEST(execute_call, imported_function_call_void)
 
     const auto module = parse(wasm);
 
-    bool called = false;
-    const auto host_foo = [&called](Instance&, const Value*, int) {
+    static bool called = false;
```
axic, line 366: Hmm, why not just capture this in the function below?

test/unittests/execute_call_test.cpp
```diff
@@ -652,13 +663,13 @@ TEST(execute_call, call_initial_depth)
     const auto wasm = from_hex("0061736d01000000010401600000020b01036d6f6403666f6f0000");
 
     auto module = parse(wasm);
-    auto host_foo = [](Instance& /*instance*/, const Value*, int depth) -> ExecutionResult {
```
axic, line 655: Oh actually why do we drop the explicit return type?

test/unittests/execute_call_test.cpp
```diff
@@ -806,15 +821,14 @@ TEST(execute_call, call_imported_infinite_recursion)
         "0061736d010000000105016000017f020b01036d6f6403666f6f0000030201000a0601040010000b");
 
     const auto module = parse(wasm);
-    int counter = 0;
-    auto host_foo = [&counter](Instance& instance, const Value* args, int depth) {
-        EXPECT_LE(depth, TestCallStackLimit - 1);
+    static int counter = 0;
```
axic, line 935: Some as the other case of static variable.

lib/fizzy/instantiate.cpp
gumb0: This is `fizzy` namespace, unnamed ends at line 303.

It can't be in header, because it uses `execute`, that's why it's in this file.

test/unittests/execute_call_test.cpp
gumb0: Lambda's capture can't be used, because only lambdas without capture can be converted into function pointer.

Another option is to pass it via `host_context`, but I decided against it, see https://github.com/wasmx/fizzy/pull/716/files/5baa695073c325728d1e60ee321ef94bf77e7248#r580939050

Also a possible evolution of this that would allow lambdas with capture is in #740

test/unittests/api_test.cpp
gumb0: Yes, it's not required / can be split out.

lib/fizzy/instantiate.cpp
axic: I mean just put it up on the top of the file or so.

test/unittests/api_test.cpp
axic: Fine with me, it is only this case so having a single PR for it would be overkill.

test/unittests/execute_call_test.cpp
axic: Sure, we just used it for clarity. But either way it is fine, the thing is typechecked in instantiate.

test/unittests/execute_call_test.cpp
```diff
@@ -489,6 +489,46 @@ TEST(execute_call, imported_function_from_another_module)
...
+        local.get 1
+        call $sub
+      )
+    )
+    */
+    const auto bin2 = from_hex(
+        "0061736d0100000001070160027f7f017f020a01026d31037375620000030201000a0a0108002000200110000"
+        "b");
+
+    const auto func_idx = fizzy::find_exported_function(*module1, "sub");
```
axic, line 523: Took me a while looking up headers what is going on. So confusing we have `find_exported_function` with instance and module, returning different things.

test/unittests/api_test.cpp
```diff
@@ -217,6 +217,40 @@ TEST(api, resolve_imported_function_duplicate)
     EXPECT_THAT(execute(*instance, 1, {0_u32}), Result(42));
 }
 
+TEST(api, resolve_imported_function_duplicate_with_context)
```
axic, line 220: The point of this test to check that even though the `any` is duplicated in each ExternalFunction, they are somehow still reference to the same?

lib/fizzy/instantiate.hpp
```diff
@@ -46,9 +46,6 @@ class ExecuteFunction
     std::any m_host_context;
 
 public:
-    /// Default function constructor.
-    ExecuteFunction() noexcept = default;
```
axic: Was the constructor of the line `fizzy::ImportedFunction imported_func;` using this?

test/unittests/execute_call_test.cpp
```diff
@@ -577,6 +577,77 @@ TEST(execute_call, imported_function_from_another_module_via_host_function)
     EXPECT_THAT(execute(*instance2, 0, {44, 2}), Result(42));
 }
 
+TEST(execute_call, imported_function_with_cotext)
```
axic, line 580:
```suggestion
TEST(execute_call, imported_function_with_context)
```

test/unittests/execute_call_test.cpp
```diff
@@ -577,6 +577,77 @@ TEST(execute_call, imported_function_from_another_module_via_host_function)
...
+
+    auto instance = instantiate(*module, {{{host_func, std::move(host_context)}, host_func_type}});
+
+    EXPECT_THAT(execute(*instance, 0, {}), Result());
+    EXPECT_EQ(counter, 1);
+    EXPECT_THAT(execute(*instance, 1, {}), Result());
+    EXPECT_EQ(counter, 2);
+}
+
+TEST(execute_call, imported_functions_with_shared_cotext)
```
axic, line 611:
```suggestion
TEST(execute_call, imported_functions_with_shared_context)
```

axic, line 611: Isn't this kind of a duplicate of `resolve_imported_function_duplicate_with_context)`?

lib/fizzy/instantiate.cpp
axic: Sure, the main point is not to have it intermingled into helpers.

lib/fizzy/capi.cpp
```diff
@@ -163,32 +163,34 @@ inline std::vector<fizzy::ExternalFunction> unwrap(
...
 inline fizzy::ImportedFunction unwrap(const FizzyImportedFunction& c_imported_func)
 {
-    fizzy::ImportedFunction imported_func;
-    imported_func.module = c_imported_func.module;
-    imported_func.name = c_imported_func.name;
-
     const auto& c_type = c_imported_func.external_function.type;
-    imported_func.inputs.resize(c_type.inputs_size);
+
+    std::vector<fizzy::ValType> inputs(c_type.inputs_size);
```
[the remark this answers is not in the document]
gumb0: Here it's possible to default-construct `ValType`s.
Not much difference, but I find it a bit simpler to use `vector` constructor + `transform`, than `reserve` + `transform` + `back_inserter`.

test/unittests/execute_call_test.cpp
```diff
@@ -577,6 +577,77 @@ TEST(execute_call, imported_function_from_another_module_via_host_function)
...
+    const auto module = parse(wasm);
+
+    constexpr auto host_func = [](std::any& host_context, Instance&, const Value*,
+                                   int) -> ExecutionResult {
+        ++(*std::any_cast<int*>(host_context));
+        return Void;
+    };
+    const auto& host_func_type = module->typesec[0];
+
+    int counter = 0;
```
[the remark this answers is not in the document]
gumb0: Yes, I think that was the motivation for `uint32_t` in the below.

test/unittests/execute_call_test.cpp
gumb0: They are somewhat similar, but in resolve it's a single function that is imported twice and context is copied internally.
Here it's different functions and the user copies the context.

axic (approved):

PR #717, opened 2021-02-03
ci: Migrate Rust to cimg/rust, cache cargo

chfast:


gumb0 (approved):

circle.yml
```diff
@@ -265,6 +268,39 @@ commands:
             echo $result
             if [ "$expected" != "$result" ]; then exit 1; fi
 
+  rust_restore_cargo_cache:
+    steps:
+      - restore_cache:
+          name: "Restore cargo cache"
+          key: cargo-
```
axic, line 275: Does this mean it will pick up anything? Latest?

circle.yml
```diff
@@ -645,26 +669,19 @@ jobs:
...
-            echo 'deb http://deb.debian.org/debian buster-backports main' >> /etc/apt/sources.list
-            apt -qq update
-            apt -yq -t buster-backports install cmake --no-install-recommends
-            apt -yq install llvm-8-dev clang-8 --no-install-recommends
-            rustup toolchain install nightly-x86_64-unknown-linux-gnu
-            rustup component add llvm-tools-preview
-            cargo install cargo-binutils rustfilt
       - run:
           name: Build (debug mode)
           command: RUSTFLAGS="-Z instrument-coverage" cargo +nightly build --target x86_64-unknown-linux-gnu
```
axic, line 680: I think `--target x86_64-unknown-linux-gnu` can be removed, but it does not hurt. It was needed in chisel because it was also built to the wasm target.

axic (approved):

circle.yml
chfast: It suppose to download the latest matching this prefix. We cannot match with Cargo.lock because it does not exist here.

PR #719, opened 2021-02-04
rust: support cloning Modules

axic:


bindings/rust/src/lib.rs
```diff
@@ -20,10 +20,21 @@ pub struct Module(*const sys::FizzyModule);
 
 impl Drop for Module {
     fn drop(&mut self) {
+        debug_assert!(!self.0.is_null());
```
axic, line 23: Changed these to `debug_assert` and added it to all relevant places. A failure of this would mean a programming error in our code, and not user code. Hence `debug_assert`, i.e. it will be caught by our tests assuming we have high enough coverage.

bindings/rust/src/lib.rs
```diff
@@ -20,10 +20,21 @@ pub struct Module(*const sys::FizzyModule);
...
         unsafe { sys::fizzy_free_module(self.0) }
     }
 }
 
+impl Clone for Module {
+    fn clone(&self) -> Self {
+        debug_assert!(!self.0.is_null());
+        let ptr = unsafe { sys::fizzy_clone_module(self.0) };
+        // TODO: this can be zero in case of memory allocation error, should this be gracefully handled?
```
axic, line 32: That being said the clone trait does not support optionality.

chfast (approved):

gumb0 (approved):

PR #720, opened 2021-02-04
rust: add simple CostNonNull wrapper based on std::ptr::NonNull

axic:


axic:
@chfast @gumb0 please review

chfast (approved):
This is little smelly to me to introduce new vocabulary types in the library. But I'm leaving Rust API design to others.

PR #572, opened 2020-10-05
Execution thread shared stack space

chfast:

PR #726, opened 2021-02-04
Introduce ThreadContext instead of execution depth

chfast:
Replace depth param in execute() functions with ThreadContext&.
The ThreadContext encapsulates the depth control from now on and is
extensible. In future it can support metering and better stack space
allocation.

Call depth control semantic is preserved, except the depth is now passed by reference (aka `int& depth`) so additional context guard is needed in places where depth is being bumped.

In C the ThreadContext is opaque what is good for API, but it is impossible to dump the depth inside a C host function.

This is baseline for PoC thread context extensions proposed in #529 and #572. Rebasing these is impossible, but the rest of the changes from there should be re-applied later.

### DIFF
--- a/bindings/rust/src/lib.rs
+++ b/bindings/rust/src/lib.rs
@@ -313,18 +313,11 @@ impl Instance {
     ///
     /// An invalid index, invalid inputs, or invalid depth can cause undefined behaviour.
     ///
-    /// The `depth` argument sets the initial call depth. Should be set to `0`.
-    ///
     /// # Safety
-    /// This function expects a valid `func_idx`, appropriate number of `args`, and valid `depth`.
-    pub unsafe fn unsafe_execute(
-        &mut self,
-        func_idx: u32,
-        args: &[Value],
-        depth: i32,
-    ) -> ExecutionResult {
+    /// This function expects a valid `func_idx` and appropriate number of `args`.
+    pub unsafe fn unsafe_execute(&mut self, func_idx: u32, args: &[Value]) -> ExecutionResult {
         ExecutionResult {
-            0: sys::fizzy_execute(self.0.as_ptr(), func_idx, args.as_ptr(), depth),
+            0: sys::fizzy_execute(self.0.as_ptr(), func_idx, args.as_ptr()),
         }
     }
 
@@ -338,15 +331,8 @@ impl Instance {
     ///
     /// An error is returned if the function can not be found, inappropriate number of arguments are passed,
     /// or the supplied types are mismatching.
-    ///
-    /// The `depth` argument sets the initial call depth. Should be set to `0`.
     // TODO: consider different approach: Result<TypedValue, Error> where Error::Trap is used for traps
-    pub fn execute(
-        &mut self,
-        name: &str,
-        args: &[TypedValue],
-        depth: i32,
-    ) -> Result<TypedExecutionResult, ()> {
+    pub fn execute(&mut self, name: &str, args: &[TypedValue]) -> Result<TypedExecutionResult, ()> {
         let func_idx = self.find_exported_function_index(&name);
         if func_idx.is_none() {
             return Err(());
@@ -369,7 +355,7 @@ impl Instance {
         // Translate to untyped raw values.
         let args: Vec<Value> = args.iter().map(|v| v.into()).collect();
 
-        let ret = unsafe { self.unsafe_execute(func_idx, &args, depth) };
+        let ret = unsafe { self.unsafe_execute(func_idx, &args) };
         Ok(TypedExecutionResult {
             result: ret.0,
             value_type: func_type.output,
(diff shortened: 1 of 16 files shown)


lib/fizzy/instantiate.hpp
```diff
@@ -21,8 +21,31 @@ namespace fizzy
 struct ExecutionResult;
 struct Instance;
 
+class ThreadContext
+{
+public:
+    int depth = 0;
```
axic, line 27: So there's no way to set a different starting depth?

lib/fizzy/instantiate.hpp
```diff
@@ -21,8 +21,31 @@ namespace fizzy
 struct ExecutionResult;
 struct Instance;
 
+class ThreadContext
```
axic, line 24: Why do we go with the term thread?  Is this future proofing for multithreaded wasm?  Is some kind of locking planned, given the memory (to begin with) is still part of the instance and not the thread.

lib/fizzy/instantiate.hpp
chfast: You can `ctx.depth = 1`, as in tests.

lib/fizzy/instantiate.hpp
chfast: It has nothing to do with multithreading. I is just execution thread context (in principle, you can create multiple thread context and start multiple call stacks in parallel). I understand the name may be confusing although WABT uses the same name for this concept. We can consider alternative names.

gumb0, line 24: Can the declaration be moved to `execute.hpp`, as `ExecutionResult` ?

chfast: Maybe.

lib/fizzy/instantiate.hpp
axic: Should we expose it? In any case if we it should be just a separate PR for the C API (and then Rust). It is fine to hide it in this PR.

bindings/rust/src/lib.rs
```diff
@@ -298,12 +293,7 @@ impl Instance {
     ///
     /// The `depth` argument sets the initial call depth. Should be set to `0`.
```
axic, line 294: Please remove these lines too, here and in `unsafe_execute`.

include/fizzy/fizzy.h
```diff
@@ -20,6 +20,8 @@ typedef struct FizzyModule FizzyModule;
 /// The opaque data type representing an instance (instantiated module).
 typedef struct FizzyInstance FizzyInstance;
 
+typedef struct FizzyThreadContext FizzyThreadContext;
```
axic, line 23: Should have a comment, see FizzyInstance above.

lib/fizzy/instantiate.hpp
```diff
@@ -21,8 +21,31 @@ namespace fizzy
...
+{
+public:
+    int depth = 0;
+
+    void acquire() noexcept { ++depth; }
```
axic, line 29: Do these need to be public? Are these supposed to be used directly or only via the Guard?

lib/fizzy/instantiate.hpp
axic, line 24: Should have some level of documentaiton.

lib/fizzy/execute.hpp
```diff
@@ -49,7 +49,15 @@ constexpr ExecutionResult Trap{false};
...
 ///                     behaviour (including crash) happens.
-/// @param  depth       The call depth (indexing starts at 0). Can be left at the default setting.
+/// @param  ctx         The thread context.
 /// @return             The result of the execution.
-ExecutionResult execute(Instance& instance, FuncIdx func_idx, const Value* args, int depth = 0);
+ExecutionResult execute(
+    Instance& instance, FuncIdx func_idx, const Value* args, ThreadContext& ctx);
+
+/// Execute a function from an instance starting at default depth of 0.
+inline ExecutionResult execute(Instance& instance, FuncIdx func_idx, const Value* args)
```
axic, line 58: Is this version used heavily in the tests?

test/unittests/capi_test.cpp
```diff
@@ -318,7 +318,9 @@ TEST(capi, find_exported_function)
     EXPECT_EQ(function.type.output, FizzyValueTypeI32);
     EXPECT_NE(function.context, nullptr);
     ASSERT_NE(function.function, nullptr);
-    EXPECT_THAT(function.function(function.context, instance, nullptr, 0), CResult(42_u32));
+
+    // FIXME: thread context missing.
```
axic, line 322: This should be fixed before merging I guess?

axic:
Looks great in general!

lib/fizzy/instantiate.hpp
axic: I think it would be a good name, but given wasm multithreading this may be confusing. Would this be applicable during that? Would something like `ExecutionContext` be less confusing, or equally confusing as it is partial too?

axic:
I think there should be a benchmark run after the comments are addressed and before this is merged, to see what is the effect, if any.

test/unittests/capi_test.cpp
chfast: This is not fixable if ThreadContext to stay opaque in C.

lib/fizzy/execute.hpp
chfast: Yes.

lib/fizzy/instantiate.hpp
chfast: > In C the ThreadContext is opaque what is good for API, but it is impossible to dump the depth inside a C host function.

axic: Well we would need C++ helpers for it first, right? :)

lib/fizzy/execute.hpp
```diff
@@ -41,6 +41,28 @@ constexpr ExecutionResult Void{true};
...
+
+    public:
+        explicit Guard(ThreadContext& ctx) noexcept : m_thread_context{ctx} {}
+        ~Guard() noexcept { --m_thread_context.depth; }
+    };
+
+public:
+    int depth = 0;
+
+    Guard bump_call_depth() noexcept
```
gumb0, line 59: Hm, this returns private class, I wouldn't think it's possible...

gumb0: Maybe call it `increment_call_depth`, bump sounds a bit informal it seems.

gumb0: Just an idea: it could be execute member function of `ThreadContext` bumping depth and redirecting to normal `execute`
```cpp
    ExecutionResult execute(Instance& instance, FuncIdx func_idx, const Value* args) noexcept;
```
then `Guard` can be completely hidden from user.

But maybe confusingly too many execute variants then.

test/unittests/execute_call_test.cpp
```diff
@@ -479,8 +481,9 @@ TEST(execute_call, imported_function_from_another_module)
     const auto func_idx = fizzy::find_exported_function(*module1, "sub");
     ASSERT_TRUE(func_idx.has_value());
 
-    auto sub = [&instance1, func_idx](Instance&, const Value* args, int depth) -> ExecutionResult {
-        return fizzy::execute(*instance1, *func_idx, args, depth);
+    auto sub = [&instance1, func_idx](Instance&, const Value* args, ThreadContext& ctx) {
+        // ThreadContext is bumped here.
```
gumb0, line 485: I was a bit confused by this comment. I think you mean at this point it was already bumped by the caller?

chfast: Probably should be "is not bumped". And it should be according to the current semantic.

chfast:
I will rename it to `ExecutionContext`.

gumb0:
> I will rename it to `ExecutionContext`.

What I don't like about `ExecutionContext` is that `void* context` in C API also can be thought of as "execution context". But I don't have a better name.

chfast:
Replaced by #766.

PR #728, opened 2021-02-05
rust: module inspection prototype

axic:


bindings/rust/src/lib.rs
```diff
@@ -48,22 +48,30 @@ pub fn validate<T: AsRef<[u8]>>(input: T) -> bool {
 
 /// A parsed and validated WebAssembly 1.0 module.
 // NOTE: cannot use NonNull here given this is *const
-pub struct Module(*const sys::FizzyModule);
+pub struct Module {
+    ptr: *const sys::FizzyModule,
+    owned: bool,
```
axic, line 53: So I think these are the options for having inspection on the module (which is still available through an instance):
a) either we have this owned flag here and disallow instantiation of a non-owned module
b) introduce a `ModuleInspection` or some other abstraction which has it, and that can be obtained as `module.inspect()` or `instance.inspect()` (this can be seen in one of the earlier commits)

chfast: This look ok, although fragile. And I'd rather name the flag `instantiated` so it is clear it can be instantiated once.

Can't Rust magic memory annotations handle this?

axic: > Can't Rust magic memory annotations handle this?

I would one workaround, where I maintain an "instance of module" in the instance and return the reference on it. Since `instantiate()` requires to own the input, a reference can't be passed so it will not compile.

i.e.
```
struct Instance {
  instance: NonNull<sys::FizzyInstance>,
  module: Module,
}

impl Instance {
    pub fn get_module(&self) -> &Module {
       &self.module
    }
}

// Then this will be a compilation error: instance.get_module().instantiate() 
```

gumb0: Not necessary to save module in `Instance`, you can call `fizzy_get_instance_module`

axic: The downside of the `ModuleInspection` option is that need to assign its lifetime bound to the underlying `Module`/`Instance`.

axic: > Not necessary to save module in Instance, you can call fizzy_get_instance_module

That is what is being done in this PR. For that one needs to extend the `Module` as in this PR.

An alternative approach is to save the module in the instance and return it as a reference.

gumb0: I imagined I guess some combination of your approaches: don't change `Module`, but have `Instance::get_module(&self) -> Module ` that will call `fizzy_get_instance_module`, wrap in `Module` and call `core::mem::forget` before returning it.

PR #730, opened 2021-02-08
rust: update bindgen dependency

axic:


axic:
Bindgen after 0.54 dropped support for Clang3.8 (does not matter too much) and Rust <1.40 (perhaps this is a too new requirement), while it does not seem to offer any benefit to what we need right now.

PR #732, opened 2021-02-09
test: Empty memory support

gumb0:


test/unittests/capi_test.cpp
```diff
@@ -145,6 +145,18 @@ TEST(capi, has_memory)
 
     fizzy_free_module(module_no_memory);
 
+    /* wat2wasm
+      (memory 0)
+    */
+    const auto wasm_memory_empty = from_hex("0061736d010000000503010000");
```
axic, line 151: Why is this duplicated in two test cases?

gumb0: One test here checks module inspection function `fizzy_module_has_memory`

The test below checks instance memory access functions.

axic (approved):

include/fizzy/fizzy.h
```diff
@@ -450,6 +450,8 @@ const FizzyModule* fizzy_get_instance_module(FizzyInstance* instance);
 /// @return             Pointer to memory data or NULL in case instance doesn't have any memory.
 ///
 /// @note    Function returns pointer to memory regardless of whether memory is exported or not.
+/// @note    For instances of the modules defined with memory of size 0 the returned pointers is not
```
chfast, line 453:
```suggestion
/// @note    For instances of the modules defined with memory of size 0 the returned pointers are not
```

chfast (approved):

include/fizzy/fizzy.h
axic: Well, shouldn't it be just `pointer is not`, since it is a single pointer?

PR #735, opened 2021-02-12
rust: use specific error for traps in execute

axic:


bindings/rust/src/lib.rs
```diff
@@ -413,30 +435,34 @@ impl Instance {
...
         if expected_types != supplied_types {
-            return Err(());
+            return Err("argument type mistmatch".into());
         }
 
         // Translate to untyped raw values.
         let args: Vec<Value> = args.iter().map(|v| v.into()).collect();
 
         let ret = unsafe { self.unsafe_execute(func_idx, &args, depth) };
+        let ret: ExecutionResult = ret;
```
axic, line 462: So in this case we could do the output type validation here and only return `Option<TypedValue>`.

So the final signature would be `Result<Option<TypedValue>, Error>`.

chfast (commented):
There are no examples using this, so can't tell if it is useful or not. Can't comment if this is rusty enough.

axic:
> There are no examples using this, so can't tell if it is useful or not. Can't comment if this is rusty enough.

I thought I pushed the test changes. Anyway, it removes one level of indirection.

```
let ret = instance.execute("test", [])?; // this would bubble up errors
if let Some(ret) = ret {
  println!("Returned {}", ret.as_u32());
}
```

axic:
Not sure about the last commit, but otherwise this feels more native to Rust.

bindings/rust/src/lib.rs
```diff
@@ -893,35 +800,23 @@ mod tests {
         // Successful execution.
         let result = instance.execute("foo", &[]);
         assert!(result.is_ok());
-        let result = result.unwrap();
-        assert!(!result.trapped());
-        assert!(result.value().is_some());
-        assert_eq!(result.value().unwrap().as_u32().unwrap(), 42);
+        assert_eq!(result.unwrap().unwrap().as_u32().unwrap(), 42);
```
chfast, line 803: Idiomatic Rust you say. I think I can see it now...

axic: That's only tests. See the example on the top of the file.

bindings/rust/src/lib.rs
```diff
@@ -365,6 +365,23 @@ impl ExecutionResult {
             None
         }
     }
+
+    pub fn typed_value(&self, expected_type: sys::FizzyValueType) -> Option<TypedValue> {
+        assert!(!self.0.trapped);
+        if self.0.has_value {
```
gumb0, line 371: Now the decision what to return is based on both `expected_type` and `self.0.has_value`

Maybe better change this to `if expected_type != sys::FizzyValueTypeVoid` and assert inside `assert!(self.0.has_value)`. Or even make it part of one bigger `match expected_type`.

Also maybe rename `expected_type` to just `type`, otherwise it looks like it's mainly not a conversion, but a check / test.

bindings/rust/src/lib.rs
```diff
@@ -365,33 +365,20 @@ impl ExecutionResult {
...
-    pub fn value(&self) -> Option<TypedValue> {
-        if self.result.has_value {
-            assert!(!self.result.trapped);
-            assert!(self.value_type != sys::FizzyValueTypeVoid);
-            Some(match self.value_type {
-                sys::FizzyValueTypeI32 => TypedValue::U32(unsafe { self.result.value.i32 }),
-                sys::FizzyValueTypeI64 => TypedValue::U64(unsafe { self.result.value.i64 }),
-                sys::FizzyValueTypeF32 => TypedValue::F32(unsafe { self.result.value.f32 }),
-                sys::FizzyValueTypeF64 => TypedValue::F64(unsafe { self.result.value.f64 }),
+    pub fn typed_value(&self, expected_type: sys::FizzyValueType) -> Option<TypedValue> {
```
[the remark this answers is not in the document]
gumb0: I like it, but see comment below.

bindings/rust/src/lib.rs
axic: Pushed a commit flipping conditions, but it doesn't make much of a difference. `assert!`'s are always in the binary, `debug_assert!`s aren't.

`type` is a keyword, hence `expected_type`.

gumb0 (approved):

PR #737, opened 2021-02-16
Improve exception handling in memory.grow

gumb0:


lib/fizzy/execute.cpp
```diff
@@ -840,7 +840,11 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, const Value* args,
                     throw std::bad_alloc();
                 memory->resize(new_pages * PageSize);
             }
-            catch (std::bad_alloc const&)
+            catch (const std::bad_alloc&)
+            {
+                ret = static_cast<uint32_t>(-1);
+            }
+            catch (const std::length_error&)
```
gumb0, line 847: Doesn't look possible to cover it, for me `std::string::max_size()` returns 2^46-1 bytes, we can't allocate more than that with wasm.

chfast: I don't think this is worth having considering [`max_size()` is not very useful](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2012/n3440.html#197).
It may be just better to add a unit test that checks if wasm memory max size is not smaller than `std::string::max_size()`.
Alternatively, you can catch `std::exception` here.

chfast: On 32-bit build the value is 2^31-1.

chfast (approved):

lib/fizzy/execute.cpp
```diff
@@ -840,7 +840,7 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, const Value* args,
                     throw std::bad_alloc();
                 memory->resize(new_pages * PageSize);
             }
-            catch (std::bad_alloc const&)
+            catch (const std::exception&)
```
chfast, line 843: Add a comment that `std::bad_alloc` and `std::legth_error` are expected.

chfast (approved):

lib/fizzy/execute.cpp
chfast: The ... catches everything so it guarantees noexcept outside. The std::exception does not, so there will be another compiler generated check ending in std::terminate or something.

So my final call is to use ... and also add an assert that the exception is one of the two expected types.

Although, checking the exception type may be tricky. Probably a helper is needed as presented in https://en.cppreference.com/w/cpp/error/rethrow_exception

gumb0: Well then in the helper `catch std::length_error` will still be not covered, it's simpler then to have here `catch (bad_alloc)` + `catch (length_error)` + `catch (...) { assert(false); } `.

gumb0: > Well then in the helper `catch std::length_error` will still be not covered, it's simpler then to have here `catch (bad_alloc)` + `catch (length_error)` + `catch (...) { assert(false); } `.

But yeah `execute` will be smaller if extra catches go into helper, and helper will should be omitted in release build...

lib/fizzy/execute.cpp
```diff
@@ -499,6 +499,25 @@ inline bool invoke_function(const FuncType& func_type, uint32_t func_idx, Instan
     return true;
 }
 
+[[maybe_unused]] bool is_resize_exception(std::exception_ptr exception)
```
axic, line 502: Nice! Perhaps you want to add a comment here explaining our use case.

lib/fizzy/execute.cpp
```diff
@@ -840,8 +859,9 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, const Value* args,
                     throw std::bad_alloc();
```
axic, line 859: Should this use `std::bad_alloc` or `std::length_error` ?

chfast: The `bad_alloc` is better but not perfect fit out of these two.

chfast: Actually, here I'd see another refactoring: do not use `throw`.
```
uint32_t ret = (new_pages <= instance.memory_pages_limit) ? static_cast<uint32_t>(cur_pages) : static_cast<uint32_t>(-1);
try {
  memory->resize(new_pages * PageSize);
} catch (...) { ... }

Comment (gumb0) [reply lib/fizzy/execute.cpp:862]: It shouldn't go to `resize` though, if it's over the limit.
I can make a version with `if`, but it looks somewhat uglier.

Comment (gumb0) [reply lib/fizzy/execute.cpp]: > The `bad_alloc` is better but not perfect fit out of these two.

Why it it better and what's the "fit" here? I'd say requirement for `memory.grow` to fail in case requested size is out of bounds is similar to requirement for `basic_string` to throw `length_error` in case `count > max_size()`.
`bad_alloc` represents failure of some lower level allocation mechanism. 

But I think in practice it doesn't matter and I'm fine with either.

File: lib/fizzy/execute.cpp (hunk 3 above)
Comment (gumb0) [reply lib/fizzy/execute.cpp]: Added a comment.

**Review by chfast — APPROVED** [review]
Not very beautiful, but I think this is fine.

It is rather visible that some `Memory` abstraction type would be nice. This can be done together with "guarded pages" later.

There is also an issue on 32-bit architectures where string/bytes can only allocate up to 2GB of memory.

File: lib/fizzy/execute.cpp [hunk 5]
```diff
@@ -833,16 +855,21 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, const Value* args,
...
-            uint32_t ret = static_cast<uint32_t>(cur_pages);
-            try
-            {
-                if (new_pages > instance.memory_pages_limit)
-                    throw std::bad_alloc();
-                memory->resize(new_pages * PageSize);
-            }
-            catch (std::bad_alloc const&)
-            {
+            uint32_t ret;
```
axic, line 858: Hmm, wondering if the following would be more readable:
```
uint32_t = static_cast<uint32_t>(-1);
if (new_pages <= instance.memory_pages_limit)
{
  try
  {
    memory->resize(new_pages * PageSize);
    ret = static_cast<uint32_t>(cur_pages);
  }
  catch ...
}
```

axic, line 858: Hmm, wondering if the following would be more readable:
```
uint32_t = static_cast<uint32_t>(-1);
if (new_pages <= instance.memory_pages_limit)
{
  try
  {
    memory->resize(new_pages * PageSize);
    ret = static_cast<uint32_t>(cur_pages);
  }
  catch ...
}
```

chfast: It looks fine, I don't mind.

The best is to wrap this into `uint32_t Memory::grow(something);`. Then you eliminate `ret` housekeeping.

gumb0: I can move it into a simple helper for now.

lib/fizzy/execute.cpp
```diff
@@ -829,22 +874,7 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, const Value* args,
...
-                if (new_pages > instance.memory_pages_limit)
-                    throw std::bad_alloc();
-                memory->resize(new_pages * PageSize);
-            }
-            catch (std::bad_alloc const&)
-            {
-                ret = static_cast<uint32_t>(-1);
-            }
-            stack.push(ret);
+            stack.push(grow_memory(*memory, delta, instance.memory_pages_limit));
```
axic, line 877: At this point could just use the stack pointer and no pop/push is needed 🙊

lib/fizzy/execute.cpp
```diff
@@ -109,6 +109,51 @@ inline bool store_into_memory(
...
+    {
+        return true;
+    }
+    catch (...)
+    {
+        return false;
+    }
+}
+
+inline uint32_t grow_memory(
```
axic, line 135: If we are going with this helper, please document it. What is important to document that it returns page count prior to expansion.

axic (approved):
I wonder how this affects benchmarks as it could mean quite a different binary layout 🙊

chfast (approved):

PR #740, opened 2021-02-24
Generic constructor for ExecuteFunction

gumb0:
This is one way to simplify host function definitions, allowing to skip `host_context` parameter and using lambdas with captures.

Another way could be having another host function pointer type like `HostFunctionWithoutContextPtr` suggested in https://github.com/wasmx/fizzy/pull/716#pullrequestreview-592257080 (it would not allow lambdas with captures.)

But I think leaving C++ API with none of these might be the least confusing, and most consistent with C API.

### DIFF
--- a/lib/fizzy/execute.hpp
+++ b/lib/fizzy/execute.hpp
@@ -41,6 +41,16 @@ constexpr ExecutionResult Void{true};
 /// Shortcut for execution that resulted in a trap.
 constexpr ExecutionResult Trap{false};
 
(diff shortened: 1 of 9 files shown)


axic:
@gumb0 do we still want this? It is mostly an improvement for testing since the C++ API is not public.

gumb0:
> @gumb0 do we still want this? It is mostly an improvement for testing since the C++ API is not public.

I don't see much benefit, will close it.

PR #743, opened 2021-02-26
rust: return string as an error

axic:
This is the string version of #677.

### DIFF
--- a/bindings/rust/src/lib.rs
+++ b/bindings/rust/src/lib.rs
@@ -68,10 +68,10 @@ impl Clone for Module {
 }
 
 /// Parse and validate the input according to WebAssembly 1.0 rules.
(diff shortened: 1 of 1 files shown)


bindings/rust/src/lib.rs
```diff
@@ -840,30 +848,78 @@ mod tests {
...
+                instance.checked_memory_slice(0, 65536).err().unwrap(),
+                "no memory is available"
+            );
+            assert_eq!(
+                instance.checked_memory_slice_mut(0, 65536).err().unwrap(),
+                "no memory is available"
+            );
+            assert_eq!(
+                instance.checked_memory_slice(65535, 1).err().unwrap(),
+                "no memory is available"
```
axic, line 869: I am not sold we need to check in every case, but at least it is clear.

gumb0 (approved):

chfast (approved):

PR #744, opened 2021-03-02
parser: Fix types for out-of-bound checks

chfast:
Fixes build issues on 32-bit architectures.

### DIFF
--- a/lib/fizzy/parser.cpp
+++ b/lib/fizzy/parser.cpp
@@ -269,7 +269,8 @@ parser_result<std::string> parse_string(const uint8_t* pos, const uint8_t* end)
     uint32_t size;
     std::tie(size, pos) = leb128u_decode<uint32_t>(pos, end);
 
-    if ((end - pos) < size)
+    assert(pos <= end);
+    if (static_cast<size_t>(end - pos) < size)
         throw parser_error{"unexpected EOF"};
 
     if (!utf8_validate(pos, pos + size))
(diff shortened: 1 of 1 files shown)


lib/fizzy/parser.cpp
```diff
@@ -269,7 +269,8 @@ parser_result<std::string> parse_string(const uint8_t* pos, const uint8_t* end)
     uint32_t size;
     std::tie(size, pos) = leb128u_decode<uint32_t>(pos, end);
 
-    if ((end - pos) < size)
+    assert(pos <= end);
+    if (static_cast<size_t>(end - pos) < size)
```
gumb0, line 273: What does the cast solve?

gumb0:
Is utf8 fix relatd to 32bit?

lib/fizzy/parser.cpp
chfast: Pointer diff is int, size_t is unsigned int.

chfast:
> Is utf8 fix related to 32bit?

The type change is.

lib/fizzy/utf8.cpp
```diff
@@ -108,7 +109,7 @@ bool utf8_validate(const uint8_t* pos, const uint8_t* end) noexcept
 
         // At this point need to read at least one more byte.
         assert(required_bytes > 1);
-        if ((end - pos) < required_bytes - 1)
+        if ((end - pos) < (required_bytes - 1))
```
gumb0, line 112: Not sure if makes sense, but perhaps it would be simpler, to just decrement all `required_bytes` literals now, and use `required_bytes` here without subtraction.

chfast: I don't really want to change more than needed here. I believe we should replace it with something more performant anyway in the future.

gumb0 (approved):

lib/fizzy/utf8.cpp
axic, line 112: I'd put this change into the previous commit.

axic:
Is there any measurable speed difference for the utf8 test cases?

axic:
We should also find some utf8 validation stress test and include it.

axic (approved):

PR #745, opened 2021-03-02
Add 32-bit build

chfast:
Depends on #744.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -151,7 +151,7 @@ commands:
             lcov --capture --directory . --output-file coverage.lcov --exclude='/usr/include/*' --exclude="$HOME/.hunter/*" --exclude="$PWD/_deps/*"
             lcov --zerocounters --directory .
             rm -rf coverage
-            genhtml coverage.lcov --output-directory coverage --title Fizzy
+            genhtml coverage.lcov --output-directory coverage --title $CIRCLE_PROJECT_REPONAME
 
   collect_coverage_clang:
     description: "Collect coverage data (Clang)"
@@ -482,6 +482,19 @@ jobs:
           min_time: "0.01"
       - spectest
 
+  x86-linux:
+    docker:
+      - image: ethereum/cpp-build-env:17-gcc-11-multilib
+    steps:
+      - install_testfloat
+      - checkout
+      - build:
+          build_type: Debug
+          cmake_options: -DCMAKE_TOOLCHAIN_FILE=~/project/cmake/toolchains/32bit.cmake
+      - test
+      # TODO: Disabled, because fixes to NaN payload handling in the tool are needed.
+      # - spectest
+
   arm64-linux:
     executor: linux-gcc-latest
     steps:
@@ -733,6 +746,7 @@ workflows:
       - spectest:
           requires:
             - coverage-clang
+      - x86-linux
       - arm64-linux
       - bindings-rust
       - bindings-rust-asan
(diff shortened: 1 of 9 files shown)


axic:
@chfast Can we get this merged by disabling FP tests on 32-bit, for now?

chfast:
Yes.

cmake/ProjectWabt.cmake
```diff
@@ -53,7 +64,7 @@ ExternalProject_Add(wabt
     ${build_command}
     INSTALL_COMMAND ""
     BUILD_BYPRODUCTS ${wabt_library}
-)
+    )
```
chfast, line 67: To be fixed.

circle.yml
```diff
@@ -482,6 +482,19 @@ jobs:
...
+    docker:
+      - image: ethereum/cpp-build-env:15-gcc-10-multilib
+    steps:
+      - install_testfloat
+      - checkout
+      - build:
+          build_type: Debug
+          cmake_options: -DCMAKE_TOOLCHAIN_FILE=~/project/cmake/toolchains/32bit.cmake
+      - test
+      # TODO: Disabled, because fixes to NaN payload handling in the tool are needed.
```
axic, line 495: As long as you are 100% sure it is only a tooling fix (and no fix/change needed in the interpreter), I'm fine merging this.

chfast: I'm not sure, but this is the first thing to fix.

test/unittests/execute_floating_point_test.cpp
```diff
@@ -1145,29 +1148,26 @@ TEST(execute_floating_point, f32_store)
...
-        {-1.0f, "cc000080bfcc"_bytes},
-        {std::nextafter(1.0f, 0.0f), "ccffff7f3fcc"_bytes},
-        {std::nextafter(-1.0f, 0.0f), "ccffff7fbfcc"_bytes},
-        {FP32::nan(FP32::canon), "cc0000c07fcc"_bytes},
-        {-FP32::nan(FP32::canon), "cc0000c0ffcc"_bytes},
-        {FP32::nan(FP32::canon + 1), "cc0100c07fcc"_bytes},
-        {-FP32::nan(FP32::canon + 1), "cc0100c0ffcc"_bytes},
-        {FP32::nan(1), "cc0100807fcc"_bytes},
-        {-FP32::nan(1), "cc010080ffcc"_bytes},
+    const std::tuple<float, bytes> test_cases[]
```
gumb0, line 1151: maybe `clang-format off` here?

test/unittests/execute_floating_point_test.cpp
```diff
@@ -1360,6 +1372,10 @@ TEST(execute_floating_point, f64_add_off_by_one)
     _FPU_SETCW(fpu_cw);
     */
 
+#if defined(__i386__) && !defined(__SSE_MATH__)
```
gumb0, line 1375: why not at the top of the test?

chfast: To execute as much as possible. Can be at the top too.

gumb0: On top it would be more clear that long comment there explains the reason for skipping IMHO.

gumb0 (approved):

test/unittests/execute_floating_point_test.cpp
chfast: I agree. Will do it.

PR #746, opened 2021-03-02
Construct Value only from uint32_t/uint64_t

chfast:
This disables Value constructors for unsigned integers for types other
than uint32_t/uint64_t. This requires users to use these types.
This can cause portability issues on macOS/Linux if type like
unsigned long is used. This also helps in 32-bit port.

### DIFF
--- a/lib/fizzy/value.hpp
+++ b/lib/fizzy/value.hpp
@@ -23,14 +23,13 @@ union Value
 
     /// Converting constructors from integer types
     ///
-    /// We need to support {signed,unsigned} x {32,64} integers. However, due to uint64_t being
-    /// defined differently in different implementations we need to avoid the alias and provide
-    /// constructors for unsigned long and unsigned long long independently.
-    constexpr Value(unsigned int v) noexcept : i32{v} {}
-    constexpr Value(unsigned long v) noexcept : i64{v} {}
-    constexpr Value(unsigned long long v) noexcept : i64{v} {}
-    constexpr Value(int64_t v) noexcept : i64{static_cast<uint64_t>(v)} {}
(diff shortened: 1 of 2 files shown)


axic (approved):

lib/fizzy/value.hpp
```diff
@@ -26,11 +26,10 @@ union Value
     /// We need to support {signed,unsigned} x {32,64} integers. However, due to uint64_t being
```
gumb0, line 26: Comment is outdated.

axic (approved):

lib/fizzy/value.hpp
```diff
@@ -23,14 +23,13 @@ union Value
...
-    /// We need to support {signed,unsigned} x {32,64} integers. However, due to uint64_t being
-    /// defined differently in different implementations we need to avoid the alias and provide
-    /// constructors for unsigned long and unsigned long long independently.
-    constexpr Value(unsigned int v) noexcept : i32{v} {}
-    constexpr Value(unsigned long v) noexcept : i64{v} {}
-    constexpr Value(unsigned long long v) noexcept : i64{v} {}
-    constexpr Value(int64_t v) noexcept : i64{static_cast<uint64_t>(v)} {}
+    /// Only fixed-size integer types are supported: {signed,unsigned} x {32,64}.
+    /// Some issues are expected when uint64_t is not unsigned long.
+    /// Then Value cannot be constructed out of unsigned long literals.
```
gumb0, line 28: "some issues" sounds a bit confusing to me, maybe this:
```suggestion
    /// On the platforms where uint64_t is not unsigned long, it is expected that
    /// Value cannot be constructed out of unsigned long literals.
```

gumb0 (approved):

gumb0 (approved):

PR #747, opened 2021-03-02
memory.grow 32-bit fix

gumb0:
Depends on #748 and #749.

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -879,6 +879,7 @@ ExecutionResult execute(Instance& instance, FuncIdx func_idx, const Value* args,
         }
         case Instr::memory_size:
         {
+            assert(memory->size() % PageSize == 0);
             stack.push(static_cast<uint32_t>(memory->size() / PageSize));
             break;
         }
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -139,15 +139,23 @@ inline uint32_t grow_memory(
     bytes& memory, uint32_t delta_pages, uint32_t memory_pages_limit) noexcept
 {
     const auto cur_pages = memory.size() / PageSize;
-    assert(cur_pages <= size_t(std::numeric_limits<int32_t>::max()));
-    const auto new_pages = cur_pages + delta_pages;
-    assert(new_pages >= cur_pages);
+    // These assertions are guaranteed by allocation in instantiate and this function for subsequent
+    // increases.
+    assert(memory.size() % PageSize == 0);
+    assert(memory_pages_limit <= MaxMemoryPagesLimit);
+    assert(cur_pages <= memory_pages_limit);
+
+    const auto new_pages = uint64_t{cur_pages} + delta_pages;
     if (new_pages > memory_pages_limit)
         return static_cast<uint32_t>(-1);
 
     try
     {
-        memory.resize(new_pages * PageSize);
+        // new_pages <= memory_pages_limit <= MaxMemoryPagesLimit guarantees multiplication
+        // will not overflow uint32_t.
+        assert(new_pages * PageSize <= std::numeric_limits<uint32_t>::max());
+        static_assert(sizeof(size_t) >= sizeof(uint32_t));
+        memory.resize(static_cast<size_t>(new_pages) * PageSize);
         return static_cast<uint32_t>(cur_pages);
     }
     catch (...)


lib/fizzy/execute.cpp
```diff
@@ -140,14 +140,15 @@ inline uint32_t grow_memory(
 {
     const auto cur_pages = memory.size() / PageSize;
```
axic, line 141: This function looks like a mess now with assertions every second line. Not sure yet, but can we somehow clean this mess up a bit?

axic: Also while we're speaking assertions, there could be one for `memory.size() % PageSize == 0` :wink:

lib/fizzy/execute.cpp
```diff
@@ -140,14 +140,15 @@ inline uint32_t grow_memory(
...
-    const auto new_pages = cur_pages + delta_pages;
+    const auto new_pages = uint64_t{cur_pages} + delta_pages;
     assert(new_pages >= cur_pages);
     if (new_pages > memory_pages_limit)
         return static_cast<uint32_t>(-1);
 
     try
     {
-        memory.resize(new_pages * PageSize);
+        assert(static_cast<size_t>(new_pages * PageSize) > new_pages);
```
chfast, line 150: Not sure what this checks. 32-bit overflow?

gumb0: Yes, I'll add a test where this fails, and I think `memory_pages_limit` should be capped for 32-bit during instantiate.

This assertion should probably be better changed to testing assumptions about `memory_pages_limit`.

lib/fizzy/execute.cpp
```diff
@@ -140,14 +140,15 @@ inline uint32_t grow_memory(
 {
     const auto cur_pages = memory.size() / PageSize;
     assert(cur_pages <= size_t(std::numeric_limits<int32_t>::max()));
-    const auto new_pages = cur_pages + delta_pages;
+    const auto new_pages = uint64_t{cur_pages} + delta_pages;
```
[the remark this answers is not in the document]
gumb0: Why? It would be redundant as operands are promoted to biggest type, it's similar to what we did here:
https://github.com/wasmx/fizzy/blob/0c663e480e0fcc2b91ece9301671ca4593f78cf9/lib/fizzy/execute.cpp#L74

lib/fizzy/execute.cpp
chfast: Then probably `assert(static_cast<size_t>(new_pages * PageSize) == (new_pages * PageSize))` is better.

Isn't memory capped to 4GB?

gumb0: I checked in the spec, and yes, it is, but it looks like we impose this limit only in parser for memory limits definition in module. It's possible to set `memory_pages_limit` higher, then `memory.grow` will allow it to grow above.

lib/fizzy/execute.cpp
```diff
@@ -138,16 +138,21 @@ inline bool store_into_memory(
 inline uint32_t grow_memory(
     bytes& memory, uint32_t delta_pages, uint32_t memory_pages_limit) noexcept
 {
+    // Guaranteed by instantiate allocation + resize in grow_memory.
+    assert(memory.size() % PageSize == 0);
```
gumb0, line 142: BTW for imported memory this is not guaranteed. Niether that its size is inside the limits (only its declared size is checked).
I guess instantiate should check that, too.

lib/fizzy/execute.cpp
gumb0: I tried to clean up by making assertions simpler and adding some comments.

lib/fizzy/execute.cpp
gumb0: Removed this assertion in favor of checking `memory_pages_limit < 4GB`. If that one is true, this must not overflow.

lib/fizzy/execute.cpp
```diff
@@ -138,16 +138,23 @@ inline bool store_into_memory(
...
     const auto cur_pages = memory.size() / PageSize;
-    assert(cur_pages <= size_t(std::numeric_limits<int32_t>::max()));
-    const auto new_pages = cur_pages + delta_pages;
-    assert(new_pages >= cur_pages);
+    // Guaranteed by instantiate allocation + resize in grow_memory.
+    assert(cur_pages <= MemoryPagesValidationLimit);
```
gumb0, line 145: Probably `MemoryPagesValidationLimit` should be renames, because it will be used not only for validation.

lib/fizzy/execute.cpp
axic: Hmm, do you have a reference in the spec which allows imports to be not pagesize-aligned?

gumb0: By "not guaranteed" I meant that our instantiate currently doesn't enforce this, but I think it should.

In the spec it is required in the valid external memory definition https://webassembly.github.io/spec/core/exec/modules.html#xref-exec-runtime-syntax-externval-mathsf-mem-a:

image

and in memory definition https://webassembly.github.io/spec/core/exec/runtime.html#memory-instances:

image

axic: It is unfortunate that the spectest architecture can not describe such test cases (i.e. imported memory not a multiple of pagesize)

gumb0: > BTW for imported memory this is not guaranteed. Niether that its size is inside the limits (only its declared size is checked).

I've added the alignment check in #749 
The check for size being inside limits was already done apparently.

lib/fizzy/execute.cpp
```diff
@@ -138,16 +138,23 @@ inline bool store_into_memory(
...
     if (new_pages > memory_pages_limit)
         return static_cast<uint32_t>(-1);
 
+    // FIXME: Guarantee this by instantiate.
+    assert(memory_pages_limit <= MemoryPagesValidationLimit);
     try
     {
-        memory.resize(new_pages * PageSize);
+        // new_pages <= memory_pages_limit <= MemoryPagesValidationLimit guarantees multiplication
+        // will not overflow size_t.
```
chfast, line 156: Maybe it is better to say it will not overflow uint32.

chfast (approved):

axic:
I think we should merge this after #748 and #749.

lib/fizzy/execute.cpp
axic, line 145: Name will change with #748.

lib/fizzy/execute.cpp
```diff
@@ -138,16 +138,23 @@ inline bool store_into_memory(
...
-    assert(new_pages >= cur_pages);
+    // Guaranteed by instantiate allocation + resize in grow_memory.
+    assert(cur_pages <= MemoryPagesValidationLimit);
+
+    const auto new_pages = uint64_t{cur_pages} + delta_pages;
     if (new_pages > memory_pages_limit)
         return static_cast<uint32_t>(-1);
 
+    // FIXME: Guarantee this by instantiate.
+    assert(memory_pages_limit <= MemoryPagesValidationLimit);
```
axic, line 152: Perhaps you want to move this before the if above instead?

lib/fizzy/execute.cpp
```diff
@@ -139,15 +139,20 @@ inline uint32_t grow_memory(
...
+    assert(cur_pages <= MaxMemoryPagesLimit);
+
+    const auto new_pages = uint64_t{cur_pages} + delta_pages;
     if (new_pages > memory_pages_limit)
         return static_cast<uint32_t>(-1);
 
     try
     {
-        memory.resize(new_pages * PageSize);
+        // new_pages <= memory_pages_limit <= MaxMemoryPagesLimit guarantees multiplication
```
axic, line 153: Not sure we need this entire comment now with the asserts above?

lib/fizzy/execute.cpp
```diff
@@ -139,15 +139,20 @@ inline uint32_t grow_memory(
     bytes& memory, uint32_t delta_pages, uint32_t memory_pages_limit) noexcept
 {
     const auto cur_pages = memory.size() / PageSize;
-    assert(cur_pages <= size_t(std::numeric_limits<int32_t>::max()));
-    const auto new_pages = cur_pages + delta_pages;
-    assert(new_pages >= cur_pages);
+    // Guaranteed by allocation in instantiate and this function for subsequent increases.
+    assert(memory.size() % PageSize == 0);
+    assert(memory_pages_limit <= MaxMemoryPagesLimit);
+    assert(cur_pages <= MaxMemoryPagesLimit);
```
axic, line 145: This could also be `cur_pages <= memory_pages_limit` for clarity.

lib/fizzy/execute.cpp
gumb0: To explain why we are sure it's safe not to check for overflow here?

axic: I mean now the `uint64_t{}` conversion and the assertions explain it well enough?

lib/fizzy/execute.cpp
```diff
@@ -139,15 +139,20 @@ inline uint32_t grow_memory(
     bytes& memory, uint32_t delta_pages, uint32_t memory_pages_limit) noexcept
 {
     const auto cur_pages = memory.size() / PageSize;
-    assert(cur_pages <= size_t(std::numeric_limits<int32_t>::max()));
-    const auto new_pages = cur_pages + delta_pages;
-    assert(new_pages >= cur_pages);
+    // Guaranteed by allocation in instantiate and this function for subsequent increases.
```
axic, line 142:
```suggestion
    // These assertions are guaranteed by allocation in instantiate and this function for subsequent increases.
```

lib/fizzy/execute.cpp
gumb0: I would leave it, it's like an inferential step between assertions and conlusion that overlow is not possible. And it shows that we didn't forget to think about it :)

axic (approved):

lib/fizzy/execute.cpp
axic: Actually would we need a `static_assert(sizeof(size_t) >= sizeof(uint32_t));` here for total clarity? 🙊

lib/fizzy/execute.cpp
```diff
@@ -139,15 +139,22 @@ inline uint32_t grow_memory(
...
+
+    const auto new_pages = uint64_t{cur_pages} + delta_pages;
     if (new_pages > memory_pages_limit)
         return static_cast<uint32_t>(-1);
 
     try
     {
-        memory.resize(new_pages * PageSize);
+        // new_pages <= memory_pages_limit <= MaxMemoryPagesLimit guarantees multiplication
+        // will not overflow uint32_t.
```
chfast, line 155: Hm... to be even more pedantic, you can add `assert(new_pages * PageSize <= std::numaric_limits<uint32_t>::max())`.

chfast (approved):

PR #748, opened 2021-03-03
Cap hard memory limit at 4GB

gumb0:


chfast (commented):
The "Rename MemoryPagesValidationLimit -> MemoryPagesMaxLimit" commit should go first.

chfast (commented):
I'm still considering using `0xffffffff` limit value for "unlimited". Not sure this change helps. Maybe it is better to report hard limit in parser?

gumb0:
> I'm still considering using `0xffffffff` limit value for "unlimited". 

What would it mean? Given that spec limits it at 4GB.

> Not sure this change helps. 

It helps to simplify `memory.grow` (otherwise you would need to check both instantiate hard limit and spec hard limit there).

> Maybe it is better to report hard limit in parser?

What does this mean, pass it to parser instead of instantiate?

chfast:
> What would it mean? Given that spec limits it at 4GB.

Using `0xffffffff` for the case where `max` value is omitted in `limits`.



> What does this mean, pass it to parser instead of instantiate?

Make modules with max higher than 4GB invalid.

gumb0:
> > What would it mean? Given that spec limits it at 4GB.
> 
> Using `0xffffffff` for the case where `max` value is omitted in `limits`.
> 

This doesn't seem related, it's about replacing `optional` in `Limits` with magic value, isn't it?

> > What does this mean, pass it to parser instead of instantiate?
> 
> Make modules with max higher than 4GB invalid.

They are already invalid of course https://github.com/wasmx/fizzy/blob/56ffca78a74ea0bf94afaf437d33839a56ea610f/lib/fizzy/parser.cpp#L260-L262

But modules with min/max higher than Fizzy's hard limit are not invalid, because we pass this limit to instantiate only.
(I think it's more spec-compliant to consider them valid, and fail only when allocation would go over the Fizzy's limit)

chfast (approved):

lib/fizzy/limits.hpp
```diff
@@ -13,8 +13,8 @@ constexpr uint32_t PageSize = 65536;
 
 /// The maximum memory page limit as defined by the specification.
 /// It is only possible to address 4 GB (32-bit) of memory.
-constexpr uint32_t MemoryPagesValidationLimit = (4 * 1024 * 1024 * 1024ULL) / PageSize;
-static_assert(MemoryPagesValidationLimit == 65536);
+constexpr uint32_t MemoryPagesMaxLimit = (4 * 1024 * 1024 * 1024ULL) / PageSize;
```
axic, line 16: Sorry for being late to the party. I think we should make the "max" and the "default" variable following the same scheme, either `MemoryPagesMaxLimit`/`MemoryPagesDefaultLimit`, or `MaxMemoryPagesLimit`/`DefaultMemoryPagesLimit`.

@chfast @gumb0 what do you think?

lib/fizzy/instantiate.cpp
```diff
@@ -174,6 +174,13 @@ std::tuple<bytes_ptr, Limits> allocate_memory(const std::vector<Memory>& module_
     static const auto bytes_delete = [](bytes* b) noexcept { delete b; };
     static const auto null_delete = [](bytes*) noexcept {};
 
+    if (memory_pages_limit > MemoryPagesMaxLimit)
+    {
+        throw instantiate_error{"hard memory limit cannot exceed " +
```
axic, line 179: Nit: I would mention "during instantiate" or something like that in the commit log for this change.

axic, line 179: @gumb0 Is this now properly covered by spectests or you are planning to submit some cases?

test/unittests/execute_test.cpp
```diff
@@ -693,6 +694,12 @@ TEST(execute, memory_grow_custom_hard_limit)
         EXPECT_THAT(execute(*instance, 0, {input}), Result(expected));
     }
 
+    {
+        const auto instance_huge_hard_limit = instantiate(*module, {}, {}, {}, {}, 65536);
+        EXPECT_THAT(execute(*instance_huge_hard_limit, 0, {65536}), Result(-1));
```
axic, line 699: What is the goal here? According to the other test, 65536 total is acceptable. Why don't we add something like instantiate with 0 and enlarge to 65536.

lib/fizzy/limits.hpp
gumb0: Sounds good, I would change to `MaxMemoryPagesLimit`/`DefaultMemoryPagesLimit`.

lib/fizzy/instantiate.cpp
gumb0: It's not covered and can't be, as "hard memory limit" is not defined in any precise way in the spec, it's just an internal limit which can be arbitrary set by implementations.

chfast: It is not easy to find, but you can get this information from codecov. You need to go to single commit view -> Files, and then only select "spectests" flag.

https://codecov.io/gh/wasmx/fizzy/src/7d7bbfce98e8fca2927e539807beb8ffb4e8bac4/lib/fizzy/instantiate.cpp

image

test/unittests/execute_test.cpp
gumb0: > What is the goal here? 

To test the check at high values of hard limit.

> According to the other test, 65536 total is acceptable. Why don't we add something like instantiate with 0 and enlarge to 65536.

It would allocate 4GB, but I think we want to keep unit tests light-weight.
(there is a spec test that does this)

chfast (approved):

lib/fizzy/instantiate.cpp
axic: > It is not easy to find, but you can get this information from codecov. 

I know, I went through that and listed the 10% missing cases in the chat last week. But that is only reflected if we upgrade out spectest branch, and we haven't done that in this PR :)

test/unittests/execute_test.cpp
```diff
@@ -734,6 +741,14 @@ TEST(execute, memory_grow_custom_hard_limit)
         EXPECT_THAT(execute(*instance_max_limit, 0, {input}), Result(expected));
     }
 
+    {
+        bytes memory(PageSize, 0);
+        const auto instance_huge_hard_limit =
+            instantiate(*module_imported, {}, {}, {{&memory, {1, std::nullopt}}}, {}, 65536);
+        EXPECT_THAT(execute(*instance_huge_hard_limit, 0, {65536}), Result(-1));
```
axic, line 748: Well perhaps leave a comment then here that we should test for increasing by `65535` but don't want to spend too much time in unittests (to allocate 4gb) as spectests covers it.

axic (approved):
I'm fine with this, though not 100% convinced our test cases are the best.

test/unittests/execute_test.cpp
gumb0: Actually I checked and there're no spec tests that allocate 4GB, only one that sets the upper limit to 4GB (but doesn't allocate anything).

axic: It is hard to test for it when VMs are allowed to return an error.

gumb0: Added a comment here.

axic (approved):

PR #749, opened 2021-03-04
Check that imported memory size is multiple of page size

gumb0:


chfast (approved):

lib/fizzy/instantiate.cpp
```diff
@@ -113,6 +113,9 @@ void match_imported_memories(const std::vector<Memory>& module_imported_memories
             throw instantiate_error{"provided imported memory has a null pointer to data"};
 
         const auto size = imported_memories[0].data->size();
+        if (size % PageSize != 0)
+            throw instantiate_error{"provided imported memory size must be multiple of page size"};
```
axic, line 117: @gumb0 are you planning a test for upstream? Can this be tested? Can you note this in our list at least? Is this properly part of the spec?

gumb0: It's not possible to make such memory in spec tests.

axic: Added it to our tracking document.

axic (approved):

PR #755, opened 2021-03-11
Fix lower limit of exported memory after grow

gumb0:
This fixes the case from https://github.com/WebAssembly/spec/pull/1293


axic:
I think we should not merge this until the spectest is merged.

axic:
This was merged last April.

chfast (approved):

gumb0:
Related spec tests are backported to Fizzy spectests in https://github.com/wasmx/wasm-spec/pull/3

axic (approved):

chfast (approved):

PR #756, opened 2021-03-12
Drop default depth param from execute()

chfast:
In preparation for landing `ExecutionContext`, the default value for depth param is dropped in C++ and the param is completely removed from C and Rust (therefore C and Rust are not able to legally call wasm functions from host functions temporarily).

PR #757, opened 2021-03-12
test: Add call depth unit tests

chfast:
Depends on #756.

These test inspect the behavior of call depth limiting.

This duplicates a number of checks from "execute_call". Some tests from there may be removed if desired.

This implements test cases from https://notes.ethereum.org/@chfast/fizzy-call-depth.


chfast:
> This duplicates a number of checks from "execute_call". Some tests from there may be removed if desired.

They are all "replaced" in the current version.

gumb0:
Add new file to CI wat2wasm4tests check.

gumb0:
> Add new file to CI wat2wasm4tests check.

Ah nevermind, we run it for all test files already.

test/unittests/execute_call_test.cpp
```diff
@@ -531,55 +531,6 @@ TEST(execute_call, imported_function_from_another_module_via_func_idx)
     EXPECT_THAT(execute(*instance2, 0, {44, 2}), Result(42));
 }
 
-TEST(execute_call, imported_function_from_another_module_via_host_function)
```
gumb0, line 534: I think this test might be ok to leave, it was not about call depth. It just demonstrates technique of passing another instance + function index into host function.

Optionally host function can be changed to bump the depth, I think it's not very important here.

gumb0:
The new tests look ok to me. I didn't check very thoroughly if they cover everything of the old tests.

test/unittests/execute_call_test.cpp
chfast: Restored.

gumb0 (approved):

lib/fizzy/execute.hpp
```diff
@@ -49,8 +49,15 @@ constexpr ExecutionResult Trap{false};
...
 ///                     the expected number of input parameters of the function, otherwise undefined
 ///                     behaviour (including crash) happens.
-/// @param  depth       The call depth (indexing starts at 0). Can be left at the default setting.
+/// @param  depth       The call depth (indexing starts at 0).
 /// @return             The result of the execution.
 ExecutionResult execute(
-    Instance& instance, FuncIdx func_idx, const Value* args, int depth = 0) noexcept;
+    Instance& instance, FuncIdx func_idx, const Value* args, int depth) noexcept;
+
+/// Execute a function from an instance starting at default depth of 0.
```
axic, line 57: I'd add "Arguments and behaviour is the same as `execute`."

lib/fizzy/execute.hpp
```diff
@@ -49,8 +49,15 @@ constexpr ExecutionResult Trap{false};
...
+inline ExecutionResult execute(Instance& instance, FuncIdx func_idx, const Value* args) noexcept
+{
+    const int depth = 0;
```
axic, line 60: Not sure it makes a difference, but constexpr?

bindings/rust/src/lib.rs
```diff
@@ -691,7 +681,7 @@ mod tests {
         )
         */
         let input = hex::decode(
-        "0061736d010000000105016000017f030201000404017000000504010101020606017f0041000b07180403666f6f00000267310300037461620100036d656d02000a06010400412a0b").unwrap();
+            "0061736d010000000105016000017f030201000404017000000504010101020606017f0041000b07180403666f6f00000267310300037461620100036d656d02000a06010400412a0b").unwrap();
```
axic, line 684: Not sure what introduced this indent change?

test/unittests/execute_call_depth_test.cpp
```diff
@@ -0,0 +1,327 @@
...
+    (func $leaf (result i32) (i32.const 1))
+    */
+    const auto wasm = from_hex(
+        "0061736d010000000105016000017f020a0104686f73740166000003030200000a0b02040010000b040041010"
+        "b");
+
+    static int recorded_depth;
+    constexpr auto host_f = [](std::any&, Instance& instance, const Value*, int depth) noexcept {
+        recorded_depth = depth;
+        return fizzy::execute(instance, 2, {}, depth + 1);
```
axic, line 213:
```suggestion
        return fizzy::execute(instance, 2 /* $leaf */, {}, depth + 1);
```

Or leave a comment that this is calling `$leaf`.

test/unittests/execute_call_depth_test.cpp
```diff
@@ -0,0 +1,327 @@
...
+    )
+    */
+    const auto wasm = from_hex(
+        "0061736d0100000001040160000002110104686f737407636f756e746572037e01030201000a0d010b00230042"
+        "017c240010000b");
+
+    Value counter;
+    auto instance = instantiate(parse(wasm), {}, {}, {}, {{&counter, {ValType::i64, true}}});
+
+    // When starting from depth 0, the $f is expected to be called CallStackLimit times.
```
axic, line 355:
```suggestion
    // When starting from depth 0, the $f is expected to be called DepthLimit times.
```
?

Technically I guess the comment is correct, just confusing to see the expectation refer to a different const.

test/unittests/execute_call_depth_test.cpp
```diff
@@ -0,0 +1,327 @@
...
+
+    recorded_depth = -1000;
+    EXPECT_THAT(execute(*instance, 1, {}, DepthLimit), Traps());
+    EXPECT_EQ(recorded_depth, -1000);
+}
+
+
+/// Infinite recursion
+
+TEST(execute_call_depth, execute_internal_function_infinite_recursion)
```
axic, line 336: Hmm, can you add a case similar to this where recursion starts at the start function? Generally also cases where start function is present and calls other things.

bindings/rust/src/lib.rs
chfast: CLion.

test/unittests/execute_call_depth_test.cpp
chfast: Added.

axic (approved):

PR #762, opened 2021-03-15
Add explicit check for imported limits validity

gumb0:
- This improves error message for the case when import's max is below min: `provided imported table doesn't fit provided limits` => `provided import's min limit is above import's max limit`.
- Also it will allow to simplify imported memory size checks in #755 (to leave only the check `memory.size() == min.limit` in instantiate).


gumb0:
Split tests into a separate commit to show the error message change.

chfast (approved):

axic (approved):

PR #764, opened 2021-03-22
Rich error reporting prototype

gumb0:


include/fizzy/fizzy.h
```diff
@@ -14,6 +14,19 @@
 extern "C" {
 #endif
 
+typedef enum FizzyErrorCode
```
gumb0, line 17: These should better go into a separate header I think.

lib/fizzy/capi.cpp
```diff
@@ -343,15 +343,36 @@ inline FizzyExportDescription wrap(const fizzy::Export& exp) noexcept
...
         return true;
     }
-    catch (...)
+    catch (const fizzy::exception& e)
+    {
+        // TODO set_error helper
+        if (error != nullptr)
+        {
+            error->code = e.code;
+            snprintf(error->message, std::size(error->message), "%s", e.what());
```
axic, line 365: Wouldn't `strlcpy(error->message, e.what(), std::size(error->message))` be safer? (Assuming `e.what()` is a `char*` and `std::size` is equivalent to `sizeof(error->message)`)

See: https://linux.die.net/man/3/strlcpy

I think evmc already uses this in the loader and has an implementation for Windows?

axic: But @chfast may be right in the sense, that this complicates matters a bit, e.g. in case of truncation we may want to have a trailing `...` which makes this worse (at least the output buffer size is set by us, within `FizzyError`, but still).

gumb0: I used here `snprintf`, because it's standard. 
evmc has its own implementation of `strlcpy` that truncates to empty string in case it doesn't fit https://github.com/ethereum/evmc/blob/c39de40f025673575412e71afdf9f72f8c6314ba/lib/loader/loader.c#L52
and also uses `vsnprintf` for error messages
https://github.com/ethereum/evmc/blob/c39de40f025673575412e71afdf9f72f8c6314ba/lib/loader/loader.c#L85

include/fizzy/fizzy.h
```diff
@@ -6,6 +6,7 @@
 /// @file
 #pragma once
 
+#include "fizzy_errors.h"
```
gumb0, line 9: Making this `<fizzy/fizzy_errors.h>` breaks the Rust build, not sure yet how to fix.

test/unittests/capi_test.cpp
```diff
@@ -15,9 +15,44 @@ static constexpr FizzyExternalFn NullFn = nullptr;
...
-    EXPECT_FALSE(fizzy_validate(wasm_prefix, sizeof(wasm_prefix)));
+    FizzyError error;
+    EXPECT_FALSE(fizzy_validate(wasm_prefix, sizeof(wasm_prefix), &error));
+    EXPECT_EQ(error.code, FIZZY_ERROR_PARSER_INVALID_MODULE_PREFIX);
+    EXPECT_STREQ(error.message, "invalid wasm module prefix");
+}
+
+TEST(capi, validate_long_duplicate_export_name)
+{
+    // clang-format off
```
gumb0, line 28: I have not found the way to make multi-line name strings in wat, I think it's not possible.

include/fizzy/fizzy.h
axic: Why do we want to separate it out? Also what is the error Rust gives?

include/fizzy/fizzy.h
```diff
@@ -218,7 +225,7 @@ typedef struct FizzyImportedGlobal
 /// @param  wasm_binary         Pointer to module binary data.
 /// @param  wasm_binary_size    Size of the module binary data.
 /// @return                     true if module is valid, false otherwise.
-bool fizzy_validate(const uint8_t* wasm_binary, size_t wasm_binary_size);
+bool fizzy_validate(const uint8_t* wasm_binary, size_t wasm_binary_size, FizzyError* error);
```
axic, line 228: Missing doctext.

include/fizzy/fizzy_errors.h
```diff
@@ -0,0 +1,22 @@
...
+#pragma once
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef enum FizzyErrorCode
+{
+    FIZZY_SUCCESS = 0,
+    FIZZY_ERROR_PARSER_INVALID_MODULE_PREFIX,
```
axic, line 16: How deep we want to go with error reporting?

lib/fizzy/capi.cpp
```diff
@@ -7,10 +7,45 @@
...
+inline void set_success(FizzyError* error) noexcept
+{
+    if (error != nullptr)
+    {
+        error->code = FIZZY_SUCCESS;
+        error->message[0] = '\0';
+    }
+}
+
+inline size_t strlcpy_s(char* dest, const char* src, size_t size) noexcept
```
axic, line 24: Can we just make this a template over the fixed-size array so that we can have static_asserts for lengths? Not feeling safe about `assert(size > 1)` and `size - 1`.

include/fizzy/fizzy_errors.h
gumb0: You mean having only 3 codes for parser, validation and instantiate vs a code for each error?

I think a code for each error is useful, but as a first version we could make a simple version with 3 errors, then later go deeper.

include/fizzy/fizzy.h
gumb0: I wanted to separate it because it is used in `lib/fizzy/exceptions.hpp`.

But now I'm thinking it might be not needed for the fist version, if we just make our 3 exception classes mean 3 possible error codes.

gumb0: Error was like this:
```
 --- stderr
  fizzy/include/fizzy/fizzy.h:10:10: fatal error: 'fizzy/fizzy_errors.h' file not found
  fizzy/include/fizzy/fizzy.h:10:10: fatal error: 'fizzy/fizzy_errors.h' file not found, err: true
  thread 'main' panicked at 'Unable to generate bindings: ()', bindings/rust/build.rs:33:10
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

axic: Probably need to add some include path into `build.rs`.

axic:
@gumb0 can this be closed then?

gumb0:
Closing in favor of #772

PR #769, opened 2021-03-23
rust: support rich errors

axic:
Depends on #772.

### DIFF
--- a/bindings/rust/src/lib.rs
+++ b/bindings/rust/src/lib.rs
@@ -37,8 +37,52 @@
 
 mod sys;
 
-use std::ffi::CString;
+use std::ffi::{CStr, CString};
 use std::ptr::NonNull;
+
+/// A safe container for handling the low-level FizzyError struct.
+struct FizzyErrorBox(Box<sys::FizzyError>);
+
+impl FizzyErrorBox {
+    /// Create a safe, boxed, and zero initialised container.
+    fn new() -> Self {
+        FizzyErrorBox {
+            0: Box::new(sys::FizzyError {
+                code: 0,
+                message: [0i8; 256],
+            }),
+        }
+    }
+
+    /// Return a pointer passable to low-level functions (validate, parse, instantiate).
+    ///
+    /// # Safety
+    /// The returned pointer is still onwed by this struct, and thus not valid after this struct goes out of scope.
+    unsafe fn as_mut_ptr(&mut self) -> *mut sys::FizzyError {
+        &mut *self.0
+    }
+
+    /// Return the underlying error code.
+    // TODO: represent the errors as proper Rust enums
+    fn code(&self) -> u32 {
+        self.0.code
+    }
+
+    /// Return an owned String copy of the underlying message.
+    fn message(&self) -> String {
+        unsafe {
+            CStr::from_ptr(self.0.message.as_ptr())
+                .to_string_lossy()
+                .into_owned()
+        }
+    }
+}
+
+impl std::fmt::Display for FizzyErrorBox {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "{} [{}]", self.code(), self.message())
+    }
+}
 
 /// Parse and validate the input according to WebAssembly 1.0 rules. Returns true if the supplied input is valid.
 pub fn validate<T: AsRef<[u8]>>(input: T) -> bool {
@@ -452,6 +496,15 @@ impl Instance {
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    #[test]
+    fn error_box() {
+        let mut err = FizzyErrorBox::new();
+        assert_ne!(unsafe { err.as_mut_ptr() }, std::ptr::null_mut());
+        assert_eq!(err.code(), 0);
+        assert_eq!(err.message(), "");
+        assert_eq!(format!("{}", err), "0 []");
+    }
 
     #[test]
     fn value_conversions() {
--- a/bindings/rust/src/lib.rs
+++ b/bindings/rust/src/lib.rs
@@ -126,16 +126,19 @@ impl Clone for Module {
 
 /// Parse and validate the input according to WebAssembly 1.0 rules.
 pub fn parse<T: AsRef<[u8]>>(input: &T) -> Result<Module, String> {
+    let mut err = FizzyErrorBox::new();
     let ptr = unsafe {
         sys::fizzy_parse(
             input.as_ref().as_ptr(),
             input.as_ref().len(),
-            std::ptr::null_mut(),
+            err.as_mut_ptr(),
         )
     };
     if ptr.is_null() {
-        return Err("parsing failure".to_string());
+        debug_assert!(err.code() != 0);
+        return Err(err.message());
     }
+    debug_assert!(err.code() == 0);
     Ok(Module { 0: ptr })
 }
 
@@ -707,7 +710,7 @@ mod tests {
             parse(&[0x00, 0x61, 0x73, 0x6d, 0x01, 0x00, 0x00, 0x01])
                 .err()
                 .unwrap(),
-            "parsing failure"
+            "invalid wasm module prefix"
         );
     }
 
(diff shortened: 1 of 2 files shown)


bindings/rust/src/lib.rs
```diff
@@ -42,7 +42,20 @@ use std::ptr::NonNull;
 
 /// Parse and validate the input according to WebAssembly 1.0 rules. Returns true if the supplied input is valid.
 pub fn validate<T: AsRef<[u8]>>(input: T) -> bool {
-    unsafe { sys::fizzy_validate(input.as_ref().as_ptr(), input.as_ref().len()) }
+    let err = Box::new(sys::FizzyError {
+        code: 0,
+        message: [0i8; 256],
+    });
+    let err = Box::into_raw(err); // drop ownership
```
axic, line 49: It is possible to merge these two lines, but ideally if we there would be a way to retrieve the pointer without drop/claim ownership that would be nicer. Obviously this can be pulled out into a helper too.

bindings/rust/src/lib.rs
```diff
@@ -42,7 +42,19 @@ use std::ptr::NonNull;
 
 /// Parse and validate the input according to WebAssembly 1.0 rules. Returns true if the supplied input is valid.
 pub fn validate<T: AsRef<[u8]>>(input: T) -> bool {
-    unsafe { sys::fizzy_validate(input.as_ref().as_ptr(), input.as_ref().len()) }
+    let mut err = Box::new(sys::FizzyError {
+        code: 0,
+        message: [0i8; 256],
+    });
+    let ret =
+        unsafe { sys::fizzy_validate(input.as_ref().as_ptr(), input.as_ref().len(), &mut *err) };
```
axic, line 50: Apparently `&mut *err` is the suggested option: https://doc.rust-lang.org/std/primitive.pointer.html#1-coerce-a-reference-t-or-mutable-reference-mut-t

bindings/rust/src/lib.rs
axic: Pulled into a helper.

bindings/rust/src/lib.rs
```diff
@@ -40,14 +40,65 @@ mod sys;
 use std::ffi::CString;
 use std::ptr::NonNull;
 
+/// A safe container for handling the low-level FizzyError struct.
+struct CErrorContainer(Box<sys::FizzyError>);
```
axic, line 44: Any better name for this? CE looks like the OE letter.

bindings/rust/src/lib.rs
```diff
@@ -40,14 +40,66 @@ mod sys;
...
+    fn message(&self) -> String {
+        unsafe {
+            std::ffi::CStr::from_ptr(self.0.message.as_ptr())
+                .to_string_lossy()
+                .into_owned()
+        }
+    }
+}
+
+impl std::fmt::Display for CErrorContainer {
```
axic, line 81: This is not currently used, but it enables things like `println!("{:?}", err)` or `err.to_string()`. Probably should remove.

bindings/rust/src/lib.rs
```diff
@@ -40,14 +40,66 @@ mod sys;
...
+    let ret = unsafe {
         sys::fizzy_validate(
             input.as_ref().as_ptr(),
             input.as_ref().len(),
-            std::ptr::null_mut(),
+            err.as_mut_ptr(),
         )
+    };
+    if ret {
+        debug_assert!(err.code() == 0);
```
axic, line 98: These assertions show the current error reporting system is not prefect, as there are two ways to tell failure, error code or the return value.

bindings/rust/src/lib.rs
gumb0: `FizzyErrorContrainer`, `FizzyErrorBox`

bindings/rust/src/lib.rs
gumb0: Sounds useful, can there be a test for it?

bindings/rust/src/lib.rs
```diff
@@ -40,14 +40,66 @@ mod sys;
...
+/// A safe container for handling the low-level FizzyError struct.
+struct CErrorContainer(Box<sys::FizzyError>);
+
+impl CErrorContainer {
+    /// Create a safe, boxed, and zero initialised container.
+    fn new() -> Self {
+        CErrorContainer {
+            0: Box::new(sys::FizzyError {
+                code: 0,
+                message: [0i8; 256],
```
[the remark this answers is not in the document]
axic: The 0 in `0i8` means it initialises with zero. Weird notation, but the numerical value (can be non-zero) before the type means the default value it is initialised with. It is needed to be explicitly defined here.

bindings/rust/src/lib.rs
gumb0: For validate theoretically we could return error code instead of `bool` (and optional error message), but for parse it seems inevitable: we have to return a pointer to module / instance in any case, and when we also return an error code this pointer is going to be NULL.

axic: Do we want to add assertions into the capi for this?

bindings/rust/src/lib.rs
axic: Added.

bindings/rust/src/lib.rs
axic: Changed.

bindings/rust/src/lib.rs
```diff
@@ -40,14 +40,66 @@ mod sys;
...
 
+/// A safe container for handling the low-level FizzyError struct.
+struct FizzyErrorBox(Box<sys::FizzyError>);
+
+impl FizzyErrorBox {
+    /// Create a safe, boxed, and zero initialised container.
+    fn new() -> Self {
+        FizzyErrorBox {
+            0: Box::new(sys::FizzyError {
+                code: 0,
```
gumb0, line 51: can you use here `sys::FIZZY_SUCCESS`?

bindings/rust/src/lib.rs
gumb0: > Do we want to add assertions into the capi for this? Or at least extend the tests? And document it?

Tests already have checks for both return value and error for both failure and success cases.

Assertions in C API - not sure where and which ones (it seems kind of obvious there that error is set when `false` or `nullptr` is returned)

Document it - something like`Returned error code is FIZZY_SUCCESS iff return value is true`? Sounds somewhat obvious, too...

bindings/rust/src/lib.rs
```diff
@@ -453,6 +513,15 @@ impl Instance {
 mod tests {
     use super::*;
 
+    #[test]
+    fn error_box() {
+        let mut err = FizzyErrorBox::new();
+        assert_ne!(unsafe { err.as_mut_ptr() }, std::ptr::null_mut());
+        assert_eq!(err.code(), 0);
+        assert_eq!(err.message(), "");
+        assert_eq!(format!("{}", err), "0 []");
```
gumb0, line 522: Is it possible to have a test also for `format!`  of failure?

bindings/rust/src/lib.rs
axic: I prefer keeping zeroes, because all I want to do here is zero initialised it, i.e. `memset(&err, 0, sizeof(err))`.

bindings/rust/src/lib.rs
axic: Not really. That would mean doing some pointer magic to set the contents.

bindings/rust/src/lib.rs
```diff
@@ -622,18 +682,20 @@ mod tests {
...
         // Too short
-        assert_eq!(validate(&[0x00]), false);
-        // Valid
         assert_eq!(
-            validate(&[0x00, 0x61, 0x73, 0x6d, 0x01, 0x00, 0x00, 0x00]),
-            true
+            validate(&[0x00]).err().unwrap(),
+            "invalid wasm module prefix"
         );
+        // Valid
```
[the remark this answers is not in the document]
axic: I would skip this refactoring in this PR.

gumb0 (approved):

bindings/rust/src/lib.rs
axic: I guess documentation is fine just saying that `error code must be FIZZY_SUCCESS` if a pointer is returned and `must NOT be FIZZY_SUCCESS` otherwise?

chfast (approved):

PR #770, opened 2021-03-24
cmake: Simplify compiler options with Cable

chfast:


CMakeLists.txt
```diff
@@ -37,21 +38,22 @@ if(is_big_endian)
...
 
 cable_configure_compiler()
-add_compile_options(
+cable_add_compile_options(
     -Wcast-qual
     -Wcast-align
     -Wmissing-declarations
-    $<$<COMPILE_LANGUAGE:CXX>:-Wextra-semi>
-    $<$<COMPILE_LANGUAGE:CXX>:-Wold-style-cast>
+    IF_SUPPORTED
```
axic, line 45: So anything below the `IF_SUPPORTED` magic is only enabled if the compiler supports it?

How about `ONLY_IF_SUPPORTED`?

CMakeLists.txt
```diff
@@ -37,21 +38,22 @@ if(is_big_endian)
 endif()
 
 cable_configure_compiler()
-add_compile_options(
+cable_add_compile_options(
     -Wcast-qual
     -Wcast-align
     -Wmissing-declarations
-    $<$<COMPILE_LANGUAGE:CXX>:-Wextra-semi>
-    $<$<COMPILE_LANGUAGE:CXX>:-Wold-style-cast>
```
axic, line 45: Does `COMPILE_LANGUAGE:CXX` means if the file is cpp or if the compiler is run in c++ mode?

cmake/CableCompileOptions.cmake
```diff
@@ -0,0 +1,84 @@
...
+        message("Checking ${flag}")
+        string(MAKE_C_IDENTIFIER ${flag} flag_id)
+
+        set(supported_in_all_languages TRUE)
+
+        foreach(lang ${languages})
+            set(result_var "${lang}${flag_id}")
+
+            # Check if the flag works in the lang's compiler.
+            # In CMake 3.18+ cmake_language(CALL ...) can be used.
```
axic, line 54: This means the file does not require 3.18+? What version is required, can we put a min version on the top?

CMakeLists.txt
chfast: Yes, these are optional and enabled if compiler of a particular language supports it.

CMakeLists.txt
chfast: It means C++ compilation determined by CMake by the source file extension (the fact that this is the same tool for C and C++ is only  coincidence).

cmake/CableCompileOptions.cmake
chfast: I don't know what the minimal CMake version required is. Hard to tell without testing it with multiple versions. And I don't think it is possible to put `cmake_minimum_required()` in this file because this will affect the project that loads this file.

CMakeLists.txt
```diff
@@ -37,21 +38,22 @@ if(is_big_endian)
...
-cable_add_cxx_compiler_flag_if_supported(-Wnewline-eof)
-cable_add_cxx_compiler_flag_if_supported(-Wsuggest-destructor-override)
-cable_add_cxx_compiler_flag_if_supported(-Wunreachable-code-break)
-cable_add_cxx_compiler_flag_if_supported(-Wduplicated-cond)
-cable_add_cxx_compiler_flag_if_supported(-Wduplicate-enum)
-cable_add_cxx_compiler_flag_if_supported(-Wlogical-op)
-cable_add_cxx_compiler_flag_if_supported(-Wno-unknown-attributes)
 
 if(CMAKE_CXX_COMPILER_ID MATCHES Clang)
     option(WEVERYTHING "Enable almost all compiler warnings" OFF)
```
axic, line 59: Would it make sense having another helper/case, i.e.
```
cable_add_compile_option(
  -Weverything
  IF_PRESENT # If the first option(s) works try adding all these?
  -Wno-c++98-compat
  ...
```

axic (approved):

CMakeLists.txt
gumb0: Not very easy to guess what it does IMHO

chfast: What exactly?

gumb0: @axic's suggestion. Your change looks fine I think.

axic:
Since the holdup here is to check if this works with Solidity, there we'd need something like this:
```
cable_add_compile_options(
        IF_SUPPORTED
        -Wimplicit-fallthrough
        # Prevent the path of the source directory from ending up in the binary via __FILE__ macros.
        -fmacro-prefix-map=${CMAKE_SOURCE_DIR}=/solidity
        # -Wpessimizing-move warns when a call to std::move would prevent copy elision
        # if the argument was not wrapped in a call.  This happens when moving a local
        # variable in a return statement when the variable is the same type as the
        # return type or using a move to create a new object from a temporary object.
        -Wpessimizing-move
        # -Wredundant-move warns when an implicit move would already be made, so the
        # std::move call is not needed, such as when moving a local variable in a return
        # that is different from the return type.
        -Wredundant-move
)
```

Would this work?

axic:
And how would you simplify the following using the new option:
```
        eth_add_cxx_compiler_flag_if_supported(-fstack-protector-strong have_stack_protector_strong_support)
        if(NOT have_stack_protector_strong_support)
                eth_add_cxx_compiler_flag_if_supported(-fstack-protector)
        endif()
```

chfast:
> Would this work?

This work.

> And how would you simplify the following using the new option:
> 
> ```
>         eth_add_cxx_compiler_flag_if_supported(-fstack-protector-strong have_stack_protector_strong_support)
>         if(NOT have_stack_protector_strong_support)
>                 eth_add_cxx_compiler_flag_if_supported(-fstack-protector)
>         endif()
> ```

This cannot be replaced.

axic:
> This cannot be replaced.

Wouldn't this be a correct one?
```
        if(have_stack_protector_strong_support)
                add_compile_options(-fstack-protector-strong)
        else()
                cable_add_compile_options(IF_SUPPORTED -fstack-protector)
        endif()
```

chfast:
This is not documented, but `-fstack-protector` overrides the `-fstack-protector-strong`. Still you can safely do `-fstack-protector -fstack-protector-strong`.
https://godbolt.org/z/xPsW179eK

So simplified version would be
```cmake
cable_add_compile_options(IF_SUPPORTED -fstack-protector -fstack-protector-strong)
```

chfast:
Other good news is that `-fstack-protector-strong` is supported in all GCC/Clang versions that support `-std=c++17` so you can ignore `-fstack-protector` if you want `-fstack-protector-strong`.

axic:
Here are the potential changes/simplifications in Solidity: https://github.com/ethereum/solidity/tree/cmake-options

PR #771, opened 2021-03-25
Expose ExecutionContext to C

chfast:


test/unittests/capi_test.cpp
```diff
@@ -330,9 +330,8 @@ TEST(capi, find_exported_function)
     EXPECT_EQ(function.type.output, FizzyValueTypeI32);
     EXPECT_NE(function.context, nullptr);
     ASSERT_NE(function.function, nullptr);
-    fizzy::ExecutionContext ctx;
-    auto* const c_ctx = reinterpret_cast<FizzyExecutionContext*>(&ctx);
-    EXPECT_THAT(function.function(function.context, instance, nullptr, c_ctx), CResult(42_u32));
+    FizzyExecutionContext ctx{};
```
chfast, line 333: This is in theory undefined behavior because C++ expects `fizzy::ExecutionContext` object being created. But it will work in practice. There is also fancier way of handling this with additional `fizzy_execution_context_init(FizzyExecutionContext*)`.

axic: If we were to support this, I'd vote for `fizzy_execution_context_init`.

PR #773, opened 2021-03-26
test: Add out-of-memory tests

chfast:


test/unittests/oom_test.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+#if defined(__SANITIZE_ADDRESS__) || __has_feature(address_sanitizer)
+#define ASAN 1
+#else
+#define ASAN 0
+#endif
+
+namespace
+{
+constexpr uint32_t OSMemoryLimit = 1024 * 1024;
```
gumb0, line 28: What unit is this?

test/unittests/oom_test.cpp
```diff
@@ -0,0 +1,92 @@
...
+bool restore_memory_limit() noexcept;
+
+#if !ASAN && defined(__linux__) && (defined(__x86_64__) || defined(__i386__))
+
+rlimit orig_limit;
+
+bool try_set_memory_limit(size_t size) noexcept
+{
+    [[maybe_unused]] int err;
+    err = getrlimit(RLIMIT_AS, &orig_limit);
```
gumb0, line 40:
```suggestion
    [[maybe_unused]] int err = getrlimit(RLIMIT_AS, &orig_limit);
```

test/unittests/oom_test.cpp
gumb0: I guess it's bytes. Add a comment or call it `OSMemoryLimitBytes`.

test/unittests/oom_test.cpp
axic, line 40: We could also have some stack tests using `RLIMIT_STACK`, i.e. having a known limit.

axic, line 40: I wonder how `RLIMIT_AS` is different from `RLIMIT_DATA` (the latter being available on macos too):
```
       RLIMIT_AS
              This is the maximum size of the process's virtual memory
              (address space).  The limit is specified in bytes, and is
              rounded down to the system page size.  This limit affects
              calls to brk(2), mmap(2), and mremap(2), which fail with
              the error ENOMEM upon exceeding this limit.  In addition,
              automatic stack expansion fails (and generates a SIGSEGV
              that kills the process if no alternate stack has been made
              available via sigaltstack(2)).  Since the value is a long,
              on machines with a 32-bit long either this limit is at
              most 2 GiB, or this resource is unlimited.
```

```
       RLIMIT_DATA
              This is the maximum size of the process's data segment
              (initialized data, uninitialized data, and heap).  The
              limit is specified in bytes, and is rounded down to the
              system page size.  This limit affects calls to brk(2),
              sbrk(2), and (since Linux 4.7) mmap(2), which fail with
              the error ENOMEM upon encountering the soft limit of this
              resource.
```

chfast: I don't know.

test/unittests/oom_test.cpp
```diff
@@ -25,7 +25,7 @@
 
 namespace
 {
-constexpr uint32_t OSMemoryLimit = 1024 * 1024;
+constexpr uint32_t OSMemoryLimit = 2 * 1024 * 1024;
```
chfast, line 28: If this is enough only for running single test, make it bigger, e.g. 10M.

gumb0: No, for me it passes everything with this.

gumb0: Single test was passing with 1Mb.

test/unittests/oom_test.cpp
chfast: It is `RLIMIT_DATA` but it does not apply to `mmap` on macos. Please test it and report back.

test/unittests/oom_test.cpp
```diff
@@ -0,0 +1,118 @@
...
+    */
+    const auto wasm = from_hex("0061736d010000000503010020");
+
+    const auto is_limited = try_set_memory_limit(OSMemoryLimitBytes);
+
+    if (is_limited)
+        EXPECT_THROW(instantiate(parse(wasm), {}, {}, {}, {}, MaxMemoryPagesLimit), std::bad_alloc);
+    else
+        EXPECT_NO_THROW(instantiate(parse(wasm), {}, {}, {}, {}, MaxMemoryPagesLimit));
+}
```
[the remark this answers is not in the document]
chfast: It should.

test/unittests/oom_test.cpp
axic: I tested this on macOS. Added `__APPLE__` next to `__linux__`. Everything compiles fine, but it does not work.

1. Tried setting not only `cur`, but `max`. No change.
2. Apparently `RLIMIT_AS` is also available on macOS, just not part of the man page. ([ref](https://github.com/apple/darwin-xnu/blob/main/bsd/sys/resource.h#L474)). Tried, compiles, but no change.
3. It seems that setrlimit/getrlimit is properly implemented in macOS ([ref](https://github.com/apple/darwin-xnu/blob/main/bsd/kern/kern_resource.c#L937)), but `RLIMIT_DATA` or `RLIMIT_AS` is actually not used by anything later on. `RLIMIT_FSIZE/CPU/NPROC/NOFILE/CORE/STACKMEMLOCK` are used in places, so those may be enforced.

See also https://stackoverflow.com/a/6169866 for potential solutions.

test/unittests/oom_test.cpp
```diff
@@ -0,0 +1,118 @@
...
+#define ASAN 0
+#endif
+
+namespace
+{
+/// The OS-specific memory limit in bytes.
+/// Exact meaning depends on OS.
+/// - For Linux this is virtual address space limit of the process (i.e. fizzy-unittests).
+///   The value must be multiple of the OS page size (4096) and bigger than the current usage
+///   at the point of setting the limit (otherwise it has not effect).
```
gumb0, line 36:
```suggestion
///   at the point of setting the limit (otherwise it returns success, but has not effect).
```

gumb0 (approved):

axic:
Can we squash some of these commits?

axic (approved):

PR #772, opened 2021-03-25
capi: Reporting error code and message

gumb0:
A simpler alternative to #764, not introducing any error codes on C++ side for now.

PR #775, opened 2021-03-29
capi: mark all external functions noexcept

axic:
Addresses https://github.com/wasmx/fizzy/pull/772#discussion_r603309139.
Depends on #776.

### DIFF
--- a/include/fizzy/fizzy.h
+++ b/include/fizzy/fizzy.h
@@ -10,6 +10,13 @@
 #include <stddef.h>
 #include <stdint.h>
 
+/// Safe way of marking a function with `noexcept` C++ specifier.
+#ifdef __cplusplus
+#define FIZZY_NOEXCEPT noexcept
+#else
+#define FIZZY_NOEXCEPT
+#endif
+
 #ifdef __cplusplus
 extern "C" {
 #endif
@@ -81,8 +88,8 @@ typedef struct FizzyExecutionContext FizzyExecutionContext;
 /// @note
 /// External functions implemented in C++ must be non-throwing, i.e. the effect of any exception
 /// escaping the function is std::terminate being called.
-typedef FizzyExecutionResult (*FizzyExternalFn)(
-    void* host_ctx, FizzyInstance* instance, const FizzyValue* args, FizzyExecutionContext* ctx);
+typedef FizzyExecutionResult (*FizzyExternalFn)(void* host_ctx, FizzyInstance* instance,
+    const FizzyValue* args, FizzyExecutionContext* ctx) FIZZY_NOEXCEPT;
 
 /// Value type.
 typedef uint8_t FizzyValueType;
@@ -243,7 +250,8 @@ typedef struct FizzyImportedGlobal
 /// @param  error               Pointer to store detailed error information at. Can be NULL if error
 ///                             information is not required.
 /// @return                     true if module is valid, false otherwise.
-bool fizzy_validate(const uint8_t* wasm_binary, size_t wasm_binary_size, FizzyError* error);
+bool fizzy_validate(
+    const uint8_t* wasm_binary, size_t wasm_binary_size, FizzyError* error) FIZZY_NOEXCEPT;
 
 /// Parse binary module.
 ///
@@ -253,7 +261,7 @@ bool fizzy_validate(const uint8_t* wasm_binary, size_t wasm_binary_size, FizzyEr
 ///                             information is not required.
 /// @return                     non-NULL pointer to module in case of success, NULL otherwise.
 const FizzyModule* fizzy_parse(
-    const uint8_t* wasm_binary, size_t wasm_binary_size, FizzyError* error);
+    const uint8_t* wasm_binary, size_t wasm_binary_size, FizzyError* error) FIZZY_NOEXCEPT;
 
 /// Free resources associated with the module.
 ///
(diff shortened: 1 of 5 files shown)


lib/fizzy/capi.cpp
```diff
@@ -576,17 +578,17 @@ FizzyInstance* fizzy_resolve_instantiate(const FizzyModule* c_module,
     }
 }
 
-void fizzy_free_instance(FizzyInstance* instance)
+void fizzy_free_instance(FizzyInstance* instance) noexcept
```
axic, line 581: Are we saving anything with marking all these noexcept?

If I understand correctly:
- before this, if used in C, an exception just causes UB (abort effectively), while in C++ it is passed through
- after this, if used in C and C++, an exception is a clear abort (however we generate exception handling code in each of these functions?)

lib/fizzy/capi.cpp
```diff
@@ -473,7 +475,7 @@ bool fizzy_find_exported_function_index(
 }
 
 bool fizzy_find_exported_function(
-    FizzyInstance* instance, const char* name, FizzyExternalFunction* out_function)
+    FizzyInstance* instance, const char* name, FizzyExternalFunction* out_function) noexcept
```
gumb0, line 478: `wrap` call below can throw `bad_alloc`, we should add `catch(...) { return false; } `  for now, later adding rich error reporting to it or getting rid of dynamic allocation somehow.

gumb0: I wonder why clang-tidy didn't complain about this...

lib/fizzy/capi.cpp
```diff
@@ -404,65 +405,66 @@ const FizzyModule* fizzy_clone_module(const FizzyModule* module)
...
+    const FizzyModule* module, uint32_t export_idx) noexcept
 {
     return wrap(unwrap(module)->exportsec[export_idx]);
 }
 
 bool fizzy_find_exported_function_index(
-    const FizzyModule* module, const char* name, uint32_t* out_func_idx)
+    const FizzyModule* module, const char* name, uint32_t* out_func_idx) noexcept
 {
     const auto optional_func_idx = fizzy::find_exported_function(*unwrap(module), name);
```
gumb0, line 469: `fizzy::find_exported_function` is not marked `noexcept`, I'm not sure exactly why, perhaps because `std::make_optional` in it is not `noexcept` (it can throw in case underlying object's contructor throws, which is not our case, we have `optional<uint32_t>`)

I would expect here that clang-tidy would complain that `fizzy_find_exported_function_index` can't be made `noexcept` either... not sure why it doesn't.

gumb0: I think we should try making `find_exported_*` functions on C++ side `noexcept`.

lib/fizzy/capi.cpp
axic: Added.

lib/fizzy/capi.cpp
axic: Fixed by #776.

test/utils/fizzy_c_engine.cpp
```diff
@@ -29,7 +29,7 @@ class FizzyCEngine final : public WasmEngine
 namespace
 {
 FizzyExecutionResult env_adler32(
-    void*, FizzyInstance* instance, const FizzyValue* args, FizzyExecutionContext*)
+    void*, FizzyInstance* instance, const FizzyValue* args, FizzyExecutionContext*) noexcept
```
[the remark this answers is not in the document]
axic: Addressed. It seems to show some minor speed improvement too?

lib/fizzy/capi.cpp
```diff
@@ -205,7 +205,7 @@ inline FizzyExternalFunction wrap(fizzy::ExternalFunction external_func)
 {
     static constexpr FizzyExternalFn c_function =
         [](void* host_ctx, FizzyInstance* instance, const FizzyValue* args,
-            FizzyExecutionContext* c_ctx) -> FizzyExecutionResult {
+            FizzyExecutionContext* c_ctx) noexcept -> FizzyExecutionResult {
```
gumb0, line 208: It's a good change, but the function can throw `bad_alloc`.

Should it perhaps trap in case new context can't be allocated?

axic: Well the underlying host function API in C++ is noexcept, so it can't do much.

Not sure if trap or `std::terminate` is nicer :)

The latter may help the programmer to find the issue more easily?

axic: Should really try to make these functions dynamic allocation free.

gumb0 (approved):
I'm fine with leaving exported functions to `std::terminate` in case of context allocation failure for now.

include/fizzy/fizzy.h
```diff
@@ -10,6 +10,13 @@
 #include <stddef.h>
 #include <stdint.h>
 
+/// Safe way of marking a function with `noexcept` C++ specifier.
+#ifdef __cplusplus
+#define FIZZY_NOEXCEPT noexcept
```
chfast, line 15: I think `NOEXCEPT` would be fine provided it is wrapped with `#ifndef NOEXCEPT`.

axic: Are you keen on this? This is a public header, so who knows what projects could define `NOEXCEPT` as. This feels a bit safer -- less like to clash.

chfast (approved):

PR #776, opened 2021-03-31
api: find_exported_* improvements

gumb0:


lib/fizzy/instantiate.cpp
```diff
@@ -305,7 +305,8 @@ ExternalGlobal find_imported_global(const std::string& module, const std::string
     return {it->value, module_global_type};
 }
 
-std::optional<uint32_t> find_export(const Module& module, ExternalKind kind, std::string_view name)
+std::optional<uint32_t> find_export(
```
axic, line 308: Are we sure that none of the constructors used would throw through `std::make_optional`? Can clang-tidy verify that at all?

lib/fizzy/instantiate.cpp
```diff
@@ -535,7 +539,8 @@ std::optional<ExternalTable> find_exported_table(Instance& instance, std::string
     return ExternalTable{instance.table.get(), instance.table_limits};
```
axic, line 539: Can `.get()` not throw?

chfast (approved):

lib/fizzy/instantiate.cpp
gumb0: It's `noexcept` https://en.cppreference.com/w/cpp/memory/unique_ptr/get

lib/fizzy/instantiate.cpp
axic: Perhaps we could add a test marked noexcept constructing all those, just to have an early indication if we brake something that clang-tidy can't detect.

gumb0: I've added `noexcept` to some constructors more, and I'm pretty sure, all `External*` objects use either explicitly `noexcept` or default move-constructors, which are noexcept implicitly.

gumb0: > Perhaps we could add a test marked noexcept constructing all those, just to have an early indication if we brake something that clang-tidy can't detect.

How would we detect it in test, if clang-tidy doesn't detect it? Well setting `terminate_handler` might work...

gumb0: If you're worried about `make_optional` only, it's used only with `uint32_t`.
Other functions construct an object and then directly construct `optional` return value.

axic: I'm just not convinced anymore that anything wrong is detected when we mark stuff noexcept 😕

gumb0: Yeah now it seems to me too, that clang-tidy doesn't check much.

axic: Also apparently `std::find_if` can throw: https://en.cppreference.com/w/cpp/algorithm/find

lib/fizzy/instantiate.cpp
```diff
@@ -504,7 +507,8 @@ std::optional<ExternalFunction> find_exported_function(Instance& instance, std::
         ExecuteFunction(instance, idx), instance.module->get_function_type(idx)};
```
axic, line 507: Here we use `*opt_index`. It is weird that `operator*` is not marked noexcept, but docs say "throws nothing". Compared to that `->value()` explicitly checks and throws:

- https://en.cppreference.com/w/cpp/utility/optional/operator*
- https://en.cppreference.com/w/cpp/utility/optional/value

lib/fizzy/instantiate.cpp
```diff
@@ -504,7 +507,8 @@ std::optional<ExternalFunction> find_exported_function(Instance& instance, std::
...
 }
 
-std::optional<ExternalGlobal> find_exported_global(Instance& instance, std::string_view name)
+std::optional<ExternalGlobal> find_exported_global(
```
axic, line 510: This function also uses `operator*`.

lib/fizzy/instantiate.cpp
gumb0: I guess it makes sense to use manual loop instead of `find_if` then.

lib/fizzy/instantiate.cpp
gumb0: I found some explanation, but it's not clear to me https://stackoverflow.com/a/63913833

lib/fizzy/instantiate.cpp
gumb0: > Also apparently `std::find_if` can throw: https://en.cppreference.com/w/cpp/algorithm/find

Actually, it's only overloads with `ExecutionPolicy` can throw.

gumb0: I think regular `find_if` is not `noexcept`, because the predicate function is allowed to throw.

I think it's safe to leave `noexcept` with `find_if` here, but I made labmda `noexcept`, too.

axic: > > Also apparently std::find_if can throw: https://en.cppreference.com/w/cpp/algorithm/find
>
> Actually, it's only overloads with ExecutionPolicy can throw.

I am not sure, that description is kind of confusing. Makes me think it can throw `bad_alloc` in any case.

lib/fizzy/instantiate.cpp
axic: C++ is weird.

axic (approved):

lib/fizzy/instantiate.cpp
chfast: First of all `noexcept` is not like `const` that all internal function calls must be `noexcept`. Usage of `find_if` looks fine to me.

> Makes me think it can throw `bad_alloc` in any case.

Only overloads with `ExecutionPolicy` which we don't use.

`clang-tidy` only reports obvious mistakes (as direct usage of `throw` inside `noexcept`). We can dig deeper if you know the check name that fails.

chfast (approved):

gumb0:
> clang-tidy only reports obvious mistakes (as direct usage of throw inside noexcept). We can dig deeper if you know the check name that fails.

Weird that it does detect a call to `bytes::resize()` in `grow_memory` (`bugprone-exception-escape` check), but doesn't detect `make_unique` there https://github.com/wasmx/fizzy/pull/775/files#r603979857

PR #777, opened 2021-03-31
Execution context stack space

chfast:
This implements segmented stack space inside ExecutionContext. I will add more documentation how it works later.
Replaces #529 and #572.

Generally, having unlimited stack space is PITA, but doable.

Before this lands I propose to generalize `ExecutionContext`:
- move it to `execution_context.hpp`,
- rename `Guard` to something like `LocalContext`,
- rename `increment_call_depth()` to something like `create_local_context()`.

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -579,8 +579,7 @@ ExecutionResult execute(
 
     const auto local_ctx = ctx.create_local_context(required_stack_space);
 
-    OperandStack stack(
-        args, args_count, code.local_count, static_cast<size_t>(code.max_stack_height));
+    OperandStack stack(args, args_count, code.local_count, local_ctx.stack_space);
 
     const uint8_t* pc = code.instructions.data();
 
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -565,6 +565,7 @@ ExecutionResult execute(
         return Trap;
 
     const auto& func_type = instance.module->get_function_type(func_idx);
+    const auto args_count = func_type.inputs.size();
 
     assert(instance.module->imported_function_types.size() == instance.imported_functions.size());
     if (func_idx < instance.imported_functions.size())
@@ -573,10 +574,13 @@ ExecutionResult execute(
     const auto& code = instance.module->get_code(func_idx);
     auto* const memory = instance.memory.get();
 
-    const auto local_ctx = ctx.create_local_context();
+    const auto required_stack_space =
+        args_count + code.local_count + static_cast<size_t>(code.max_stack_height);
 
-    OperandStack stack(args, func_type.inputs.size(), code.local_count,
-        static_cast<size_t>(code.max_stack_height));
+    const auto local_ctx = ctx.create_local_context(required_stack_space);
+
+    OperandStack stack(
+        args, args_count, code.local_count, static_cast<size_t>(code.max_stack_height));
 
     const uint8_t* pc = code.instructions.data();
 
--- a/lib/fizzy/execution_context.hpp
+++ b/lib/fizzy/execution_context.hpp
@@ -42,10 +42,8 @@ class ExecutionContext
             if (required_stack_space <= m_shared_ctx.free_stack_space)
             {
                 // Must be a segment of default size or required_stack_space is 0.
-                const auto offset =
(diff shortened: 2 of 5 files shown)


chfast:
Unit tests will be updated later.

lib/fizzy/execute.cpp
```diff
@@ -573,10 +574,12 @@ ExecutionResult execute(
     const auto& code = instance.module->get_code(func_idx);
     auto* const memory = instance.memory.get();
 
-    const auto local_ctx = ctx.create_local_context();
+    const auto required_stack_space =
+        args_count + code.local_count + static_cast<size_t>(code.max_stack_height);
 
-    OperandStack stack(args, func_type.inputs.size(), code.local_count,
-        static_cast<size_t>(code.max_stack_height));
+    const auto local_ctx = ctx.create_local_context(required_stack_space);
```
gumb0, line 580: Should handle somehow that this can throw now.

lib/fizzy/execution_context.hpp
```diff
@@ -4,38 +4,87 @@
...
+
+            prev_free_stack_space = m_shared_ctx.free_stack_space;
+
+            if (required_stack_space <= m_shared_ctx.free_stack_space)
+            {
+                // Must be a segment of default size or required_stack_space is 0.
+                const auto offset =
+                    DefaultStackSpaceSegmentSize - m_shared_ctx.free_stack_space;
+                stack_space = m_shared_ctx.stack_space_segment + offset;
+                prev_free_stack_space = m_shared_ctx.free_stack_space;
```
gumb0, line 48: This looks duplicated.

lib/fizzy/execution_context.hpp
```diff
@@ -4,38 +4,87 @@
...
+                m_shared_ctx.free_stack_space -= required_stack_space;
+            }
+            else
+            {
+                prev_stack_space_segment = m_shared_ctx.stack_space_segment;
+                const auto new_segment_size =
+                    std::max(DefaultStackSpaceSegmentSize, required_stack_space);
+                m_shared_ctx.stack_space_segment = new Value[new_segment_size];
```
gumb0, line 56: It would allow a performance degrading behaviour like growing stack to almost `DefaultStackSpaceSegmentSize`, then calling a function in a loop that goes over the limit, allocates new segment and immediately deallocates after return, and that's repeated in a loop.
Should this be a concern? I can imagine mitigation with increasing allocation size each time, and keeping the last allocation size in `m_shared_ctx`.

lib/fizzy/execution_context.hpp
```diff
@@ -4,38 +4,87 @@
...
 
-        ~LocalContext() noexcept { --m_shared_ctx.depth; }
+        ~LocalContext() noexcept
+        {
+            --m_shared_ctx.depth;
+
+            m_shared_ctx.free_stack_space = prev_free_stack_space;
+            if (prev_stack_space_segment != nullptr)
+            {
+                assert(m_shared_ctx.stack_space_segment == stack_space);
```
gumb0, line 69: Maybe also
```
assert(stack_space != m_shared_ctx.first_stack_space_segment);
```

lib/fizzy/execute.cpp
chfast: It throws before too.

lib/fizzy/execution_context.hpp
chfast: It is a concern, but the same as before (where stack space was dynamically allocated in some cases).
I don't think the increasing allocation size is the solution - it will still continue to deallocating new segment, now bigger each time.
The solution I considered is never deallocate segments. They work as linked list of buffers. However, such solution might require more housekeeping.

lib/fizzy/execute.cpp
gumb0: Ah right.

chfast: To be precise: now `create_local_context()` throws, previously `OperandStack` was throwing (now `noexcept`).

Someone should handle this separately. But there is also easy solution: have fixed shared stack space. In case of out-of-space we can trap the execution.

gumb0: Catching `bad_alloc` and trapping would also be easy

chfast: But this is non-deterministic execution.

axic:
> Replaces #529 and #572.

Can we close those two?

PR #779, opened 2021-04-01
Execution context generalization

chfast:
Before #777  lands I propose to generalize `ExecutionContext`:
- move it to `execution_context.hpp`,
- rename `Guard` to something like `LocalContext`,
- rename `increment_call_depth()` to something like `create_local_context()`.

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -573,7 +573,7 @@ ExecutionResult execute(
     const auto& code = instance.module->get_code(func_idx);
     auto* const memory = instance.memory.get();
 
-    const auto local_ctx = ctx.increment_call_depth();
+    const auto local_ctx = ctx.create_local_context();
 
     OperandStack stack(args, func_type.inputs.size(), code.local_count,
         static_cast<size_t>(code.max_stack_height));
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -573,7 +573,7 @@ ExecutionResult execute(
     const auto& code = instance.module->get_code(func_idx);
     auto* const memory = instance.memory.get();
 
-    const auto ctx_guard = ctx.increment_call_depth();
+    const auto local_ctx = ctx.increment_call_depth();
 
     OperandStack stack(args, func_type.inputs.size(), code.local_count,
         static_cast<size_t>(code.max_stack_height));
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -537,7 +537,6 @@ inline bool invoke_function(const FuncType& func_type, uint32_t func_idx, Instan
     assert(stack.size() >= num_args);
     const auto call_args = stack.rend() - num_args;
 
-    const auto ctx_guard = ctx.increment_call_depth();
     const auto ret = execute(instance, func_idx, call_args, ctx);
     // Bubble up traps
     if (ret.trapped)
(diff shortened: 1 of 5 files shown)


gumb0:
Nit: the commit message `Make ExecutionContext::Guard non-copyable/non-movable` I think is wrong, as it's still movable, isnt't it?

lib/fizzy/execute.cpp
```diff
@@ -574,6 +573,8 @@ ExecutionResult execute(
     const auto& code = instance.module->get_code(func_idx);
     auto* const memory = instance.memory.get();
 
+    const auto local_ctx = ctx.create_local_context();
```
gumb0, line 576: Does moving this action from `invoke` to `execute` change the depth that host functions see?

chfast:
> Nit: the commit message `Make ExecutionContext::Guard non-copyable/non-movable` I think is wrong, as it's still movable, isnt't it?

To my understanding, move constructor is not implicitly generated if you declare a copy constructor.

gumb0:
> > Nit: the commit message `Make ExecutionContext::Guard non-copyable/non-movable` I think is wrong, as it's still movable, isnt't it?
> 
> To my understanding, move constructor is not implicitly generated if you declare a copy constructor.

I think only if you define a custom one, but I might be wrong.

chfast:
> > > Nit: the commit message `Make ExecutionContext::Guard non-copyable/non-movable` I think is wrong, as it's still movable, isnt't it?
> > 
> > 
> > To my understanding, move constructor is not implicitly generated if you declare a copy constructor.
> 
> I think only if you define a custom one, but I might be wrong.

https://godbolt.org/z/14osqo3rG

lib/fizzy/execute.cpp
chfast: No. We record the value in many tests: https://github.com/wasmx/fizzy/blob/faddd97f55ab36cb3ff5214e604707076b0e4f08/test/unittests/execute_call_depth_test.cpp#L65.

gumb0: I thought so, but why?
Before: `call` opcode calls `invoke_function`, it increments depth, calls `execute`, it calls host function.
After:  `call` opcode calls `invoke_function`, calls `execute`, it calls host function. (increment only if it's not imported)

chfast: Before: `execute(depth: D)`, `call` opcode, `invoke_function()`: increment depth, `execute(depth: D+1)`.
After: `execute(depth: D): increment depth, `call` opcode, `invoke_function()`, `execute(depth: D+1)`.

So this changes the value of the `depth` only inside the `execute()`.

This is needed because later I want to get access the `ExecutionContext` shared stack space in the same place where `depth` is incremented.

I was not able to also move the depth check to the same place easily. This may be handled later, e.g. by trapping inside the `invoke_function()`.

gumb0:
> > > > Nit: the commit message `Make ExecutionContext::Guard non-copyable/non-movable` I think is wrong, as it's still movable, isnt't it?
> > > 
> > > 
> > > To my understanding, move constructor is not implicitly generated if you declare a copy constructor.
> > 
> > 
> > I think only if you define a custom one, but I might be wrong.
> 
> https://godbolt.org/z/14osqo3rG

I would add explicit deletions of move contructor and move assignment then.

chfast:
> I would add explicit deletions of move contructor and move assignment then.

OK.

gumb0 (approved):

axic (approved):

PR #780, opened 2021-04-07
capi: Add memory_pages_limit parameter to fizzy_instantiate

gumb0:


include/fizzy/fizzy.h
```diff
@@ -410,6 +410,7 @@ bool fizzy_module_has_start_function(const FizzyModule* module) FIZZY_NOEXCEPT;
 /// @param  imported_globals           Pointer to the imported globals array. Can be NULL iff
 ///                                    @p imported_globals_size equals 0.
 /// @param  imported_globals_size      Size of the imported global array. Can be zero.
+/// @param  memory_pages_limit         Hard limit for memory growth in pages. Cannot be above 65536.
```
axic, line 413: Hmm, would it be a good idea allowing `0` for allowing "default setting"? Or perhaps someone actually wants to disable memory, and then `0` makes sense for that? If so, perhaps `uint32_t(-1)` to signal "use default"?

gumb0: I'm not sure, not a fan of special magic values.

axic: What is annoying is hardcoding 4096 in multiple places, like in rust.

gumb0: We could define default constant in `fizzy.h`, then it could be reused in tests and in rust.

axic: Yeah I think we should export the "default".

chfast: I agree here. Looks good otherwise.

test/utils/fizzy_c_engine.cpp
```diff
@@ -11,6 +11,8 @@
 
 namespace fizzy::test
 {
+constexpr uint32_t DefaultMemoryPagesLimit = 4096;  // 256MB
```
axic, line 14: Why not use the one from the header?

lib/fizzy/limits.hpp
```diff
@@ -18,6 +18,7 @@ static_assert(MaxMemoryPagesLimit == 65536);
 
 /// The default hard limit of the memory size (256MB) as number of pages.
 constexpr uint32_t DefaultMemoryPagesLimit = (256 * 1024 * 1024ULL) / PageSize;
+static_assert(DefaultMemoryPagesLimit == 4096);
```
chfast, line 21: Why this is defined twice then?

test/utils/fizzy_c_engine.cpp
gumb0: Forgot about this one.

lib/fizzy/limits.hpp
gumb0: Because it doesn't worth to introduce a dependency form C++ to C API only for this yet.

test/utils/fizzy_c_engine.cpp
gumb0: Fixed.

lib/fizzy/limits.hpp
axic: Can add a `static_assert(::DefaultMemoryPagesLimit == fizzy::DefaultMemoryPagesLimit)` in `capi_test.cpp`?

gumb0: > Can add a `static_assert(::DefaultMemoryPagesLimit == fizzy::DefaultMemoryPagesLimit)` in `capi_test.cpp`?

Added.

axic (approved):

lib/fizzy/limits.hpp
chfast: > Because it doesn't worth to introduce a dependency form C++ to C API only for this yet.

I don't agree here. This case returns from time to time.

chfast (approved):

PR #781, opened 2021-04-07
rust: Overhaul errors to use an enum

axic:
Depends on #783.
Based on #677.

### DIFF
--- a/bindings/rust/src/lib.rs
+++ b/bindings/rust/src/lib.rs
@@ -128,7 +128,7 @@ impl std::fmt::Display for FizzyErrorBox {
 }
 
 /// Parse and validate the input according to WebAssembly 1.0 rules. Returns true if the supplied input is valid.
-pub fn validate<T: AsRef<[u8]>>(input: T) -> Result<(), String> {
+pub fn validate<T: AsRef<[u8]>>(input: T) -> Result<(), Error> {
     let mut err = FizzyErrorBox::new();
     let ret = unsafe {
         sys::fizzy_validate(
@@ -142,7 +142,7 @@ pub fn validate<T: AsRef<[u8]>>(input: T) -> Result<(), String> {
         Ok(())
     } else {
         debug_assert!(err.code() != 0);
-        Err(err.message())
+        Err(err.error().unwrap())
     }
 }
 
@@ -727,11 +727,14 @@ mod tests {
     #[test]
     fn validate_wasm() {
         // Empty
-        assert_eq!(validate(&[]).err().unwrap(), "invalid wasm module prefix");
+        assert_eq!(
+            validate(&[]).err().unwrap(),
+            Error::MalformedModule("invalid wasm module prefix".to_string())
+        );
         // Too short
         assert_eq!(
             validate(&[0x00]).err().unwrap(),
-            "invalid wasm module prefix"
+            Error::MalformedModule("invalid wasm module prefix".to_string())
         );
         // Valid
         assert!(validate(&[0x00, 0x61, 0x73, 0x6d, 0x01, 0x00, 0x00, 0x00]).is_ok());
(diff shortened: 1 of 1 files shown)


bindings/rust/src/lib.rs
```diff
@@ -76,6 +99,24 @@ impl FizzyErrorBox {
...
+    fn error(&self) -> Option<Error> {
+        match self.0.code {
+            sys::FizzyErrorCode_FIZZY_SUCCESS => None,
+            sys::FizzyErrorCode_FIZZY_ERROR_MALFORMED_MODULE => {
+                Some(Error::MalformedModule(self.message()))
+            }
+            sys::FizzyErrorCode_FIZZY_ERROR_INVALID_MODULE => {
+                Some(Error::InvalidModule(self.message()))
+            }
+            sys::FizzyErrorCode_FIZZY_ERROR_INSTANTIATION_FAILED => {
```
axic, line 113: These variable names look nasty, but at least it works.

bindings/rust/src/lib.rs
```diff
@@ -76,6 +99,24 @@ impl FizzyErrorBox {
                 .into_owned()
         }
     }
+
+    /// Return a translated error object.
+    fn error(&self) -> Option<Error> {
+        match self.0.code {
+            sys::FizzyErrorCode_FIZZY_SUCCESS => None,
+            sys::FizzyErrorCode_FIZZY_ERROR_MALFORMED_MODULE => {
+                Some(Error::MalformedModule(self.message()))
```
axic, line 108: Could also think getting rid of `code()` and `message()` and only having `error()`.

bindings/rust/src/lib.rs
```diff
@@ -40,6 +40,29 @@ mod sys;
 use std::ffi::CString;
 use std::ptr::NonNull;
 
+#[derive(Eq, PartialEq, Debug, Clone)]
+pub enum Error {
+    MalformedModule(String),
```
axic, line 45: Would be nicer using `&str` with a lifetime, but maybe better not getting `FIzzyErrorBox` involved into lifetime management?

bindings/rust/src/lib.rs
axic: Changed in #783.

bindings/rust/src/lib.rs
```diff
@@ -40,6 +40,32 @@ mod sys;
...
 use std::ptr::NonNull;
 
+/// The various kinds of errors, which can be returned by any of the interfaces.
+#[derive(Eq, PartialEq, Debug, Clone)]
+pub enum Error {
+    MalformedModule(String),
+    InvalidModule(String),
+    InstantiationFailed(String),
+    Other(String),
+    FunctionNotFound,
```
axic, line 50: The top 4 errors are 1-to-1 match with the C ones, and the below 5 are Rust specific.

gumb0 (approved):
Looks fine to me.

bindings/rust/src/lib.rs
chfast: No idea what this mean.

bindings/rust/src/lib.rs
```diff
@@ -684,11 +725,14 @@ mod tests {
     #[test]
     fn validate_wasm() {
         // Empty
-        assert_eq!(validate(&[]).err().unwrap(), "invalid wasm module prefix");
+        assert_eq!(
+            validate(&[]).err().unwrap(),
+            Error::MalformedModule("invalid wasm module prefix".to_string())
```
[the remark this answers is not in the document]
gumb0: (only for this case where error message is repeated. Repeating just an enum like `Error::InvalidMemoryOffsetOrSize` is fine)

chfast (approved):

gumb0 (approved):

axic:
> No idea what this mean.https://github.com/wasmx/fizzy/blob/master/include/fizzy/fizzy.h#L24-L39

bindings/rust/src/lib.rs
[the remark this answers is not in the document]
axic: True, I'm slightly leaning towards keeping it as-is because it is easier to change if underlying errors change, and it is only an error message.

PR #782, opened 2021-04-07
rust: add support for imported functions

axic:


lib/fizzy/capi.cpp
```diff
@@ -617,6 +619,11 @@ FizzyInstance* fizzy_instantiate(const FizzyModule* module,
         set_success(error);
         return wrap(instance.release());
     }
+    catch (fizzy::instantiate_error const& e)
+    {
+        std::cout << e.what() << std::endl;
```
axic, line 624: These was my error reporting prior to us adding proper support 😅

bindings/rust/src/lib.rs
```diff
@@ -96,16 +96,70 @@ impl Drop for Instance {
...
+    println!("host fuction called!");
+    //unimplemented!()
+    sys::FizzyExecutionResult {
+        trapped: false,
+        has_value: false,
+        value: sys::FizzyValue { i32: 0 },
+    }
+}
+
+fn create_function_import_list(
```
axic, line 122: Still need to figure out what is the best API defining host functions, but at least this semi-finished translation layer works.

PR #784, opened 2021-04-12
capi: Add a note about error code and returned value relation

gumb0:
Addresses comment https://github.com/wasmx/fizzy/pull/769#discussion_r605143971

### DIFF
--- a/include/fizzy/fizzy.h
+++ b/include/fizzy/fizzy.h
@@ -24,10 +24,15 @@ extern "C" {
 /// Error codes.
 typedef enum FizzyErrorCode
 {
+    /// Success.
     FizzySuccess = 0,
+    /// Malformed module.
     FizzyErrorMalformedModule,
+    /// Invalid module.
     FizzyErrorInvalidModule,
+    /// Instantiation failed.
     FizzyErrorInstantiationFailed,
+    /// Other error.
     FizzyErrorOther
 } FizzyErrorCode;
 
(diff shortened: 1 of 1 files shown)


include/fizzy/fizzy.h
```diff
@@ -250,6 +255,9 @@ typedef struct FizzyImportedGlobal
 /// @param  error               Pointer to store detailed error information at. Can be NULL if error
 ///                             information is not required.
 /// @return                     true if module is valid, false otherwise.
+///
+/// @note   FizzyError::code must be ::FizzySuccess if function returns `true` and must not be
```
gumb0, line 259: Does this not sound like it's a requirement to a user's input?

axic: Yeah I think so. Would suggest something like "FizzyError::code will be ::FizzySuccess if functions...".

axic (approved):
Still suggest to use "will be" instead of "must be".

chfast (approved):

include/fizzy/fizzy.h
```diff
@@ -260,6 +268,9 @@ bool fizzy_validate(
 /// @param  error               Pointer to store detailed error information at. Can be NULL if error
 ///                             information is not required.
 /// @return                     non-NULL pointer to module in case of success, NULL otherwise.
+///
+/// @note   FizzyError::code will be ::FizzySuccess if function returns non-NULL
+///         will and must not be ::FizzySuccess otherwise.
```
axic, line 273: Nit: the other places used will not be.

PR #785, opened 2021-04-16
Floating-point arithmetic on 386/x87 (32-bit)

chfast:


README.md
```diff
@@ -163,6 +163,31 @@ In the GNU C Library the rounding mode can be controlled via the [`fesetround` a
 
 If strict compliance is sought with WebAssembly, then the user of Fizzy must ensure to keep the default rounding mode.
 
+### x87 FPU
+
+On Intel i386 architecture (32-bit) the [x87](https://en.wikipedia.org/wiki/X87) FPU is used by default to perform floating-point operations.
```
axic, line 168:
```suggestion
On Intel i386 architecture (32-bit) the [x87](https://en.wikipedia.org/wiki/X87) FPU ISA is used by default to perform floating-point operations.
```

or 

```suggestion
On Intel i386 architecture (32-bit) an [x87](https://en.wikipedia.org/wiki/X87)-compatible FPU is used by default to perform floating-point operations.
```

README.md
```diff
@@ -163,6 +163,31 @@ In the GNU C Library the rounding mode can be controlled via the [`fesetround` a
...
+The FPU is claimed to be IEEE-754 compliant, but there is one gotcha. The operations are executed with so-called _internal precision_ and the results are rounded to the target precision in the end [[1]](#1).
```
axic, line 169:
```suggestion
The FPU is claimed to be IEEE-754 compliant, but there is one gotcha. The operations are executed with so-called _internal precision_ and the results are rounded to the target precision at the end [[1]](#1).
```

README.md
```diff
@@ -163,6 +163,31 @@ In the GNU C Library the rounding mode can be controlled via the [`fesetround` a
...
+By default the precision is set to [80-bit extended precision](https://en.wikipedia.org/wiki/Extended_precision) (except for VC++ runtime [[2]](#2)).
+Unfortunately, this causes problems for 64-bit double precision operations (`f64.add`, `f64.sub`, `f64.mul`, `f64.div`) — the results may be different than when computed with double precision directly.
+
+The FPU precision can be dynamically modified by using compiler intrinsics [[1]](#1) but this has similar issues to controlling the rounding mode and there exist no C/C++ standard way of doing so.
+
+In the Fizzy implementation we decided it is not worth to fight with the x87 FPU quirks.
+Floating-point operations were never our top priorities.
+We decided to opt-in for using SSE2 instructions to implement WebAssembly floating-pointing instructions
+also in x86 32-bit builds (this is default for 64-builds). I.e. SSE2 instructions set is required.
```
axic, line 178:
```suggestion
also in x86 32-bit builds (this is default for 64-builds). This means for strict WebAssembly compliance the SSE2 instructions set is required.
```

README.md
```diff
@@ -163,6 +163,31 @@ In the GNU C Library the rounding mode can be controlled via the [`fesetround` a
...
+This option is controlled by [`-msse2 -mfpmath=sse`][x86-Options] compiler options and one can always override
+this decision if wants to experiment with x87 FPU.
```
axic, line 180:
```suggestion
this decision if wants to experiment with the x87 FPU.
```

README.md
```diff
@@ -163,6 +163,31 @@ In the GNU C Library the rounding mode can be controlled via the [`fesetround` a
...
+
+For the record, we also tried to use [`-mpc64`][x86-Options] GCC compiler option which is suppose to set the FPU to 64-bit double precision.
```
axic, line 182:
```suggestion
For the record, we also tried to use [`-mpc64`][x86-Options] GCC compiler option which is supposed to set the FPU to 64-bit double precision.
```

README.md
```diff
@@ -163,6 +163,31 @@ In the GNU C Library the rounding mode can be controlled via the [`fesetround` a
...
+This attempt was unsuccessful for unknown reason.
```
axic, line 183:
```suggestion
This attempt was unsuccessful for unknown reasons.
```

test/testfloat/CMakeLists.txt
```diff
@@ -30,6 +30,11 @@ if(TESTFLOAT_GEN)
         list(APPEND IGNORE_LIST ui64_to_f64/min)
     endif()
 
+    if(CMAKE_SYSTEM_PROCESSOR STREQUAL x86_64 AND CMAKE_SIZEOF_VOID_P EQUAL 4 AND CMAKE_CXX_COMPILER_ID MATCHES GNU)
+        # TODO: GCC 32-bit i386 build produces -0 for input 0, https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100119.
```
axic, line 34: Supposedly fixed in gcc 12.

axic:
Rebased on top of #745.

test/testfloat/CMakeLists.txt
axic: Still doesn't work:
```
138/183 Test #103: fizzy/testfloat/ui32_to_f64/min ......................***Failed    0.01 sec
FAILURE: 8000000000000000 <- 00000000
         0000000000000000 (expected)
FAILURE: 8000000000000000 <- 00000000
         0000000000000000 (expected)
FAILURE: 8000000000000000 <- 00000000
         0000000000000000 (expected)
```

circle.yml
```diff
@@ -754,8 +754,10 @@ workflows:
       - spectest:
           requires:
             - coverage-clang
+      - arm64-linux:
```
axic, line 757: @chfast do we need this CI change here?

axic (approved):

README.md
```diff
@@ -163,6 +163,28 @@ In the GNU C Library the rounding mode can be controlled via the [`fesetround` a
...
 If strict compliance is sought with WebAssembly, then the user of Fizzy must ensure to keep the default rounding mode.
 
+### x87 FPU
+
+On the 32-bit Intel i386 architecture an [x87](https://en.wikipedia.org/wiki/X87)-compatible FPU is used by default to perform floating-point operations.
+The FPU is claimed to be IEEE-754 compliant, but there is one gotcha. The operations are executed with so-called *internal precision* and the results are rounded to the target precision at the end [[1]](#1).
+By default the precision is set to [80-bit extended precision](https://en.wikipedia.org/wiki/Extended_precision) (except for VC++ runtime [[2]](#2)).
+Unfortunately, this causes problems for 64-bit double precision operations (`f64.add`, `f64.sub`, `f64.mul`, `f64.div`) as the results may be different than when computed with double precision directly.
+
+The FPU precision can be dynamically modified by using compiler intrinsics [[1]](#1), but this has similar issues to controlling the rounding mode and there exist no C/C++ standard way of doing so.
```
gumb0, line 173:
```suggestion
The FPU precision can be dynamically modified by using compiler intrinsics [[1]](#1), but this has similar issues to controlling the rounding mode and there exists no C/C++ standard way of doing so.
```

test/testfloat/CMakeLists.txt
axic: Ran it on the CI, which I believed had 12, but now on a second look it is 11.

chfast:
With optimized build in GCC 12 I have these failures:
```
The following tests FAILED:
        126 - fizzy/testfloat/ui32_to_f64/min (Failed)
        182 - fizzy/testfloat/f64_ceil/min (Failed)
        190 - fizzy/testfloat/f64_trunc/min (Failed)
        194 - fizzy/testfloat/f64_nearest/min (Failed)
        421 - fizzy/unittests/execute_floating_point_types/f64.  # TypeParam = double.ceil (Failed)
        423 - fizzy/unittests/execute_floating_point_types/f64.  # TypeParam = double.trunc (Failed)
        424 - fizzy/unittests/execute_floating_point_types/f64.  # TypeParam = double.nearest (Failed)
```

axic:
@chfast let's shelve this then for now

CMakeLists.txt
```diff
@@ -71,6 +71,12 @@ if(CMAKE_CXX_COMPILER_ID MATCHES Clang)
     endif()
 endif()
 
+if(CMAKE_SYSTEM_PROCESSOR STREQUAL x86_64 AND CMAKE_SIZEOF_VOID_P EQUAL 4)
```
axic, line 74: `x86_64`?

chfast: Maybe it's different on true 386, but I have not way of checking.

axic: I'm sure it should be `i386` on any non-64-bit x86 (at some point they tried to retrofit `ia32` though).

axic: Some search suggest that it may return the following: `x86`, `i386`, `i686`.

See https://boringssl.googlesource.com/boringssl/+/refs/heads/2272/CMakeLists.txt and http://www.autoscool-clermont.com/blum/wp-content/themes/matrix/inc/kirki/assets/js/vendor/codemirror/mode/cmake/index.html (which has `set(X86_ALIASES x86 i386 i686 x86_64 amd64)`).

axic, line 74: How about sorting it out with:
```suggestion
# TODO: potentially include x86, i386, i686 here
if(CMAKE_SYSTEM_PROCESSOR STREQUAL x86_64 AND CMAKE_SIZEOF_VOID_P EQUAL 4)
```

axic: Also just realised that basically this check here enables "32-bit mode support" as opposed to supporting x86. Maybe just documenting that too is good.

chfast: Please make the required changes yourself.

PR #786, opened 2021-04-16
test: Add f64.{add,sub} off by one unit tests

chfast:
The i386 x87 FPU performs all arithmetic with 80-bit extended precision
by default and in the end rounds it to the required type. This causes
issues for f64 types because results may be different than when doing
computation with 64-bit precision. f32 is not affected.
These tests presents examples of results which are different in such
case.

### DIFF
--- a/test/unittests/execute_floating_point_test.cpp
+++ b/test/unittests/execute_floating_point_test.cpp
@@ -1282,3 +1282,108 @@ TEST(execute_floating_point, f64_store_overflow)
     // Offset is 0x7fffffff + 0x80000001 => 0x100000000
     EXPECT_THAT(execute(*instance, 0, {0x80000001}), Traps());
 }
+
+TEST(execute_floating_point, f64_add_round_to_even)
+{
+    // This test presents how IEEE-754 "round to nearest, ties to even" works.
+    // This rounding mode is required by WebAssembly.
+
+    struct TestCase
+    {
+        double a;
+        double b;
+        double expected_sum;
+    };
+
+    constexpr TestCase test_cases[]{
+        // = - no rounding, ^ - round up, v - round down.
+        {0x1p0, 0x1.0000000000000p0, /* 0x2.0000000000000p0 = */ 0x1.0000000000000p1},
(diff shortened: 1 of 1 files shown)


test/unittests/execute_floating_point_test.cpp
```diff
@@ -1282,3 +1282,108 @@ TEST(execute_floating_point, f64_store_overflow)
...
+        EXPECT_THAT(execute(*instance, 0, {t.a, t.b}), Result(t.expected_sum));
+    }
+}
+
+TEST(execute_floating_point, f64_add_off_by_one)
+{
+    // The i386 x87 FPU performs all arithmetic with 80-bit extended precision by default
+    // and in the end rounds it to the required type. This causes issues for f64 types
+    // because results may be different than when doing computation with 64-bit precision.
+    // f32 is not affected.
```
gumb0, line 1336: Why is f32 not affected?

test/unittests/execute_floating_point_test.cpp
```diff
@@ -1282,3 +1282,108 @@ TEST(execute_floating_point, f64_store_overflow)
...
+    (func (param f64 f64) (result f64)
+      (f64.add (local.get 0) (local.get 1))
+    )
+    */
+    const auto wasm = from_hex("0061736d0100000001070160027c7c017c030201000a0901070020002001a00b");
+
+    auto instance = instantiate(parse(wasm));
+
+    constexpr auto a = 0x1.0008000008000p60;
+    constexpr auto b = 0x1.0000000081fffp07;
```
gumb0, line 1350: If you know the easy way to do this, maybe it would be helpful to comment with "de-normalized" exponent 60 value of this.

I mean, as I understand, it's something like this:
```suggestion
    constexpr auto a = 0x1.0008000008000p60;
    constexpr auto b = 0x1.0000000081fffp07; // 0x0.00000000000008...p60; 
```

gumb0 (approved):

test/unittests/execute_floating_point_test.cpp
chfast: I don't really know what exactly is going on here. Maybe if we convert this also to `long double` we may see what's going on.

PR #788, opened 2021-04-19
capi: better OOM handling

gumb0:


axic (approved):
Looks okay to me, but not fully convinced we want to expose this as a separate error.

PR #790, opened 2021-04-21
WASI tests

gumb0:


chfast:
If this way is easier, I'm fine with this. Do you have estimate how much uvwasi methods are we are going to need?

gumb0:
> If this way is easier, I'm fine with this. Do you have estimate how much uvwasi methods are we are going to need?

It seems it's about one uvwasi funciton per one WASI function + some serialization functions that we may or may not want to mock.

tools/wasi/CMakeLists.txt
```diff
@@ -7,8 +7,8 @@ include(ProjectUVWASI)
 add_library(wasi)
 add_library(fizzy::wasi ALIAS wasi)
 target_include_directories(wasi PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
-target_link_libraries(wasi PUBLIC fizzy::fizzy-internal PRIVATE uvwasi::uvwasi)
-target_sources(wasi PRIVATE wasi.cpp wasi.hpp)
+target_link_libraries(wasi PUBLIC fizzy::fizzy-internal uvwasi::uvwasi)
```
gumb0, line 10: Tests require now to include uvwasi's `<wasi_types.h>` to implement mock interface.

Is this ok change or should there be a better way?

gumb0:
Obeservations about serizaliation part of uvwasi:

1. Deserialization functions for `fd_write` / `fd_read` check passed pointers against the bounds passed in `end` parameter:
https://github.com/nodejs/uvwasi/blob/adda15515cbb5a478d0ddb85602d987278a43384/src/wasi_serdes.c#L191-L192

So I think passing the memory size here is correct:
https://github.com/wasmx/fizzy/blob/2ea01bcd77d6a67080f6f81608502435563bc72f/tools/wasi/wasi.cpp#L45

2. It doesn't check bounds when deserializing "buffer pointer" and "buffer size" values themselves:
https://github.com/nodejs/uvwasi/blob/adda15515cbb5a478d0ddb85602d987278a43384/src/wasi_serdes.c#L188-L189
We could check these pointers ourselves at our side of `fd_write` / `fd_read`.

tools/wasi/uvwasi.hpp
```diff
@@ -0,0 +1,36 @@
...
+#include <wasi_types.h>
+#include <memory>
+
+namespace fizzy::wasi
+{
+/// UVWASI interface.
+class UVWASI
+{
+public:
+    virtual ~UVWASI();
```
chfast, line 16: ` = default`.

tools/wasi/wasi.hpp
```diff
@@ -13,6 +13,9 @@ struct Instance;
 
 namespace wasi
 {
+class UVWASI;
+extern UVWASI* uvwasi;
```
chfast, line 17: Is this needed for mocking?

chfast (commented):
This looks good.

Some additional questions:
1. This seems not using GMock. Is the include even needed?
2. We considered before to create separate unit tests executable for wasi, e.g. `fizzy-unittests-wasi` to decouple this subproject and its dependencies. This would also allow us to create separate directory for wasi unit tests and split these into more files.

gumb0:
> 1. This seems not using GMock. Is the include even needed?

It's needed only for `ElementsAre`

tools/wasi/wasi.hpp
gumb0: Yes.

Alternatively maybe It could be declared in `wasi_test.cpp` and another time in `wasi.cpp`

gumb0:
> 2\. We considered before to create separate unit tests executable for wasi, e.g. `fizzy-unittests-wasi` to decouple this subproject and its dependencies. This would also allow us to create separate directory for wasi unit tests and split these into more files.

Good idea, I will try that next, maybe in a separate PR.

tools/wasi/uvwasi.hpp
gumb0: This produces `error: 'UVWASI' has no out-of-line virtual method definitions; its vtable will be emitted in every translation unit [clang-diagnostic-weak-vtables,-warnings-as-errors]`

https://app.circleci.com/pipelines/github/wasmx/fizzy/6667/workflows/39a77e98-002f-42e8-a916-19acec64fcbc/jobs/58758/parallel-runs/0/steps/0-111

tools/wasi/uvwasi.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+    uvwasi_errno_t environ_sizes_get(
+        uvwasi_size_t* environ_count, uvwasi_size_t* environ_buf_size) noexcept final
+    {
+        return uvwasi_environ_sizes_get(&m_state, environ_count, environ_buf_size);
+    }
+};
+}  // namespace
+
+UVWASI::~UVWASI() {}
```
axic, line 63: Need to call `uvwasi_destroy` here.

chfast (approved):

gumb0:
@axic I have removed the merge commit from here, the last commit is now a single commit from #796.

axic:
@gumb0 can you rebase this?

gumb0:
Rebased.

chfast (approved):

test/unittests/wasi_test.cpp
```diff
@@ -177,3 +177,82 @@ TEST_F(wasi_mocked_test, fd_write)
...
+    EXPECT_FALSE(mock_uvwasi->init_called);
+    EXPECT_FALSE(mock_uvwasi->write_fd.has_value());
+    EXPECT_TRUE(mock_uvwasi->write_data.empty());
+
+    std::ostringstream err;
+    EXPECT_TRUE(wasi::run(*mock_uvwasi, *instance, 0, nullptr, err)) << err.str();
+
+    EXPECT_TRUE(mock_uvwasi->init_called);
+    ASSERT_TRUE(mock_uvwasi->write_fd.has_value());
+    EXPECT_EQ(*mock_uvwasi->write_fd, 1);
```
axic: Based on the tests, `write_fd` should be a vector too.

test/unittests/wasi_test.cpp
```diff
@@ -177,3 +177,82 @@ TEST_F(wasi_mocked_test, fd_write)
...
+}
+
+TEST_F(wasi_mocked_test, fd_write_invalid_input)
+{
+    /* wat2wasm
+      (func (import "wasi_snapshot_preview1" "fd_write") (param i32 i32 i32 i32) (result i32))
+      (memory (export "memory") 1)
+      (data (i32.const 0)    "\00\00\01\00")    ;; buf ptr - out of memory bounds
+      (data (i32.const 0x04) "\04\00\00\00")    ;; buf len
+      (global (mut i32) (i32.const 0))
```
axic: Shouldn't this be a different value to differentiate from `global.set 0`?

test/unittests/wasi_test.cpp
```diff
@@ -177,3 +177,82 @@ TEST_F(wasi_mocked_test, fd_write)
...
+    // nwritten
+    EXPECT_EQ(instance->memory->substr(0x1c, 4), from_hex("0c000000"));
+}
+
+TEST_F(wasi_mocked_test, fd_write_invalid_input)
+{
+    /* wat2wasm
+      (func (import "wasi_snapshot_preview1" "fd_write") (param i32 i32 i32 i32) (result i32))
+      (memory (export "memory") 1)
+      (data (i32.const 0)    "\00\00\01\00")    ;; buf ptr - out of memory bounds
```
axic: So this only checks the offset against pages?

test/unittests/wasi_test.cpp
```diff
@@ -256,3 +274,123 @@ TEST_F(wasi_mocked_test, fd_write_invalid_input)
...
+}
+
+TEST_F(wasi_mocked_test, fd_read_invalid_input)
+{
+    /* wat2wasm
+      (func (import "wasi_snapshot_preview1" "fd_read") (param i32 i32 i32 i32) (result i32))
+      (memory (export "memory") 1)
+      (data (i32.const 0)    "\00\00\01\00")    ;; buf ptr - out of memory bounds
+      (data (i32.const 0x04) "\04\00\00\00")    ;; buf len
+      (global (mut i32) (i32.const 0))
```
axic: Same with the global here.

test/unittests/wasi_test.cpp
gumb0: Not sure what you mean, WASI `fd_write` calls `uvwasi_fd_write` only once, with one `fd` argument, and this is what the mock saves in `write_fd` and the test checks.

test/unittests/wasi_test.cpp
gumb0: I think the test tries to check that error value returned from uvwasi is correctly propagated. I don't think we should test all possible errors that uvwasi returns.

test/unittests/wasi_test.cpp
gumb0: We care in this test that the final value of the global is not 0.

(Final `global.set 0` instruction updates it to a value returned from `fd_write`)

test/unittests/wasi_test.cpp
axic: In the current tests, but eventually can imagine multiple calls.

test/unittests/wasi_test.cpp
axic: Ah okay didn't register that it was `set`.

test/unittests/wasi_test.cpp
axic: I meant this only checks against exceeding memory pages?

gumb0: The test tries to get the data to write from the address out of memory bounds (0x10000 = 65536), so this is just one possible way to make `fd_write` fail.

axic (approved):
Okay for now I think.

test/unittests/wasi_test.cpp
gumb0: Then it would rather be vector like `vector<pair<fd, write_data>>` to record all calls... not sure it's worth to complicate it now.

PR #792, opened 2021-05-05
wasi: Add UVWASI interface

gumb0:
Pulled out of #790

### DIFF
--- a/tools/wasi/uvwasi.cpp
+++ b/tools/wasi/uvwasi.cpp
@@ -1,0 +1,69 @@
+// Fizzy: A fast WebAssembly interpreter
+// Copyright 2021 The Fizzy Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "uvwasi.hpp"
+#include <uvwasi.h>
+
+namespace fizzy::wasi
+{
+namespace
+{
+class UVWASIImpl final : public UVWASI
+{
+    /// UVWASI state.
+    uvwasi_t m_state{};
+
+public:
+    uvwasi_errno_t init(uvwasi_size_t argc, const char** argv) noexcept final
+    {
+        // Initialisation settings.
+        // TODO: Make const after https://github.com/nodejs/uvwasi/pull/155 is merged.
+        uvwasi_options_t options = {
+            3,           // sizeof fd_table
+            0, nullptr,  // NOTE: no remappings
+            argc, argv,
+            nullptr,  // TODO: support env
+            0, 1, 2,
+            nullptr,  // NOTE: no special allocator
+        };
+        return uvwasi_init(&m_state, &options);
+    }
+
+    uvwasi_errno_t proc_exit(uvwasi_exitcode_t exit_code) noexcept final
+    {
+        return uvwasi_proc_exit(&m_state, exit_code);
+    }
+
+    uvwasi_errno_t fd_write(uvwasi_fd_t fd, const uvwasi_ciovec_t* iovs, uvwasi_size_t iovs_len,
+        uvwasi_size_t* nwritten) noexcept final
+    {
+        return uvwasi_fd_write(&m_state, fd, iovs, iovs_len, nwritten);
+    }
+
+    uvwasi_errno_t fd_read(uvwasi_fd_t fd, const uvwasi_iovec_t* iovs, uvwasi_size_t iovs_len,
+        uvwasi_size_t* nread) noexcept final
+    {
+        return uvwasi_fd_read(&m_state, fd, iovs, iovs_len, nread);
+    }
+
+    uvwasi_errno_t fd_prestat_get(uvwasi_fd_t fd, uvwasi_prestat_t* buf) noexcept final
+    {
+        return uvwasi_fd_prestat_get(&m_state, fd, buf);
+    }
+
+    uvwasi_errno_t environ_sizes_get(
+        uvwasi_size_t* environ_count, uvwasi_size_t* environ_buf_size) noexcept final
+    {
+        return uvwasi_environ_sizes_get(&m_state, environ_count, environ_buf_size);
+    }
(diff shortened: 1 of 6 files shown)


tools/wasi/uvwasi.cpp
```diff
@@ -0,0 +1,71 @@
...
+{
+namespace
+{
+class UVWASIImpl final : public UVWASI
+{
+    /// UVWASI state.
+    uvwasi_t m_state{};
+
+public:
+    ~UVWASIImpl() final { uvwasi_destroy(&m_state); }
```
gumb0, line 18: Apparently we'll need a flag to check if `uvwasi_init` was already called sucessfully.

tools/wasi/wasi.cpp
```diff
@@ -152,6 +152,13 @@ bool run(bytes_view wasm_binary, int argc, const char* argv[], std::ostream& err
         {ns, "environ_get", {ValType::i32, ValType::i32}, ValType::i32, return_enosys},
     };
 
+    auto module = parse(wasm_binary);
+    auto imports = resolve_imported_functions(*module, wasi_functions);
+    return instantiate(std::move(module), std::move(imports), {}, {}, {}, MaxMemoryPagesLimit);
+}
```
axic, line 156: We had an `assert(instance != nullptr)` in the old place. Worth keeping here or `unique_ptr` ensures it?

tools/wasi/wasi.hpp
```diff
@@ -26,6 +31,18 @@ std::optional<bytes> load_file(std::string_view file, std::ostream& err) noexcep
 /// @todo Make noexcept.
 bool load_and_run(int argc, const char** argv, std::ostream& err);
 
+/// Instantiates a module that imports WASI functions.
+std::unique_ptr<Instance> instantiate(bytes_view wasm_binary);
```
axic, line 35: What is the reason behind splitting?

gumb0: To access instance in tests after `run`.

tools/wasi/wasi.cpp
gumb0: `unique_ptr` doesn't care, but `instantiate` reports errors by throwing, so I don't think it's useful.

tools/wasi/uvwasi.hpp
```diff
@@ -0,0 +1,36 @@
...
+#include <wasi_types.h>
+#include <memory>
+
+namespace fizzy::wasi
+{
+/// UVWASI interface.
+class UVWASI
+{
+public:
+    virtual ~UVWASI();
```
chfast, line 16:
```suggestion
    virtual ~UVWASI() noexcept;
```

chfast (changes requested):

tools/wasi/uvwasi.hpp
gumb0: Destructors are `noexcept` by default.

chfast: I didn't know this.

tools/wasi/wasi.cpp
```diff
@@ -152,32 +150,25 @@ bool run(bytes_view wasm_binary, int argc, const char* argv[], std::ostream& err
...
-    };
+    auto module = parse(wasm_binary);
+    auto imports = resolve_imported_functions(*module, wasi_functions);
+    return instantiate(std::move(module), std::move(imports), {}, {}, {}, MaxMemoryPagesLimit);
+}
 
-    const uvwasi_errno_t uvwasi_err = uvwasi_init(&state, &options);
+bool run(Instance& instance, int argc, const char* argv[], std::ostream& err)
+{
+    if (!uvwasi)
```
chfast, line 160: I don't think this is good idea, for many reasons (e.g. not thread-safe, hidden dependencies). Why just not add `UVWASI&` param?

gumb0: This is a good idea.

Do you want also to pass it into WASI functions via `host_context` to make it thread-safe?

chfast: I don't see why this would be needed. This object will be bound to the imported functions, right?
By "not-thread-safe" I mean the initialization. I think this should be propagated up to the `main()` function which should invoke `create_uvwasi()`. And `fizzy-unittest-wasi` will execute `create_mocked_uvwasi()` if needed.

tools/wasi/uvwasi.cpp
```diff
@@ -0,0 +1,81 @@
...
+class UVWASIImpl final : public UVWASI
+{
+    /// UVWASI state.
+    uvwasi_t m_state{};
+    bool m_inited = false;
+
+public:
+    ~UVWASIImpl() final
+    {
+        if (m_inited)
```
gumb0, line 21: Also suggested a change to uvwasi to handle this https://github.com/nodejs/uvwasi/pull/163

But it won't help the case with calling `init` twice, so we need a flag anyway.

tools/wasi/wasi.cpp
gumb0: > I don't see why this would be needed. This object will be bound to the imported functions, right?

What do you mean by bound? The way to bind it in current API is to pass it via `host_context`.

See the latest version.

chfast: It looks good.

chfast (approved):

tools/wasi/uvwasi.cpp
```diff
@@ -13,10 +13,20 @@ class UVWASIImpl final : public UVWASI
 {
     /// UVWASI state.
     uvwasi_t m_state{};
+    bool m_inited = false;
```
axic, line 16: Perhaps `m_initialised` is more correct, but fine either way.

axic (approved):

PR #796, opened 2021-05-11
cmake: Split WASI unit tests into a separate executable

gumb0:


axic:
What is the reason for this?

gumb0:
> What is the reason for this?

@chfast's point 2 here https://github.com/wasmx/fizzy/pull/790#pullrequestreview-651129672

PR #797, opened 2021-05-12
ci: upgrade homebrew to avoid delays due to bintray

axic:


circle.yml
```diff
@@ -25,7 +25,10 @@ commands:
     steps:
       - run:
           name: "Install system dependencies"
-          command: HOMEBREW_NO_AUTO_UPDATE=1 HOMEBREW_NO_INSTALL_CLEANUP=1 brew install cmake ninja
+          command: |
+            # Update homebrew to hopefully >=3.2 so that reliance on bintray is removed
+            HOMEBREW_NO_AUTO_UPDATE=1 HOMEBREW_NO_INSTALL_CLEANUP=1 brew update
+            HOMEBREW_NO_AUTO_UPDATE=1 HOMEBREW_NO_INSTALL_CLEANUP=1 brew install cmake ninja
```
chfast, line 31:
```suggestion
            HOMEBREW_NO_INSTALL_CLEANUP=1 brew install cmake ninja
```
You can just drop "no auto update".

axic: Thanks, I think this change does not really help much, it seemed to be slower even comparing to building some of the packages in the current case. Also CircleCI announced they will update the images, so I guess we can just close this PR.

chfast: Yes, homebrew update is slow, that's why it was disabled.

PR #800, opened 2021-06-04
test: Split capi_test.cpp

gumb0:


test/unittests/capi_test.cpp
```diff
@@ -206,6 +206,7 @@ TEST(capi, clone_module)
     fizzy_free_module(module1);
 }
 
+// capi_module_test.cpp
```
gumb0: I should better just rename test suites to `capi_module` etc.

chfast (approved):

axic:
Just checked and no changes to `capi` were merged since this PR was created, so rebasing it should be safe.

test/unittests/capi_test.cpp
```diff
@@ -402,7 +402,7 @@ TEST(capi, get_export_description)
     fizzy_free_module(module);
 }
 
-TEST(capi, export_name_after_instantiate)
+TEST(capi_exports, export_name_after_instantiate)
```
axic: Shouldn't these be part of `capi_module`?

chfast: Not sure what the classification is, but it exports a thing.

chfast (approved):

test/unittests/capi_test.cpp
gumb0: I don't remember exactly anyomore either, but I guess I put exports-related tests into a separate file, because there're quite a few of them.
And they some of the functions tested there operate on a instance, too.

gumb0:
Rebased.

axic:
Validated the tests by reconstructing the original file. Seems identical.

axic (approved):

axic (approved):
Need to rebase though.

PR #804, opened 2021-12-20
Upgrade dependencies and CI config

chfast:


test/bench_internal/parser_benchmarks.cpp
```diff
@@ -100,4 +100,4 @@ static void parse_string(benchmark::State& state)
 
     state.SetItemsProcessed(static_cast<int64_t>(size));
 }
-BENCHMARK(parse_string)->RangeMultiplier(2)->Range(16, 4 * 1024);
+BENCHMARK(parse_string)->RangeMultiplier(2)->Range(16, 4096);
```
axic, line 103: Is this change needed after `-bugprone-implicit-widening-of-multiplication-result` ?

test/utils/wasm3_engine.cpp
```diff
@@ -124,7 +124,7 @@ WasmEngine::Result Wasm3Engine::execute(
 {
     unsigned ret_valid;
     uint64_t ret_value;
-    IM3Function function = reinterpret_cast<IM3Function>(func_ref);
+    auto function = reinterpret_cast<IM3Function>(func_ref);  // NOLINT(performance-no-int-to-ptr)
```
axic, line 127: Why was this needed?

lib/fizzy/instantiate.cpp
```diff
@@ -203,7 +203,7 @@ std::tuple<bytes_ptr, Limits> allocate_memory(const std::vector<Memory>& module_
         }
 
         // NOTE: fill it with zeroes
-        bytes_ptr memory{new bytes(memory_min * PageSize, 0), bytes_delete};
+        bytes_ptr memory{new bytes(uint32_t{memory_min * PageSize}, 0), bytes_delete};
```
axic, line 206: @gumb0 can you double check?

CMakeLists.txt
```diff
@@ -66,7 +66,6 @@ if(CMAKE_CXX_COMPILER_ID MATCHES Clang)
             -Wno-double-promotion
             -Wno-float-equal
             -Wno-padded
-            -Wno-return-std-move-in-c++11
```
axic, line 69: What was this causing?

test/bench_internal/parser_benchmarks.cpp
chfast: No.

test/utils/wasm3_engine.cpp
chfast: This is probably new clang-tidy warning (as I upgraded to clang-tidy-13). Pointer to integer cast blocks some optimizations. https://clang.llvm.org/extra/clang-tidy/checks/performance-no-int-to-ptr.html.

CMakeLists.txt
chfast: This flag does not exist any more.

lib/fizzy/instantiate.cpp
chfast: Note that this is non-functional change. The `uint32_t{}` explicitly states that the multiplication is done with 32-bit precision. This will not compile if any of multiplication operands has bigger type.

test/bench_internal/parser_benchmarks.cpp
gumb0: why change then?

lib/fizzy/instantiate.cpp
gumb0: Is this change still needed with disabled `bugprone-implicit-widening-of-multiplication-result`?

chfast: It is not. I started "fixing" these but it turned out there are too many cases for constant multiplication. I reported the bug about the check but it is not worth it in current form.

gumb0: Well turns out we have a bug on master here, because trying to allocate 4Gb actually overflows and allocates 0.

https://github.com/wasmx/fizzy/pull/805

.codespell-whitelist
```diff
@@ -1 +1,2 @@
 mut
+crate
```
gumb0, line 2: Also not sure why, it's a normal English word?

chfast: I don't know. The codespell suggests crate -> create.

lib/fizzy/instantiate.cpp
chfast: The `uint32_t{}` is my proposed why of silencing the warning. I can revert these 2 changes from this PR, especially if there is a fix now in a side PR.

gumb0: I would revert them if warning is disabled anyway.

gumb0 (approved):

PR #805, opened 2021-12-22
Fix overflows with memory of max size

gumb0:
Calculating memory in bytes to be allocated as `memory_min * PageSize` overflows 32 bit when trying to allocate 4 Gb and 0 bytes is actually allocated.

Probably this should be a spec test instead of unit test, unit test crashes on CI anyway.

`memory.grow` looks to not have a problem, but a test would be nice, too.

### DIFF
--- a/lib/fizzy/instantiate.cpp
+++ b/lib/fizzy/instantiate.cpp
@@ -203,7 +203,7 @@ std::tuple<bytes_ptr, Limits> allocate_memory(const std::vector<Memory>& module_
         }
 
         // NOTE: fill it with zeroes
-        bytes_ptr memory{new bytes(memory_min * PageSize, 0), bytes_delete};
+        bytes_ptr memory{new bytes(uint64_t{memory_min} * PageSize, 0), bytes_delete};
         return {std::move(memory), module_memories[0].limits};
     }
     else if (imported_memories.size() == 1)
--- a/lib/fizzy/instantiate.cpp
+++ b/lib/fizzy/instantiate.cpp
@@ -121,7 +121,8 @@ void match_imported_memories(const std::vector<Memory>& module_imported_memories
 
         const auto min = imported_memories[0].limits.min;
         const auto& max = imported_memories[0].limits.max;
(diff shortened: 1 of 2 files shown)


gumb0:
> Probably this should be a spec test instead of unit test, unit test crashes on CI anyway.

Probably it's not for the spec tests either, there're no tests currently that would allocate 4Gb, only one declaring 4Gb as max memory 
https://github.com/WebAssembly/spec/blob/ffb5e3b40db7a45237d4e2fc972c1552ac4abef5/test/core/memory.wast#L8

gumb0:
Importing 4Gb memory looks to be also affected https://github.com/wasmx/fizzy/blob/bbf7638d698e1405752a5a271025b3bc4757d6c1/lib/fizzy/instantiate.cpp#L124

lib/fizzy/instantiate.cpp
```diff
@@ -199,11 +200,12 @@ std::tuple<bytes_ptr, Limits> allocate_memory(const std::vector<Memory>& module_
             (memory_max.has_value() && *memory_max > memory_pages_limit))
         {
             throw instantiate_error{"cannot exceed hard memory limit of " +
-                                    std::to_string(memory_pages_limit * PageSize) + " bytes"};
+                                    std::to_string(uint64_t{memory_pages_limit} * PageSize) +
```
gumb0, line 203: This actually will never overflow, because module memory with min or max > 65536 will be an invalid (malformed) module (so will fail earlier in parser), but I think it's better to cast here, too, for consistency. But no test possible.

gumb0:
Now I wonder if it would be better to define `PageSize` as `uint64_t` instead of casting every time.

gumb0:
> `memory.grow` looks to not have a problem, but a test would be nice, too.

It does have a problem on 32-bit build though. But there it's part of a bigger problem: max hard memory limit of 65536 pages cannot be used on 32-bit system at all, as there's no way to allocate 65536 pages.

gumb0:
Closing in favor of #807

PR #807, opened 2021-12-30
Fix handling memory of max size on 32-bit systems: change hard memory limit

gumb0:
To support 32-bit systems I think it's better not to add another check for `size_t` overflow, but instead rely on existing check against hard memory limit and cap this limit to below 4 Gb (e.g. to 65535 pages) for 32-bit systems.

This way the logic is simpler in `allocate_memory` and `grow_memory` (I leave onle assertions for `size_t` overflow in these places), there's no additional check to test, the only complication is that some unit tests have different expectations for 32-bit build.
(Also `(module (memory 0 65536))` spec test won't pass on 32-bit)

Hard memory limit could be thought of as maximum possible memory size that system supports, then for 32-bit it shoud be below 4 Gb.

### DIFF
--- a/lib/fizzy/limits.hpp
+++ b/lib/fizzy/limits.hpp
@@ -4,12 +4,19 @@
 
 #pragma once
 
+#include <cstddef>
 #include <cstdint>
 
 namespace fizzy
 {
 /// The page size as defined by the WebAssembly 1.0 specification.
 constexpr uint32_t PageSize = 65536;
+
+/// Convert memory size in pages to size in bytes.
+inline constexpr uint64_t memory_pages_to_bytes(uint32_t pages) noexcept
+{
+    return pages * PageSize;
+}
 
 /// The maximum memory page limit as defined by the specification.
 /// It is only possible to address 4 GB (32-bit) of memory.
--- a/lib/fizzy/limits.hpp
+++ b/lib/fizzy/limits.hpp
@@ -4,8 +4,10 @@
 
 #pragma once
 
+#include <algorithm>
 #include <cstddef>
 #include <cstdint>
+#include <limits>
 
 namespace fizzy
 {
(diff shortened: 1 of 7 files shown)


lib/fizzy/limits.hpp
```diff
@@ -4,17 +4,30 @@
...
 #include <cstdint>
+#include <limits>
 
 namespace fizzy
 {
 /// The page size as defined by the WebAssembly 1.0 specification.
 constexpr uint32_t PageSize = 65536;
 
+/// Convert memory size in pages to size in bytes.
+inline constexpr uint64_t memory_pages_to_bytes(uint32_t pages) noexcept
```
gumb0, line 18: Any better place for this helper?

lib/fizzy/limits.hpp
```diff
@@ -4,17 +4,31 @@
...
+{
+    return uint64_t{pages} * PageSize;
+}
+
 /// The maximum memory page limit as defined by the specification.
```
gumb0, line 23: Maybe these comments could be improved.

chfast (approved):
Looks good to be although I'm not sure if having the same limit for all architectures wouldn't be cleaner. Having the `allocate()` helper could check if the requested memory size would fit into `size_t`. Otherwise the allocation should fail imminently.

gumb0:
> Looks good to be although I'm not sure if having the same limit for all architectures wouldn't be cleaner. Having the `allocate()` helper could check if the requested memory size would fit into `size_t`. Otherwise the allocation should fail imminently.

Yeah I think it would be possible, the 2nd helper will be called in 2 places: `allocate_memory` and `grow_memory` where assertions currently are. 

But in `allocate_memory` if the check looks like `if (!valid_allocation(size)) throw(...)` throwing lines cannot be covered in 64-bit build, unless we put the throw into the 3d helper...

chfast:
> But in `allocate_memory` if the check looks like `if (!valid_allocation(size)) throw(...)` throwing lines cannot be covered in 64-bit build, unless we put the throw into the 3d helper...

This can be in `if constexpr` section. And we can also collect coverage for 32-bit builds. If these are the only problems, I'd try to do it.

gumb0:
Closing in favor of #808

PR #808, opened 2022-01-07
Max memory 32bit

gumb0:
Alternative to https://github.com/wasmx/fizzy/pull/807


lib/fizzy/limits.hpp
```diff
@@ -17,6 +19,21 @@ inline constexpr uint64_t memory_pages_to_bytes(uint32_t pages) noexcept
     return uint64_t{pages} * PageSize;
 }
 
+/// Check if the integral value can be casted without an overflow to a narrower type.
+template <typename TypeTo, typename TypeFrom,
+    typename =
+        typename std::enable_if_t<std::is_integral_v<TypeFrom> && std::is_integral_v<TypeTo> &&
+                                  sizeof(TypeFrom) >= sizeof(TypeTo)>>
+inline constexpr bool can_narrow_cast(TypeFrom value) noexcept
```
gumb0, line 27: This is not about limit anymore, move to a new header?

lib/fizzy/execute.cpp
```diff
@@ -152,15 +152,8 @@ inline uint32_t grow_memory(
...
-    if constexpr (sizeof(size_t) < sizeof(uint64_t))
-    {
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wtautological-type-limit-compare"
-        // NOLINTNEXTLINE(clang-diagnostic-tautological-type-limit-compare)
-        if (new_bytes > std::numeric_limits<size_t>::max())
-            return static_cast<uint32_t>(-1);
-#pragma clang diagnostic pop
-    }
+    if (!can_narrow_cast<size_t>(new_bytes))
```
chfast, line 155: The GSL introduced `narrow_cast` and `narrow`. The `narrow` throws exception if conversion changes value. The `narrow_cast` is just an alias for `static_cast` so it always succeeds. So probably better name would be `can_narrow` or `can_safely_narrow`.

chfast (approved):
Clean up git history before merge.

PR #809, opened 2022-01-10
Fix overflows with memory of max size

gumb0:
The first part of max-memory-related fixes, fixing overflows happening on both 32-bit and 64-bit.

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -145,17 +145,19 @@ inline uint32_t grow_memory(
     assert(memory_pages_limit <= MaxMemoryPagesLimit);
     assert(cur_pages <= memory_pages_limit);
 
-    const auto new_pages = uint64_t{cur_pages} + delta_pages;
-    if (new_pages > memory_pages_limit)
+    const auto new_pages_u64 = uint64_t{cur_pages} + delta_pages;
+    if (new_pages_u64 > memory_pages_limit)
         return static_cast<uint32_t>(-1);
+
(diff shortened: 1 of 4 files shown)


lib/fizzy/execute.cpp
```diff
@@ -145,17 +145,19 @@ inline uint32_t grow_memory(
...
 
+    const auto new_pages = static_cast<uint32_t>(new_pages_u64);
+
     try
     {
-        // new_pages <= memory_pages_limit <= MaxMemoryPagesLimit guarantees multiplication
+        // new_pages <= memory_pages_limit <= MaxMemoryPagesLimit guarantees memory_pages_to_bytes
         // will not overflow uint32_t.
-        assert(new_pages * PageSize <= std::numeric_limits<uint32_t>::max());
+        assert(memory_pages_to_bytes(new_pages) <= std::numeric_limits<uint32_t>::max());
```
gumb0, line 158: This assertion is fixed in follow-up PRs.

lib/fizzy/limits.hpp
```diff
@@ -4,13 +4,20 @@
 
 #pragma once
 
+#include <cstddef>
```
chfast, line 7: Why is this needed?

chfast (approved):

lib/fizzy/limits.hpp
gumb0: Leftover from the version returning size_t, will remove.

PR #818, opened 2022-05-23
ci: Reenable disabled sanitizer options in sanitizers-gcc

axic:
These bugs have been fixed.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -264,8 +264,7 @@ commands:
           working_directory: ~/build
           command: |
             set +e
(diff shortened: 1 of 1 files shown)


gumb0 (approved):

chfast (approved):

chfast (commented):
The bad part is that the compiler is auto-upgraded?

axic:
> The bad part is that the compiler is auto-upgraded?

We have that problem generally?

PR #821, opened 2022-05-23
capi: Add static assertions for FizzyValueType* vs. fizzy:ValType::*

axic:


lib/fizzy/cxx23/utility.hpp
```diff
@@ -0,0 +1,18 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <type_traits>
+
+namespace fizzy
+{
+/// C++14 version of the proposed C++23 feature
+/// https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p1682r3.html
```
gumb0, line 12: I'd prefer a link to https://en.cppreference.com/w/cpp/utility/to_underlying, it's more readable

test/unittests/cxx23_utility_test.cpp
```diff
@@ -0,0 +1,35 @@
...
+    EXPECT_EQ(to_underlying(A8::A), 0xff);
+
+    enum class A32 : uint32_t
+    {
+        A = 0xff,
+        B = 0xffffffff
+    };
+    EXPECT_EQ(to_underlying(A32::A), 0xff);
+    EXPECT_EQ(to_underlying(A32::B), 0xffffffff);
+}
```
gumb0, line 35: Should we test it also for regular C `enum`?

axic: Added.

gumb0 (approved):

test/unittests/cxx23_utility_test.cpp
```diff
@@ -0,0 +1,41 @@
...
+using namespace fizzy;
+using namespace testing;
+
+TEST(cxx23_utility, to_underlying)
+{
+    enum class A
+    {
+        A = 0xff
+    };
+    EXPECT_EQ(to_underlying(A::A), 0xff);
```
chfast, line 20: Add `static_assert(std::is_same_v<decltype(to_underlying(A::A)), int>);

axic: Specs said it is *at least* `int`. So this may fail on some platforms?

gumb0: It will be bigger than `int` only if some values don't fit `int`

chfast: For scoped enumerations (`enum class`) the implicit underlying type is `int`. In case value do not fit, this is compile error. https://godbolt.org/z/1x8ffaadW.

test/unittests/cxx23_utility_test.cpp
```diff
@@ -0,0 +1,45 @@
...
+    };
+    static_assert(std::is_same_v<decltype(to_underlying(A32::A)), uint32_t>);
+    EXPECT_EQ(to_underlying(A32::A), 0xff);
+    EXPECT_EQ(to_underlying(A32::B), 0xffffffff);
+
+    enum C
+    {
+        A = 0xff
+    };
+    static_assert(std::is_same_v<decltype(to_underlying(C::A)), unsigned int>);
```
axic, line 43: Is this stable?

chfast: No. This can be any integer type.

test/unittests/cxx23_utility_test.cpp
chfast: I just wanted to know if this returns the same type as in the enum declaration. If you want to be explicit, you can define it as `enum class A : int`.

chfast (approved):

axic:
> I just wanted to know if this returns the same type as in the enum declaration. If you want to be explicit, you can define it as enum class A : int.

The others are explicit, intentionally left in the implicit one.

chfast (approved):

PR #822, opened 2022-05-23
rust: Add host function test in rust

axic:


bindings/rust/src/lib.rs
```diff
@@ -1310,4 +1310,31 @@ mod tests {
             "invalid offset or size"
         );
     }
+
+    #[test]
+    fn execute_host_function() {
```
gumb0, line 1315: Maybe test name should reflect that it tests error condition with missing import.

chfast (approved):
Fine from me, but apply @gumb0 's suggestion.

bindings/rust/src/lib.rs

PR #826, opened 2022-05-28
Ensure stricter templates for helpers in execution

axic:
Covers floating point helpers: fabs, fneg, fcopysign


chfast (commented):
Please build it locally first before asking for review.

PR #828, opened 2022-05-31
Another set of floating point optimisation attemps

axic:


lib/fizzy/execute.cpp
```diff
@@ -335,27 +339,42 @@ inline constexpr T popcnt(T value) noexcept
...
-inline bool signbit(double value) noexcept
+inline bit_cast_constexpr bool signbit(double value) noexcept
 {
     return (bit_cast<uint64_t>(value) & F64SignMask) != 0;
 }
 
+template <typename T>
+T isnan(T value) noexcept = delete;
+
+inline bit_cast_constexpr bool isnan(float value) noexcept
```
axic, line 355: The creation of this was motivated by making it constexpr. This could be the default in C++23: https://open-std.org/JTC1/SC22/WG21/docs/papers/2019/p0533r5.pdf

PR #832, opened 2022-06-20
ci: Separate spectests preparation

chfast:
The spectests preparation requires pre-built WAST tools so do it once
per pipeline on x64 architecture. The test files are then available
in CircleCI workspace.

image

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -82,7 +82,7 @@ commands:
         default: ""
       build_type:
         type: enum
-        enum: [ "Debug", "Release", "RelWithDebInfo", "Coverage" ]
+        enum: ["Debug", "Release", "RelWithDebInfo", "Coverage"]
       cmake_options:
         type: string
         default: ""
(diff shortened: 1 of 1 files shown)


circle.yml
```diff
@@ -230,6 +230,27 @@ commands:
...
 
+  fetch_spectests:
+    description: "Download and preprocess the WebAssembly spec tests"
+    steps:
+      - install_wabt
+      - run:
+          name: "Download spectest files"
+          working_directory: "~"
+          command: |
+            git clone https://github.com/wasmx/wasm-spec --branch w3c-1.0-tests-backported-20210212 --depth 1 --single-branch
```
axic, line 241: The previous version extracted to `wasm_spec` specifically, but this also uses the same directory, I guess that is the internal directory name?

chfast: It will extract to `~/wasm-spec` because this is the repo name (as previously, just not explicit this time).

axic (approved):
Looks good to me.

Issue #621, opened 2020-10-22
Prototype metering


axic:
Implement metering based on [the ewasm docs](https://github.com/ewasm/design/blob/master/metering.md). Should have a separate gas table and go with everything set at 1 in the beginning, but should experiment with [the numbers from here](https://github.com/ewasm/design/blob/master/determining_wasm_gas_costs.md) later.

We want to implement multiple metering methods and compare them via benchmarking.

The different methods:
1) Naive runtime metering — each instruction in the execution loop checks for cost vs limit
2) Block-based runtime metering — the basic blocks are summed up and there is a pseudo opcode which is injected *or* each block has a cost immediate
3) Injected metering — a call to a `useTicks` function is injected

The injection is also implemented by "[sentinel](https://github.com/ewasm/sentinel-rs)" but better to implement it here.

axic:
Native runtime metering was implemented and released in 0.8.0.

axic:
The metering implemented in #626 is missing the variable cost for `memory_grow`, should add it.

PR #844, opened 2023-01-06
Support metering cost for memory pages

axic:
Addresses https://github.com/wasmx/fizzy/issues/621#issuecomment-1298594860.

### DIFF
--- a/lib/fizzy/execute.cpp
+++ b/lib/fizzy/execute.cpp
@@ -898,8 +898,15 @@ ExecutionResult execute(
         }
         case Instr::memory_grow:
         {
-            stack.top() =
-                grow_memory(*memory, stack.top().as<uint32_t>(), instance.memory_pages_limit);
+            const auto delta_pages = stack.top().as<uint32_t>();
+
+            if constexpr (MeteringEnabled)
+            {
+                if ((ctx.ticks -= get_grow_memory_cost(delta_pages)) < 0)
+                    goto trap;
+            }
+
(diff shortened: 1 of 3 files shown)


lib/fizzy/instructions.hpp
```diff
@@ -31,4 +31,10 @@ const uint8_t* get_instruction_max_align_table() noexcept;
 /// execution metering.
 const int16_t* get_instruction_cost_table() noexcept;
 
+/// Calculates the cost of of memory expansion. Currently set at 65536 per page (1 per byte).
```
axic, line 34:
```suggestion
/// Calculates the cost of memory expansion. Currently set at 65536 per page (1 per byte).
```

chfast (approved):

axic:
@gumb0 any comments here?

lib/fizzy/execute.cpp
```diff
@@ -898,6 +898,12 @@ ExecutionResult execute(
         }
         case Instr::memory_grow:
         {
+            if constexpr (MeteringEnabled)
+            {
+                if ((ctx.ticks -= get_grow_memory_cost(stack.top().as<uint32_t>())) < 0)
```
gumb0, line 903: I would maybe define `const auto delta = stack.top().as<uint32_t>();` to use here and below

gumb0 (approved):

lib/fizzy/execute.cpp
axic: Using `delta_pages` as that is used in both `grow_memory` and `get_grow_memory_cost`.

PR #853, opened 2023-03-10
ci: Add native Apple Silicon (m1) executor

### LINKED ISSUE #709: Support Apple Silicon natively
While due to Rosetta fizzy will just work on macOS, it would make sense considering supporting dual-binary builds with Apple Silicon.

From https://developer.apple.com/documentation/apple_silicon/addressing_architectural_differences_in_your_macos_code:
image

But probably stdc++ hides this difference for us.

CMake features: https://cmake.org/cmake/help/latest/envvar/CMAKE_APPLE_SILICON_PROCESSOR.html and https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_ARCHITECTURES.html

And a helpful tip (from https://cutecoder.org/software/detecting-apple-silicon-shell-script/) to force architecture selection: `arch -arm64 fizzy-bench`

axic:
Fixes #709.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -546,6 +546,20 @@ jobs:
           min_time: "0.01"
       - spectest
 
+  m1-native:
+    machine:
+      resource_class: macos.m1.large.gen1
+    steps:
+      - install_macos_deps
+      - install_testfloat
+      - checkout
+      - build:
+          build_type: Release
+      - test
+      - benchmark:
+          min_time: "0.01"
+      - spectest
+
   benchmark:
     machine:
       image: ubuntu-1604:201903-01
@@ -810,6 +824,9 @@ workflows:
       - x86-linux
       # - arm64-linux
       - arm64-native:
+          requires:
+            - fetch-spectests
+      - m1-native:
           requires:
             - fetch-spectests
       - bindings-rust


circle.yml
```diff
@@ -546,6 +546,20 @@ jobs:
           min_time: "0.01"
       - spectest
 
+  m1-native:
+    machine:
+      resource_class: macos.m1.large.gen1
```
axic, line 551: This task failed: https://app.circleci.com/pipelines/github/wasmx/fizzy/7204/workflows/2386fdfd-78ad-4576-b2b0-781280e6b41b/jobs/67260

> Job was rejected because resource class macos.m1.large.gen1, image default is not a valid resource class

According to https://circleci.com/docs/using-macos/#available-resource-classes unfortunately we can't use these:
> The macos.x86.metal.gen1 resource requires a minimum 24-hour lease. See the [Dedicated Host for macOS](https://circleci.com/docs/dedicated-hosts-macos/) page to learn more about this resource class.
>
> The large and macos.m1.large.gen1 resource classes are only available for customers with an annual contract. [Open a support ticket](https://support.circleci.com/hc/en-us/requests/new) if you would like to learn more about our annual plans.

